PR #29, opened 2015-05-13
Add custom transport option, document AcceptType, UserAgent and Transport to be advanced options

harshavardhana:


harshavardhana:
LGTM

PR #47, opened 2015-05-18
Provide a new function to add user agent with appname, version and list of comments in accordance with

harshavardhana:


harshavardhana:
LGTM

PR #54, opened 2015-05-20
Increase calculated part size for multiparts heuristically using objectsize and maximum parts allowed, rather than using default 5MB

harshavardhana:


harshavardhana:
LGTM

PR #109, opened 2015-07-10
Adding excerpts from aws developers - regarding ignored headers

harshavardhana:


fkautz:
Implemented in all libraries.

Issue #167, opened 2015-09-30
Absolute paths regression with swift3 on openstack kilo


dudymas:
## Summary
I'm currently working with swift3 and auth version2 on openstack kilo, and I noticed that the mc client was reporting 500's when making any request to the s3 endpoint. The issue appears to be related to the absolute paths used by the minio client.

## Analysis

Normally a GET to swift's S3 would go like this:
```
GET /bucket/object/path HTTP/1.1
Host: 10.0.0.0
```

Instead, I'm seeing this from mc:
```
GET https://10.0.0.0/bucket/object/path HTTP/1.1
Host: 10.0.0.0
```

While this works with many s3 implementations, the swift3 implementation is rather conservative and the juno/kilo release of swift doesn't appear to know how to map those routes.

To show this, I rigged up a simple HAProxy config that strips the path out, leaving the rest of the request the same:
```
global
  log 127.0.0.1 local1 notice
  maxconn 256

defaults
  log global
  mode http
  timeout connect 15000ms
  timeout client 15000ms
  timeout server 25000ms
  option http-server-close
  option httplog

listen https-swift-in
  bind *:443 ssl crt /certs/server-bundle.pem
  reqrep ^(^[^\ :]*)\ https://[^\ /]+(/.*) \1\ \2
  server server1 10.0.0.1:8080 maxconn 32
```

Note the 'reqrep' . That did the trick and mc was happily grabbing details and downloading files from the bucket/container.
I don't know how to test delete with the current mc client yet (I just started using it yesterday ;-) ) But that should work too.

harshavardhana:
> I don't know how to test delete with the current mc client yet (I just started using it yesterday ;-) ) But that should work too.

We do not have delete implemented, since we just generally think it is a bad idea to provide delete through a CLI tool that could lead to data loss and it becomes even more evident with ``mc`` like tool. 

If one has to Delete explicitly do it through a management console. Where it is a scheduled and controlled delete which can be easily audited.

harshavardhana:
This issue should be fixed by now, closing this - re-open if seen again.

PR #168, opened 2015-10-01
Remove host from url.Opaque

dudymas:
## Summary
This addresses #167 

## Analysis
The `Opaque` in the `URL` structure ultimately lead to the issue in #167, which had the request not just allowing unicode urls (the expected result), but also not scrubbing -anything-... including the host info (unexpected).

While RFC 2616 allows this ( see http://www.w3.org/Protocols/rfc2616/rfc2616-sec5.html#sec5.1.2 )the result affects openstack's conservative swift3 middleware. The middleware seems to identify and parse the parts of the url correctly, but otherwise doesn't know how to treat the scheme and host details in the URI and fails rather than guess (after all, the middleware doesn't even know that the host details are its own).

## Regressions
This should work fine, because the .Host field in the http.Request had ought to keep special characters (In fact, I think Url.Opaque only exists for avoiding fanciness in the Request URI), however, care should be taken all the same. If a future version of http messes with 'Host', I'm thinking the only real solution would be to dig even deeper into the Url object.

## Notes
I've already run go test on this, however I haven't been able to run the test suite for mc due to gopkg.in/check.v1 not being vendored in.

Should I also make a PR to minio/mc for adding the check.v1 to vendor?

### DIFF
--- a/request.go
+++ b/request.go
@@ -159,9 +159,9 @@ func httpNewRequest(method, urlStr string, body io.Reader) (*http.Request, error
 		if err != nil {
 			return nil, err
 		}
-		uEncoded.Opaque = "//" + uEncoded.Host + separator + bucketName + separator + encodedObjectName
+		uEncoded.Opaque = separator + bucketName + separator + encodedObjectName
 	} else {
-		uEncoded.Opaque = "//" + uEncoded.Host + separator + bucketName
+		uEncoded.Opaque = separator + bucketName
 	}
 	rc, ok := body.(io.ReadCloser)
 	if !ok && body != nil {
@@ -417,12 +417,9 @@ func (r *request) getSignedHeaders() string {
 func (r *request) getCanonicalRequest(hashedPayload string) string {
 	r.req.URL.RawQuery = strings.Replace(r.req.URL.Query().Encode(), "+", "%20", -1)
 
-	// get path URI from Opaque
-	uri := strings.Join(strings.Split(r.req.URL.Opaque, "/")[3:], "/")
-
 	canonicalRequest := strings.Join([]string{
 		r.req.Method,
-		"/" + uri,
+		r.req.URL.Opaque,
 		r.req.URL.RawQuery,
 		r.getCanonicalHeaders(),
 		r.getSignedHeaders(),


harshavardhana:
> I've already run go test on this, however I haven't been able to run the test suite for mc due to gopkg.in/check.v1 not being vendored in.
> 
> Should I also make a PR to minio/mc for adding the check.v1 to vendor?

gopkg.in is vendored in - https://github.com/minio/mc/tree/master/vendor/gopkg.in/check.v1 - there must be some other problem. 

```
               {
                        "canonical": "gopkg.in/check.v1",
                        "comment": "",
                        "local": "vendor/gopkg.in/check.v1",
                        "revision": "11d3bc7aa68e238947792f30573146a3231fc0f1",
                        "revisionTime": "2015-07-29T10:04:31+02:00"
                }
```

harshavardhana:
> This should work fine, because the .Host field in the http.Request had ought to keep special characters (In fact, I think Url.Opaque only exists for avoiding fanciness in the Request URI), however, care should be taken all the same. If a future version of http messes with 'Host', I'm thinking the only real solution would be to dig even deeper into the Url object.

Yes i agree.

harshavardhana:
Thanks for the PR, it looks good. There is one more change needed i.e https://github.com/minio/minio-go/blob/master/request.go#L421

Move the uri block into "r.req.URL.Opaque"

harshavardhana:
Following change will do.

```
@@ -417,12 +417,9 @@ func (r *request) getSignedHeaders() string {
 func (r *request) getCanonicalRequest(hashedPayload string) string {
        r.req.URL.RawQuery = strings.Replace(r.req.URL.Query().Encode(), "+", "%20", -1)

-       // get path URI from Opaque
-       uri := strings.Join(strings.Split(r.req.URL.Opaque, "/")[3:], "/")
-
        canonicalRequest := strings.Join([]string{
                r.req.Method,
-               "/" + uri,
+               r.req.URL.Opaque,
```

dudymas:
ah man, I totally forgot about those v4 signatures. This won't be necessary on legacy, right?

harshavardhana:
> ah man, I totally forgot about those v4 signatures. This won't be necessary on legacy, right?

I will have to see..

harshavardhana:
LGTM

PR #250, opened 2015-12-06
cleanup: Major rewrite - most of the components 

### LINKED ISSUE #241: Disallow any other endpoint other than https://s3.amazonaws.com

### LINKED ISSUE #243: make_bucket - needs to have location argument

### LINKED ISSUE #251: bucket name needs to be complaint with amazon s3
Please refer http://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html.  In addition to this,

* for make bucket, bucket name with '.' should not be allowed due to SSL certificate verification error.
* for make bucket, bucket name with "[A-Z]" should not be allowed due to problem in virtual host style

harshavardhana:
Fixes #251 
Fixes #243 
Fixes #241


harshavardhana:
ping @krishnasrinivas - a large patch sorry :-). 

But there was no way around. Take your time and review this. I will be adding more comments for any tricky parts of the code base.

harshavardhana:
ping @minio/core-team - should look at it as well.

krishnasrinivas:
will review this.

harshavardhana:
> api, err := New(endpoint, accessKeyID, secretAccessKey, inSecure)
> api.config.Signature = SignatureV2

No. This is a bad API design - internal client object states should never be modified in this manner. The reasoning behind NewV2(), NewV4() is to add a call which doesn't do automatic signature verification when the user knows that his/her server is V2 or V4 compatible.

harshavardhana:
> typo here. this should be ErrInvalidObjectName()

Typo indeed.

harshavardhana:
> Are the above checks for '.' needed because the check below covers it.

Looks like those are not needed, will remove them.

harshavardhana:
> we can call it isAmazon() because S3 is the protocol

That is correct.

harshavardhana:
> we can remove the endpointURL in API{} as we already have the info in config.Endpoint
> you use "a.endpointURL" only in putObject in api.go so it will be easy to remove it and instead use a.config.Endpoint there

makes sense, i don't know why i saved it as part of API as well.

harshavardhana:
> if you observe the calls below we do:
> 1. getTargetURL
> 2. getRegion
> 3. newRequest

Yes that is what i am doing right now - in-fact previously they were inside newRequest(). The reason to move them up was i wanted to see how they look separately.

harshavardhana:
> here we are holding the lock for the entire duration of the call (which accesses remote server).
> it's enough to lock to protect the critical region which is one line:
> a.regionMap[bucketName] = location

Correct, let me make that change.

harshavardhana:
> a.mutex.Lock()
>         // Search through regionMap and return if found.
>         region, ok := a.regionMap[bucketName]
>         a.mutex.Unlock()
>         if ok {
>             return region,nil
>         }

Will see how to improve it.

harshavardhana:
> this will return error when prefix is ""
> we can have a new function isValidPrefix()

Let me think about it.

harshavardhana:
> if you observe the calls below we do:
> 1. getTargetURL
> 2. getRegion
> 3. newRequest
> Yes that is what i am doing right now - in-fact previously they were inside newRequest(). The reason to move them up was i wanted to see how they look separately.

Ah now i see why, the problem is separation of the contexts. What happened earlier was we needed to figure out what was the bucketName and objectName for the call. 

What we can in-fact do is put them as part of the requestMetadata and construct the URL and region inside.

krishnasrinivas:
@harshavardhana can you rebase? `This branch is out-of-date with the base branch`

harshavardhana:
> @harshavardhana can you rebase? This branch is out-of-date with the base branch

Done!

PR #255, opened 2015-12-11
putObject: Improve PutObject to return length.

harshavardhana:


put-object.go
```diff
@@ -180,26 +189,19 @@ func (a API) putNoChecksum(bucketName, objectName string, data io.ReadSeeker, si
...
 	if int64(len(dataBytes)) != size {
-		msg := fmt.Sprintf("Data read ‘%s’ is not equal to expected size ‘%s’",
-			strconv.FormatInt(int64(len(dataBytes)), 10), strconv.FormatInt(size, 10))
-		return ErrorResponse{
-			Code:       "UnexpectedShortRead",
-			Message:    msg,
-			BucketName: bucketName,
-			Key:        objectName,
-		}
+		return int64(len(dataBytes)), ErrUnexpectedEOF(int64(len(dataBytes)), size, bucketName, objectName)
```
abperiasamy: return int64(len(dataBytes)), ErrUnexpectedEOF(int64(len(dataBytes)), size, bucketName, objectName)
should be
return 0, ErrUnexpectedEOF(int64(len(dataBytes)), size, bucketName, objectName)

Because this size is reserved for PutObject's written len use.

error-response.go
```diff
@@ -104,6 +106,41 @@ func BodyToErrorResponse(errBody io.Reader) error {
 	return errorResponse
 }
 
+// ErrEntityTooLarge input size is larger than supported maximum.
+var ErrEntityTooLarge = func(totalSize int64, bucketName, objectName string) error {
```
abperiasamy: Rewrite it as 
func ErrEntityTooLarge(totalSize int64, bucketName, objectName string) error {

same goes for all other errors.

PR #257, opened 2015-12-12
request: Add missing custom transport

harshavardhana:


s3-api.go
```diff
@@ -111,9 +111,10 @@ func (a API) putBucketRequest(bucketName, acl, region string) (*Request, error)
 	// Initialize request metadata.
 	var rmetadata requestMetadata
```
abperiasamy: rmetadata should be renamed to reqMetadata

s3-api.go
```diff
@@ -413,9 +417,10 @@ func (a API) listObjectsRequest(bucketName, objectPrefix, objectMarker, delimite
 
 	// Initialize a new request.
 	r, err := newRequest("GET", targetURL, requestMetadata{
```
abperiasamy: rename r to req

s3-multipart-api.go
```diff
@@ -126,10 +127,11 @@ func (a API) initiateMultipartRequest(bucketName, objectName, contentType string
 	multipartHeader.Set("Content-Type", contentType)
 
 	rmetadata := requestMetadata{
```
abperiasamy: rename here too.

PR #260, opened 2015-12-12
endpoint: Validate endpoint properly using isDomainName and isValidIP

harshavardhana:


helpers.go
```diff
@@ -34,6 +35,63 @@ var maxParts = int64(10000)
 // maxPartSize - maximum part size for a single multipart upload operation.
 var maxPartSize int64 = 1024 * 1024 * 1024 * 5
 
+// isDomainName validates input string if it conforms with
+// RFC 1035 and RFC 3696 for domain standards.
+func isDomainName(host string) bool {
```
abperiasamy: Use regex here.

helpers.go
```diff
@@ -34,6 +35,34 @@ var maxParts = int64(10000)
...
+	// See RFC 1035, RFC 3696.
+	if len(host) == 0 {
+		// if string length '0' return false.
+		return false
+	}
+	if len(host) > 255 {
+		// if string larger than 255 characters return false.
+		return false
+	}
+	if host[(len(host)-1):] == "-" || host[:1] == "-" {
```
abperiasamy: This check is redundant. Regex below will catch this case as well.

helpers.go
```diff
@@ -34,6 +35,34 @@ var maxParts = int64(10000)
 // maxPartSize - maximum part size for a single multipart upload operation.
 var maxPartSize int64 = 1024 * 1024 * 1024 * 5
 
+// isDomainName validates if input string is a valid domain name.
+func isDomainName(host string) bool {
+	// See RFC 1035, RFC 3696.
+	if len(host) == 0 {
```
abperiasamy: Regex can check for 1-255 range as well.

helpers.go
```diff
@@ -34,6 +35,50 @@ var maxParts = int64(10000)
...
+	// host cannot start or end with "-"
+	if host[len(host)-1:] == "-" || host[:1] == "-" {
+		return false
+	}
+	// host cannot start or end with "_"
+	if host[len(host)-1:] == "_" || host[:1] == "_" {
+		return false
+	}
+	// host cannot start with a "."
+	if host[:1] == "." {
```
abperiasamy: Cannot end with "." as well.

helpers.go
```diff
@@ -34,6 +35,50 @@ var maxParts = int64(10000)
...
+		return false
+	}
+	for _, r := range host {
```
abperiasamy: use strings.ContainsAny

harshavardhana:
> Cannot end with "." as well.

Domains can end with "." - those are called absolute URLs

harshavardhana:
> Domains can end with "." - those are called absolute domains.

http://www.dns-sd.org/trailingdotsindomainnames.html - for example a request to "https://www.example.com." is as valid as "https://www.example.com"

harshavardhana:
> http://www.dns-sd.org/trailingdotsindomainnames.html - for example a request to "https://www.example.com." is as valid as "https://www.example.com"

You are right since this will be for our purposes, i do not see we need to be Browser like generic in validating domain names.

PR #264, opened 2015-12-14
putBucket: Make bucket request should not use location constraint.

harshavardhana:
Make bucket request should use 'us-east-1' as part of signature.
Region is set as part of location constraint.

### DIFF
--- a/api-put-bucket.go
+++ b/api-put-bucket.go
@@ -32,10 +32,6 @@ func (a API) MakeBucket(bucketName string, acl BucketACL, region string) error {
 		return ErrInvalidArgument("Unrecognized ACL " + acl.String())
 	}
 
-	if region == "" {
-		region = "us-east-1"
-	}
-
 	req, err := a.makeBucketRequest(bucketName, string(acl), region)
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


harshavardhana:
ping @krishnasrinivas

krishnasrinivas:
@harshavardhana this change is correct, but still bucket creation in standard us-east-1 fails:

```
krishna@escape:~/gocode/src/github.com/minio/minio-go/examples/s3$ go run makebucket.go 
2015/12/14 11:45:11 The specified location-constraint is not valid
exit status 1
krishna@escape:~/gocode/src/github.com/minio/minio-go/examples/s3$ 
```

observation is: the api does not expect XML body when the region is us-east-1

this change fixes the issue:
```
+++ b/api-put-bucket.go
@@ -32,8 +32,8 @@ func (a API) MakeBucket(bucketName string, acl BucketACL, region string) error {
                return ErrInvalidArgument("Unrecognized ACL " + acl.String())
        }
 
-       if region == "" {
-               region = "us-east-1"
+       if region == "us-east-1" {
+               region = ""
        }
 
        req, err := a.makeBucketRequest(bucketName, string(acl), region)
@@ -70,7 +70,7 @@ func (a API) makeBucketRequest(bucketName, acl, region string) (*Request, error)
```

krishnasrinivas:
LG

PR #265, opened 2015-12-14
minio-go: Move request into Client, rename API as Client.

harshavardhana:
- This cleanup merges newRequest into Client and uses httpClient
  instance properly in a single Client object context.
- This cleanup also ensures that we use getBucketLocation, inside
  request constructor to avoid the overkill of each requests handling
  them.
- ListBuckets now replies a list of buckets not an iterator.
- Remove mocked tests, we will write them through functional tests.


api-get.go
```diff
@@ -19,17 +19,26 @@ import (
 //  public-read - owner gets full access, others get read access.
 //  public-read-write - owner gets full access, others get full access too.
 //  authenticated-read - owner gets full access, authenticated users get read access.
-func (a API) GetBucketACL(bucketName string) (BucketACL, error) {
+func (a Client) GetBucketACL(bucketName string) (BucketACL, error) {
```
abperiasamy: a -> c

bucket-cache.go
```diff
@@ -144,16 +95,52 @@ func (a API) getBucketLocation(bucketName string) (string, error) {
...
+	if region == "" {
+		region = "us-east-1"
+	}
+
+	// region can be 'EU' convert it to meaningful 'eu-west-1'.
+	if region == "EU" {
+		region = "eu-west-1"
+	}
+
+	a.bucketRgnCache.Set(bucketName, region)
```
abperiasamy: bucketRgnCache -> bucketLocCache

bucket-cache.go
```diff
@@ -144,16 +95,52 @@ func (a API) getBucketLocation(bucketName string) (string, error) {
 		return "", err
 	}
 
-	// location is empty will be 'us-east-1'.
-	if locationConstraint == "" {
-		return "us-east-1", nil
+	region := locationConstraint
```
abperiasamy: region -> location

api-get.go
```diff
@@ -175,12 +154,12 @@ type objectReadSeeker struct {
...
 
 // newObjectReadSeeker wraps getObject request returning a io.ReadSeeker.
-func newObjectReadSeeker(api API, bucket, object string) *objectReadSeeker {
+func newObjectReadSeeker(client Client, bucket, object string) *objectReadSeeker {
 	return &objectReadSeeker{
 		mutex:      new(sync.Mutex),
 		reader:     nil,
 		isRead:     false,
-		api:        api,
+		client:     client,
```
abperiasamy: This implementation is not right. Client should not be embedded inside readseeker. Client should feed readseeker via go routine.

harshavardhana:
> we need not pass objectStatCh as it's already available in the scope

It is a practice, i would want to follow.  Not rely on scoping alone.

abperiasamy:
Only pending items are  readSeeker rewrite, fputObject and fgetObject.

PR #268, opened 2015-12-17
cleanup: listObjectPartsRecursive() was adding complexity. removing it makes code much simpler.

krishnasrinivas:


abperiasamy:
I dont see how it simplifies the code. It will lead to duplicate code every time you use listObjectsParts. I would just rename listObjectPartsRecursive -> listObjectParts and listObjectsParts -> listObjectPartsQuery.

krishnasrinivas:
The go routine management in listObjectPartsRecursive was looking complex and can be avoided, I changed it first to:
```
func (c Client) listObjectPartsRecursive(bucketName, objectName, uploadID string) func()([]objectPartMetadata, isTruncated, bool) {
	var nextPartNumberMarker int
	return func()([]objectPartMetadata, isTruncated, bool) {
		listObjPartsResult, err := c.listObjectParts(bucketName, objectName, uploadID, nextPartNumberMarker, 1000)
		if err != nil {
			return []objectPartMetadata{}, err
		}
		nextPartNumberMarker = listObjPartsResult.NextPartNumberMarker
		return listObjPartsResult.ObjectParts, listObjPartsResult, err
	}
}
```

so this can be done:
list := listObjectPartsRecursive(bucketName, objectName, uploadID)
so list() will fetch the next list of parts. 
But listObjectPartsRecursive() looked too small so moved the functionality to the caller itself. Note that code duplication is very small. Compare it to complexity of listObjectPartsRecursive's go routine

abperiasamy:
Using recursive approach doesnt is still not equivalent. Caller has to perform the query on top where ever you call the function. It doesnt simplify, it only adds more complexity. Channel based approach is idiomatic and simpler. Complexity is reduced to a single function.

krishnasrinivas:
There is no recursion anywhere. Let me send another diff where listObjectPartsRecursive() returns a function instead of a channel. See what you think about it.

harshavardhana:
> There is no recursion anywhere. Let me send another diff where listObjectPartsRecursive() returns a function instead of a channel. See what you think about it.

How will you list all the parts without querying the server again? for the function returned - caller has to do 'for' until IsTruncated is false? -  A channel based is more cleaner in that case. 

This is inherently an issue of all channel based go-routines which need to have an exit strategy, since they can leak without control.  Including our ListObjects, ListIncompleteUploads are affected by this issue - i sent another patch #269 which should address this. 

Done channel is largely used to cleanly exit go-routines, which we didn't use and also didn't notice since 'mc' is a short while operation but it all the more necessary now that we observe it.

krishnasrinivas:
> This is inherently an issue of all channel based go-routines which need to have an  
> exit strategy, since they can leak without control. Including our ListObjects,
> ListIncompleteUploads are affected by this issue - i sent another patch #269 which 
> should address this.

Yes, was about to bring this up in tomorrow's meeting.

check the new patch... felt it looks simpler than the go routine based approach.

harshavardhana:
> Yes, was about to bring this up in tomorrow's meeting.
> 
> check the new patch... felt it looks simpler than the go routine based approach.

It is simple enough but the recursion is still outside i.e every usage of listParts requires you to write a this style of iterator.  This style of approach would have been nice if Golang had python yield equivalent.  I will defer to @abperiasamy on his views as well. 

In general we cannot apply the same approach to ListObjects and ListIncompleteUploads since those are caller facing and it requires us to do explicit channels.

krishnasrinivas:
There is another way, hold on.

krishnasrinivas:
> In general we cannot apply the same approach to ListObjects and
> ListIncompleteUploads since those are caller facing and it requires us to do explicit channels.

that is true. We have to make it easy for the consumers of ListObjects and ListIncompleteUploads

krishnasrinivas:
we can follow the same pattern of passing done channel to the go routine and hide the complexity inside the go routine. closing the pr.

PR #269, opened 2015-12-17
list: Add done channels for listObjects and listIncomplete uploads.

harshavardhana:
This is needed to callers to have an exit strategy, when reading from a channel.

### DIFF
--- a/api-put-object.go
+++ b/api-put-object.go
@@ -225,36 +225,33 @@ func (c Client) putParts(bucketName, objectName, uploadID string, data io.ReadSe
 	defer cleanupStaleTempfiles("multiparts$")
 
 	// total data read and written to server. should be equal to 'size' at the end of the call.
-	var totalWritten int64
-
-	// Seek offset where the file will be seeked to.
-	var seekOffset int64
+	var totalUploadedSize int64
 
 	// Starting part number. Always part '1'.
 	partNumber := 1
 	completeMultipartUpload := completeMultipartUpload{}
-	doneCh := make(chan bool, 1)
-	defer close(doneCh)
-	for objPart := range c.listObjectPartsRecursive(bucketName, objectName, uploadID, doneCh) {
+	// Done channel is used to communicate with the go routine inside listObjectParts.
+	// It is necessary to close dangling routines inside once we break out of the loop.
+	doneCh := make(chan struct{})
+	for objPart := range c.listObjectParts(bucketName, objectName, uploadID, doneCh) {
 		if objPart.Err != nil {
+			close(doneCh)
 			return 0, objPart.Err
 		}
 		// Verify if there is a hole i.e one of the parts is missing
-		// Break and start uploading that part.
+		// Break and start uploading from this part.
 		if partNumber != objPart.PartNumber {
-			// Close listObjectParts channel.
-			doneCh <- true
+			// Close listObjectParts channel by communicating that we are done.
+			close(doneCh)
 			break
 		}
 		var completedPart completePart
 		completedPart.PartNumber = objPart.PartNumber
 		completedPart.ETag = objPart.ETag
 		completeMultipartUpload.Parts = append(completeMultipartUpload.Parts, completedPart)
-		// Add seek Offset for future Seek to skip entries.
-		seekOffset += objPart.Size
-		// Save total written to verify later.
-		totalWritten += objPart.Size
-		// Increment lexically to verify holes in next iteration.
+		// Save total uploaded size which will be incremented later.
+		totalUploadedSize += objPart.Size
+		// Increment additively to verify holes in next iteration.
 		partNumber++
 	}
 
@@ -281,9 +278,9 @@ func (c Client) putParts(bucketName, objectName, uploadID string, data io.ReadSe
 	// Allocate a new wait group.
 	wg := new(sync.WaitGroup)
 
-	// Seek to the new offset if greater than '0'
-	if seekOffset > 0 {
-		if _, err := data.Seek(seekOffset, 0); err != nil {
+	// Seek to the total uploaded size obtained after listing all the parts.
+	if totalUploadedSize > 0 {
+		if _, err := data.Seek(totalUploadedSize, 0); err != nil {
 			return 0, err
 		}
 	}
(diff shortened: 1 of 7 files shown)


harshavardhana:
> It's better to return done channel as the consumers of our API will not have to create a buffered channel themselves.

Actually it works either way, the problem was that should be it be treated as an input or output. The problem by treating it as an output is that it is inconsistent to handle it cleanly. 

A iterator doesn't convey that it will return a doneCh where you should write to it, it only makes sense if its a send only channel. 

A done channel here is an input much like bucketName and objectName - which is more cleaner than an output argument. 

```
counterCh, doneCh := funcCounter()
    for v := range counterCh {
        if v > 4 {
            doneCh <- struct{}{}
            break
        }
    }
``` 

v/s 

```
    doneCh := make(chan struct{}, 1)
    defer close(doneCh)
    for v := range funcCounter(doneCh) {
        if v > 4 {
            doneCh <- struct{}{}
            break
        }
    }
```

harshavardhana:
While this is a simpler approach, i think we have to do this differently.  

By not changing the API but by providing a Struct which implements 

``Next(), End()``` style functions. 

This Next() and End() is an interface which all iterators adhere to. 

```
Next() <- chan interface{}
End() internally sends a 'done <- true' to close the underlying go-routine. 
```

harshavardhana:
More arguments towards taking in a done channel 

Fundamentally a caller knows better than the function itself when the go-routine should quit, so the caller should have the responsibility. This also lets us do shortcuts like this:

```
    for v := range funcCounter(nil) {
     .... 
    }
```

Since the caller knew the channel would be drained to exhaustion, it didn’t need to make a done channel at all. In Go, nil channel never sends, but there’s nothing illegal or panic-inducing about listening to a nil channel, so funcCounter() works just fine.

Furthermore, we can reuse the same quit channel for multiple functions. If we were sending the done signal with done <- struct{}{}, we’d have to send one message for each go-routine that reuses the same done channel and needs to be closed. However because closed channels are always “sending” their blank value, we can stop all of the go-routines that are listening to the same done channel by just doing close(done) once in the application. 

I think the generic iterator style we can keep for some other day.

harshavardhana:
Counter closed. 
```
package main

import (
	"fmt"
	"time"
)

func main() {
	doneCh := make(chan struct{}, 1)
	defer close(doneCh)
	counterCh := funcCounter(doneCh)
	for v := range counterCh {
		if v > 4 {
			doneCh <- struct{}{}
			break
		}
	}
	_, ok := <-counterCh
        fmt.Println("Counter closed: ", !ok)
	time.Sleep(1 * time.Hour)
}

func funcCounter(doneCh <-chan struct{}) <-chan int {
	counterCh := make(chan int, 1)
	go func() {
		defer close(counterCh)
		for i := 0; i < 20; i++ {
			fmt.Print("Writing: ", i)
			select {
			case <-doneCh:
				fmt.Println("I am finished.")
				return
			case counterCh <- i:
				fmt.Println("Written")
			}
		}
	}()
	return counterCh
}
```

Counter closed and drained. 

```
package main

import (
	"fmt"
	"time"
)

func main() {
	counterCh := funcCounter(nil)
	for range counterCh {
	// Counter fully drained. 
	}
	_, ok := <-counterCh
        fmt.Println("Counter channel closed ", !ok)
	time.Sleep(1 * time.Hour)
}

func funcCounter(doneCh <-chan struct{}) <-chan int {
	counterCh := make(chan int, 1)
	go func() {
		defer close(counterCh)
		for i := 0; i < 20; i++ {
			fmt.Print("Writing: ", i)
			select {
			case <-doneCh:
				fmt.Println("I am finished.")
				return
			case counterCh <- i:
				fmt.Println("Written")
			}
		}
	}()
	return counterCh
}
```


Counter not closed and drained. 

```
package main

import (
	"fmt"
	"time"
)

func main() {
	counterCh := funcCounter(nil)
	for v := range counterCh {
               if v > 4 { 
                  // Leaving a dangling counter channel. 
                  break
               } 
	}
	_, ok := <-counterCh
        fmt.Println("Counter channel closed ", !ok)
	time.Sleep(1 * time.Hour)
}

func funcCounter(doneCh <-chan struct{}) <-chan int {
	counterCh := make(chan int, 1)
	go func() {
		defer close(counterCh)
		for i := 0; i < 20; i++ {
			fmt.Print("Writing: ", i)
			select {
			case <-doneCh:
				fmt.Println("I am finished.")
				return
			case counterCh <- i:
				fmt.Println("Written")
			}
		}
	}()
	return counterCh
}
```

krishnasrinivas:
ok, [here](https://blog.golang.org/pipelines) too it takes done channel as argument, looks like it's an accepted pattern. Also if we expect done channel always to be closed we need not have to make a buffered channel in the caller:
```go
func main() {
    // Set up a done channel that's shared by the whole pipeline,
    // and close that channel when this pipeline exits, as a signal
    // for all the goroutines we started to exit.
    done := make(chan struct{})
    defer close(done)
```

harshavardhana:
> ok, here too it takes done channel as argument, looks like it's an accepted pattern. Also if we expect done channel always to be closed we need not have to make a buffered channel in the caller:

Yes that is what we followed as well. We don't have to make it buffered.

api-put-object.go
```diff
@@ -225,36 +225,33 @@ func (c Client) putParts(bucketName, objectName, uploadID string, data io.ReadSe
...
 		if objPart.Err != nil {
 			return 0, objPart.Err
 		}
 		// Verify if there is a hole i.e one of the parts is missing
-		// Break and start uploading that part.
+		// Break and start uploading from this part.
 		if partNumber != objPart.PartNumber {
-			// Close listObjectParts channel.
-			doneCh <- true
+			// Close listObjectParts channel by communicating that we are done.
```
abperiasamy: defer close about should take care of all return

api-put-object.go
```diff
@@ -225,36 +225,33 @@ func (c Client) putParts(bucketName, objectName, uploadID string, data io.ReadSe
...
 	partNumber := 1
 	completeMultipartUpload := completeMultipartUpload{}
-	doneCh := make(chan bool, 1)
+	// Done channel is used to communicate with the go routine inside listObjectParts.
+	// It is necessary to close dangling routines inside once we break out of the loop.
+	doneCh := make(chan struct{})
 	defer close(doneCh)
-	for objPart := range c.listObjectPartsRecursive(bucketName, objectName, uploadID, doneCh) {
+	for objPart := range c.listObjectParts(bucketName, objectName, uploadID, doneCh) {
 		if objPart.Err != nil {
```
abperiasamy: close doneCh here too.

abperiasamy: Examples should also be updated in the same pull req to be consistent.

harshavardhana:
> Examples should also be updated in the same pull req to be consistent.

Ah correct let me do.

harshavardhana:
Pushed the new changes.. fixed examples as well.

krishnasrinivas:
LG

PR #270, opened 2015-12-19
    getObject: Re implement getObjectPartial as readSeeker

harshavardhana:


harshavardhana:
Not finished yet, do not merge.

harshavardhana:
This is ready for review.

harshavardhana:
> Now GetPartialObject signature and its usage differs from the other libraries. We can make it take length, offset arguments to fix it.
> 

This is acceptable, this is only minio-go. 

> I think we don't need to implement Seeker anymore as we have fGetObject and fPutObject now - mc can make use of them instead for cloud-to-cloud copy. fGetObject (which is resumable) can first download the object to a temporary file on local FS, then we can use fPutObject(which is again resumable) to upload. The downside of this is mc will use temporary files, but the advantage is our library will be simpler.
> 

That is bad idea there is no way we can afford that for a 5TB object. 

> If I remember right, Seeker implementation was driven by mc's requirement of resumable cloud-to-cloud copy, is it useful in any other situation?

minio-go is a special case, if it is 'mc'  it is might be still required in other situations. For other languages there are limitations with in the language so the reason to avoid this.

harshavardhana:
> This go routine leaks as close(dataCh) never gets called unless there is an error.
> We can implement Close() call and close the dataCh there.

The problem is there is no io.ReadSeekerCloser in Golang, we cannot implement a Close() call and expose it. I will see if there is any other way.

harshavardhana:
> It's not right to close doneCh during error conditions as any subsequent Read() will block.
> A Seek() failure should not mean that no more operations is possible on the stream.
> We should implement Close() and do close(doneCh) there.

A Seek() failure means it is not seekable and it should fail , but blocking is correct it should not block. Removing the 'doneCh' will help.

harshavardhana:
> This is not error according to documentation:
>  Seeking to any positive offset is legal, but the behavior of subsequent I/O operations on the underlying object is implementation-dependent.

Yes but seeking outside of boundary conditions is not.

harshavardhana:
> But in our case I think it's better to return error as we are implementing a http stream and not a FS stream. A seek beyond object size would mean the application is doing something wrong. But in case of FS the application can issue a write and create a hole in between which is a valid case.

Let me talk to @abperiasamy  about this.

krishnasrinivas:
> > Now GetPartialObject signature and its usage differs from the other libraries. We 
> > can make it take length, offset arguments to fix it.
> This is acceptable, this is only minio-go.

better to check with AB once as this involves api change.

harshavardhana:
> func (r *objectReadSeeker) Read(p []byte) (int, error) {
>     r.mutex.Lock()
>     defer r.mutex.Unlock()
> 
>     if (r.httpReader == nil) {
>         r.httpReader, _, err := r.c.getObject(bucketName, objectName, 0, 0)
>     }
> 
>     if r.prevOffset != r.currentOffset {
>         r.httpReader, _, err = r.c.getObject(bucketName, objectName, msg.currentOffset, 0)
>     }
> 
>     n, err := r.httpReader.Read(p)
> 
>     r.currentOffset += n
>     r.prevOffset = r.currentOffset

This is exactly what should be avoided, having a client inside a Read call.  Current implementation already does this.

krishnasrinivas:
> This is exactly what should be avoided, having a client inside a Read call. Current implementation already does this.

is this how you did initially and ab found problems?
what is the problem with this approach?

harshavardhana:
> better to check with AB once as this involves api change.

That we already agreed this is only for minio-go since it is a special requirement.

harshavardhana:
> is this how you did initially and ab found problems?
> what is the problem with this approach?

A io.ReadSeeker shouldn't do HTTP connections internally. A getObject internally has a bucketCache which also has locks inside, inheriting those locks inside a ReadSeeker is almost calling for bugs which can never be found. 

The right approach is a channel feeding the ReadSeeker when the Reads and Seek come in.  It is clean and no more blind copies of client object.

krishnasrinivas:
> no more blind copies of client object.

no, what if r.c is a pointer to the client object, so there is no copying of client object.

harshavardhana:
> no, what if r.c is a pointer to the client object, so there is no copying of client object.

Still it is not the right approach.

harshavardhana:
It is not about if a pointer is passed around of client or not, it is not Read() to implement or use another HTTP client inside.

harshavardhana:
NOTE: Close() is not implemented yet, so a successful Read() operation might leak.  All other errors will not leak anymore even subsequent Reads upon errors etc.

Leaking operation
```
readSeeker, _, _ := r.GetPartialObject(bucket, object)
totalAmountToRead := 1024 * 1024
for {
   readSeeker.Seek(1024, 0)
   ... make([]byte, 1024)
   readSeeker.Read(p)
   if totalAmountToRead == totalAmountRead {
       break
   } 
```

Right now implementing a timer thread to manage this go-routine, and kill it when dangling.

abperiasamy:
It is not the correct approach. All it takes is a control channel to communicate seek intent and the read method will exclusively read from the req.Body.

harshavardhana:
Closing, this should be done in a separate one.

PR #271, opened 2015-12-19
Implement FGetObject and FPutObject

### LINKED ISSUE #267: Implement fputobject fgetobject

harshavardhana:
Fixes #267

### DIFF
--- a/api-fget-object.go
+++ b/api-fget-object.go
@@ -17,21 +17,23 @@
 package minio
 
 import (
-	"errors"
 	"io"
 	"os"
 	"path/filepath"
 )
 
-// FPutObject - put object a file.
-func (c Client) FPutObject(bucketName, objectName, fileName string) (string, error) {
-	return "", errors.New("Not implemented yet")
-}
-
 // FGetObject - get object to a file.
-func (c Client) FGetObject(bucketName, objectName, fileName string) error {
+func (c Client) FGetObject(bucketName, objectName, filePath string) error {
+	// Input validation.
+	if err := isValidBucketName(bucketName); err != nil {
+		return err
+	}
+	if err := isValidObjectName(objectName); err != nil {
+		return err
+	}
+
 	// Verify if destination already exists.
-	st, err := os.Stat(fileName)
+	st, err := os.Stat(filePath)
 	if err == nil {
 		// If the destination exists and is a directory.
 		if st.IsDir() {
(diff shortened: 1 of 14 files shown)


harshavardhana:
FPutObject is not implemented yet, do not merge.  It will be subsequently added right here.

harshavardhana:
FPutObject is also implemented, but requires a lot of testing, i will confirm once it starts working properly.

harshavardhana:
FPutObject and FGetObject are implemented and working - please review.

harshavardhana:
> should we auto detect contentType as we have the file extension type?

No it is not that straightforward, it is better for user to set.

api-fput-object.go
```diff
@@ -0,0 +1,204 @@
...
+		completedPart.ETag = objPart.ETag
+		completeMultipartUpload.Parts = append(completeMultipartUpload.Parts, completedPart)
+		// Save total uploaded size which will be incremented later.
+		totalUploadedSize += objPart.Size
+		// Increment additively to verify holes in next iteration.
+		partNumber++
+	}
+
+	// Return.
+	return partNumber, totalUploadedSize, completeMultipartUpload, nil
```
abperiasamy: partNumber should be partsCount

api-fput-object.go
```diff
@@ -0,0 +1,204 @@
...
+		// Save the new upload id.
+		uploadID = initMultipartUploadResult.UploadID
+	}
+	return uploadID, nil
+}
+
+// getMultipartStat gather next part number, total uploaded size and list of parts.
+// TODO: reduce number of results.
+func (c Client) getMultipartStat(bucketName, objectName, uploadID string) (int, int64, completeMultipartUpload, error) {
+	// Input validation.
```
abperiasamy: returning 4 variables is fine, but in cases like these int and int64 does not convey what it stands for. Either we can document them or declare variables here (preferred approach).

api-fget-object.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	// Seek to current position for incoming reader.
+	objectReader, objectStat, err := c.getObject(bucketName, objectName, st.Size(), 0)
+	if err != nil {
+		return err
+	}
+
+	// Write to the part file.
+	if _, err = io.CopyN(filePart, objectReader, objectStat.Size); err != nil {
```
abperiasamy: must verify if written != st.Size and return errShortRead

PR #272, opened 2015-12-22
cleanup: Fix fput object, put object and getObjectPartial behavior.

harshavardhana:
- Remove usage of channels and section/part managers.
- Implement ReaderAt interface rather than io.ReadSeeker
- PutObject takes only a Reader and verifies md5sum before upload.
- FPutObject verifies md5sum before upload.


api-fput-object.go
```diff
@@ -139,46 +91,90 @@ func (c Client) FPutObject(bucketName, objectName, filePath, contentType string)
...
-	// gather next part number to be uploaded, total uploaded size and list of all parts uploaded.
-	partNumber, totalUploadedSize, completeMultipartUpload, err := c.getMultipartStat(bucketName, objectName, uploadID)
+	// total data read and written to server. should be equal to 'size' at the end of the call.
+	var totalUploadedSize int64
+
+	// Complete multipart upload.
+	var completeMultipartUpload completeMultipartUpload
+
+	// Fetch previously upload parts and save the total size.
+	objectParts, err := c.listObjectParts(bucketName, objectName, uploadID)
```
abperiasamy: returns partSize, partsInfo, err.

api-fput-object.go
```diff
@@ -139,46 +91,90 @@ func (c Client) FPutObject(bucketName, objectName, filePath, contentType string)
...
 	if err != nil {
 		return 0, err
 	}
+	for _, part := range objectParts {
```
abperiasamy: completedParts should be called uploadedParts.

api-get.go
```diff
@@ -130,122 +129,165 @@ func (c Client) GetBucketACL(bucketName string) (BucketACL, error) {
...
+		defer close(responseCh)
+
+		// Loop through the incoming control messages and read data.
+		for {
+			select {
+			// When the done channel is closed exit our routine.
+			case <-doneCh:
+				return
+			// Control message.
+			case msg := <-requestCh:
```
abperiasamy: call msg as req.Offset and req.Buffer

api-get.go
```diff
@@ -130,122 +129,165 @@ func (c Client) GetBucketACL(bucketName string) (BucketACL, error) {
...
-	offset     int64
-	bucketName string
-	objectName string
+	// User allocated and defined.
+	requestCh  chan<- requestMessage
+	responseCh <-chan responseMessage
+	doneCh     chan<- struct{}
+	objectSize int64
+
+	currentErr error
```
abperiasamy: Rename reqCh, resCh, objSize, prevErr

api-get.go
```diff
@@ -130,122 +129,165 @@ func (c Client) GetBucketACL(bucketName string) (BucketACL, error) {
...
+// response message container to reply back for the request.
+type responseMessage struct {
+	Size  int
+	Error error
 }
 
-// objectReadSeeker container for io.ReadSeeker.
-type objectReadSeeker struct {
+// request message container to communicate with internal go-routine.
+type requestMessage struct {
```
abperiasamy: Rename as readAtRequest and readAtResponse

api.go
```diff
@@ -320,9 +320,13 @@ type CloudStorageClient interface {
 	ListIncompleteUploads(bucket, prefix string, recursive bool, doneCh <-chan struct{}) <-chan ObjectMultipartStat
 
 	// Object Read/Write/Stat operations.
-	GetObject(bucketName, objectName string) (io.ReadSeeker, error)
-	GetPartialObject(bucketName, objectName string, offset, length int64) (io.ReadSeeker, error)
-	PutObject(bucketName, objectName string, data io.ReadSeeker, size int64, contentType string) (int64, error)
+	GetObject(bucketName, objectName string) (io.ReadCloser, ObjectStat, error)
+	PutObject(bucketName, objectName string, data io.Reader, size int64, contentType string) (int64, error)
```
abperiasamy: use named return variables here.

api-fput-object.go
```diff
@@ -139,49 +111,102 @@ func (c Client) FPutObject(bucketName, objectName, filePath, contentType string)
...
 		}
 
 		// Save successfully uploaded size.
-		totalUploadedSize += part.Size
+		totalUploadedSize += partMdata.Size
 
-		// Save successfully uploaded part metadatc.
-		completeMultipartUpload.Parts = append(completeMultipartUpload.Parts, complPart)
+		// Save successfully uploaded part metadata.
+		completeMultipartUpload.Parts[partNumber-1] = complPart
```
krishnasrinivas: array index is past it's length

api-get.go
```diff
@@ -130,122 +129,166 @@ func (c Client) GetBucketACL(bucketName string) (BucketACL, error) {
...
+	}
+
+	// Create request channel.
+	reqCh := make(chan readAtRequest)
+	// Create response channel.
+	resCh := make(chan readAtResponse)
+	// Create done channel for exit strategy.
+	doneCh := make(chan struct{})
+
+	go func() {
```
krishnasrinivas: go routine leakage happens when there is no error

harshavardhana:
> array index is past it's length

> go routine leakage happens when there is no error

That leak is expected. Treating it like fd leak for the time being, a caller has to read till EOF.

harshavardhana:
> That leak is expected. Treating it like fd leak for the time being, a caller has to read till EOF.

In-fact going to do recover()

harshavardhana:
> length := min(len(req.Buffer), objectStat.Size - req.Offset)

Yes correct.

harshavardhana:
> Write must return a non-nil error if it returns n < len(p)

Read this! write will never return a nil error if it has written less, we are verifying doubly that it should be exactly same even if the underlying implementation is buggy.

krishnasrinivas:
> > Write must return a non-nil error if it returns n < len(p)

>Read this! write will never return a nil error if it has written less, we are verifying 
>doubly that it should be exactly same even if the underlying implementation is buggy.

I see, so golang Write does more than what write(2) does.

harshavardhana:
All the fixes have went in, ready for review and examples updated as well.

api-fput-object.go
```diff
@@ -128,9 +85,53 @@ func (c Client) FPutObject(bucketName, objectName, filePath, contentType string)
...
 	fileSize := fileStat.Size()
-	var enableSha256Sum bool
-	if !c.signature.isV2() {
-		enableSha256Sum = true
+	if fileSize > int64(maxMultipartPutObjectSize) {
+		return 0, ErrInvalidArgument("Input file size is bigger than the supported maximum of 5TB.")
+	}
+
+	// NOTE: S3 doesn't allow anonymous multipart requests.
+	if isAmazonEndpoint(c.endpointURL) && c.anonymous {
```
abperiasamy: Better to call google check before this check.

api-fput-object.go
```diff
@@ -139,49 +141,96 @@ func (c Client) FPutObject(bucketName, objectName, filePath, contentType string)
...
+		completedPart.ETag = partInfo.ETag
+		completedPart.PartNumber = partInfo.PartNumber
+		uploadedPartsMap[partInfo.PartNumber] = completedPart
+		totalUploadedSize += partInfo.Size
+		if partInfo.Size >= prevMaxPartSize {
+			prevMaxPartSize = partInfo.Size
+		}
+	}
 
 	// Calculate the optimal part size for a given file size.
```
abperiasamy: rename calculatePartSize -> optimalPartSize

api-fput-object.go
```diff
@@ -139,49 +141,96 @@ func (c Client) FPutObject(bucketName, objectName, filePath, contentType string)
...
+	var completeMultipartUpload completeMultipartUpload
+
+	// Fetch previously upload parts and save the total size.
+	partsInfo, err := c.listObjectParts(bucketName, objectName, uploadID)
 	if err != nil {
 		return 0, err
 	}
+	var uploadedPartsMap = make(map[int]completePart)
+	var prevMaxPartSize int64
+	for _, partInfo := range partsInfo {
```
abperiasamy: Requires code comment.

api-fput-object.go
```diff
@@ -139,49 +141,96 @@ func (c Client) FPutObject(bucketName, objectName, filePath, contentType string)
...
-	partNumber, totalUploadedSize, completeMultipartUpload, err := c.getMultipartStat(bucketName, objectName, uploadID)
+	// total data read and written to server. should be equal to 'size' at the end of the call.
+	var totalUploadedSize int64
+
+	// Complete multipart upload.
+	var completeMultipartUpload completeMultipartUpload
+
+	// Fetch previously upload parts and save the total size.
+	partsInfo, err := c.listObjectParts(bucketName, objectName, uploadID)
 	if err != nil {
```
abperiasamy: let listObjectParts return map instead of list.

harshavardhana:
> let listObjectParts return map instead of list.

Both are different structs, i still have to convert it back to a different one.

api-get.go
```diff
@@ -130,122 +131,193 @@ func (c Client) GetBucketACL(bucketName string) (BucketACL, error) {
...
-}
-
-// Size returns the size of the object.
-func (r *objectReadSeeker) Size() (int64, error) {
-	objectSt, err := r.client.StatObject(r.bucketName, r.objectName)
-	r.stat = objectSt
-	return r.stat.Size, err
+// The behavior of Close after the first call returns error for
+// subsequent Close() calls.
+func (r *objectReadAtCloser) Close() (err error) {
```
abperiasamy: maintain a bool isClosed

api-fput-object.go
```diff
@@ -139,49 +141,100 @@ func (c Client) FPutObject(bucketName, objectName, filePath, contentType string)
...
 		return 0, err
 	}
+	// Save uploaded parts in a map.
+	var uploadedPartsMap = make(map[int]completePart)
+	// Previous maximum part size
+	var prevMaxPartSize int64
+	// Loop through all parts to save them in a local map.
+	for _, partInfo := range partsInfo {
+		var completedPart completePart
```
abperiasamy: requires code comment. Particularly explain that how part size chosen based on max part size.

api-fput-object.go
```diff
@@ -139,49 +141,100 @@ func (c Client) FPutObject(bucketName, objectName, filePath, contentType string)
...
-		// On any error return.
-		if part.Err != nil {
-			return 0, part.Err
+	// Loop through until EOF.
+	for totalUploadedSize < fileSize {
+		// Create a hash multiwriter.
+		hashMD5 = md5.New()
+		hashWriter := io.MultiWriter(hashMD5)
+		if c.signature.isV4() {
+			hashSha256 = sha256.New()
```
abperiasamy: This can be turned into a function to compute hash without disturbing the seek offset.

api-get.go
```diff
@@ -130,122 +131,193 @@ func (c Client) GetBucketACL(bucketName string) (BucketACL, error) {
...
+		// Loop through the incoming control messages and read data.
+		for {
+			select {
+			// When the done channel is closed exit our routine.
+			case <-doneCh:
+				return
+			// Request message.
+			case req := <-reqCh:
+				length := math.Min(float64(len(req.Buffer)), float64(objectStat.Size-req.Offset))
+				httpReader, _, err := c.getObject(bucketName, objectName, req.Offset, int64(length))
```
abperiasamy: requires code comment.

api-get.go
abperiasamy: also requires mutex locking

api-get.go
```diff
@@ -130,122 +131,193 @@ func (c Client) GetBucketACL(bucketName string) (BucketACL, error) {
...
+// input source, ReadAt may return either err == EOF or err == nil.
+//
+// If ReadAt is reading from an input source with a seek offset,
+// ReadAt should not affect nor be affected by the underlying
+// seek offset.
+//
+// Clients of ReadAt can execute parallel ReadAt calls on the
+// same input source.
+//
+// Implementations must not retain p.
```
abperiasamy: This is official defn. Let ReadAt always set error = io.EOF at EOF, even when n = len(b).

ReadAt reads len(b) bytes from the File starting at byte offset off. It returns the number of bytes read and the error, if any. ReadAt always returns a non-nil error when n < len(b). At end of file, that error is io.EOF.

api-put-object-partial.go
```diff
@@ -0,0 +1,190 @@
...
+func (c Client) PutObjectPartial(bucketName, objectName string, data ReadAtCloser, size int64, contentType string) (n int64, err error) {
+	// Input validation.
+	if err := isValidBucketName(bucketName); err != nil {
+		return 0, err
+	}
+	if err := isValidObjectName(objectName); err != nil {
+		return 0, err
+	}
+
+	// NOTE: Google Cloud Storage multipart Put is not compatible with Amazon S3 APIs.
```
abperiasamy: remove this comment.

examples/play/getobjectpartial.go
```diff
@@ -0,0 +1,79 @@
...
+	defer localfile.Close()
+
+	st, err := localFile.Stat()
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	var readAtOffset = st.Size()
+	for {
+		readAtBuffer := make([]byte, 5*1024*1024)
```
abperiasamy: should be outside loop. Reuse allocated buffer.

examples/play/getobjectpartial.go
```diff
@@ -0,0 +1,79 @@
...
+	// This boolean value is the last argument for New().
+
+	// New returns an Amazon S3 compatible client object. API copatibality (v2 or v4) is automatically
+	// determined based on the Endpoint value.
+	s3Client, err := minio.New("play.minio.io:9002", "Q3AM3UQ867SPQQA43P2F", "zuf+tfteSlswRu7BJ86wekitnifILbZam1KYY3TG", false)
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	reader, stat, err := s3Client.GetObjectPartial("bucket-name", "objectName")
```
abperiasamy: stat is declared but not used anywhere.

examples/s3/getobjectpartial.go
```diff
@@ -0,0 +1,79 @@
...
+		log.Fatalln(err)
+	}
+	defer localfile.Close()
+
+	st, err := localFile.Stat()
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	var readAtOffset = st.Size()
```
abperiasamy: here too.

examples/s3/getobjectpartial.go
```diff
@@ -0,0 +1,79 @@
...
+	// This boolean value is the last argument for New().
+
+	// New returns an Amazon S3 compatible client object. API copatibality (v2 or v4) is automatically
+	// determined based on the Endpoint value.
+	s3Client, err := minio.New("s3.amazonaws.com", "YOUR-ACCESS-KEY-HERE", "YOUR-SECRET-KEY-HERE", false)
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	reader, stat, err := s3Client.GetObjectPartial("bucket-name", "objectName")
```
abperiasamy: here too.

multipart-constants.go
```diff
@@ -0,0 +1,34 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package minio
+
+/// Multipart upload defaults.
+
+// minimumPartSize minimum part size per object after which putObject behaves internally as multipart.
```
abperiasamy: in the comment it is preferred to also note 5GB 5TB ..

harshavardhana:
> This can be turned into a function to compute hash without disturbing the seek offset.

This is really hard to do, since our put-object is just a reader and putobject partial is just a readAt with no Seeker.  There is no easier way to converge them for all these functions unless we make a function take an interface{} and do reflection to implement functionality for each functions. 

Since each have different styles it became cumbersome to address all cases.

api-fput-object.go
```diff
@@ -128,60 +86,178 @@ func (c Client) FPutObject(bucketName, objectName, filePath, contentType string)
...
+
+// sectionSums carries checksums for sectionReader.
+type sectionSums struct {
+	MD5Sum    []byte
+	Sha256Sum []byte
+	Size      int64
+}
+
+// sectionSumReader - calculates MD5 and Sha256 for an input section reader.
+func (c Client) sectionSumReader(sectionReader *io.SectionReader) (sectionSums, error) {
```
abperiasamy: sectionSumReader like naming convention is used interface names. Also let sectionreader be inside this function. Make this function as computeHash and return the values directly as variables. It is ok to return 4 variables here in this case. It will be more readable.

api-get.go
```diff
@@ -130,122 +131,176 @@ func (c Client) GetBucketACL(bucketName string) (BucketACL, error) {
...
+	}
+
+	// Create request channel.
+	reqCh := make(chan readAtRequest)
+	// Create response channel.
+	resCh := make(chan readAtResponse)
+	// Create done channel.
+	doneCh := make(chan struct{})
+
+	go func() {
```
abperiasamy: Add a comment here for the go routine about what it does.

api-put-object.go
```diff
@@ -178,28 +174,46 @@ func (c Client) putSmallObject(bucketName, objectName string, data io.ReadSeeker
...
-	if err := isValidObjectName(objectName); err != nil {
-		return 0, err
+// copyMetadata carries checksums for input data.
+type copyMetadata struct {
+	MD5Sum    []byte
+	Sha256Sum []byte
+	Size      int64
+}
+
+// ioCopyStreamSum - calculates Md5sum and Sha256sum for upto partSize amount of bytes.
```
abperiasamy: here too return values directly without struct.

api-put-object.go
```diff
@@ -178,28 +174,46 @@ func (c Client) putSmallObject(bucketName, objectName string, data io.ReadSeeker
...
+	// Seek back to beginning.
+	if _, err := writer.Seek(0, 0); err != nil {
+		return copyMetadata{}, err
 	}
 
-	// Initiate multipart upload.
-	return c.putParts(bucketName, objectName, uploadID, data, size)
+	cpMetadata := copyMetadata{}
+	cpMetadata.Size = size
+	cpMetadata.MD5Sum = hashMD5.Sum(nil)
```
abperiasamy: md5sum of nil?

PR #275, opened 2015-12-25
bucket-cache: Invalidate bucket location upon RemoveBucket

### LINKED ISSUE #273: region caching needs to take care cache invalidation
As the library caches bucket region, we would need to take care below cases
* if a bucket is deleted after the bucket gets cached, we would need to invalidate the cache if we get the request for the same bucket
* if a bucket is deleted and recreated after the bucket gets cached, we would need to invalidate the cache if we get region mismatch errors from amazon s3

harshavardhana:
Additionally saves the location from MakeBucket request.

Fixes #273

### DIFF
--- a/api-put-bucket.go
+++ b/api-put-bucket.go
@@ -55,6 +55,11 @@ func (c Client) MakeBucket(bucketName string, acl BucketACL, location string) er
 		return ErrInvalidArgument("Unrecognized ACL " + acl.String())
 	}
 
+	// If location is empty, treat is a default region 'us-east-1'.
+	if location == "" {
+		location = "us-east-1"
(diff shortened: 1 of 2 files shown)


harshavardhana:
A retry loop is pending on this one, need to talk to AB before i make those changes.

krishnasrinivas:
> A retry loop is pending on this one, need to talk to AB before i make those changes.

In today's call AB suggested not to retry, instead invalidate bucket cache and return error message for that call. For the next call we lookup the region after the cache miss. Since it will be rare for this to happen, this will be a good solution.

krishnasrinivas:
LG

PR #278, opened 2015-12-28
tracing: Add HTTP tracing.

### LINKED ISSUE #276: implement tracing
implement apis:
TraceOn(traceFilePath)
TraceOff()
This should trace the HTTP calls made similar to mc --debug and also should trace any useful information - like if the region is getting used from cache, region map invalidation, if during resume of putObject/fPutObject it should log if it is continuing from where it left off or reuploading etc.
In the functional tests tracing should be 'on'. We should regularly monitor the functional test log to make sure that it is printing what is expected.

harshavardhana:
Current implementation only dumps Request and Response
headers. In time this will print more information method and
method arguments.

Fixes #276

### DIFF
--- a/api.go
+++ b/api.go
@@ -19,10 +19,14 @@ package minio
 import (
 	"encoding/base64"
 	"encoding/hex"
+	"fmt"
 	"io"
 	"net/http"
+	"net/http/httputil"
 	"net/url"
+	"os"
 	"runtime"
+	"strings"
 	"time"
 )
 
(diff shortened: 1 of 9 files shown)


harshavardhana:
ping @krishnasrinivas

api.go
```diff
@@ -159,6 +167,26 @@ func (c *Client) SetCustomTransport(customHTTPTransport http.RoundTripper) {
 	}
 }
 
+// TraceOn - turns on http request tracing.
```
abperiasamy: Enable HTTP tracing.

api.go
```diff
@@ -159,6 +167,26 @@ func (c *Client) SetCustomTransport(customHTTPTransport http.RoundTripper) {
...
+	}
+	// Sets a new output stream.
+	c.traceOutput = outputStream
+
+	// Enable tracing.
+	c.isTraceEnabled = true
+	return nil
+}
+
+// TraceOff - turns off http request tracing.
```
abperiasamy: Disable HTTP tracing.

PR #281, opened 2015-12-29
getObjectPartial: ReadAt shouldn't return io.EOF for each read operat…

harshavardhana:
…ion.

ReadAt should return io.EOF only at the end of file has reached not
for reach bufferred reads.

### DIFF
--- a/api-get.go
+++ b/api-get.go
@@ -17,6 +17,7 @@
 package minio
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


harshavardhana:
> this will not work unless all data has been read. For example if we do:
> (for object sized 10, p[] sized 1, and we have not read bytes 0-8)
> n, err := obj.ReadAt(p, 9, 1)
> this will not return EOF but it should.
> 
> Your fix on minio#280 or my fix both will work fine.

There is more to this than just handing,  http.Read is not returning all the bytes requested. Verifying why.

harshavardhana:
> There is more to this than just handing, http.Read is not returning all the bytes requested. Verifying why.

The reason is that http Read will never return large buffers if requested, it is for us to loop through the stream - so the right approach is to use io.CopyN() to copy into our ReadAt() buffer.

api-get.go
```diff
@@ -193,9 +194,9 @@ func (c Client) GetObjectPartial(bucketName, objectName string) (ReadAtCloser, O
 					}
 					return
 				}
-				size, err := httpReader.Read(req.Buffer)
+				size, err := io.CopyN(req.Buffer, httpReader, int64(length))
 				resCh <- readAtResponse{
-					Size:  size,
+					Size:  int(size),
```
abperiasamy: make it int64

PR #289, opened 2015-12-31
putObject: Do not uploads all parts, these were being uploaded previo…

harshavardhana:
…usly.

The issue happened since the ETag verification failed since S3 sends back
ETags with quotes character in it.


harshavardhana:
ping @krishnasrinivas - take a look here.

api-list.go
```diff
@@ -393,6 +394,8 @@ func (c Client) listObjectParts(bucketName, objectName, uploadID string) (partsI
 		}
 		// Append to parts info.
 		for _, part := range listObjPartsResult.ObjectParts {
+			// Trim off the odd double quotes from ETag.
+			part.ETag = strings.Trim(part.ETag, "\"")
```
abperiasamy: It is always a good practice to use TrimPrefix and TrimSuffix to trim quotes.

PR #290, opened 2016-01-02
Fix constants on 32 bit machines

fd0:
This uses the Go keyword 'const' to really define the values as constants. Since constants in Go are not restricted by size, this allows compiling and using the library on 32 bit machines. For variables, int in the default type, so that on 32 bit machines at least maxSinglePutObjectSize and maxMultipartPutObjectSize overflow, and Go refuses to compile.

Example (before):

    $ GOARCH=386 go test -v
    # github.com/minio/minio-go
    ./constants.go:32: constant 5368709120 overflows int
    ./constants.go:35: constant 5497558138880 overflows int
    FAIL	github.com/minio/minio-go [build failed]

### DIFF
--- a/constants.go
+++ b/constants.go
@@ -20,19 +20,19 @@ package minio
 
 // minimumPartSize - minimum part size 5MiB per object after which
 // putObject behaves internally as multipart.
-var minimumPartSize int64 = 1024 * 1024 * 5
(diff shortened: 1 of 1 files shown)


harshavardhana:
LGTM

PR #297, opened 2016-01-07
Add coverage

harshavardhana:


harshavardhana:
Coveralls will not work for us..

PR #300, opened 2016-01-07
getObject: Close() should return error only on second attempt.

harshavardhana:


fd0:
LGTM

abperiasamy:
isClosed is redundant. Then you will also have to additionally check for isClosed in other methods. You can set o=nil on close instead. Let close return prevErr for non-EOF cases alone.

PR #301, opened 2016-01-10
getReaderSize: Generalize getReaderSize to use reflection over Method Names.

harshavardhana:
getReaderSize can be generalized to fetch values based on capabilities.

### DIFF
--- a/api-put-object.go
+++ b/api-put-object.go
@@ -17,42 +17,66 @@
 package minio
 
 import (
-	"bytes"
 	"io"
 	"io/ioutil"
 	"net/http"
 	"os"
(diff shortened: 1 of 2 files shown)


GarimaKapoor:
Store reflected values in variables and check them for nil before accessing its value. This way you can avoid panic-recover. --ab

harshavardhana:
> Store reflected values in variables and check them for nil before accessing its value. This way you can avoid panic-recover. --ab

The panic will occur for cases where reflect.Int() is called, if the underlying value is not int. So panic can still happen since we are using reflect.

PR #306, opened 2016-01-16
putObjectReadAt: Always upload parts in order.

harshavardhana:
This change also reduces the chance of significant random i/o on the source.

### DIFF
--- a/api-put-object-common.go
+++ b/api-put-object-common.go
@@ -43,6 +43,24 @@ func isReadAt(reader io.Reader) (ok bool) {
 	return
 }
 
+// shouldUploadPart - verify if part should be uploaded.
+func shouldUploadPart(objPart objectPart, objectParts map[int]objectPart) bool {
+	// If part not found should upload the part.
+	uploadedPart, found := objectParts[objPart.PartNumber]
+	if !found {
+		return true
+	}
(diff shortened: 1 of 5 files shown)


api-put-object-common.go
```diff
@@ -43,6 +43,19 @@ func isReadAt(reader io.Reader) (ok bool) {
 	return
 }
 
+// isPartUploaded - true if part is already uploaded.
+func isPartUploaded(objPart objectPart, objectParts map[int]objectPart) bool {
```
abperiasamy: isPartUploaded -  true if part is already uploaded.
This gives an illusion that this part is really uploaded if ret value is true. Instead this function should be more like isPartSelected.

api-put-object-readat.go
```diff
@@ -25,73 +25,16 @@ import (
...
-						readAtOffset:    int64(uploadedPart.PartNumber-1) * partSize,
-						missingPartSize: partSize,
-					}
-				}
-			}
-		}
-		// Increment part number here.
-		partNumber++
+// isPartUploadedReadAt - true if part is already uploaded.
+func isPartUploadedReadAt(objPart objectPart, objectParts map[int]objectPart) bool {
```
abperiasamy: isPartSelectedReadAt

PR #310, opened 2016-01-20
putObjectReadAt: use sectionReader itself rather than looping over ReadAt.

harshavardhana:


api-presigned.go
```diff
@@ -21,9 +21,9 @@ import (
 	"time"
 )
 
-// PresignedGetObject - Returns a presigned URL to access an object without credentials.
+// presignedMethodObject - Returns a presigned URL for an input 'method'.
 // Expires maximum is 7days - ie. 604800 and minimum is 1.
-func (c Client) PresignedGetObject(bucketName, objectName string, expires time.Duration) (string, error) {
+func (c Client) presignedMethodObject(method, bucketName, objectName string, expires time.Duration) (string, error) {
```
abperiasamy: rename to presignURL

api-presigned.go
```diff
@@ -35,10 +35,15 @@ func (c Client) PresignedGetObject(bucketName, objectName string, expires time.D
 		return "", err
 	}
 
+	// Defaults to "GET"
+	if method == "" {
+		method = "GET"
+	}
+
```
abperiasamy: Return invalid argument error if method is empty.

api-presigned.go
abperiasamy: func (c Client) presignedMethodObject(op string, bucketName string, objectName string, expires time.Duration) (url string, e error) {

api-presigned.go
```diff
@@ -50,33 +55,16 @@ func (c Client) PresignedGetObject(bucketName, objectName string, expires time.D
 	return req.URL.String(), nil
 }
 
+// PresignedGetObject - Returns a presigned URL to access an object without credentials.
+// Expires maximum is 7days - ie. 604800 and minimum is 1.
+func (c Client) PresignedGetObject(bucketName, objectName string, expires time.Duration) (string, error) {
```
abperiasamy: same here too. each arg has its own type. return var is named.

api-presigned.go
```diff
@@ -118,24 +106,36 @@ func (c Client) PresignedPostPolicy(p *PostPolicy) (map[string]string, error) {
 	}
 
 	// Add date policy.
-	p.addNewPolicy(policyCondition{
+	err = p.addNewPolicy(policyCondition{
```
abperiasamy: when you have only err (and or temp variable) returned, better to evalaute if here..

api_functional_v2_test.go
```diff
@@ -181,6 +181,109 @@ func TestRemovePartiallyUploadedV2(t *testing.T) {
 	}
 }
 
+// Tests resumable put object cloud to cloud.
```
abperiasamy: Even test code requires comment. Description of what each test is trying to achieve.

PR #313, opened 2016-01-22
api: Implement PutObjectWithProgress reader.

harshavardhana:


api-put-object-file.go
```diff
@@ -174,6 +174,12 @@ func (c Client) putObjectMultipartFromFile(bucketName, objectName string, fileRe
 			return 0, err
 		}
 
+		var reader io.Reader
+		reader = sectionReader
+		if progress != nil {
```
abperiasamy: requires code comment.

api-put-object-progress.go
```diff
@@ -0,0 +1,122 @@
...
+	if err != nil {
+		blackHolePool.Put(bufp)
+		return 0, err
+	}
+	return written, nil
+}
+
+// PutObjectWithProgress - With progress.
+func (c Client) PutObjectWithProgress(bucketName, objectName string, reader io.Reader, progress io.Reader, contentType string) (n int64, err error) {
+	// Input validation.
```
abperiasamy: progress reader should be the last arg.

api-put-object-progress.go
```diff
@@ -0,0 +1,122 @@
...
+ */
+
+package minio
+
+import (
+	"io"
+	"sync"
+)
+
+var blackHolePool = sync.Pool{
```
abperiasamy: This logic is not required.

hook-reader.go
```diff
@@ -0,0 +1,47 @@
...
+		if herr != io.EOF {
+			return n, herr
+		}
+	}
+	return n, err
+}
+
+// newHook returns a io.Reader which implements hookReader that
+// reports the data read from the source to the hook.
+func newHook(source, hook io.Reader) io.Reader {
```
abperiasamy: return source as is, if hook is nil. This will avoid handling 'else' case.

hook-reader.go
```diff
@@ -0,0 +1,47 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package minio
+
+import "io"
+
+type hookReader struct {
```
abperiasamy: Add comment: hookReader hooks additional reader in the source stream. It is useful for making progress bars. Second reader is appropriately notified about the exact number of bytes read from the primary source on each Read operation.

examples/s3/putobject-progress.go
```diff
@@ -0,0 +1,61 @@
...
+
+	objectInfo, err := reader.Stat()
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	progress := pb.New64(objectInfo.Size)
+	progress.Start()
+
+	n, err := s3Client.PutObjectWithProgress("my-bucketname", "my-objectname-progress", reader, progress, "application/octet-stream")
```
abperiasamy: Add comment: progress reader is notified as PutObject makes progress with the read. For partial resume put object, progress reader is appropriately advanced.

api-put-object-progress.go
```diff
@@ -0,0 +1,122 @@
...
+	New: func() interface{} {
+		b := make([]byte, 8*1024*1024)
+		return b
+	},
+}
+
+func ioProgressN(writer io.Writer, reader io.Reader, size int64) (written int64, err error) {
```
abperiasamy: This function doesnt make sense. It should be replaced with:  io.CopyN(ioutil.Discard, source, size).

api-put-object-progress.go
```diff
@@ -0,0 +1,108 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package minio
+
+import "io"
+
+// PutObjectWithProgress - With progress.
```
abperiasamy: Regular PutObject should now become a wrapper to this call with last arg as nil.

PR #314, opened 2016-01-22
make BucketInfo and ObjectInfo JSON-aware

balamurugana:


harshavardhana:
> This is not javascript style avoid underscore and hyphens - change this to 'creationDate'

I see this was modified, LGTM.

PR #318, opened 2016-01-26
Remove duplicate code

fwessels:


harshavardhana:
> This code is needed to copy the object to local file. Why do you think this is duplicate?

Ah now i see.

PR #320, opened 2016-01-28
copyBuffer: Implement copyBuffer to support old Golang versions.

harshavardhana:
Golang versions < 1.5.x do not implement io.CopyBuffer, all
dependent applications will fail if they wish to build their
binaries for linux distributions which are still shipping older
versions of Golang.


abperiasamy:
This is an interim solution. This copyBuffer approach will be replaced by streaming io.Copy over a single GetObject body.

PR #325, opened 2016-02-05
Splitting the main interface into multiple smaller ones

arschles:
Having a single, all-encompassing interface is ok, but it’s often useful to pass around a smaller chunk of functionality, to reduce the surface area of functionality that a function gets access to. Additionally, passing smaller interfaces around makes it much easier to write unit tests.

This PR is a first try at splitting up the `CloudStorageClient` interface.

Note that the same tests fail here as on master (see https://github.com/minio/minio-go/issues/326).

### DIFF
--- a/api.go
+++ b/api.go
@@ -491,26 +491,17 @@ func (c Client) makeTargetURL(bucketName, objectName, bucketLocation string, que
 // CloudStorageClient - Cloud Storage Client interface.
 type CloudStorageClient interface {
 	// Bucket Read/Write/Stat operations.
-	MakeBucket(bucketName string, cannedACL BucketACL, location string) error
-	BucketExists(bucketName string) error
-	RemoveBucket(bucketName string) error
-	SetBucketACL(bucketName string, cannedACL BucketACL) error
-	GetBucketACL(bucketName string) (BucketACL, error)
+	BucketOperator
+	BucketACLOperator
+	ObjectOperator
+	FileOperator
 
-	ListBuckets() ([]BucketInfo, error)
-	ListObjects(bucket, prefix string, recursive bool, doneCh <-chan struct{}) <-chan ObjectInfo
 	ListIncompleteUploads(bucket, prefix string, recursive bool, doneCh <-chan struct{}) <-chan ObjectMultipartInfo
 
 	// Object Read/Write/Stat operations.
-	GetObject(bucketName, objectName string) (reader *Object, err error)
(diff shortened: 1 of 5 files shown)


harshavardhana:
> Having a single, all-encompassing interface is ok, but it’s often useful to pass around a smaller chunk of functionality, to reduce the surface area of functionality that a function gets access to. Additionally, passing smaller interfaces around makes it much easier to write unit tests.
> 
> This PR is a first try at splitting up the CloudStorageClient interface.

We are never going to test functions outside of 'New()' in this manner,  This makes only sense when you bring in 'New()' for each of these interfaces or they have a way to be instantiated independently and they can be independently usable which is not the case.

Everytime you need to test any of these methods you are going to do 'New()' and that replies back CloudStorageClient - so you know what methods to expect.

Also the interface is kept largely for function signature verification, i.e some one trying to fix or edit a function gets an error so that the API is is not made accidentally incompatible.

arschles:
>We are never going to test functions outside of 'New()' in this manner, This makes only sense when you bring in 'New()' for each of these interfaces or they have a way to be instantiated independently and they can be independently usable which is not the case.

Makes sense, but others may use them in their code for a few reasons:

- It's significantly easier to unit test a `func doSomething(bOp BucketOperator)` than a `func doSomething(csl CloudStorageClient)`
- Large interfaces are often considered an anti-pattern in Go if they're not composed of smaller ones
- Passing around `CloudStorageClient`s around to code that only uses a few methods on it violates the principle of least responsibility

>Everytime you need to test any of these methods you are going to do 'New()' and that replies back CloudStorageClient - so you know what methods to expect.

True for internal testing, but again, not for consumers of this packages

>Also the interface is kept largely for function signature verification, i.e some one trying to fix or edit a function gets an error so that the API is is not made accidentally incompatible.

I don't believe this PR breaks function signature verification, unless I misunderstand what you mean.

harshavardhana:
> Makes sense, but others may use them in their code for a few reasons:
> 
> It's significantly easier to unit test a func doSomething(bOp BucketOperator) than a func doSomething(csl CloudStorageClient)
> Large interfaces are often considered an anti-pattern in Go if they're not composed of smaller ones
> Passing around CloudStorageClients around to code that only uses a few methods on it violates the principle of least responsibility

Yes you are right, but the fundamental problem was the interface itself was not necessary, so we removed it.

arschles:
Ah, I see. Thanks for clarifying!

PR #335, opened 2016-02-29
request: Add expect header with 100-continue

### LINKED ISSUE #101: http: Verify if we need to get http client support for 100 Continue
https://github.com/golang/go/issues/3665 - waiting on this Go 1.6

harshavardhana:
Fixes #101

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,8 @@ env:
 - ARCH=i686
 
 go:
-- 1.5.2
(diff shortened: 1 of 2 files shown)


harshavardhana:
> 100-continue is used by all aws-sdk's and our server also implements it. I don't remember when we decided not to do it.
> 

Lets discuss this again and merge later.

PR #337, opened 2016-03-05
retry: Implement transparent retry logic for all requests.

### LINKED ISSUE #330: request: Add a transparent retry logic for network errors

harshavardhana:
Retry is transparent and retries upto maximum of '5' attempts
in binomial time ie.

  1st attempt after 1sec.
  2nd attempt after 2sec.
  ...
  5th attempt after 16sec.

Fixes #330


retry.go
```diff
@@ -0,0 +1,95 @@
...
+
+package minio
+
+import (
+	"math"
+	"time"
+)
+
+// maxRetries is the maximum number of retries before stopping.
+var maxRetries = 5
```
abperiasamy: MaxRetries = 3

Make 3 as default and public variable.

retry.go
```diff
@@ -0,0 +1,95 @@
...
+
+// ErrMaxRetriesReached - Maximum retries reached.
+func ErrMaxRetriesReached(message string) error {
```
abperiasamy: This error should never be exposed. Users should only see the real error such as connection refused or host unreachable.

retry.go
```diff
@@ -0,0 +1,60 @@
...
+	"RequestTimeout":        {},
+	"Throttling":            {},
+	"ThrottlingException":   {},
+	"RequestLimitExceeded":  {},
+	"RequestThrottled":      {},
+	"InternalError":         {},
+	"ExpiredToken":          {},
+	"ExpiredTokenException": {},
+}
+
```
abperiasamy: hide this inside isS3CodeRetryable using switch case.

PR #338, opened 2016-03-06
Use a trivial implementation of min for int64 instead of Math.min

ckemper67:
Math.min works on float64 and the conversion limits us to 53 bits. See
http://mrekucci.blogspot.com/2015/07/dont-abuse-mathmax-mathmin.html
for an explanation of the potential issues.

Note that this is not likely to cause any problems in the
foreseeable future, so this is more a matter of principle of not using
floats for integer calculations.

### DIFF
--- a/api-get.go
+++ b/api-get.go
@@ -20,7 +20,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"math"
 	"net/http"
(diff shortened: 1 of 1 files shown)


harshavardhana:
Thanks for your contribution @ckemper67

PR #339, opened 2016-03-06
retry: Enhance retry to support cases of restricted bucket policy.

### LINKED ISSUE #332: request: Add support of redirected to the correct endpoint
When bucket located in non-Standard US region and policy not allow getBucketLocation then request send us-east-1 and s3 responds with
>The bucket you are attempting to access must be addressed using the specified endpoint. Please send all future requests to this endpoint.

In such case minio-go may try to retry and redirect all future request to correct endpoint suggested by s3.

https://github.com/restic/restic/issues/388#issuecomment-182318912

harshavardhana:
This is necessary if GetBucketLocation() is explicitly disabled
for a user policy, fallback to the errorResponse Region and attempt
the request again if necessary. - fixes #332

An example of one such policy

```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": "s3:ListAllMyBuckets",
            "Resource": "arn:aws:s3:::*"
        },
        {
            "Effect": "Allow",
            "Action": [
                "s3:GetObject",
                "s3:GetObjectAcl",
                "s3:GetObjectVersion",
                "s3:GetObjectVersionAcl",
                "s3:ListBucket"
            ],
            "Resource": [
                "arn:aws:s3:::jarjarbing",
                "arn:aws:s3:::jarjarbing/*"
            ]
        }
    ]
}
```

Clearly user with this policy can list objects on ``jarjarbing`` bucket. But
cannot possibly use our library in this case. This patch addresses this by
using the information returned in the error response and retry the request.

### DIFF
--- a/api.go
+++ b/api.go
@@ -339,7 +339,7 @@ func (c Client) do(req *http.Request) (*http.Response, error) {
 		if ok && strings.Contains(urlErr.Err.Error(), "EOF") {
 			return nil, fmt.Errorf("Connection closed by foreign host %s. Retry again.", urlErr.URL)
 		}
-		return resp, err
+		return nil, err
 	}
 
 	// Response cannot be non-nil, report if its the case.
@@ -373,6 +373,8 @@ func (c Client) executeMethod(method string, metadata requestMetadata) (res *htt
 	// error until maxRetries have been exhausted, retry attempts are
 	// performed after waiting for a given period of time in a
 	// binomial fashion.
+	//
+	// FIXME: Handle network related errors.
 	for range newRetryTimer(MaxRetry, time.Second) {
 		if isRetryable {
 			// Seek back to beginning for each attempt.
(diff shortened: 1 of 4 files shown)


api.go
```diff
@@ -397,15 +397,21 @@ func (c Client) executeMethod(method string, metadata requestMetadata) (res *htt
 		res, err = c.do(req)
 		if err != nil {
 			// Fail quickly here, no need to retry.
-			return nil, err
+			return res, err
 		}
```
abperiasamy: when ever err is non-nil, res is undefined. Why would you return res instead of nil?

harshavardhana: > when ever err is non-nil, res is undefined. Why would you return res instead of nil?
> Also local network errors, server disconnects needs retry as well.

Network errors will do separately, it is not part of this patch.  Be 'res' nil or non nil we shouldn't define the value. I am returning everything that is sent by the lower level call to .Do()

abperiasamy: nil is returned deliberately when error is set. It is the correct coding style. We cannot assume people write clean code with proper error checks all the time.

abperiasamy: Also add a FIXME till the network errors are handled.

PR #341, opened 2016-03-06
Cleaned up the retry logic and introduced retry jitter.

ckemper67:
Added support for computing the backoff time with jitter according to
https://www.awsarchitectureblog.com/2015/03/backoff.html

Added a test for the backoff time computation.

### DIFF
--- a/api.go
+++ b/api.go
@@ -384,7 +384,7 @@ func (c Client) executeMethod(method string, metadata requestMetadata) (res *htt
 	// error until maxRetries have been exhausted, retry attempts are
 	// performed after waiting for a given period of time in a
 	// binomial fashion.
-	for range newRetryTimer(MaxRetry, time.Second) {
+	for range newRetryTimer(MaxRetry, time.Second, time.Second*30, MaxJitter) {
 		if isRetryable {
 			// Seek back to beginning for each attempt.
 			if _, err = bodySeeker.Seek(0, 0); err != nil {
--- a/api_unit_test.go
+++ b/api_unit_test.go
@@ -26,6 +26,7 @@ import (
 	"os"
 	"strings"
 	"testing"
+	"time"
 )
 
 type customReader struct{}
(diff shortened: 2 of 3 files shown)


abperiasamy:
Exponential backoff with jitter is meant for optimistic concurrency control, not for retry logic. Binomial delay is correct approach for retry.

harshavardhana:
> Exponential backoff with jitter is meant for optimistic concurrency control, not for retry logic. Binomial delay is correct approach for retry.

http://docs.aws.amazon.com/general/latest/gr/api-retries.html - This is something @ckemper67 shared about why Jitter is necessary. It seems valid.

ckemper67:
Here is another explanation from Google's Cloud Storage Documentation: https://cloud.google.com/storage/docs/exponential-backoff

abperiasamy:
Makes sense. Couple of suggestions..
1) I think it is better to reduce the default jitter by half or increase default max retries to 5. Main purpose    of bringing retries was to give sufficient time to reestablish lost network connections. 
2) Can you also add seeding, may using init() for once or inside this function is also fine.
    rand.Seed(time.Now().UTC().UnixNano())

ckemper67:
I decided to change the max retries. My main reason is that the article was pretty clear that is was better to use jitter over the whole interval. But I also decided to expose the jitter in the newRetryTime method to make it easier to change.

Why do you want to add seeding? It doesn't really matter if we get the same random sequence  every time does it? And we don't need a source of entropy as you typically need for encyption.

abperiasamy:
Without seed, all the servers will hit the server with the same random sequence. The idea of random jitter is to introduce tiny random delays between concurrent requests to avoid collision. Let me read the spec again.

ckemper67:
Ah I can see your point, let me think about that a bit. I think it assumes
that they all start synchronized to begin with and perform the same
actions, but it may be better to add seeding to guarantee to spread that
out some more.

ckemper67:
PTAL

I decided to go with my own instance of a random number generator that I
can initialize to the current time.

BTW: Let me know if I should squash the commits on the branch. I kept the
corrections from the review in a separate commit so far.

harshavardhana:
@ckemper67 since they are related.  It makes sense to squash them. Thanks for your contribution.

abperiasamy:
this PR is ready after squashing. Awesome contribution @ckemper67

ckemper67:
Cool thanks. I am at work now, so I will do it tonight.

In the meantime: I was not sure if I understand the code correctly because
I am still trying to wrap my head around how to use channels effectively.

It looks to me like the code will now wait for 1s before it executes the
first try. Can you confirm that that is not the case? Otherwise we should
change the retry timer to return 0 for the first attempt.

abperiasamy:
@ckemper67 you are right. It would sleep before first attempt. This bug is there in the master branch as well. :)

Send first and then sleep (swap the order). It should fix the problem

attemptCh <- i + 1 // Attempts start from 1.
time.Sleep(exponentialBackoffWait(unit, i, cap, jitter))

harshavardhana:
Merged..

PR #347, opened 2016-03-10
api: GetObject now only fetches the Reader when offset changes.

### LINKED ISSUE #322: api: Re-work GetObject() to not fetch new readers for each Read() operation.
Current code *minio.Object 

* Read(p []byte) - causes the underlying request to fetch new readers for all ranges.  If p buffer is small there will be a flood of Get requests which is not necessary.  

To avoid this we started using io.CopyBuffer which is a newer API and leads to build failures.  But the right fix is to keep the reader until the Read() finishes.

harshavardhana:
Optimized GetObject() now fetches new Reader only if there
is an offset change.

Fixes #322

### DIFF
--- a/api-get.go
+++ b/api-get.go
@@ -140,8 +140,9 @@ func (c Client) GetObject(bucketName, objectName string) (*Object, error) {
 	if err := isValidObjectName(objectName); err != nil {
 		return nil, err
 	}
-	// Send an explicit info to get the actual object size.
-	objectInfo, err := c.StatObject(bucketName, objectName)
+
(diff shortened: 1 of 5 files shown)


harshavardhana:
> That optimization is not necessary, it is cleaner to indicate the ReadAt() behavior as is. ReadAt use case is rare, so it is okay.

Actually it is fine to do it.

harshavardhana:
@abperiasamy - don't merge this just yet. I have few things to discuss before this can be changed.

PR #351, opened 2016-03-11
rand: Implement locked source to protect from races.

harshavardhana:


api.go
```diff
@@ -124,6 +125,29 @@ func New(endpoint string, accessKeyID, secretAccessKey string, insecure bool) (*
 	return clnt, nil
 }
 
+// lockedRandSource provides protected rand source, implements rand.Source interface.
+type lockedRandSource struct {
+	lk  sync.Mutex
+	src rand.Source
+}
+
```
abperiasamy: change lockedRandSource to lockedRand and add a new method NewSource()

PR #358, opened 2016-03-18
include region specific validation for aws s3 endpoints

smothiki:
refer https://github.com/deis/builder/issues/266


harshavardhana:
@smothiki - as discussed this patch is not needed, let me know how your validation goes with new 'mc'.

PR #365, opened 2016-03-24
api: Simplify copyObject API, return only error if any

harshavardhana:


api-put-object-copy.go
```diff
@@ -0,0 +1,76 @@
...
+
+package minio
+
+import (
+	"net/http"
+	"strings"
+)
+
+// supportCopyObjectHeaders - list of supported headers
+var supportedCopyObjectHeaders = http.Header{
```
abperiasamy: This can simply be a const string array. I would even move it inside CopyObject method if it is not useful outside in the future.

harshavardhana: > This can simply be a const string array. I would even move it inside CopyObject method if it is not useful outside in the future.

A string array requires for us to loop for each keys, to validate if its the right one. It is better to stick with map.

api-put-object-copy.go
```diff
@@ -0,0 +1,76 @@
...
+package minio
+
+import (
+	"net/http"
+	"strings"
+)
+
+// CopyObject - copy a source object into a new object with the provided name in the provided bucket
+func (c Client) CopyObject(bucketName string, objectName string, xAmzHeaders http.Header) error {
+	// Input validation.
```
abperiasamy: type CopyCondition int16
const  (
CopyAlways CopyCondition = iota
CopyIfMatchETag
CopyIfNoMatchETag
CopyIfUnmodifiedSince
CopyIfModifiedSince
)
- func (c Client) CopyObject(bucketName string, objectName string, xAmzHeaders http.Header) error 
+ func (c Client) CopyObject(bucketName string, objectName string, objectSource string, condt CopyCondition) error

harshavardhana: > type CopyCondition int16
> const (
> CopyAlways CopyCondition = iota
> CopyIfMatchETag
> CopyIfNoMatchETag
> CopyIfUnmodifiedSince
> CopyIfModifiedSince
> )

The conditions are not true false, they have values. that is why it is a map of key/values. So this approach cannot handle this.

PR #366, opened 2016-03-24
putObject: using buffer.Bytes() causes memory copy. Avoid it.

harshavardhana:
In Current code it was done so that we could create a Seeker, so that
if the request fails we could retry with the same body buffer.

But memory allocation doubles for example in case of uploading a
1GB file a maximum size of 550MB buffer in case of putObjectStream()
leads to 1GB usage this is bad for tiny distros and embedded systems.

Ideally it would have preferred if bytes.NewBuffer() gave us a Seeker()
too, but we will fix that with our own implementation.


harshavardhana:
This also fixes a bug - https://github.com/minio/mc/issues/1641

PR #368, opened 2016-03-28
Tests: tests for utils.go

hackintoshrao:
Good coverage and structuring of tests for utils.go is necessary for start with tests fr functions in the repo. Since functions from utils is used in most of the other functionalities its not possible to write tests for them unless tests for utils are covered and structured.


utils_test.go
```diff
@@ -0,0 +1,368 @@
...
+
+import (
+	"fmt"
+	"net/url"
+	"testing"
+	"time"
+)
+
+func TestGetEndpointURL(t *testing.T) {
+
```
harshavardhana: Remove spaces, add comments for tests - on what it does.

utils_test.go
```diff
@@ -0,0 +1,368 @@
...
+		for _, valueToEncode := range testCase.valueToEncode {
+			urlValues.Add(testCase.queryKey, valueToEncode)
+		}
+		result := queryEncode(urlValues)
+		if testCase.result != result {
+			t.Errorf("Test %d: Expected queryEncode result to be \"%s\", but found it to be \"%s\" instead", i+1, testCase.result, result)
+		}
+	}
+}
+
```
harshavardhana: Same for all the test functions implemented here.. comments.

utils_test.go
hackintoshrao: I'll add a comment before the test function ?? The detailed commenting style which I had done for Minio server tests is not necessary here since the functionalities are fairly simple.

harshavardhana:
LGTM

PR #372, opened 2016-04-05
Tests: Tests for MakeBucketRequest

hackintoshrao:
- Test for makeBucketRequest
- Test for `MakeBucket` is covered in the functional tests, since `makeBucketRequest` forms the core of `MakeBucket` there is no need for unit test for it. Since tests running on S3 is time consuming its better to just keep them in functional tests.

### DIFF
--- a/api-put-bucket_test.go
+++ b/api-put-bucket_test.go
@@ -1,0 +1,270 @@
+/*
+ * Minio Go Library for Amazon S3 Compatible Cloud Storage (C) 2015, 2016 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


api-put-bucket_test.go
```diff
@@ -0,0 +1,268 @@
...
+	"encoding/base64"
+	"encoding/hex"
+	"encoding/xml"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"testing"
+)
+
+func createExpectedRequest(c *Client, bucketName string, location string, req *http.Request) (*http.Request, error) {
```
harshavardhana: Add a comment here.

api-put-bucket_test.go
```diff
@@ -0,0 +1,268 @@
...
+		{"localhost:9000", "my-access-key", "my-secret-key", false},
+		// one of acess-key and secret-key are empty.
+		{"localhost:9000", "", "my-secret-key", false},
+
+		// endpoint amazon s3.
+		{"s3.amazonaws.com", "", "", false},
+		{"s3.amazonaws.com", "my-access-key", "my-secret-key", false},
+		{"s3.amazonaws.com", "my-acess-key", "", false},
+
+		// endpoint google bucket storage.
```
harshavardhana: google cloud storage not 'bucket storage'

harshavardhana:
Everything else looks good. Just those two comments.

hackintoshrao:
Will push in a minute.

Issue #375, opened 2016-04-08
Panic in signv2


klingtnet:
The call to `sort.Strings(resourceList)` in function `writeCanonicalizedResource` from [`api-signature-v2.go`](https://github.com/minio/minio-go/blob/master/request-signature-v2.go#L274) can panic. Here is the stacktrace:

```
panic: runtime error: index out of range

goroutine 892 [running]:
panic(0x8c4e80, 0xc820010080)
/usr/lib/go/src/runtime/panic.go:464 +0x3e6
sort.(*StringSlice).Less(0xc8221f3b40, 0x0, 0xffffffffffffffff, 0x1)
<autogenerated>:13 +0x114
sort.doPivot(0x7f4c2eaee8a0, 0xc8221f3b40, 0x0, 0x14, 0xc8221f3b40, 0x18)
/usr/lib/go/src/sort/sort.go:128 +0x27b
sort.quickSort(0x7f4c2eaee8a0, 0xc8221f3b40, 0x0, 0x14, 0x9)
/usr/lib/go/src/sort/sort.go:195 +0xa3
sort.Sort(0x7f4c2eaee8a0, 0xc8221f3b40)
/usr/lib/go/src/sort/sort.go:229 +0x74
sort.Strings(0xbeb9e0, 0x14, 0x14)
/usr/lib/go/src/sort/sort.go:305 +0x86
/cephy/vendor/github.com/minio/minio-go.writeCanonicalizedResource(0xc8202c2d20, 0x9587f8, 0x3, 0xc820271100, 0x964de0, 0x8, 0x1, 0x1, 0xc82041fa10, 0x0, ...)
src/cephy/vendor/github.com/minio/minio-go/request-signature-v2.go:274 +0x92
/cephy/vendor/github.com/minio/minio-go.getStringToSignV2(0x9587f8, 0x3, 0xc820271100, 0x964de0, 0x8, 0x1, 0x1, 0xc82041fa10, 0x0, 0x0, ...)
src/cephy/vendor/github.com/minio/minio-go/request-signature-v2.go:187 +0xb8
/cephy/vendor/github.com/minio/minio-go.signV2(0x9587f8, 0x3, 0xc820271100, 0x964de0, 0x8, 0x1, 0x1, 0xc82041fa10, 0x0, 0x0, ...)
src/cephy/vendor/github.com/minio/minio-go/request-signature-v2.go:155 +0x1d6
[...]

PR #376, opened 2016-04-08
Do not sort requestList in each call and add an unit test to check if it is already correctly sorted

vadmeste:
An optimization which corrects this bug https://github.com/minio/minio-go/issues/375 (concurrent routines trying to sort resourceList)

### DIFF
--- a/request-signature-v2.go
+++ b/request-signature-v2.go
@@ -235,7 +235,8 @@ func writeCanonicalizedHeaders(buf *bytes.Buffer, req http.Request) {
 	}
 }
 
-// Must be sorted:
+// The following list is already sorted and should always be, otherwise we could
+// have signature-related issues
 var resourceList = []string{
 	"acl",
 	"location",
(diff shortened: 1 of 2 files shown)


request-signature-v2_test.go
```diff
@@ -0,0 +1,35 @@
+/*
+ * Minio Go Library for Amazon S3 Compatible Cloud Storage (C) 2015 Minio, Inc.
```
harshavardhana: Add '2016'

request-signature-v2.go
```diff
@@ -243,13 +243,13 @@ var resourceList = []string{
...
 	"response-cache-control",
 	"response-content-disposition",
 	"response-content-encoding",
-	"requestPayment",
+	"response-content-language",
+	"response-content-type",
+	"response-expires",
 	"torrent",
 	"uploadId",
 	"uploads",
```
harshavardhana: Mention that this list is sorted already and mention that new fields added here should make sure to add at lexical location to this list. Otherwise can have signature related issues later.

PR #383, opened 2016-04-19
Tests: Api error response tests added

hackintoshrao:


api-error-response_test.go
```diff
@@ -0,0 +1,279 @@
...
+func TestHttpRespToErrorResponse(t *testing.T) {
+	// 'formulateAPIErrorResponse' formulates ErrorResponse for given APIError.
+	// provides a encodable populated response values.
+	formulateAPIErrorResponse := func(err APIError, bucketName string) ErrorResponse {
+		var data = ErrorResponse{}
+		data.Code = err.Code
+		data.Message = err.Description
+
+		data.BucketName = bucketName
+		// TODO implement this in future
```
harshavardhana: what is this to be implemented in future?

api-error-response_test.go
```diff
@@ -0,0 +1,279 @@
...
+	"strconv"
+	"testing"
+)
+
+// Tests validate the Error generator function for http response with error.
+func TestHttpRespToErrorResponse(t *testing.T) {
+	// 'formulateAPIErrorResponse' formulates ErrorResponse for given APIError.
+	// provides a encodable populated response values.
+	formulateAPIErrorResponse := func(err APIError, bucketName string) ErrorResponse {
+		var data = ErrorResponse{}
```
harshavardhana: don't call it 'var data' call it 'var errResp'

api-error-response_test.go
```diff
@@ -0,0 +1,279 @@
...
+	encodeErr := func(response interface{}) []byte {
+		var bytesBuffer bytes.Buffer
+		bytesBuffer.WriteString(xml.Header)
+		encode := xml.NewEncoder(&bytesBuffer)
+		encode.Encode(response)
+		return bytesBuffer.Bytes()
+	}
+
+	// `formulteAPIErrorResponse` Mocks XML error response from the server.
+	formulteAPIErrorResponse := func(APIErr APIError, bucketName string) *http.Response {
```
harshavardhana: Looks like you have a name conflict here,  you avoided that by introducing a typo. This would be incorrect call this 

'genHTTPErrorResponse' and 'genAPIErrorResponse' - rather than using words like 'formulate' a better word would be just 'create' or 'gen',  pick easier names.

api-error-response_test.go
```diff
@@ -0,0 +1,279 @@
...
+			Key:        objectName,
+			RequestID:  resp.Header.Get("x-amz-request-id"),
+			HostID:     resp.Header.Get("x-amz-id-2"),
+			Region:     resp.Header.Get("x-amz-bucket-region"),
+		}
+		return errResp
+	}
+
+	// Generate invalid argument error.
+	formualteInvalidError := func(message string) error {
```
harshavardhana: you can just use existing 'ErrInvalidArgument' , why need to repeat this?

api-error-response_test.go
hackintoshrao: Was just a task list... Missed out. Removing.

api-error-response_test.go
hackintoshrao: It'll be good if tests are as self contained as possible. This will enable detection bugs due to changes in the source. Yes, there is a separate functional tests for `ErrInvalidArgument` too, but this too is not a bad way to test is what I believe, it'll act as one additional layer of test.

PR #387, opened 2016-04-22
list: Fix listObjects to avoid large bufferred channels to facilitate…

### LINKED ISSUE #386: Endless waiting in ListObjects() with doneCh
I did some simple test for ListObjects() and got endless waiting when I attempt send value into doneCh. Note: backet contains more then 14 objects (15).
```
func testListBackets() {
	doneCh := make(chan struct{})
	defer close(doneCh)

	n := 0
	objectCh := s3cli.ListObjects("backet", "", true, doneCh)
	for object := range objectCh {
		if object.Err != nil {
			fmt.Println(object.Err)
			return
		}
		n++
		fmt.Println(n, object)
		if n == 14 {
			doneCh <- struct{}{}
		}
		// waiting here ........................................................
		fmt.Println(n) // Println does not print 14
	}
}
```

harshavardhana:
… doneCh.

If we send 1000 bufferred entries simultaneously in bufferred channel, the
top level caller while using doneCh to limit the entries will have no way
to communicate back on how much a ListObjects should be listing the entries
for.

The usual scenario is when one wishes to only get N entires from ListObjects,
It is better for us to send 1 entry at a time so that as and when the
doneCh is called we exit the loop without accumulating more entries than
was needed.

Following program talks about how this issue can happen.
```
func testListBackets() {
    doneCh := make(chan struct{}, 1)
    defer close(doneCh)

    n := 0
    objectCh := s3cli.ListObjects("backet", "", true, doneCh)
    for object := range objectCh {
        if object.Err != nil {
            fmt.Println(object.Err)
            return
        }
        n++
        fmt.Println(n, object)
        if n == 14 {
            doneCh <- struct{}{}
        }
        // waiting here ........................................................
        fmt.Println(n) // Println does not print 14
    }
}
```

Here the program ends up printing 15th object as well even after doneCh indicating
the routine to exit.

Fixes #386

### DIFF
--- a/api-list.go
+++ b/api-list.go
@@ -77,7 +77,7 @@ func (c Client) ListBuckets() ([]BucketInfo, error) {
 //
 func (c Client) ListObjects(bucketName, objectPrefix string, recursive bool, doneCh <-chan struct{}) <-chan ObjectInfo {
 	// Allocate new list objects channel.
-	objectStatCh := make(chan ObjectInfo, 1000)
+	objectStatCh := make(chan ObjectInfo)
 	// Default listing is delimited at "/"
 	delimiter := "/"
 	if recursive {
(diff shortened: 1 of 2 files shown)


api-list.go
```diff
@@ -77,7 +77,7 @@ func (c Client) ListBuckets() ([]BucketInfo, error) {
 //
 func (c Client) ListObjects(bucketName, objectPrefix string, recursive bool, doneCh <-chan struct{}) <-chan ObjectInfo {
 	// Allocate new list objects channel.
-	objectStatCh := make(chan ObjectInfo, 1000)
+	objectStatCh := make(chan ObjectInfo, 1)
```
krishnasrinivas: If this is a buffer channel of 1 entry then the application for loop will read one extra entry unless it does an explicit break (but if we have to mandate a break then this chan can be buffered for 1000 entries too). It should be unbuffered channel
objectStatCh := make(chan ObjectInfo)

PR #390, opened 2016-04-25
retry: Retry api requests for two more cases.

harshavardhana:
- TLS handshake timeout.
- TCP dial timeout.

Fixes - https://github.com/minio/mc/issues/1693

### DIFF
--- a/retry.go
+++ b/retry.go
@@ -73,13 +73,24 @@ func (c Client) newRetryTimer(maxRetry int, unit time.Duration, cap time.Duratio
 // isNetErrorRetryable - is network error retryable.
 func isNetErrorRetryable(err error) bool {
 	switch err.(type) {
-	case *net.DNSError, *net.OpError, net.UnknownNetworkError:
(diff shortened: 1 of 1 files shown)


retry.go
```diff
@@ -82,6 +85,15 @@ func isNetErrorRetryable(err error) bool {
...
+	if _, ok := err.(net.Error); ok {
+		if strings.Contains(err.Error(), "net/http: TLS handshake timeout") {
+			// If error is - tlsHandshakeTimeoutError, retry.
+			return true
+		} else if strings.Contains(err.Error(), "i/o timeout") {
+			// If error is - tcp timeoutError, retry.
+			return true
+		}
+	}
 	return false
```
abperiasamy: All of these can be combined into switch case.

harshavardhana:
> All of these can be combined into switch case.

It cannot be, one is a type these are strings - they cannot be made types - internal types are not exposed.

PR #393, opened 2016-05-10
api: Change PresignedPostPolicy to return back proper URL

harshavardhana:
The reason to do this is because bucket names can be in
various different regions and for users it becomes convenient
if we return back the proper endpoint which can be safely used
without worrying about virtual host style or path style.

We have enough information to provide a proper URL, this is also
consistent with other Presigned operations as it would be an
expected behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,9 +85,9 @@ func main() {
 * [FGetObject(bucketName, objectName, filePath) error](examples/s3/fgetobject.go)
 
 ### Presigned Operations.
-* [PresignedGetObject(bucketName, objectName, time.Duration, url.Values) (string, error)](examples/s3/presignedgetobject.go)
-* [PresignedPutObject(bucketName, objectName, time.Duration) (string, error)](examples/s3/presignedputobject.go)
-* [PresignedPostPolicy(NewPostPolicy()) (map[string]string, error)](examples/s3/presignedpostpolicy.go)
+* [PresignedGetObject(bucketName, objectName, time.Duration, url.Values) (*url.URL, error)](examples/s3/presignedgetobject.go)
+* [PresignedPutObject(bucketName, objectName, time.Duration) (*url.URL, error)](examples/s3/presignedputobject.go)
+* [PresignedPostPolicy(NewPostPolicy()) (*url.URL, map[string]string, error)](examples/s3/presignedpostpolicy.go)
 
 ### Bucket Policy Operations.
 * [SetBucketPolicy(bucketName, objectPrefix, BucketPolicy) error](examples/s3/setbucketpolicy.go)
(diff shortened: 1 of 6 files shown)


krishnasrinivas:
this is needed for other libs too.

harshavardhana:
> this is needed for other libs too.

Yes that was the question i asked you for javascript.

api-presigned.go
```diff
@@ -84,36 +84,42 @@ func (c Client) presignURL(method string, bucketName string, objectName string,
...
-func (c Client) PresignedPutObject(bucketName string, objectName string, expires time.Duration) (url string, err error) {
+func (c Client) PresignedPutObject(bucketName string, objectName string, expires time.Duration) (urlStr string, err error) {
 	return c.presignURL("PUT", bucketName, objectName, expires, nil)
 }
 
-// PresignedPostPolicy - Returns POST form data to upload an object at a location.
-func (c Client) PresignedPostPolicy(p *PostPolicy) (map[string]string, error) {
+// PresignedPostPolicy - Returns POST urlString, form data to upload an object.
+func (c Client) PresignedPostPolicy(p *PostPolicy) (urlStr string, formData map[string]string, err error) {
 	// Validate input arguments.
```
abperiasamy: return URL as a native type.

harshavardhana:
> return URL as a native type.

There is no reason to return as `*url.URL` - it shouldn't be a mutable data type, it is easier to return as immutable string.  There is no reason for users to further modify this value.

abperiasamy:
Returning as a native typed object has nothing to do with mutability. Native URL type is the right type here.

krishnasrinivas:
returning a string here is more intuitive, as this string needs to be passed to the client. Returning native URL will need more explanation on how it should be converted to a string and passed to the client.

If we return native URL here we should do the same for presignedGetObject and presignedPutObject - that will be unintuitive for users.

krishnasrinivas:
for presignedGET/presignedPUT [aws-sdk-go apis](http://docs.aws.amazon.com/sdk-for-go/latest/v1/developerguide/common-examples.title.html) return string as well.

abperiasamy:
@krishnasrinivas AWS SDK APIs are poorly designed.

README.md
```diff
@@ -87,7 +87,7 @@ func main() {
 ### Presigned Operations.
 * [PresignedGetObject(bucketName, objectName, time.Duration, url.Values) (string, error)](examples/s3/presignedgetobject.go)
 * [PresignedPutObject(bucketName, objectName, time.Duration) (string, error)](examples/s3/presignedputobject.go)
-* [PresignedPostPolicy(NewPostPolicy()) (map[string]string, error)](examples/s3/presignedpostpolicy.go)
+* [PresignedPostPolicy(NewPostPolicy()) (string, map[string]string, error)](examples/s3/presignedpostpolicy.go)
```
abperiasamy: should be *url.URL

PR #398, opened 2016-05-15
api: Do not add 100-continue for Google Cloud Storage.

harshavardhana:
There are some known problems with this, avoid setting
'100-continue' if the endpoint is Google Cloud Storage.

Ref : https://github.com/restic/restic/issues/520

### DIFF
--- a/api.go
+++ b/api.go
@@ -568,10 +568,15 @@ func (c Client) newRequest(method string, metadata requestMetadata) (req *http.R
 		req.Body = ioutil.NopCloser(metadata.contentBody)
 	}
 
-	// set 'Expect' header for the request.
-	req.Header.Set("Expect", "100-continue")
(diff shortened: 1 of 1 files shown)


api.go
```diff
@@ -568,10 +568,14 @@ func (c Client) newRequest(method string, metadata requestMetadata) (req *http.R
...
 	}
 
-	// set 'Expect' header for the request.
-	req.Header.Set("Expect", "100-continue")
+	// Skip setting 'expect' header for Google Cloud Storage, there
+	// are some known issues - https://github.com/restic/restic/issues/520
+	if !isGoogleEndpoint(c.endpointURL) {
+		// Set 'Expect' header for the request.
+		req.Header.Set("Expect", "100-continue")
+	}
```
abperiasamy: Add a FIXME: here. We can keep reviewing all FIXMEs once a while.

PR #400, opened 2016-05-18
api: getReaderSize - Supports Len() function

harshavardhana:
This is needed for certain packages which implement Len() instead
of Size() example *bytes.Buffer, *strings.Reader etc.

### DIFF
--- a/api-put-object.go
+++ b/api-put-object.go
@@ -27,78 +27,94 @@ import (
 	"strings"
 )
 
+// toInt - converts go value to its integer representation based
+// on the value kind if it is an integer.
+func toInt(value reflect.Value) (size int64) {
+	size = -1
+	if value.IsValid() {
+		switch value.Kind() {
+		case reflect.Int:
+			fallthrough
+		case reflect.Int8:
+			fallthrough
+		case reflect.Int16:
+			fallthrough
+		case reflect.Int32:
+			fallthrough
+		case reflect.Int64:
+			size = value.Int()
+		}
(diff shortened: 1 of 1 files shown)


krishnasrinivas:
had hit this requirement just a couple of hours ago when I was trying to PutObject a `bytes.NewBuffer([]byte) `stream

api-put-object.go
```diff
@@ -27,78 +27,94 @@ import (
 	"strings"
 )
 
+// toInt - converts go value to its integer representation based
+// on the value kind if it is an integer.
+func toInt(value reflect.Value) (size int64) {
```
abperiasamy: You can directly do int() on reflect.ValueOf and let the compiler do the type validation.

harshavardhana:
> You can directly do int() on reflect.ValueOf and let the compiler do the type validation.

That is not possible since reflect.Value is a proper type - this is what you get. 

https://play.golang.org/p/SeGQG1u8V-
```
prog.go:23: cannot convert result[0] (type reflect.Value) to type int
```

You have to know the Kind() to figure out and call the appropriate conversion function like value.Int().

abperiasamy:
In that case why not use reflect.Value.Int()?
https://play.golang.org/p/bZPNIGLi8X

harshavardhana:
> In that case why not use reflect.Value.Int()?
> https://play.golang.org/p/bZPNIGLi8X

In that case if the Value is not Int, it would panic.

PR #404, opened 2016-05-20
Fixed part size

longnguyen11288:
With maxMultipartPutObjectSize set to 5TiB and dividing by maxPartsCount of 10000, you get 525mb chunks which... doesn't work for some providers who support s3. 

I believe intent was 5MiB chucks

### DIFF
--- a/api-put-object-common.go
+++ b/api-put-object-common.go
@@ -68,7 +68,7 @@ func shouldUploadPart(objPart objectPart, objectParts map[int]objectPart) bool {
 // NOTE: Assumption here is that for any object to be uploaded to any S3 compatible
 // object storage it will have the following parameters as constants.
 //
-//  maxPartsCount - 10000
+//  maxPartsCount - 1000000
 //  minPartSize - 5MiB
 //  maxMultipartPutObjectSize - 5TiB
 //
--- a/api_unit_test.go
+++ b/api_unit_test.go
@@ -341,14 +341,14 @@ func TestPartSize(t *testing.T) {
 	if err != nil {
 		t.Fatal("Error: ", err)
 	}
-	if totalPartsCount != 9987 {
-		t.Fatalf("Error: expecting total parts count of 9987: got %v instead", totalPartsCount)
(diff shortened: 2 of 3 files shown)


krishnasrinivas:
@longnguyen11288 max parts allowed is 10000 as mentioned here http://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadUploadPart.html

Which provider who support S3 does not allow 525MB part size?

longnguyen11288:
Scality. How should I work around this?

krishnasrinivas:
How big is the file that you are uploading?
Are you uploading a file that is from local file system? or are you uploading from a stream directly to scality?

thomasmmitchell:
It is from a stream out of bzip. The input file size to bzip is is arbitrary, but a particular example we've been using is around 3 gigs. But because its a input stream, `func optimalPartInfo()` is defaulting the file size to 5 TiB... and therefore is using maximum size parts, which scality doesn't like.

longnguyen11288:
Stream and file is 1.5gb

krishnasrinivas:
Since minio-go is not able to determine the input stream size it uploads in part size of 525MB which scality is not supporting. As a work around you can write the stream to a local file system first and then upload this file, that should make minio-go use smaller part size (5MB).

harshavardhana:
> With maxMultipartPutObjectSize set to 5TiB and dividing by maxPartsCount of 10000, you get 525mb chunks which... doesn't work for some providers who support s3.
> 
> I believe intent was 5MiB chucks

Thanks for the patch. 

After looking at the patch this is in violation of S3 protocol. Changing this breaks S3 compatibility. 525MB is really small compared to maximum allowed upto 5GB.  The real fix should happen at the server side to support proper S3 semantics. 

Here are some quick facts about multipart limits which minio-go uses these values for http://docs.aws.amazon.com/AmazonS3/latest/dev/qfacts.html

harshavardhana:
Closing as ***WorkingAsIntended***.

Issue #406, opened 2016-05-23
Access Policy: Policy not removed for multiple policies in same bucket


mkash32:
```
$ mc access readonly myminio/mybucket/dir
Access permission for ‘myminio/mybucket/dir’ is set to 'readonly'
$ mc access readonly myminio/mybucket/folder
Access permission for 'myminio/mybucket/folder' is set to 'readonly'
$ mc access none myminio/mybucket/dir
Access permission for ‘myminio/mybucket/dir’ is set to 'none'
$ mc access none myminio/mybucket/folder
Access permission for ‘myminio/mybucket/folder’ is set to 'none'
```
Although it is reported that `myminio/mybucket/folder` is set to none (in the last line), the previous access policy of 'readonly' is still set (and it can still be accessed by anonymous user). This bug only occurs when we set the access policy of the prefixes to the same setting.

PR #407, opened 2016-05-25
Access Policy: Policy Removal Fix

mkash32:
Fixes: https://github.com/minio/minio-go/issues/406 (Bug in access policy removal)

* Added boolean in the remove policy functions to prevent the function from removing more than one Bucket Resource Statement.
* Test cases added to bucket-policy unit tests


hackintoshrao:
@harshavardhana : This looks good... The input Statement for which it was failing is added to the test case.

Issue #409, opened 2016-05-26
Access Policy: Setting policy of subdirectory in bucket 


mkash32:
Refer to https://github.com/minio/minio/issues/1705

harshavardhana:
Is this fixed @mkash32 @hackintoshrao ?

mkash32:
@harshavardhana Working on it, will send the pr today.

PR #410, opened 2016-05-26
Access Policy: Resource Matching

mkash32:
Fixes: https://github.com/minio/minio-go/issues/409

* Included the resource match function (from minio server). Literal string equality -> matching resource pattern using resourceMatch function.
* Unit tests added for resourceMatch (taken from minio server) and isPolicyDefinedForObjectPrefix (newly added function)
* Test cases associated with matching and the changes added.

### DIFF
--- a/bucket-policy.go
+++ b/bucket-policy.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"sort"
+	"strings"
 )
 
 // maximum supported access policy size.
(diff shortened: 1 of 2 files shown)


mkash32:
@hackintoshrao Can you take a look?

hackintoshrao:
@harshavardhana :  The `ResoureMatch` function from the server side is now used to interpret the policy statement coming from server.  
LGTM.

PR #414, opened 2016-06-07
Access Policy: Added Pseudo None Statements

mkash32:


harshavardhana:
This is not correct we shouldn't add new policy statements to satisfy `none` - `none` means there is no policy. Also what is the issue that this patch fixes?. 

Additionally this also breaks s3 compatibility - how will this work with s3?

mkash32:
closed in favor of #415

PR #419, opened 2016-06-14
CompleteMultipartUpload API may contain error response in resp.Body

### LINKED ISSUE #413: multipart: complete multipart upload should process white space characters after 200 OK
http://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadComplete.html:
`
Processing of a Complete Multipart Upload request could take several minutes to complete. After Amazon S3 begins processing the request, it sends an HTTP response header that specifies a 200 OK response. While processing is in progress, Amazon S3 periodically sends whitespace characters to keep the connection from timing out. Because a request could fail after the initial 200 OK response has been sent, it is important that you check the response body to determine whether the request succeeded.
`

Not sure how server indicates a failure after 200 OK with white space characters, needs to be figured out.

krisis:
ref: http://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadComplete.html

Closes #413

### DIFF
--- a/api-put-object-multipart.go
+++ b/api-put-object-multipart.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"encoding/hex"
 	"encoding/xml"
+	"errors"
 	"io"
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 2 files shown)


api-put-object-multipart.go
```diff
@@ -351,11 +352,31 @@ func (c Client) completeMultipartUpload(bucketName, objectName, uploadID string,
...
+		// xml parsing failure due to presence an ill-formed xml fragment
 		return completeMultipartUploadResult, err
+	} else if completeMultipartUploadResult.Bucket == "" {
+		// xml's Decode method ignores well-formed xml that don't apply to the type of value supplied.
+		// In this case, it would leave completeMultipartUploadResult with the corresponding zero-values
+		// of the members.
+
+		// Decode completed multipart upload response on failure
+		completeMultipartUploadErr := completeMultipartUploadErr{}
+		err = xmlDecoder(bytes.NewReader(b), &completeMultipartUploadErr)
```
harshavardhana: why not use ErrorResponse struct here. Also provides 'Error()' compatible data type. so that you can avoid doing 'errors.New()'

PR #424, opened 2016-06-22
api: Skip object when on Glacier.

harshavardhana:


harshavardhana:
Closing preferred method is to use 'mc' and ignore.

PR #425, opened 2016-06-23
Added File Object Operations section in examples & API also modified …

koolhead17:
…the divied the example content by introducing Initialize Minio Client


koolhead17:
@harshavardhana modified the contents as per @GarimaKapoor suggestions. :)

Issue #429, opened 2016-07-03
Misleading error msg when setting a policy twice on a specific bucket


vadmeste:
```
$ mc ls myminio/testbucket/                                                         
[2016-07-03 14:25:40 CET] 5.5GiB 5.5gb.bin
$ mc access readonly myminio/testbucket/                                            
mc: <ERROR> Unable to set access permission ‘readonly’ for ‘myminio/testbucket/’. No policy exists on testbucket/
```

PR #430, opened 2016-07-03
Access Policy: Fixed misleading error response when same policy set twice

mkash32:
Fixes: https://github.com/minio/minio-go/issues/429

* When same policy is set twice on same prefix, return directly with nil error
 response(without sending a request to server).

### DIFF
--- a/api-put-bucket.go
+++ b/api-put-bucket.go
@@ -176,10 +176,19 @@ func (c Client) SetBucketPolicy(bucketName string, objectPrefix string, bucketPo
 	}
 	statements = append(statements, generatedStatements...)
 
-	// No change in the statements indicates an attempt of setting 'none' on a prefix
-	// which doesn't have a pre-existing policy.
+	// No change in the statements indicates either an attempt of setting 'none'
+	// on a prefix which doesn't have a pre-existing policy, or setting a policy
+	// on a prefix which already has the same policy.
 	if reflect.DeepEqual(policy.Statements, statements) {
-		return ErrNoSuchBucketPolicy(fmt.Sprintf("No policy exists on %s/%s", bucketName, objectPrefix))
+		// If policy being set is 'none' return an error, otherwise return nil to
+		// prevent the unnecessary request from being sent
+		var err error
+		if bucketPolicy == BucketPolicyNone {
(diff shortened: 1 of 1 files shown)


api-put-bucket.go
```diff
@@ -176,10 +176,17 @@ func (c Client) SetBucketPolicy(bucketName string, objectPrefix string, bucketPo
...
+	// on a prefix which doesn't have a pre-existing policy, or setting a policy
+	// on a prefix which already has the same policy.
 	if reflect.DeepEqual(policy.Statements, statements) {
-		return ErrNoSuchBucketPolicy(fmt.Sprintf("No policy exists on %s/%s", bucketName, objectPrefix))
+		// If policy being set is 'none' return an error, otherwise return nil to
+		// prevent the unnecessary request from being sent
+		if bucketPolicy == BucketPolicyNone {
+			return ErrNoSuchBucketPolicy(fmt.Sprintf("No policy exists on %s/%s", bucketName, objectPrefix))
+		} else {
+			return nil
```
mkash32: @harshavardhana Should an error be returned in this case, or is it fine to just return nil directly to bypass the request to the server?

harshavardhana: > @harshavardhana Should an error be returned in this case, or is it fine to just return nil directly to bypass the request to the server?

We need to only send error if its an error, if we do know that its not an error then we are fine. 

```
var err error
if bucketPolicy == BucketPolicyNone {
        err = ErrNoSuchBucketPolicy(fmt.Sprintf("No policy exists on %s/%s", bucketName, objectPrefix))
} else {
        err = nil
}
return err
```

harshavardhana: > We need to only send error if its an error, if we do know that its not an error then we are fine.
> 
> var err error
> if bucketPolicy == BucketPolicyNone {
>         err = ErrNoSuchBucketPolicy(fmt.Sprintf("No policy exists on %s/%s", bucketName, objectPrefix))
> } else {
>         err = nil
> }
> return err

We should add a comment as well explaining why we are 'returning 'nil'

harshavardhana:
Can you vendorize this for mc @mkash32 as well ?. Thanks

mkash32:
Sure, I found another small issue as well, will post the issue and send a pr for that as well and vendorize for both changes. Will get it done by tonight.

PR #436, opened 2016-07-06
get-object: Encapsulate closer into safe closer.

harshavardhana:


utils.go
```diff
@@ -142,16 +142,26 @@ func closeResponse(resp *http.Response) {
...
-		resp.Body.Close()
+		closeReader(resp.Body)
 	}
 }
 
+// closeReader - closes the readCloser safely based on its availability,
+// also drains the reader for any remaining data.
+func closeReader(reader io.ReadCloser) error {
+	if reader != nil {
+		io.Copy(ioutil.Discard, reader)
```
krishnasrinivas: i think this will be expensive if we do Seek, i.e we are reading the entire stream. for mc cloud->cloud copy, we seek to skip uploaded parts so that we don't have to download already copied parts. So purpose of Seek will be defeated. A connection not returned to pool will be better in comparison with reading the entire stream

but there must be some way to fix both situations.

harshavardhana: > i think this will be expensive if we do Seek, i.e we are reading the entire stream. for mc cloud->cloud copy, we seek to skip uploaded parts so that we don't have to download already copied parts. So purpose of Seek will be defeated. A connection not returned to pool will be better in comparison with reading the entire stream

Actually we need to validate if io.Copy() is still needed.

harshavardhana:
Will validate the old theory if it holds good.

harshavardhana:
Closing this deprecated.

PR #437, opened 2016-07-06
client: endpointURL should be saved as string and parsed as needed.

harshavardhana:
This fix is needed to avoid the problems of passing down mutating pointers.

### DIFF
--- a/bucket-cache.go
+++ b/bucket-cache.go
@@ -147,7 +147,10 @@ func (c Client) getBucketLocationRequest(bucketName string) (*http.Request, erro
 	urlValues.Set("location", "")
 
 	// Set get bucket location always as path style.
-	targetURL := c.endpointURL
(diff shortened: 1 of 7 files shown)


bucket-cache.go
```diff
@@ -147,7 +147,7 @@ func (c Client) getBucketLocationRequest(bucketName string) (*http.Request, erro
 	urlValues.Set("location", "")
 
 	// Set get bucket location always as path style.
-	targetURL := c.endpointURL
+	targetURL := *c.endpointURL
```
krishnasrinivas: Client.endpointURL can itself hold a value instead of pointer right? There is no use of it being a pointer

harshavardhana: > Client.endpointURL can itself hold a value instead of pointer right? There is no use of it being a pointer
> It can avoid bugs like these

We need to re-do this and use string itself.

PR #443, opened 2016-07-15
Do not compute payload sha256 with https endpoint

### LINKED ISSUE #428: Disable sha256 computation for https

vadmeste:
Fixes #428 

It would be better to make the layer responsible for putting objects chooses what hash algorithms should be computed

### DIFF
--- a/api-put-object-file.go
+++ b/api-put-object-file.go
@@ -17,8 +17,11 @@
 package minio
 
 import (
+	"crypto/md5"
+	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
+	"hash"
 	"io"
 	"io/ioutil"
 	"mime"
@@ -176,10 +179,17 @@ func (c Client) putObjectMultipartFromFile(bucketName, objectName string, fileRe
 		// Get a section reader on a particular offset.
 		sectionReader := io.NewSectionReader(fileReader, totalUploadedSize, partSize)
 
-		// Calculates MD5 and SHA256 sum for a section reader.
-		var md5Sum, sha256Sum []byte
+		// Add hash algorithms that need to be calculated by computeHash()
+		// In case of a non-v4 signature or https connection, sha256 is not needed.
+		hashAlgos := make(map[string]hash.Hash)
+		hashSums := make(map[string][]byte)
+		hashAlgos["md5"] = md5.New()
+		if c.signature.isV4() && !c.secure {
+			hashAlgos["sha256"] = sha256.New()
+		}
+
 		var prtSize int64
-		md5Sum, sha256Sum, prtSize, err = c.computeHash(sectionReader)
+		prtSize, err = computeHash(hashAlgos, hashSums, sectionReader)
 		if err != nil {
 			return 0, err
 		}
@@ -191,14 +201,14 @@ func (c Client) putObjectMultipartFromFile(bucketName, objectName string, fileRe
 
 		// Verify if part should be uploaded.
 		if shouldUploadPart(objectPart{
-			ETag:       hex.EncodeToString(md5Sum),
+			ETag:       hex.EncodeToString(hashSums["md5"]),
 			PartNumber: partNumber,
 			Size:       prtSize,
 		}, partsInfo) {
 			// Proceed to upload the part.
 			var objPart objectPart
 			objPart, err = c.uploadPart(bucketName, objectName, uploadID, reader, partNumber,
-				md5Sum, sha256Sum, prtSize)
+				hashSums["md5"], hashSums["sha256"], prtSize)
 			if err != nil {
 				return totalUploadedSize, err
 			}
--- a/api-put-object-multipart.go
+++ b/api-put-object-multipart.go
@@ -18,8 +18,11 @@ package minio
 
 import (
 	"bytes"
+	"crypto/md5"
+	"crypto/sha256"
 	"encoding/hex"
 	"encoding/xml"
+	"hash"
 	"io"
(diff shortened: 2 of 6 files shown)


api-put-object-multipart.go
```diff
@@ -114,7 +117,15 @@ func (c Client) putObjectMultipartStream(bucketName, objectName string, reader i
...
-		md5Sum, sha256Sum, prtSize, rErr := c.hashCopyN(tmpBuffer, reader, partSize)
+		hashSums := make(map[string][]byte)
+		hashAlgos := make(map[string]hash.Hash)
+
+		hashAlgos["md5"] = md5.New()
+		if c.signature.isV4() && !c.secure {
+			hashAlgos["sha256"] = sha256.New()
+		}
+
+		prtSize, rErr := hashCopyN(hashAlgos, hashSums, tmpBuffer, reader, partSize)
```
krishnasrinivas: with this approach, hashCopyN and related functions (hashCopy hashCopyBuffer) can be eliminated by doing:

```
md5Writer := md5.New()
writers := []io.Writers{tmpBuffer, md5Writer}

var sha256Writer hash.Hash

if c.signature.isV4() && !c.secure {
    sha256Writer = sha256.New()
    writers = append(writers, sha256Writer)
}

wn, err := io.CopyN(io.MultiWriter(writers...), reader, partSize)
```

api.go
```diff
@@ -583,9 +588,13 @@ func (c Client) newRequest(method string, metadata requestMetadata) (req *http.R
...
-			req.Header.Set("X-Amz-Content-Sha256", hex.EncodeToString(sum256([]byte{})))
-			if metadata.contentSHA256Bytes != nil {
-				req.Header.Set("X-Amz-Content-Sha256", hex.EncodeToString(metadata.contentSHA256Bytes))
+			if !c.secure {
+				req.Header.Set("X-Amz-Content-Sha256", hex.EncodeToString(sum256([]byte{})))
+				if metadata.contentSHA256Bytes != nil {
+					req.Header.Set("X-Amz-Content-Sha256", hex.EncodeToString(metadata.contentSHA256Bytes))
+				}
+			} else {
+				req.Header.Set("X-Amz-Content-Sha256", "UNSIGNED-PAYLOAD")
```
krishnasrinivas: better flow would be:
```
shaHeader := "UNSIGNED-PAYLOAD"
if !c.secure {
    if metadata.contentSHA256Bytes == nil {
        shaHeader = hex.EncodeToString(sum256([]byte{})))
    } else {
        shaHeader = hex.EncodeToString(metadata.contentSHA256Bytes)
    }
}
req.Header.Set("X-Amz-Content-Sha256", shaHeader)
```

vadmeste: :+1:

api-put-object-multipart.go
vadmeste: > with this approach, hashCopyN and related functions (hashCopy hashCopyBuffer) can be eliminated

@krishnasrinivas, actually, hashCopyN is called from three places, I guess we have to keep it factored

krishnasrinivas: >> with this approach, hashCopyN and related functions (hashCopy hashCopyBuffer) can be eliminated

> @krishnasrinivas, actually, hashCopyN is called from three places, I guess we have to keep it factored

similar change has to be done at other places also, our own implementation of hash* functions are not providing advantage anymore. So why not use golang io package's utility functions io.Copy io.CopyN io.CopyBuffer.

api-put-object-file.go
```diff
@@ -177,9 +180,16 @@ func (c Client) putObjectMultipartFromFile(bucketName, objectName string, fileRe
 		sectionReader := io.NewSectionReader(fileReader, totalUploadedSize, partSize)
 
 		// Calculates MD5 and SHA256 sum for a section reader.
-		var md5Sum, sha256Sum []byte
+		hashAlgos := make(map[string]hash.Hash)
+		hashSums := make(map[string][]byte)
+
+		hashAlgos["md5"] = md5.New()
```
harshavardhana: Needs a comment here, why this is needed.

api-put-object-multipart.go
```diff
@@ -114,7 +117,15 @@ func (c Client) putObjectMultipartStream(bucketName, objectName string, reader i
 	for partNumber <= totalPartsCount {
 		// Calculates MD5 and SHA256 sum while copying partSize bytes
 		// into tmpBuffer.
-		md5Sum, sha256Sum, prtSize, rErr := c.hashCopyN(tmpBuffer, reader, partSize)
+		hashSums := make(map[string][]byte)
+		hashAlgos := make(map[string]hash.Hash)
+
+		hashAlgos["md5"] = md5.New()
```
harshavardhana: Needs a comment as well.

api-put-object.go
```diff
@@ -193,11 +196,19 @@ func (c Client) putObjectSingle(bucketName, objectName string, reader io.Reader,
 	if size <= -1 {
 		size = maxSinglePutObjectSize
 	}
-	var md5Sum, sha256Sum []byte
+	// var md5Sum, sha256Sum []byte
```
harshavardhana: This comment should be changed and fixed to reflect the following changes..

api.go
```diff
@@ -163,6 +166,8 @@ func privateNew(endpoint, accessKeyID, secretAccessKey string, secure bool) (*Cl
 		clnt.anonymous = true
 	}
 
+	clnt.secure = secure
```
harshavardhana: Comment?

PR #447, opened 2016-07-17
Add ListObjectsV2 API

vadmeste:


api-list.go
```diff
@@ -53,6 +53,179 @@ func (c Client) ListBuckets() ([]BucketInfo, error) {
...
+
+// listObjectsV2Query - (List Objects V2) - List some or all (up to 1000) of the objects in a bucket.
+//
+// You can use the request parameters as selection criteria to return a subset of the objects in a bucket.
+// request parameters :-
+// ---------
+// ?continuation-token - Specifies the key to start with when listing objects in a bucket.
+// ?delimiter - A delimiter is a character you use to group keys.
+// ?prefix - Limits the response to keys that begin with the specified prefix.
+// ?max-keys - Sets the maximum number of keys returned in the response body.
```
harshavardhana: what about "start-after" query param? 

http://docs.aws.amazon.com/AmazonS3/latest/API/v2-RESTBucketGET.html

````
	
If you want the API to return key names after a specific object key in your key space, you can add this parameter. Amazon S3 lists objects in UTF-8 character encoding in lexicographical order.

This parameter is valid only in your first request. In case the response is truncated, you can specify this parameter along with the continuation-token parameter, and then Amazon S3 will ignore this parameter.
```

api-s3-datatypes.go
```diff
@@ -41,6 +41,33 @@ type commonPrefix struct {
...
+	IsTruncated bool
+	MaxKeys     int64
+	Name        string
+
+	// Hold the token that will be sent in the next request to fetch the next group of keys
+	NextContinuationToken string
+
+	ContinuationToken string
+	FetchOwner        string
+	StartAfter        string
```
harshavardhana: Start after doesn't seem to be used.

api-list.go
harshavardhana: If its not needed, we need to make a mention of it in the code.

vadmeste: Currently, it is not needed. Only if we add a new parameter to ListObjectsV2 API which could be used by developers

vadmeste: But I guess we want to keep the API simple

harshavardhana:
> But I guess we want to keep the API simple

Let's mention why its not implemented.

Issue #451, opened 2016-07-21
Implement functional test for PresignedPostPolicy


harshavardhana:
Current cover profile is 29%

vadmeste:
isn't this duplicated to #453 ?

harshavardhana:
> isn't this duplicated to #453 ?

Yes , closing

Issue #455, opened 2016-07-21
Implement functional tests for listObjectParts() 


harshavardhana:
Current cover profile is 55% for api-list.go

GarimaKapoor:
@harshavardhana is this issue still valid?

harshavardhana:
Closing this..

Issue #454, opened 2016-07-21
Implement unit/functional tests for put-object-readat.go 


harshavardhana:
Current cover profile is 0%

GarimaKapoor:
@harshavardhana is this issue still valid?

harshavardhana:
Closing this..

Issue #453, opened 2016-07-21
Implement unit/function tests for post-policy.go 


harshavardhana:
Current cover profile is 0%

harshavardhana:
Is this fixed @vadmeste ?

vadmeste:
@harshavardhana, yes, coverage 81% for this file

PR #457, opened 2016-07-21
More test covers

vadmeste:
This increases the total test coverage from 61% to 74%,

One little problem is that the testing time will increase because I needed to trigger many multipart tests.. Since we increased the minimum part size to 64mb, then it will takes little more time and bandwidth

### DIFF
--- a/api_functional_v2_test.go
+++ b/api_functional_v2_test.go
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package minio_test
+package minio
 
 import (
 	"bytes"
@@ -28,8 +28,6 @@ import (
 	"os"
 	"testing"
 	"time"
-
-	"github.com/minio/minio-go"
 )
 
 // Tests bucket re-create errors.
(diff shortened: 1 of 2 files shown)


vadmeste:
Fixes #451 #453 #454 #455

api_functional_v2_test.go
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package minio_test
+package minio
```
harshavardhana: This was written on purpose.. check standard golang packages.

vadmeste: Actually, I changed it because I needed access to minPartSize, and that's important to be sure that we are triggering multipart in our tests. I still don't know if there is another way to do that

vadmeste: I found this https://golang.org/src/bytes/export_test.go which exports variables for testing purpose as written, but I also found this too: https://golang.org/src/log/log_test.go and  https://golang.org/src/log/example_test.go which implies in my sense that there are unit tests for internal code, and functional tests for the API.. It doesn't seem that there are any standard anyway.. besides, minPartSize should be private so the developer won't be able to modify it especially during operation.

harshavardhana: > I found this https://golang.org/src/bytes/export_test.go which exports variables for testing purpose as written, but I also found this too: https://golang.org/src/log/log_test.go and https://golang.org/src/log/example_test.go which implies in my sense that there are unit tests for internal code, and functional tests for the API.. It doesn't seem that there are any standard anyway.. besides, minPartSize should be private so the developer won't be able to modify it especially during operation.

Ah

Issue #438, opened 2016-07-07
copyObject: x-amz-copy-source should urlEncode the object source string.


harshavardhana:
https://github.com/minio/minio-go/blob/master/api-put-object-copy.go#L41

PR #465, opened 2016-07-26
CopyObject: urlEncode the object source string

awwalker:
Fixes: #438

### DIFF
--- a/api-put-object-copy.go
+++ b/api-put-object-copy.go
@@ -38,7 +38,7 @@ func (c Client) CopyObject(bucketName string, objectName string, objectSource st
 	}
 
 	// Set copy source.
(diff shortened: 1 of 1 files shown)


api-put-object-copy.go
```diff
@@ -38,7 +41,7 @@ func (c Client) CopyObject(bucketName string, objectName string, objectSource st
 	}
 
 	// Set copy source.
-	customHeaders.Set("x-amz-copy-source", objectSource)
+	customHeaders.Set("x-amz-copy-source", url.QueryEscape(objectSource))
```
harshavardhana: You would have to use urlEncodeName() here rather than url.QueryEscape the problem is perhaps because if "+" is a character in the objectSource name. Then it would be escaped improperly.

PR #470, opened 2016-08-02
pkg: add set package

balamurugana:
This package contains StringSet like string slice where it allows to
be a set.  StringSet has set operation methods like Intersection(),
Difference() and Union().  It also has JSON marshal/unmarshal
functions which generates/reads JSON array strings.

bucket-policy.Statement structure is modified to use StringSet than
string slice.

### DIFF
--- a/bucket-policy.go
+++ b/bucket-policy.go
@@ -17,10 +17,10 @@
 package minio
 
 import (
-	"encoding/json"
 	"fmt"
-	"sort"
 	"strings"
+
+	"github.com/minio/minio-go/pkg/set"
 )
 
 // maximum supported access policy size.
@@ -51,7 +51,7 @@ func (p BucketPolicy) isValidBucketPolicy() bool {
 
 // User - canonical users list.
 type User struct {
-	AWS []string
+	AWS set.StringSet
 }
 
 // Statement - minio policy statement
@@ -59,8 +59,8 @@ type Statement struct {
 	Sid        string
 	Effect     string
 	Principal  User                         `json:"Principal"`
-	Actions    []string                     `json:"Action"`
-	Resources  []string                     `json:"Resource"`
+	Actions    set.StringSet                `json:"Action"`
+	Resources  set.StringSet                `json:"Resource"`
 	Conditions map[string]map[string]string `json:"Condition,omitempty"`
 }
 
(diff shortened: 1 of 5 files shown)


pkg/set/stringset.go
```diff
@@ -0,0 +1,161 @@
...
+
+package set
+
+import "encoding/json"
+
+// StringSet - uses map as set of strings.
+type StringSet map[string]struct{}
+
+// keys - returns StringSet keys.
+func (set StringSet) keys() []string {
```
hackintoshrao: @balamurugana : Can you add tests for the package.

balamurugana: Done

bucket-policy.go
```diff
@@ -383,90 +324,15 @@ func removeBucketPolicyStatement(statements []Statement, bucketName string, obje
...
-			}
-			statement.Actions = []string{action}
-		}
-		// Resources can have two different types, either as []string
-		// and either as regular 'string'. We fall back to doing this
-		// since there is no other easier way to fix this.
-		err = json.Unmarshal(stmtUntyped.Resource, &statement.Resources)
-		if err != nil {
-			var resource string
-			err = json.Unmarshal(stmtUntyped.Resource, &resource)
```
harshavardhana: How are we handling when "Resource" is of only type "string" ?

bucket-policy.go
```diff
@@ -383,90 +324,15 @@ func removeBucketPolicyStatement(statements []Statement, bucketName string, obje
...
-		statement.Effect = stmtUntyped.Effect
-		statement.Conditions = stmtUntyped.Condition
-
-		// AWS user can have two different types, either as []string
-		// and either as regular 'string'. We fall back to doing this
-		// since there is no other easier way to fix this.
-		err = json.Unmarshal(stmtUntyped.Principal.AWS, &statement.Principal.AWS)
-		if err != nil {
-			var awsUser string
-			err = json.Unmarshal(stmtUntyped.Principal.AWS, &awsUser)
```
harshavardhana: How are we handling when "Principal.AWS" is of type only "string" ? - the reason is such a policy is valid and our unmarshal will fail.

pkg/set/stringset.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+// MarshalJSON - converts to JSON data.
+func (set StringSet) MarshalJSON() ([]byte, error) {
+	return json.Marshal(set.keys())
+}
+
+// UnmarshalJSON - parses JSON data and creatse new set with it.
+func (set *StringSet) UnmarshalJSON(data []byte) error {
+	sl := []string{}
```
harshavardhana: Can you add an explanation here why we do type types of unmarshalling here? one for "slice of strings" and another for just "string"

pkg/set/stringset.go
```diff
@@ -0,0 +1,165 @@
...
+			set.Add(s)
+		}
+	}
+
+	return err
+}
+
+// String - returns printable string of the set.
+func (set StringSet) String() string {
+	data, _ := set.MarshalJSON()
```
harshavardhana: In all cases error be ignored here? - if yes - this needs a comment on why 'err' is ignored.

pkg/set/stringset.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+	return nset
+}
+
+// Equals - checks whether given set is equal to current set or not.
+func (set StringSet) Equals(sset StringSet) bool {
+	if len(set) != len(sset) {
+		return false
+	}
+
```
harshavardhana: Needs a comment.

pkg/set/stringset.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+// Contains - checks if string is in the set.
+func (set StringSet) Contains(s string) bool {
+	_, ok := set[s]
+	return ok
+}
+
+// FuncMatch - returns new set containing each value who passes match function.
+func (set StringSet) FuncMatch(matchFn func(string, string) bool, matchString string) StringSet {
```
harshavardhana: We need to explain a bit about 'matchFn' here.. on why it is 'string, string'

bucket-policy.go
balamurugana: `set.StringSet` is taking care of a string or string slice automatically.

bucket-policy.go
balamurugana: `Principal.AWS` is `set.StringSet` type which is taking care of a string or string slice automatically.

pkg/set/stringset.go
balamurugana: Done

pkg/set/stringset.go
balamurugana: I have modified the method below to avoid using MarshalJSON()

```go
// String - returns printable string of the set.
func (set StringSet) String() string {
	return fmt.Sprintf("%s", set.keys())
}
```

pkg/set/stringset.go
balamurugana: Done

pkg/set/stringset.go
balamurugana: Done.

pkg/set/stringset_test.go
```diff
@@ -0,0 +1,292 @@
...
+ */
+
+package set
+
+import (
+	"strings"
+	"testing"
+)
+
+// Test NewStringSet()
```
hackintoshrao: Can these comments be more elaborate, with more context on tests, like these https://github.com/minio/minio/blob/master/object-api-getobject_test.go#L32 .

balamurugana: Done

PR #479, opened 2016-08-08
Bucket location sends UNSIGNED-PAYLOAD with https connection

### LINKED ISSUE #473: getBucketLocation should do 'UNSIGNED-PAYLOAD' for https as well
minio-go is not doing UNSINGED-PAYLOAD for getbucketlocation

vadmeste:
Fixes #473

### DIFF
--- a/bucket-cache.go
+++ b/bucket-cache.go
@@ -165,7 +165,13 @@ func (c Client) getBucketLocationRequest(bucketName string) (*http.Request, erro
 
 	// Set sha256 sum for signature calculation only with signature version '4'.
 	if c.signature.isV4() {
-		req.Header.Set("X-Amz-Content-Sha256", hex.EncodeToString(sum256([]byte{})))
+		var contentSha256 string
+		if c.secure {
+			contentSha256 = unsignedPayload
+		} else {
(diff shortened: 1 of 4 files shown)


bucket-cache.go
```diff
@@ -165,7 +165,13 @@ func (c Client) getBucketLocationRequest(bucketName string) (*http.Request, erro
 
 	// Set sha256 sum for signature calculation only with signature version '4'.
 	if c.signature.isV4() {
-		req.Header.Set("X-Amz-Content-Sha256", hex.EncodeToString(sum256([]byte{})))
+		var contentSha256 string
+		if c.secure {
+			contentSha256 = "UNSIGNED-PAYLOAD"
```
harshavardhana: Let's name this a global value as a 'const' and then use it everywhere.

vadmeste: This is part of the s3 protocol and it doesn't seem that it will be changed anytime soon :)

harshavardhana:
> This is part of the s3 protocol and it doesn't seem that it will be changed anytime soon :)

Even then when it changes it would require one place for it to be changed.

PR #481, opened 2016-08-09
Check the Server header to verify listen bucket notification support

vadmeste:
We need to check the Server header in listen bucket notification webservice call to be sure that we are talking with minio servers since this is about a custom API.

### DIFF
--- a/api-error-response.go
+++ b/api-error-response.go
@@ -233,3 +233,13 @@ func ErrNoSuchBucketPolicy(message string) error {
 		RequestID: "minio",
 	}
 }
+
+// ErrAPINotSupported - API not supported response
+// The specified API call is not supported
+func ErrAPINotSupported(message string) error {
+	return ErrorResponse{
+		Code:      "APINotSupported",
+		Message:   message,
+		RequestID: "minio",
+	}
+}
--- a/api-notification.go
+++ b/api-notification.go
@@ -134,7 +134,15 @@ func (c Client) ListenBucketNotification(bucketName string, accountArn Arn, done
 			return
 		}
 
-		// Continuously run and listen on bucket notification.
+		// Check ARN partition to verify if listening bucket is supported
+		if accountArn.Partition != "minio" {
+			notificationInfoCh <- NotificationInfo{
+				Err: ErrAPINotSupported("Listening bucket notification is specific only to `minio` partitions"),
+			}
+			return
+		}
+
+		// Continously run and listen on bucket notification.
 		for {
 			urlValues := make(url.Values)
 			urlValues.Set("notificationARN", accountArn.String())


api-error-response.go
```diff
@@ -233,3 +233,13 @@ func ErrNoSuchBucketPolicy(message string) error {
 		RequestID: "minio",
 	}
 }
+
+// ErrAPINotSupported - API not supported response
+// The specified API call is not supported
+func ErrAPINotSupported(message string) error {
+	return ErrorResponse{
+		Code:      "NoSuchBucketPolicy",
```
harshavardhana: This means there is no bucket policy.

api-notification.go
```diff
@@ -156,6 +157,14 @@ func (c Client) ListenBucketNotification(bucketName string, accountArn Arn, done
 			return
 		}
 
+		// Check if we are calling Minio server which supports listening to bucket notifications
+		if !strings.HasPrefix(resp.Header.Get("Server"), "Minio/") {
+			notificationInfoCh <- NotificationInfo{
+				Err: ErrAPINotSupported("Listening to bucket notification not supported"),
```
harshavardhana: 'Listen on bucket notification is only supported by Minio.'

api-notification.go
```diff
@@ -156,6 +157,14 @@ func (c Client) ListenBucketNotification(bucketName string, accountArn Arn, done
 			return
 		}
 
+		// Check if we are calling Minio server which supports listening to bucket notifications
+		if !strings.HasPrefix(resp.Header.Get("Server"), "Minio/") {
```
harshavardhana: This won't work if the server is running behind a proxy.

api-error-response.go
vadmeste: :+1:

api-notification.go
vadmeste: According to this https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.38, proxies are not allowed to modify the server Header

harshavardhana: > According to this https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.38, proxies are not allowed to modify the server Header

This RFC is deprecated can you check the latest?

vadmeste: > This RFC is deprecated can you check the latest?

yes, you wer right, in recent specs, the paragraph about not touching the server header is removed (compare this https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.38 and https://tools.ietf.org/html/rfc7231#page-73)

In addition, I noticed that some reverse proxies put their own names in the server field and some will remove it.
https://trac.nginx.org/nginx/ticket/538
http://cherokee-project.com/doc/modules_handlers_proxy.html
https://caddyserver.com/docs/header
https://twitter.com/caddyserver/status/714587023948931072

Clearly, users are not reluctant to remove the Server header in their proxies, so better to avoid this solution.

harshavardhana: > In addition, I noticed that some reverse proxies put their own names in the server field and some will remove it.
> https://trac.nginx.org/nginx/ticket/538
> http://cherokee-project.com/doc/modules_handlers_proxy.html
> https://caddyserver.com/docs/header
> https://twitter.com/caddyserver/status/714587023948931072
> 
> Clearly, users are not reluctant to remove the Server header in their proxies, so better to avoid this solution.

Yes it has changed.

vadmeste:
so here, what we probably could do:
1. minio-go will be aware about the server capabilities by using a parameter in the library set by the developer to indicate that we are talking with a minio server (little ugly)
2. use a different header name other than `Server`, this link doesn't recommand using custom headers in general, https://www.safaribooksonline.com/library/view/restful-web-services/9780596809140/ch01s13.html  but I don't know how much this is reliable
3. use `OPTIONS *` which is at least described in http RFC as it returns server capabilities

harshavardhana:
> 3. use OPTIONS * which is at least described in http RFC as it returns server capabilities

This is better.

harshavardhana:
> minio-go will be aware about the server capabilities by using a parameter in the library set by the developer to indicate that we are talking with a minio server (little ugly)

Yes this is even better.. Arns not having 'minio' in the beginning are invalid. Otherwise we send anyways and if we see XML data on the wire we reject such requests. 

XML data on the wire will have an xml.Header which can be read and validated.

harshavardhana:
> XML data on the wire will have an xml.Header which can be read and validated.

Without reading the entire body.

harshavardhana:
ping @vadmeste - anything pending here?

harshavardhana:
ping @vadmeste

vadmeste:
I just tested, I thought testing on ARNs is enough, but it turned out we need to check the XML header too, will do that right now

vadmeste:
@harshavardhana, there is no easy way to validate xml on the wire.. xml.Unmarshal() of the xml header returns an error. Also the server is streaming data, so we can't get all data and unmarshal them to check if it is an xml or a json.

The current state of the patch is that it is refusing ARNs with partition field different than `minio`.

harshavardhana:
> The current state of the patch is that it is refusing ARNs with partition field different than minio.

This is fine for now.

harshavardhana:
Merging once build is complete..

PR #485, opened 2016-08-11
[WIP] refactor get and set bucket policy.

balamurugana:


harshavardhana:
Need to bring back the test cases in this patch as well. Let's not do it separately.

balamurugana:
> Need to bring back the test cases in this patch as well. Let's not do it separately.

I am adding test cases now.

Issue #484, opened 2016-08-10
Allow reading only a limited range of a file


fd0:
At the moment `GetObject()` will directly initiate a GET request for the whole object. I regularly only need a small portion of the object, when calling `Seek()` and `Read()` or `ReadAt()` a second GET request with a byte range is sent to the server. The first request is then aborted as soon as `Close()` is called, but at least the beginning of the file is always sent by the server, which wastes bandwidth and memory on the client.

I'd like to be able to only request a range of a file.

fd0:
Thanks!

Issue #464, opened 2016-07-26
Tests: Running go test -race ./... causes SIGQUIT


awwalker:
```
=== RUN   TestResumableFPutObject
SIGQUIT: quit
PC=0x63a5b m=0

goroutine 0 [idle]:
runtime.mach_semaphore_wait(0x23000001003, 0x208ad000165, 0x1, 0x208ad000165, 0x0, 0x911a40, 0x57019, 0xffffffffffffffff, 0x9118e8, 0x7fff5fbff4e4, ...)
	/usr/local/Cellar/go/1.6/libexec/src/runtime/sys_darwin_amd64.s:411 +0xb
runtime.semasleep1(0xffffffffffffffff, 0x9118e8)
	/usr/local/Cellar/go/1.6/libexec/src/runtime/os1_darwin.go:423 +0xdf
runtime.semasleep.func1()
	/usr/local/Cellar/go/1.6/libexec/src/runtime/os1_darwin.go:439 +0x29
runtime.systemstack(0x7fff5fbff4e8)
	/usr/local/Cellar/go/1.6/libexec/src/runtime/asm_amd64.s:307 +0xab
runtime.semasleep(0xffffffffffffffff, 0x0)
	/usr/local/Cellar/go/1.6/libexec/src/runtime/os1_darwin.go:440 +0x36
runtime.notesleep(0x9120e8)
	/usr/local/Cellar/go/1.6/libexec/src/runtime/lock_sema.go:166 +0xed
runtime.stopm()
	/usr/local/Cellar/go/1.6/libexec/src/runtime/proc.go:1535 +0x10b
runtime.findrunnable(0xc820018000, 0x0)
	/usr/local/Cellar/go/1.6/libexec/src/runtime/proc.go:1973 +0x808
runtime.schedule()
	/usr/local/Cellar/go/1.6/libexec/src/runtime/proc.go:2072 +0x2a6
runtime.park_m(0xc8203c8180)
	/usr/local/Cellar/go/1.6/libexec/src/runtime/proc.go:2137 +0x157
runtime.mcall(0x57960)
	/usr/local/Cellar/go/1.6/libexec/src/runtime/asm_amd64.s:233 +0x5b

goroutine 1 [chan receive]:
testing.RunTests(0x762418, 0x90f840, 0x4e, 0x4e, 0x0)
	/usr/local/Cellar/go/1.6/libexec/src/testing/testing.go:583 +0xb39
[...]

PR #495, opened 2016-08-25
GetObject: allow reading only a limited range of a file

awwalker:
Fixes: #484 

This removes a test in api_unit_test.go because newObject no longer functions by taking in ObjectInfo as a parameter. 

Running $ go test -v ./... runs into #464 but testing all individual GetObject tests and PutObject tests runs just fine.

### DIFF
--- a/api-get-object.go
+++ b/api-get-object.go
@@ -36,12 +36,14 @@ func (c Client) GetObject(bucketName, objectName string) (*Object, error) {
 		return nil, err
 	}
 
-	// Start the request as soon Get is initiated.
-	httpReader, objectInfo, err := c.getObject(bucketName, objectName, 0, 0)
-	if err != nil {
-		return nil, err
-	}
+	var httpReader io.ReadCloser
+	var objectInfo ObjectInfo
+	var err error
 
+	// Create start channel.
+	startCh := make(chan firstRequest)
+	// Create first response channel.
+	firstResCh := make(chan firstReqRes)
 	// Create request channel.
 	reqCh := make(chan readRequest)
 	// Create response channel.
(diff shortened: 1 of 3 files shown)


api-get-object.go
```diff
@@ -53,25 +54,71 @@ func (c Client) GetObject(bucketName, objectName string) (*Object, error) {
...
 
 		// Loop through the incoming control messages and read data.
 		for {
 			select {
+			// This channel signals that we have received the first request for this object.
+			// Could be either: Read/ReadAt/Stat/Seek.
+			case firstReq := <-startCh:
+				if firstReq.isReadOp {
+					// First request is a Read/ReadAt call.
+					httpReader, objectInfo, err := c.getObject(bucketName, objectName, firstReq.Offset, 0)
```
harshavardhana: This has a scoping bug httpReader is set again in the local scope.

api-get-object.go
```diff
@@ -107,12 +154,28 @@ type readResponse struct {
...
+	Offset   int64  // Used for ReadAt offset.
+	Buffer   []byte // Used for ReadAt/Read requests.
+}
+
+// The first request asked must send back the objectInfo or an error to create a new Object.
+type firstReqRes struct {
+	Error      error
+	objectInfo ObjectInfo
+	Size       int
+	didRead    bool // Lets know subsequent calls whether or not httpReader has been initiated.
```
harshavardhana: The comment sentence can be worded better.

api-get-object.go
```diff
@@ -132,8 +197,29 @@ type Object struct {
...
+	beenRead bool
+}
+
+// setObjectInfo - blocks until the first request is completed and then either errors out
+// or sets the received objectInfo.
+func (o *Object) setObjectInfo() (int, error) {
+	firstResponse := <-o.firstResCh
+	// Error out here if the firstRequest caused an error.
+	if firstResponse.Error != nil {
+		return 0, firstResponse.Error
```
harshavardhana: This error should be set properly to 'o.prevErr' so that any subsequent operations on the same client object fail with the same error.

api-get-object.go
```diff
@@ -148,65 +234,107 @@ func (o *Object) Read(b []byte) (n int, err error) {
...
+			// We have reached end of file, we return io.EOF.
+			if o.currOffset >= o.objectInfo.Size {
+				return size, io.EOF
+			}
+			return size, nil
+		}
+
+		// Save any error.
+		o.prevErr = err
+		return size, err
```
harshavardhana: Since you are returning right here drop the else..

api-get-object.go
```diff
@@ -148,65 +234,107 @@ func (o *Object) Read(b []byte) (n int, err error) {
 	o.mutex.Lock()
 	defer o.mutex.Unlock()
 
-	// prevErr is previous error saved from previous operation.
-	if o.prevErr != nil || o.isClosed {
-		return 0, o.prevErr
-	}
+	// This is the first request.
+	if !o.isStarted {
+		o.isStarted = true
```
harshavardhana: Set isStarted as true only after we have successfully completed a channel handshake.

api-get-object.go
```diff
@@ -218,6 +346,21 @@ func (o *Object) Stat() (ObjectInfo, error) {
 	o.mutex.Lock()
 	defer o.mutex.Unlock()
 
+	if !o.isStarted {
+		// This is the first request.
+		o.isStarted = true
```
harshavardhana: Same as above.

api_unit_test.go
```diff
@@ -77,25 +77,6 @@ func TestGetReaderSize(t *testing.T) {
 		t.Fatalf("Reader length doesn't match got: %v, want: %v", size, len("Hello World"))
 	}
 
-	// Create request channel.
-	reqCh := make(chan readRequest)
```
harshavardhana: Do not remove this test. 

Send back 'objInfo' through a channel. and still proceed to execute getReaderSize() this would in-fact tell if the Size is appropriately set.

fd0:
Thanks for your work. I've had a look, and it's indeed much butter: For each invocation of `ReadAt()`, only one `GET` request with a byte range is sent to the server. That's a good result. Some traces:
before:
https://gist.github.com/fd0/9903492335f51cf4d68a61335a47a1af

after:
https://gist.github.com/fd0/e25881d7d660282ff0280fe09a81aaf7

Unfortunately I've just discovered something else: The `GET` request's byte range never specifies an end:
```
Range: bytes=2651761-
```
So the remainder of the file is read (depending on how fast the connection is closed) although I (as a user) never requested it.

The `io.ReaderAt` interface specifies that only `len(buf)` bytes are to be read: "ReadAt reads len(p) bytes into p starting at offset off in the underlying input source." https://godoc.org/io#ReaderAt
So I'd not expect that more is fetched.

I think we have two different use cases here: Either a user is interested in a number of sequential data chunks, in this case she would use `Seek()` followed by multiple calls to `Read()`. The code is very well suited to handle this.

The other use case is the one I'm interested in: I'd like to get a (small) part of a file, so I'm using the `ReadAt()` function with an offset and a buffer.

I'd suggest simplifying the `ReadAt()` function as follows: Execute exactly one GET request with the given byte range (offset-offset+len(buf)) and block until this request is returned. This would allow us to remove all the sync code from `ReadAt()` and make it a standalone function (sort of).

What do you think?

awwalker:
@fd0 thanks for taking a look! 

I think I have a simpler solution than making readAt a standalone function. I'll let getObject know whether or not this is a readAt request or standard read and use the (offset, offset+len(buf)) range for  readAt requests. I think this will solve your issue and let the underlying code stay the same. 

Does that sound good?

Once this change is in place there will be a downside in that for a smaller buffer the request could be aggressively long. Where each request for a separate range would initiate a new Get() when one full read would be better. To the server this could look like random i/o if they are initiate in parallel. 

Keeping a local cache and operating on that would be better. ReadAt() is good on a file system but on a network it is not ideal.

fd0:
Sounds good!

fd0:
Looks really good now! https://gist.github.com/fd0/a5b279b1eec8913440700b13fff2eee7

harshavardhana:
can you rebase this @awwalker - looks like travis build was missed.

fd0:
LGTM :+1:

api-get-object.go
```diff
@@ -107,12 +171,31 @@ type readResponse struct {
...
+	didRead    bool // Lets know subsequent calls whether or not httpReader has been initiated.
+}
+
 // Read request message container to communicate with internal
 // go-routine.
 type readRequest struct {
 	Buffer          []byte
 	Offset          int64 // readAt offset.
 	DidOffsetChange bool
+	beenRead        bool
```
harshavardhana: Needs to have a comment..

api-get-object.go
```diff
@@ -107,12 +171,31 @@ type readResponse struct {
...
+	isReadAt        bool // Determines if this request is a ReadAt request to allow "getting" only a specific range
```
harshavardhana:
```
Determines if this request is a ReadAt request to allow "getting" only a specific range
```
   ---> 

```
Determines if caller requested ReadAt to allow fetching a specific range
```

api-get-object.go
```diff
@@ -122,6 +205,8 @@ type Object struct {
 	mutex *sync.Mutex
 
 	// User allocated and defined.
+	startCh    chan<- firstRequest
```
harshavardhana: startCh --> firstReqCh to be consistent.

api-get-object.go
```diff
@@ -132,8 +217,32 @@ type Object struct {
...
 	// Previous error saved for future calls.
 	prevErr error
+
+	// Keeps track of if this object has been read yet.
+	beenRead bool
+}
+
+// setObjectInfo - blocks until the first request is completed and then either errors out
+// or sets the received objectInfo.
+func (o *Object) setObjectInfo() (int, error) {
```
harshavardhana: setObjectInfo should only set the objectInfo and return error if any. It shouldn't return size. Size is available for future operations via ``o.objectInfo``

api-get-object.go
```diff
@@ -97,8 +161,8 @@ func (c Client) GetObject(bucketName, objectName string) (*Object, error) {
 			}
 		}
 	}()
-	// Return the readerAt backed by routine.
-	return newObject(reqCh, resCh, doneCh, objectInfo), nil
+	// Create a newObject through the information sent back by firstReq.
```
harshavardhana: Can you word this differently.. ? - newObject is already created regardless of information sent back by firstReq.

Issue #482, opened 2016-08-10
api: Enable parallel uploads for multipart in PutObject


harshavardhana:

harshavardhana:
Fixed by #504

PR #504, opened 2016-09-02
PutObject (Multipart): add support for parallel upload of parts in mu…

awwalker:
Fix for: #482 and updates test cases to use random data and 4 parts in multipart tests to check parts are uploaded in the correct order.


awwalker:
@harshavardhana Ok will fix this up and resubmit...the original Appveyor problem was with using cheggaaa/pb inside the tests. What is the best way to deal with that error?

PR #505, opened 2016-09-03
putObject: Relax validating number of parts during upload.

harshavardhana:
There is always a chance in multipart upload that the new
file being uploaded file would have changed we should
allow it to be re-uploaded and completed.

### DIFF
--- a/api-put-object-readat.go
+++ b/api-put-object-readat.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"crypto/md5"
 	"crypto/sha256"
+	"fmt"
 	"hash"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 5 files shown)


api-put-object-readat.go
```diff
@@ -186,6 +186,11 @@ func (c Client) putObjectMultipartFromReadAt(bucketName, objectName string, read
 		tmpBuffer.Reset()
 	}
 
+	// Verify if we uploaded all the data.
+	if totalUploadedSize != size {
+		return totalUploadedSize, ErrUnexpectedEOF(totalUploadedSize, size, bucketName, objectName)
+	}
+
```
krishnasrinivas: `totalUploadedSize` is calculated in the for loop below

krishnasrinivas:
in putObjectMultipartFromReadAt:

```
190         for _, part := range partsInfo {                                                            
191                 var complPart completePart                                                          
192                 complPart.ETag = part.ETag                                                          
193                 complPart.PartNumber = part.PartNumber                                              
194                 totalUploadedSize += part.Size                                                      
195                 complMultipartUpload.Parts = append(complMultipartUpload.Parts, complPart)          
196         }   
```

partsInfo will have excess parts related to previously tried upload and will cause data corruption for the current object

krishnasrinivas:
the problem is there in api-put-object-multipart.go and api-put-object-file.go as well

harshavardhana:
> the problem is there in api-put-object-multipart.go and api-put-object-file.go as well

Parts are only skipped and not verified only in ReadAt() whereas both Stream and File verify md5sum and replace the parts in-place for each part number. I don't see why there would be excess?

ReadAt() can have the issue you are mentioning because without verifying total parts we can't be sure that we uploaded requested number of parts.  ReadAt() optimization was introduced for GetObject(). 

I see we can't remove the check for total parts verification since we have to be sure that we uploaded correctly otherwise would write a wrong representation of the object.

harshavardhana:
> I see we can't remove the check for total parts verification since we have to be sure that we uploaded correctly otherwise would write a wrong representation of the object.

Ah i see now previous check prevented this from happening, we need to use totalPartsCount as the main value to loop and add.

PR #506, opened 2016-09-03
putObject: Streaming uploads with size -1 should avoid erroring out f…

harshavardhana:
…or missing parts.

size == -1 should be treated specially for putObjectStream() missing parts
shouldn't error out since with '-1' number of parts are dynamic only when
the stream ends we do know that the we have finished the upload.

Fixes - minio/mc#1819

### DIFF
--- a/api-put-object-multipart.go
+++ b/api-put-object-multipart.go
@@ -94,7 +94,8 @@ func (c Client) putObjectMultipartStream(bucketName, objectName string, reader i
 	}
 
 	// If This session is a continuation of a previous session fetch all
-	// previously uploaded parts info.
+	// previously uploaded parts info and as a special case only fetch partsInfo
+	// for only known upload size.
 	if !isNew {
 		// Fetch previously uploaded parts and maximum part size.
 		partsInfo, err = c.listObjectParts(bucketName, objectName, uploadID)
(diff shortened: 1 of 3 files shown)


krishnasrinivas:
> size == -1 should be treated specially for putObjectStream() and avoid fetching
> partsInfo for previously uploaded parts since '-1' has no guarantee that the
> data has no changed. Let the client upload again and not resume for unknown sizes.

but we check if part is already uploaded by checking ETag right? so if a stream is uploaded with size `-1` with same data we will lose that optimization, right?

harshavardhana:
> but we check if part is already uploaded by checking ETag right? so if a stream is uploaded with size -1 with same data we will lose that optimization, right?

We can keep the optimization and just that we don't need to error out for '-1'  size since the total parts are dynamic in that scenario.

api-put-object-multipart.go
```diff
@@ -190,6 +191,10 @@ func (c Client) putObjectMultipartStream(bucketName, objectName string, reader i
 	for i := 1; i <= totalPartsCount; i++ {
```
krishnasrinivas: this will also cause data corruption if previous incomplete upload had more parts than the correct upload. you can use:
`for i := 1; i <= partNumber; i++ {` to avoid that situation and also the `if size < 0 {` check below will not be needed.

harshavardhana: Ah i see.

PR #510, opened 2016-09-07
putObject: use lastPartNumber instead for generating complete multipa…

harshavardhana:
…rts.

partNumber is higher than the lastPartNumber due to the loop. Use
lastPartNumber instead since the file size is not expected to change
underneath with in the call.

For streaming based stick to partNumber-1 for generating parts.

### DIFF
--- a/api-put-object-file.go
+++ b/api-put-object-file.go
@@ -172,10 +172,13 @@ func (c Client) putObjectMultipartFromFile(bucketName, objectName string, fileRe
 		return 0, err
 	}
 
+	// Used for readability, lastPartNumber is always totalPartsCount.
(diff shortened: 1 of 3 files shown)


spantaleev:
:+1: 

Seems like a previous commit (40505f5d08c7) introduced an off-by-one regression.

During the `complMultipartUpload.Parts` assembly step, it would compare against `partNumber` (which ends up being `totalPartsCount + 1`), making it look for one more (non-existing) part.

With these new changes in place, it seems like the problem has been corrected.

harshavardhana:
Yes @spantaleev thanks for noticing

PR #514, opened 2016-09-12
Add delete to subresource list

breml:
According to the current [AWS S3 API documentation](http://docs.aws.amazon.com/AmazonS3/latest/dev/RESTAuthentication.html#ConstructingTheCanonicalizedResourceElement) the delete query
string parameter must be included when you create the CanonicalizedResource
for a multi-object Delete request.

### DIFF
--- a/request-signature-v2.go
+++ b/request-signature-v2.go
@@ -257,6 +257,7 @@ func writeCanonicalizedHeaders(buf *bytes.Buffer, req http.Request) {
 // have signature-related issues
 var resourceList = []string{
 	"acl",
+	"delete",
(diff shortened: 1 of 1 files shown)


krishnasrinivas:
@harshavardhana though minio-go does not support mult-object delete, it will be good to have this change for correctness point of view, also in case we enhance minio-go to do multi object delete.

harshavardhana:
> @harshavardhana though minio-go does not support mult-object delete, it will be good to have this change for correctness point of view, also in case we enhance minio-go to do multi object delete.

:+1:

PR #515, opened 2016-09-18
Percent encode '/' (forward slash) in query string

donatello:
* '/' is not a valid character in query strings as it can be mistaken
  for a path separator.

* Added a test case for this change.

* Additional reference: See definition of UriEncode() in AWS S3 v4
  Signature Authorization documentation:
  https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html
  - relevant snippet is: `Encode the forward slash character, '/',
  everywhere except in the object key name. For example, if the object
  key name is photos/Jan/sample.jpg, the forward slash in the key name
  is not encoded.`

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -310,7 +310,22 @@ func isValidObjectPrefix(objectPrefix string) error {
 	return nil
 }
 
-// queryEncode - encodes query values in their URL encoded form.
+//expects ascii encoded strings - from output of urlEncodePath
+func percentEncodeSlash(s string) string {
+	r := ""
+	for _, k := range s {
+		if k == '/' {
+			r += "%2F"
+		} else {
+			r += string(k)
+		}
+	}
+	return r
+}
+
+// queryEncode - encodes query values in their URL encoded form. In
+// addition to the percent encoding performed by urlEncodePath() used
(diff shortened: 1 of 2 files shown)


utils.go
```diff
@@ -310,7 +310,22 @@ func isValidObjectPrefix(objectPrefix string) error {
 	return nil
 }
 
-// queryEncode - encodes query values in their URL encoded form.
+//expects ascii encoded strings - from output of urlEncodePath
+func percentEncodeSlash(s string) string {
```
harshavardhana: Can you add a separate test case for `percentEncodeSlash()`

harshavardhana:
Everything else LGTM, Thanks :+1:

utils.go
donatello: Done!

krishnasrinivas:
https://golang.org/pkg/strings/#Replace would have been simpler?

donatello:
Krishna Srinivas writes:

> https://golang.org/pkg/strings/#Replace would have been simpler?
>

You are right, thanks. I can refactor this.

--
Aditya Manthramurthy
https://minio.io

PR #522, opened 2016-09-26
Add multi delete API

vadmeste:


api-remove.go
```diff
@@ -74,6 +79,86 @@ func (c Client) RemoveObject(bucketName, objectName string) error {
 	return nil
 }
 
+// generateRemoveMultiObjects - generate the XML request for remove multi objects request
+func generateRemoveMultiObjects(objects []string) []byte {
+	rmObjects := []deleteObject{}
+	for _, obj := range objects {
+		rmObjects = append(rmObjects, deleteObject{Key: obj})
+	}
+	xmlbytes, _ := xml.Marshal(deleteMultiObjects{Objects: rmObjects, Quiet: false})
```
harshavardhana: Set quiet as true.

api-remove.go
```diff
@@ -74,6 +79,86 @@ func (c Client) RemoveObject(bucketName, objectName string) error {
...
+				Code:    obj.Code,
+				Message: obj.Message,
+			}
+		}
+	}
+	return errs
+}
+
+// MultiRemoveObjects remove multiples objects from a bucket.
+func (c Client) MultiRemoveObjects(bucketName string, objectsCh chan []string) (chan []error, error) {
```
harshavardhana: We need to make it little different 

```
func (c Client) MultiRemoveObjects(bucketName string, objectsCh chan<- string) (<-chan error) {
go func() 
       for { 
             select {
                     case obj, ok := <-objectsCh:
                      if !ok {
                           return
                       }
                      .... read 1000 and send delete, process response and return over error channel.
                      for _, resp := range response {
                                 errCh <- resp.Error 
                      }
             }
    }
}()
```

balamurugana:
As I am enhancing `mc rm` using batch delete, multi-delete in minio-go could accept a channel for input args and returns a channel as result.

api-remove.go
harshavardhana: Yes something like that. This is how ListObjects() does.

api-remove.go
```diff
@@ -74,6 +78,110 @@ func (c Client) RemoveObject(bucketName, objectName string) error {
...
+	xmlBytes, _ := xml.Marshal(deleteMultiObjects{Objects: rmObjects, Quiet: true})
+	return xmlBytes
+}
+
+// processRemoveMultiObjectsResponse - parse the remove multi objects web service
+// and return the success/failure result status for each object
+func processRemoveMultiObjectsResponse(body io.Reader, objects []string, errorCh chan objectNameError) {
+	// Parse multi delete XML response
+	rmResult := &deleteMultiObjectsResult{}
+	respBody, _ := ioutil.ReadAll(body)
```
harshavardhana: Let's not ignore the error here..

api-remove.go
```diff
@@ -74,6 +78,110 @@ func (c Client) RemoveObject(bucketName, objectName string) error {
...
+	err := xml.Unmarshal(respBody, rmResult)
```
harshavardhana: you can directly use `xmlDecoder(resp.Body, rmResult)`

api-remove.go
```diff
@@ -74,6 +78,110 @@ func (c Client) RemoveObject(bucketName, objectName string) error {
...
+	if err := isValidBucketName(bucketName); err != nil {
+		return nil, err
+	}
+	errorCh := make(chan objectNameError)
+	go func() {
+		maxEntries := 1000
+		finish := false
+		urlValues := make(url.Values)
+		urlValues.Set("delete", "")
+
```
harshavardhana: Needs a comment here..

api-remove.go
```diff
@@ -74,6 +78,110 @@ func (c Client) RemoveObject(bucketName, objectName string) error {
...
+// MultiRemoveObjects remove multiples objects from a bucket.
+// The list of objects to remove are received from objectsCh.
+// Remove failures are sent back via error channel.
+func (c Client) MultiRemoveObjects(bucketName string, objectsCh chan string) (chan objectNameError, error) {
+	// Input validation.
+	if err := isValidBucketName(bucketName); err != nil {
+		return nil, err
+	}
+	errorCh := make(chan objectNameError)
+	go func() {
```
harshavardhana: Needs a comment here..

api-remove.go
```diff
@@ -74,6 +78,110 @@ func (c Client) RemoveObject(bucketName, objectName string) error {
...
+				continue
+			}
+
+			// Process multiobjects remove xml response
+			processRemoveMultiObjectsResponse(resp.Body, batch, errorCh)
+
+			closeResponse(resp)
+		}
+		// Multi delete finished. Close error channel
+		close(errorCh)
```
harshavardhana: You can do 'defer close(errorCh)' very early on.

api-s3-datatypes.go
```diff
@@ -206,3 +206,38 @@ type createBucketConfiguration struct {
...
+	XMLName xml.Name `xml:"Delete"`
+	Quiet   bool
+	Objects []deleteObject `xml:"Object"`
+}
+
+// deletedMultiObjectsResult container for MultiObjects Delete XML response
+type deleteMultiObjectsResult struct {
+	XMLName           xml.Name           `xml:"DeleteResult"`
+	DeletedObjects    []deletedObject    `xml:"Deleted"`
+	NonDeletedObjects []nonDeletedObject `xml:"Error"`
```
harshavardhana: This should be `UnDeletedObjects`

api_functional_v4_test.go
```diff
@@ -497,6 +498,86 @@ func TestGetObjectClosedTwice(t *testing.T) {
...
+		t.Fatal("Error:", err, bucketName)
+	}
+
+	r := bytes.NewReader(bytes.Repeat([]byte("a"), 8))
+
+	// Multi remove of 1100 objects
+	nrObjects := 1100
+
+	// Prepare objects and upload them
+	objects := make([]string, nrObjects)
```
harshavardhana: Perhaps rather than buffering like this you can show that as soon as you have objectName you are writing to channel. 

```
go func() 
      defer close(objectsCh) 
       for i := 0; i < nrObjects; i++ {
 		objectName := "sample" + strconv.Itoa(i) + ".txt"
 		_, err = c.PutObject(bucketName, objectName, r, "application/octet-stream")
 		if err != nil {
 			t.Fatal("Error: PutObject shouldn't fail.")
 		}
 		objectCh <- objectName
 	}
}()
```

what do you think?

vadmeste: > what do you think?

+1

vadmeste: > what do you think?

+1

PR #531, opened 2016-10-10
Export API error codes to facilitate error inspection by developers

vadmeste:
This is an API change but looks like minor since I replaced `ErrorResponse.Code` type to `APIErrorCode`: 
- `type APIErrorCode string`

### DIFF
--- a/api-error-response.go
+++ b/api-error-response.go
@@ -38,7 +38,7 @@ import (
 // ErrorResponse - Is the typed error returned by all API operations.
 type ErrorResponse struct {
 	XMLName    xml.Name `xml:"Error" json:"-"`
-	Code       string
+	Code       APIErrorCode
 	Message    string
 	BucketName string
 	Key        string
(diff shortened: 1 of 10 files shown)


api-error-response.go
```diff
@@ -83,6 +83,32 @@ const (
 	reportIssue = "Please report this issue at https://github.com/minio/minio-go/issues."
 )
 
+type APIErrorCode string
+
+var (
+	ErrCodeNoSuchBucket          = APIErrorCode("NoSuchBucket")
```
vadmeste: Currently, developers need to read  minio-go API + S3 spec to understand errors received from minio-go library. I just wanted to make them look only at minio-go documentation with some details of the errors that they could receive. The code will still work if we receive an error code that we haven't seen before. In that case, I think the only change that we can do is adding a new exported error variable as I did above and this will be a backward compatible change. 



> Minor deviations require significant amount of code change on our end.

Can you elaborate on this?

harshavardhana: The problem with typed approach is that this is non-exhaustive it requires us to constantly track the meaning. Current approach is XML holds the value and you look for what is needed as and when needed. 

Current approach is simpler a typed approach is that we are documenting all kinds of errors that can happen and will happen and their associated meaning.  Just doing it for a subset of errors is not useful. If we do it we do it for all S3 API errors we might receive.

harshavardhana:
Closing this due to age, please re-open if still needed.

PR #533, opened 2016-10-20
Fixed race condition for ListenBucketNotification.

mikepulaski:
If you begin listening to a bucket and upload in rapid succession, you will miss notifications because `ListenBucketNotification` returns instantly. This is because`ListenBucketNotification` returns a chan for reading before the actual GET request starts.

To solve this, I added a semaphore which signals once either a preflight error occurs or after the first request has been accepted by the server.

I would have loved to add a test for this, but I wasn’t sure how to do so since this is a Minio-only feature. The other client methods seem to use AWS S3.


harshavardhana:
> If you begin listening to a bucket and upload in rapid succession, you will miss notifications because ListenBucketNotification returns instantly. This is becauseListenBucketNotification returns a chan for reading before the actual GET request starts.

The channel blocks until any data is written so until GET is successful there is no data for the caller to read where would the events be lost?,  can you provide an example?

harshavardhana:
Closing this due to age. Please re-open if still relevant.

PR #549, opened 2016-11-10
Allow to add success_action_status to policy

montoias:
Allow policy to have *success_action_status*

http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPOST.html

```
Accepts the values 200, 201, or 204 (default).

If the value is set to 200 or 204, Amazon S3 returns an empty document with a 200 or 204 status code.

If the value is set to 201, Amazon S3 returns an XML document with a 201 status code.

If the value is not set or if it is set to an invalid value, Amazon S3 returns an empty document with a 204 status code.

Type: String

```

### DIFF
--- a/post-policy.go
+++ b/post-policy.go
@@ -149,6 +149,24 @@ func (p *PostPolicy) SetContentLengthRange(min, max int64) error {
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


montoias:
Thank you @harshavardhana

PR #554, opened 2016-11-21
return an error when max retry is reached in executeMethod()

ericychoi:
DO NOT MERGE, missing tests.

### DIFF
--- a/api.go
+++ b/api.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"encoding/base64"
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -34,6 +35,8 @@ import (
 	"sync"
 	"time"
 )
+
+var MAX_RETRY_ERROR error = errors.New("max retry reached")
 
 // Client implements Amazon S3 compatible methods.
 type Client struct {
@@ -448,6 +451,8 @@ func (c Client) executeMethod(method string, metadata requestMetadata) (res *htt
 	// Indicate to our routine to exit cleanly upon return.
 	defer close(doneCh)
 
+	var retryCount int
+
 	// Blank indentifier is kept here on purpose since 'range' without
 	// blank identifiers is only supported since go1.4
 	// https://golang.org/doc/go1.4#forrange.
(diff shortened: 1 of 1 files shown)


harshavardhana:
This is not a bad idea but i think we can return the actual error rather than sending special error like retry. Right now it is sent it is just that DeleteObject() ignores it. You can modify that part of code instead.

ericychoi:
I believe `err` is not sent back to `RemoveObject()`.  Since `c.do()` will not return an error on 500.  (https://github.com/minio/minio-go/blob/master/api.go#L479)

It will return an error on closed connection or empty response, but not on a valid 500.  

Now I can inspect `resp` in `RemoveObject()` and if it contains non-2xx status code, return an error to signify that to the client.  Would you think that's better? Keep in mind then the same issue will keep happening for any other calls that use `executeMethod()` (update, for instance)

harshavardhana:
> I believe err is not sent back to RemoveObject(). Since c.do() will not return an error on 500. (https://github.com/minio/minio-go/blob/master/api.go#L479)
> 
> It will return an error on closed connection or empty response, but not on a valid 500.
> 
> Now I can inspect resp in RemoveObject() and if it contains non-2xx status code, return an error to signify that to the client. Would you think that's better? Keep in mind then the same issue will keep happening for any other calls that use executeMethod() (update, for instance)

Yes in case of executeMethod() we should save the error regardless of the response codes.

api.go
```diff
@@ -528,6 +532,11 @@ func (c Client) executeMethod(method string, metadata requestMetadata) (res *htt
 		// For all other cases break out of the retry loop.
 		break
 	}
+
+	if retryCount >= MaxRetry {
+		err = errors.New("Max retry reached")
```
dtjm: I think it's better to declare the error once at the package level, and return that instead.  

On the other hand, in this case, would you want to return the last error received?  I was thinking if you just return "Max retry reached", the caller won't have any idea what the underlying error was.

harshavardhana: Yes that is what we need here, we should only return the last error after all the retries.

ericychoi:
Created #555 as an alternative approach.

harshavardhana:
You can close this @ericychoi

harshavardhana:
This approach is no longer needed.

PR #555, opened 2016-11-22
fix max retry issue in RemoveObject()

ericychoi:


harshavardhana:
Thanks @ericychoi

harshavardhana:
Oops i see you only changed this for RemoveBucket @ericychoi - let me fix this properly.

PR #556, opened 2016-11-27
list: Don't send incomplete uploads info with err

vadmeste:
Incomplete Uploads List API continues to send object information even
when an error is encountered. This commit fixes this behavior with
sending only err in the incomplete upload list.

### DIFF
--- a/api-list.go
+++ b/api-list.go
@@ -482,6 +482,7 @@ func (c Client) listIncompleteUploads(bucketName, objectPrefix string, recursive
 						objectMultipartStatCh <- ObjectMultipartInfo{
 							Err: err,
 						}
+						continue
 					}
 				}
 				select {


harshavardhana:
Commit message is incorrect here we are still sending err on object channel ?

vadmeste:
> Commit message is incorrect here we are still sending err on object channel ?

yes.. I meant avoid sending object info when error is encountered, good enough?

harshavardhana:
> yes.. I meant avoid sending object info when error is encountered, good enough?

Yes you need to update the commit message. Currently it says something else.

harshavardhana:
Also @vadmeste can you format the commit message so that the commit line is of a set length and wrapping into next line.  Perhaps a good measurement is to 50 characters at max on the top line and explain the rest in the body.

vadmeste:
> Also @vadmeste can you format the commit message so that the commit line is of a set length and wrapping into next line. Perhaps a good measurement is to 50 characters at max on the top line and explain the rest in the body.

Done. I saw that VIM gives that 50 columns hint when writing the commit message, however, I wonder how it will be always possible to write a commit title with 50 characters only.

harshavardhana:
> Done. I saw that VIM gives that 50 columns hint when writing the commit message, however, I wonder how it will be always possible to write a commit title with 50 characters only.

It should be title should be precise and small. To explain the rest move to the body.

PR #559, opened 2016-12-06
close res.Body

Kane-Sendgrid:


harshavardhana:
res.Body is closed by the caller of executeMethod we shouldn't close inside executeMethod. Look at closeResponse wrapper for this.

Kane-Sendgrid:
Ah, I see now, it's closed way up the call stack. Closing it.

Issue #560, opened 2016-12-06
potential resource leak because res.Body is not closed


Kane-Sendgrid:
https://github.com/minio/minio-go/pull/559

harshavardhana:
@Kane-Sendgrid is there an issue still?

Kane-Sendgrid:
@harshavardhana just was checking this line - https://github.com/minio/minio-go/blob/master/api.go#L503, res.Body gets overwritten with NopCloser, what happens to original res.Body?

harshavardhana:
> @harshavardhana just was checking this line - https://github.com/minio/minio-go/blob/master/api.go#L503, res.Body gets overwritten with NopCloser, what happens to original res.Body?

You are right looks like a bug.

harshavardhana:
Fixed by #561

PR #561, opened 2016-12-08
api: res.Body should be closed properly

Kane-Sendgrid:
#560


api.go
```diff
@@ -495,6 +495,8 @@ func (c Client) executeMethod(method string, metadata requestMetadata) (res *htt
 
 		// Read the body to be saved later.
 		errBodyBytes, err := ioutil.ReadAll(res.Body)
+		// res.Body should be closed
+		res.Body.Close()
```
harshavardhana: you can use "closeResponse()" itself. there is a need to do that in Go http client.

Kane-Sendgrid: I guess won't hurt, changing

PR #562, opened 2016-12-09
Move s3 signature v4 and v2 to a separate pkg

vadmeste:
Avoid code duplication by moving code related to signture v2 and v4 in a different location

### DIFF
--- a/api-presigned.go
+++ b/api-presigned.go
@@ -20,6 +20,8 @@ import (
 	"errors"
 	"net/url"
 	"time"
+
+	"github.com/minio/minio-go/pkg/s3signer"
 )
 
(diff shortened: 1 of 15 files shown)


api-presigned.go
```diff
@@ -32,6 +35,17 @@ var supportedGetReqParams = map[string]struct{}{
 	"response-content-disposition": {},
 }
 
+// getCredential - generates Credential field for post policy
+func getCredential(accessKeyID, location string, t time.Time) string {
```
harshavardhana: Move this to signer package as well `s3signer.GetCredential`

vadmeste: > Move this to signer package as well s3signer.GetCredential

Done

PR #568, opened 2016-12-13
api: GetObject, StatObject should return more metadata.

harshavardhana:


krishnasrinivas:
Was this requested by any user? What header fields apart from the ObjectInfo fields will be useful?

harshavardhana:
> Was this requested by any user? What header fields apart from the ObjectInfo fields will be useful?

Yes this is needed.. for `mc mirror` to copy metadata to PutObject().

krishnasrinivas:
> Yes this is needed.. for mc mirror to copy metadata to PutObject().

Ah, in that case we should resp.Header.Del() all the http header values like
Transfer-Encoding
Accept-Ranges
Date
Server
Vary
X-Amz-Request-Id

so that Metadata contains only Metadata for that object.

harshavardhana:
> Ah, in that case we should resp.Header.Del() all the http header values like
> Transfer-Encoding
> Accept-Ranges
> Date
> Server
> Vary
> X-Amz-Request-Id
> 
> so that Metadata contains only Metadata for that object.

Ah yes you are right we need to get rid of these ones..

harshavardhana:
Done

api-stat.go
```diff
@@ -49,6 +49,30 @@ func (c Client) BucketExists(bucketName string) (bool, error) {
...
+// and removed from the response header map. Before
+// using them as metadata for the object.
+func cleanRespHeaders(headers http.Header, customIgnoreHeaders []string) {
+	for _, h := range customIgnoreHeaders {
+		headers.Del(h)
+	}
+	for _, h := range ignoreHeaders {
+		headers.Del(h)
+	}
+}
```
krishnasrinivas: better will be:
```
func extractMetadata(h http.Header) (metadata http.Header) {                                                                                                                      
        ignoreHeaders := map[string]bool {                                                                                                                                        
                "transfer-encoding": true,                                                                                                                                        
                "accept-ranges": true,                                                                                                                                            
                "date": true,                                                                                                                                                     
                ...                                                                                                                                                               
                ...                                                                                                                                                               
        }                                                                                                                                                                         
        for k, vv := range h {                                                                                                                                                    
                if !ignoreHeaders[k] {                                                                                                                                            
                        vv2 := make([]string, len(vv))                                                                                                                            
                        copy(vv2, vv)                                                                                                                                             
                        metadata[k] = vv2                                                                                                                                         
                }                                                                                                                                                                 
        }                                                                                                                                                                         
        return metadata                                                                                                                                                           
}
```

i.e the right function name and no mutation of resp.Header
also customIgnoreHeaders can be part of ignoreHeaders as it does not look like there will be any use for it keeping separate. If there is use for it in future then the function can be refactored.

api-stat.go
```diff
@@ -49,6 +49,31 @@ func (c Client) BucketExists(bucketName string) (bool, error) {
...
 
+// List of header keys to be filtered, usually
+// from all S3 API http responses.
+var defaultFilterKeys = []string{
+	"Transfer-Encoding",
+	"Accept-Ranges",
+	"Date",
+	"Server",
+	"Vary",
+	"x-amz-request-id",
```
krishnasrinivas: looks like minio server sends `X-Amz-Request-Id`

```
mc: <DEBUG> HTTP/1.1 200 OK
Content-Length: 2248
Accept-Ranges: bytes
Date: Tue, 13 Dec 2016 21:35:29 GMT
Etag: "1ff633d084da03ff72793c7a93787fca"
Last-Modified: Tue, 13 Dec 2016 21:31:20 GMT
Server: Minio/DEVELOPMENT.GOGET (linux; amd64)
Vary: Origin
X-Amz-Request-Id: O1BCY6HBCDTU0MY0
```

don't know about x-amz-id-2

it will be safer to make it canonical somewhere

harshavardhana: let me use the helper method .Del() instead.  which will canonicalize this properly.

PR #569, opened 2016-12-13
client: Consolidate parsing URLs.

### LINKED ISSUE #530: Parse endpoint URL once and save it in *minio.Client
Hi:
    I am just learning yours project
    I am confused with api-put-bucket.go  MakeBucket() call isValidBucketName() to validate the input arguments and call makeBucketRequest() to make a httprequest

  but come into makeBucketRequest()  again call isValidBucketName()  here is repeat ??
  and in makeBucketRequest()
  targetURL, err := url.Parse(c.endpointURL)   the client is init from the New  here to parse it why ??

harshavardhana:
This patch consolidates url parsing and parses endpoint once. 

Fixes #530

### DIFF
--- a/api.go
+++ b/api.go
@@ -41,6 +41,9 @@ import (
 type Client struct {
 	///  Standard options.
 
+	// Parsed endpoint url provided by the user.
+	endpointURL url.URL
+
 	// AccessKeyID required for authorized requests.
 	accessKeyID string
 	// SecretAccessKey required for authorized requests.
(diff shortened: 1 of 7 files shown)


api.go
```diff
@@ -41,6 +41,9 @@ import (
 type Client struct {
 	///  Standard options.
 
+	// Parsed endpoint url provided by the user.
+	endpointURL *url.URL
```
krishnasrinivas: better:
`endpointURL url.URL`
as this is never supposed to be mutated.

PR #571, opened 2016-12-15
Compile regexp only once and use it as many times wanted.

balamurugana:


pkg/s3signer/utils.go
```diff
@@ -77,6 +77,9 @@ func queryEncode(v url.Values) string {
 	return buf.String()
 }
 
+// if object matches reserved string, no need to encode them
+var ReservedObjectNames = regexp.MustCompile("^[a-zA-Z0-9-_.~/]+$")
```
harshavardhana: no need to export this..

pkg/s3signer/utils.go
```diff
@@ -85,9 +88,7 @@ func queryEncode(v url.Values) string {
 // This function on the other hand is a direct replacement for url.Encode() technique to support
 // pretty much every UTF-8 character.
 func urlEncodePath(pathName string) string {
```
harshavardhana: In-fact this function should be exposed..

api.go
```diff
@@ -278,6 +278,9 @@ type requestMetadata struct {
 	contentMD5Bytes    []byte
 }
 
+var regCred = regexp.MustCompile("Credential=([A-Z0-9]+)/")
```
harshavardhana: Add comments for these regexes.

balamurugana: Done

PR #573, opened 2016-12-16
Add S3 Acceleration endpoint via "accelerate" bucket

stanhu:
There is currently no way use [Amazon S3 Acceleration](http://docs.aws.amazon.com/AmazonS3/latest/dev/transfer-acceleration.html) without Minio defaulting to s3.amazonaws.com. This would fix https://gitlab.com/gitlab-org/omnibus-gitlab/issues/1804 for GitLab.


stanhu:
Actually, this isn't the right change. We need `bucket.s3-accelerate.amazonaws.com` and change related utilities.

PR #577, opened 2016-12-17
Fix for looking up bucket locations in S3 Acceleration endpoints

### LINKED ISSUE #576: Hard to configure bucket location for Amazon S3 Acceleration
Amazon S3 Acceleration buckets need to have the proper location (e.g. `us-west-1`), but currently the Minio Golang library  tries to figure out based on a lookup from the bucket name. However, in the S3 Acceleration case, the location lookup to `bucketname.s3-accelerate.amazonaws.com` results in a 400 Bad Request, which causes the default to be set to `us-east-1`.

One way to improve this is to allow the client to set a default bucket location for the client. Thoughts?

stanhu:
Attempting to lookup a bucket via the s3-accelerate.amazonaws.com host
returns an Error 400. Fall back to looking up with the us-east-1 endpoint.

Closes #576

### DIFF
--- a/bucket-cache.go
+++ b/bucket-cache.go
@@ -163,6 +163,13 @@ func (c Client) getBucketLocationRequest(bucketName string) (*http.Request, erro
 
 	// Set get bucket location always as path style.
 	targetURL := c.endpointURL
+
+	// Requesting a bucket location from an accelerate endpoint returns a 400,
+	// so default to us-east-1 for the lookup
+	if isAmazonS3AccelerateEndpoint(c.endpointURL) {
+		targetURL.Host = getS3Endpoint("us-east-1")
+	}
+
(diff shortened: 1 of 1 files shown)


harshavardhana:
LGTM.

stanhu:
Thanks! Could you release a new version soon?

Ehekatl:
This only work if the accelerate bucket is located in us-east-1.
Can we make minio respect aws environment variables? e.g. AWS_REGION

harshavardhana:
> This only work if the accelerate bucket is located in us-east-1.
> Can we make minio respect aws environment variables? e.g. AWS_REGION

There is no reason we need to do that you can initialize with the correct region yourself

```
minio.NewWithRegion(...., os.Getenv("AWS_REGION"))
```

PR #579, opened 2016-12-19
Move functions related to s3 urls to a new pkg

vadmeste:
For review and comments


api-notification.go
```diff
@@ -23,6 +23,8 @@ import (
 	"net/http"
 	"net/url"
 	"time"
+
+	urlutils "github.com/minio/minio-go/pkg/s3url_utils"
```
harshavardhana: don't use '_' in package name.  Just use `s3utils`.

vadmeste: > don't use '_' in package name. Just use s3utils.

Done

PR #582, opened 2016-12-21
Use consistent jargon while describing API

krisis:
Also, changed minPartSize reference to 64MiB everywhere including docs.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -9,9 +9,9 @@
 package main
 
 import (
-	"fmt"
+    "fmt"
 
-	"github.com/minio/minio-go"
+    "github.com/minio/minio-go"
 )
 
 func main() {
@@ -22,7 +22,7 @@ func main() {
         minioClient, err := minio.New("play.minio.io:9000", "Q3AM3UQ867SPQQA43P2F", "zuf+tfteSlswRu7BJ86wekitnifILbZam1KYY3TG", ssl)
         if err != nil {
                 fmt.Println(err)
-          	    return
+                return
         }
 }
 
@@ -35,9 +35,9 @@ func main() {
 package main
 
 import (
-	"fmt"
+    "fmt"
 
-	"github.com/minio/minio-go"
+    "github.com/minio/minio-go"
 )
 
 func main() {
@@ -48,7 +48,7 @@ func main() {
         s3Client, err := minio.New("s3.amazonaws.com", "YOUR-ACCESSKEYID", "YOUR-SECRETACCESSKEY", ssl)
         if err != nil {
                 fmt.Println(err)
-                return            
+                return
         }
 }
 
@@ -77,10 +77,10 @@ __Parameters__
 
 |Param   |Type   |Description   |
 |:---|:---| :---|
-|`endpoint`   | _string_  |S3 object storage endpoint.   |
-| `accessKeyID`  |_string_   | Access key for the object storage endpoint.  |
-| `secretAccessKey`  | _string_  |Secret key for the object storage endpoint.   |
-|`ssl`   | _bool_  | Set this value to 'true' to enable secure (HTTPS) access.  |
+|`endpoint`   | _string_  |S3 compatible object storage endpoint   |
+|`accessKeyID`  |_string_   |Access key for the object storage |
+|`secretAccessKey`  | _string_  |Secret key for the object storage |
+|`ssl`   | _bool_  | If 'true' API requests will be secure (HTTPS), and insecure (HTTP) otherwise  |
 
 
 ## 2. Bucket operations
@@ -94,8 +94,8 @@ __Parameters__
 
 | Param  | Type  | Description  |
 |---|---|---|
-|`bucketName`  | _string_  | Name of the bucket. | 
-| `location`  |  _string_ | Default value is us-east-1 Region where the bucket is created. Valid values are listed below:|
+|`bucketName`  | _string_  | Name of the bucket |
+| `location`  |  _string_ | Region where the bucket is to be created. Default value is us-east-1. Other valid values are listed below. |
 | | |us-east-1 |
 | | |us-west-1 |
 | | |us-west-2 |
@@ -104,7 +104,7 @@ __Parameters__
 | | | ap-southeast-1|
 | | | ap-northeast-1|
 | | | ap-southeast-2|
-| | | sa-east-1| 
+| | | sa-east-1|
 
 
 __Example__
@@ -128,30 +128,30 @@ Lists all buckets.
 
 | Param  | Type  | Description  |
 |---|---|---|
-|`bucketList`  | _[]BucketInfo_  | Lists bucket in following format shown below: | 
+|`bucketList`  | _[]BucketInfo_  | Lists of all buckets |
 
 
 | Param  | Type  | Description  |
 |---|---|---|
-|`bucket.Name`  | _string_  | bucket name. | 
-|`bucket.CreationDate`  | _time.Time_  | date when bucket was created. | 
+|`bucket.Name`  | _string_  | Name of the bucket |
+|`bucket.CreationDate`  | _time.Time_  | Date of bucket creation |
 
 
- __Example__
+__Example__
 
- 
- ```go
 
- buckets, err := minioClient.ListBuckets()
-if err != nil {
+```go
+
+buckets, err := minioClient.ListBuckets()
+    if err != nil {
     fmt.Println(err)
     return
 }
 for _, bucket := range buckets {
-  fmt.Println(bucket)
-}  
+    fmt.Println(bucket)
+}
 
- ```
+```
 
 <a name="BucketExists"></a>
 ### BucketExists(bucketName string) (found bool, err error)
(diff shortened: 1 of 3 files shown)


docs/API.md
```diff
@@ -579,28 +579,28 @@ __Parameters__
...
 
 
 |Param   |Type   |Description   |
 |:---|:---| :---|
-|`objInfo.LastModified`  | _time.Time_  |modified time stamp. |
-|`objInfo.ETag` | _string_ |etag of the object.|
-|`objInfo.ContentType` | _string_ |Content-Type of the object.|
-|`objInfo.Size` | _int64_ |size of the object.|
+|`objInfo.LastModified`  | _time.Time_  |Time when object was last modified |
+|`objInfo.ETag` | _string_ |Etag of the object|
```
harshavardhana: MD5 checksum of the object

docs/API.md
```diff
@@ -639,22 +639,22 @@ if err != nil {
...
+|`bucketName`  | _string_  |Name of the bucket  |
+|`objectsCh` | _chan string_  | Prefix of objects to be removed   |
 
 
 __Return Values__
 
 |Param   |Type   |Description   |
 |:---|:---| :---|
-|`errorCh` | _chan minio.RemoveObjectError  | read objects deletion errors  |
+|`errorCh` | _chan minio.RemoveObjectError  | Channel of errors observed during deletion.  |
```
harshavardhana: Why not "Errors observed during deletion" ?.  Channel might not be necessary.

harshavardhana: No need to have "." at the end since we are removing that anyways

docs/API.md
```diff
@@ -708,10 +708,10 @@ __Parameters__
...
 |Param   |Type   |Description   |
 |:---|:---| :---|
-|`bucketName`  | _string_  |name of the bucket.   |
-|`objectName` | _string_  |name of the object.   |
-|`expiry` | _time.Duration_  |expiry in seconds.   |
-|`reqParams` | _url.Values_  |additional response header overrides supports _response-expires_, _response-content-type_, _response-cache-control_, _response-content-disposition_.  |
+|`bucketName`  | _string_  |Name of the bucket   |
+|`objectName` | _string_  |Name of the object   |
+|`expiry` | _time.Duration_  |Expiry of presigned URL in seconds   |
+|`reqParams` | _url.Values_  |Additional response header overrides supports _response-expires_, _response-content-type_, _response-cache-control_, _response-content-disposition_.  |
```
harshavardhana: Might be missing a comma here "Additional response header overrides, supports ... "

docs/API.md
```diff
@@ -1006,36 +1007,36 @@ bucketNotification := BucketNotification{}
...
-	fmt.Println("Cannot set the bucket notification: " + err)
+    fmt.Println("Cannot set the bucket notification: " + err)
 }
 ```
 
 <a name="RemoveAllBucketNotification"></a>
 ### RemoveAllBucketNotification(bucketName string) error
 
-Remove all configured bucket notifications on a bucket.
+Remove all configured notifications on a bucket.
```
harshavardhana: "Remove all configured bucket notifications" is correct..  "notifications" what is sent by the server.

docs/API.md
```diff
@@ -1006,36 +1007,36 @@ bucketNotification := BucketNotification{}
...
+|`err` | _error_  |Standard Error  |
 
 __Example__
 
 
 ```go
 err := c.RemoveAllBucketNotification(bucketName)
 if err != nil {
-	fmt.Println("Cannot remove bucket notifications.")
+    fmt.Println("Cannot remove bucket notifications.", err)
```
harshavardhana: "Unable to remove bucket notifications."

docs/API.md
```diff
@@ -77,10 +77,10 @@ __Parameters__
 
 |Param   |Type   |Description   |
 |:---|:---| :---|
-|`endpoint`   | _string_  |S3 object storage endpoint.   |
-| `accessKeyID`  |_string_   | Access key for the object storage endpoint.  |
-| `secretAccessKey`  | _string_  |Secret key for the object storage endpoint.   |
-|`ssl`   | _bool_  | Set this value to 'true' to enable secure (HTTPS) access.  |
+|`endpoint`   | _string_  |Object storage endpoint   |
```
harshavardhana: "S3 compatible object storage endpoint"

docs/API.md
```diff
@@ -77,10 +77,10 @@ __Parameters__
...
+|`accessKeyID`  |_string_   |Access key for the object storage |
+|`secretAccessKey`  | _string_  |Secret key for the object storage |
+|`ssl`   | _bool_  | If 'true' API requests will be over HTTPS, and HTTP otherwise  |
```
harshavardhana: "If 'true' API requests will be secure (HTTPS), and insecure (HTTP) otherwise"

docs/API.md
```diff
@@ -128,30 +128,30 @@ Lists all buckets.
 
 | Param  | Type  | Description  |
 |---|---|---|
-|`bucketList`  | _[]BucketInfo_  | Lists bucket in following format shown below: | 
+|`bucketList`  | _[]BucketInfo_  | Lists buckets |
```
harshavardhana: "Lists buckets" means this is doing something. In-fact this is return value so it should be "List of all buckets"

docs/API.md
```diff
@@ -163,15 +163,15 @@ __Parameters__
...
 
 
 __Return Values__
 
 |Param   |Type   |Description   |
 |:---|:---| :---|
-|`found`  | _bool_ | indicates whether bucket exists or not  |
-|`err` | _error_  | standard error  |
+|`found`  | _bool_ | Indicates whether bucket exists or not  |
+|`err` | _error_  | Standard error  |
```
harshavardhana: "Standard Error" consistent with below changes.. or change the below changes to "Standard error"

docs/API.md
```diff
@@ -269,15 +269,15 @@ for object := range objectCh {
...
 
 __Parameters__
 
 
 |Param   |Type   |Description   |
 |:---|:---| :---|
-|`bucketName`  | _string_  |name of the bucket.   |
-| `objectPrefix` |_string_   | the prefix of the objects that should be listed. |
+|`bucketName`  | _string_  |Name of the bucket |
+| `objectPrefix` |_string_   | Prefix of objects to listed |
```
harshavardhana: "Prefix of objects to be listed"

docs/API.md
```diff
@@ -456,16 +456,16 @@ __Parameters__
 
 |Param   |Type   |Description   |
 |:---|:---| :---|
-|`bucketName`  | _string_  |name of the bucket.   |
-|`objectName` | _string_  |name of the object.   |
-|`reader` | _io.Reader_  |Any golang object implementing io.Reader.   |
+|`bucketName`  | _string_  |Name of the bucket  |
+|`objectName` | _string_  |Name of the object   |
+|`reader` | _io.Reader_  |Any Go type that implements io.Reader interface |
```
harshavardhana: "Any Go type that implements io.Reader" - interface is not needed.

docs/API.md
```diff
@@ -537,24 +537,24 @@ if err != nil {
...
 
 |Param   |Type   |Description   |
 |:---|:---| :---|
-|`bucketName`  | _string_  |name of the bucket.   |
-|`objectName` | _string_  |name of the object.   |
-|`filePath` | _string_  |file path of the file to be uploaded. |
-|`contentType` | _string_  |content type of the object.  |
+|`bucketName`  | _string_  |Name of the bucket  |
+|`objectName` | _string_  |Name of the object |
+|`filePath` | _string_  |path to file to be uploaded |
```
harshavardhana: "path" is not capitalized.

docs/API.md
```diff
@@ -579,28 +579,28 @@ __Parameters__
...
 
 |Param   |Type   |Description   |
 |:---|:---| :---|
-|`objInfo.LastModified`  | _time.Time_  |modified time stamp. |
-|`objInfo.ETag` | _string_ |etag of the object.|
-|`objInfo.ContentType` | _string_ |Content-Type of the object.|
-|`objInfo.Size` | _int64_ |size of the object.|
+|`objInfo.LastModified`  | _time.Time_  |Time when object was last modified |
+|`objInfo.ETag` | _string_ |Etag of the object|
+|`objInfo.ContentType` | _string_ |Content-Type of the object|
```
harshavardhana: "Content type of the object."  - we don't need to write the http.Header canonical name here. Just like how we don't say "Content-Length of the object" instead we say "Size of the object".

docs/API.md
harshavardhana: Actually Channel is fine, since we are using it everywhere.

docs/API.md
krisis: No, there is no need for a comma there.

docs/API.md
krisis: Done.

docs/API.md
krisis: Done.

docs/API.md
krisis: Done.

docs/API.md
krisis: Done.

docs/API.md
krisis: Done.

docs/API.md
krisis: Done.

PR #591, opened 2017-01-06
stringset: Add new method ToSlice()

balamurugana:
stringset.ToSlice() returns string slice of the set.


harshavardhana:
@balamurugana any particular reason to expose this?

balamurugana:
> @balamurugana any particular reason to expose this?

Yes. I am using it to identify duplicate endpoints.

PR #609, opened 2017-02-09
pkg/s3signer: Add signatureV4 chunked upload support.

krisis:
This PR is complete w.r.t chunked upload signatureV4 signing. Feedback welcome on API and choice of chunk size etc.

### DIFF
--- a/api-put-object-progress.go
+++ b/api-put-object-progress.go
@@ -121,3 +121,51 @@ func (c Client) PutObjectWithMetadata(bucketName, objectName string, reader io.R
 	}
 	return n, nil
 }
+
+// PutObjectStreaming using AWS streaming signature V4
+func (c Client) PutObjectStreaming(bucketName, objectName string, reader io.Reader, size int64) (n int64, err error) {
+	if size < 0 {
+		return 0, ErrInvalidArgument("Size can't be negative.")
+	}
+
+	// Check for largest object size allowed.
+	if size > int64(maxMultipartPutObjectSize) {
+		return 0, ErrEntityTooLarge(size, maxMultipartPutObjectSize, bucketName, objectName)
+	}
+
+	// Note: Streaming signature is not supported by GCS.
+	if s3utils.IsGoogleEndpoint(c.endpointURL) {
+		return 0, ErrorResponse{
+			Code:       "NotImplemented",
+			Message:    "AWS streaming signature v4 is not supported with Google Cloud Storage",
+			Key:        objectName,
+			BucketName: bucketName,
+		}
+	}
+
+	// Set signature type to streaming signature v4.
+	c.signature = ChunkedV4
+
+	if size < minPartSize && size >= 0 {
+		return c.putObjectNoChecksum(bucketName, objectName, reader, size, nil, nil)
+	}
+
+	// For all sizes greater than 5MiB do multipart.
+	n, err = c.putObjectMultipartStreamNoChecksum(bucketName, objectName, reader, size, nil, nil)
+	if err != nil {
+		errResp := ToErrorResponse(err)
+		// Verify if multipart functionality is not available, if not
+		// fall back to single PutObject operation.
+		if errResp.Code == "AccessDenied" && strings.Contains(errResp.Message, "Access Denied") {
+			// Verify if size of reader is greater than '5GiB'.
+			if size > maxSinglePutObjectSize {
+				return 0, ErrEntityTooLarge(size, maxSinglePutObjectSize, bucketName, objectName)
+			}
+			// Fall back to uploading as single PutObject operation.
+			return c.putObjectNoChecksum(bucketName, objectName, reader, size, nil, nil)
+		}
+		return n, err
+	}
+
+	return n, nil
+}
--- a/api_functional_v4_test.go
+++ b/api_functional_v4_test.go
@@ -354,6 +354,68 @@ func TestPutObjectWithMetadata(t *testing.T) {
 	}
 }
 
+// Test put object with streaming signature.
+func TestPutObjectStreaming(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping function tests for short runs")
+	}
+
+	// Seed random based on current time.
+	rand.Seed(time.Now().Unix())
+
+	// Instantiate new minio client object.
+	c, err := NewWithRegion(
+		os.Getenv("S3_ADDRESS"),
+		os.Getenv("ACCESS_KEY"),
+		os.Getenv("SECRET_KEY"),
+		mustParseBool(os.Getenv("S3_SECURE")),
+		"us-east-1",
+	)
+	if err != nil {
+		t.Fatal("Error:", err)
+	}
+
+	// Set user agent.
+	c.SetAppInfo("Minio-go-FunctionalTest", "0.1.0")
+
+	// Generate a new random bucket name.
+	bucketName := randString(60, rand.NewSource(time.Now().UnixNano()),
+		"minio-go-test")
+
+	// Make a new bucket.
+	err = c.MakeBucket(bucketName, "us-east-1")
+	if err != nil {
+		t.Fatal("Error:", err, bucketName)
+	}
+
+	// Upload an object.
+	sizes := []int64{0, 64*1024 - 1, 64 * 1024}
+	objectName := "test-object"
+	for i, size := range sizes {
+		data := bytes.Repeat([]byte("a"), int(size))
+		n, err := c.PutObjectStreaming(bucketName, objectName, bytes.NewReader(data), size)
+		if err != nil {
+			t.Fatalf("Test %d Error: %v %s %s", i+1, err, bucketName, objectName)
+		}
+
+		if n != size {
+			t.Errorf("Test %d Expected upload object size %d but got %d", i+1, size, n)
+		}
+	}
+
+	// Remove the object.
+	err = c.RemoveObject(bucketName, objectName)
+	if err != nil {
+		t.Fatal("Error:", err)
+	}
+
+	// Remove the bucket.
+	err = c.RemoveBucket(bucketName)
+	if err != nil {
+		t.Fatal("Error:", err)
+	}
+}
+
 // Test listing partially uploaded objects.
 func TestListPartiallyUploaded(t *testing.T) {
 	if testing.Short() {
--- a/pkg/s3signer/request-signature-streaming.go
+++ b/pkg/s3signer/request-signature-streaming.go
@@ -1,0 +1,292 @@
+/*
+ * Minio Go Library for Amazon S3 Compatible Cloud Storage (C) 2017 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package s3signer
+
+import (
+	"bytes"
+	"encoding/hex"
+	"fmt"
+	"io"
+	"net/http"
+	"strconv"
+	"strings"
+	"time"
+)
+
+// Reference for constants used below -
+// http://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html#example-signature-calculations-streaming
+const (
+	streamingSignAlgorithm = "STREAMING-AWS4-HMAC-SHA256-PAYLOAD"
+	streamingEncoding      = "aws-chunked"
+	streamingPayloadHdr    = "AWS4-HMAC-SHA256-PAYLOAD"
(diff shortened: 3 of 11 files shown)


pkg/s3signer/request-signature-v4.go
```diff
@@ -70,7 +70,7 @@ const (
 ///
 ///      Is skipped for obvious reasons
 ///
-var ignoredHeaders = map[string]bool{
+var v4_ignoredHeaders = map[string]bool{
```
vadmeste: This variable name needs to be changed to `v4IgnoredHeaders`

pkg/s3signer/request-signature-streaming.go
```diff
@@ -0,0 +1,242 @@
...
+
+func (s *StreamingReader) Read(buf []byte) (int, error) {
+	if s.done {
+		// return what is left from last read from baseReader.
+		// s.buf will return io.EOF once the buffer is drained.
+		return s.buf.Read(buf)
+	}
+
+	n0, err := s.buf.Read(buf)
+	if err == io.EOF {
```
vadmeste: Can you do `if err != io.EOF { return n0, err } ` ? more easier to analyze the logic of the code

vadmeste: > Can you do if err != io.EOF { return n0, err } ? more easier to analyze the logic of the code

No need, I understand why it is done this way.

vadmeste: s.buf.Read() is called in different places in this function. I think this can be simplified this way: the type of s.buf is bytes.Buffer so we can know the length of data inside. What we need is to fill s.buf whenever len(s.buf) < len(buf) and then stream the content of s.buf at the end of this function, do you think this doable @krisis ?

krisis: @vadmeste `s.buf` always one signed chunk. It is refilled only when it is empty. This keeps the number of different states `s.buf` 's contents could be in. Let me think about it.

pkg/s3signer/request-signature-v4.go
harshavardhana: golint complains about "_"

harshavardhana:
@krisis - travis CI has failed.

pkg/s3signer/request-signature-streaming.go
krisis: @vadmeste I have simplified how `s.buf` is re-filled as per your suggestion. Let me know your thoughts.

pkg/s3signer/request-signature-streaming.go
```diff
@@ -0,0 +1,260 @@
...
+var ignoredStreamingHeaders = map[string]bool{
+	"Authorization": true,
+	"User-Agent":    true,
+	"Content-Type":  true,
+}
+
+// getSignedChunkLength - calculates the length of chunk metadata
+func getSignedChunkLength(chunkDataSize int64) int64 {
+	return int64(len(fmt.Sprintf("%x", chunkDataSize))) +
+		17 + // ";chunk-signature="
```
harshavardhana: Can we have global constants for this? and also cite references where we got this defaults from and why they shouldn't be changed.

pkg/s3signer/request-signature-streaming.go
```diff
@@ -0,0 +1,260 @@
...
+	req.Header.Set("X-Amz-Content-Sha256", streamingSignAlgorithm)
+	req.Header.Set("Content-Encoding", streamingEncoding)
+	req.Header.Set("X-Amz-Date", timestamp.Format(iso8601DateFormat))
+
+	streamContentLen := getStreamLength(dataLen, int64(payloadChunkSize))
+	// Set content length with streaming signature for each chunk included.
+	req.Header.Set("Content-Length", strconv.FormatInt(streamContentLen, 10))
+	req.Header.Set("x-amz-decoded-content-length", strconv.FormatInt(dataLen, 10))
+
+	return &req, dataLen, nil
```
harshavardhana: why is `dataLen` again returned back since caller already has `dataLen`  - how is it useful?

pkg/s3signer/request-signature-streaming.go
```diff
@@ -0,0 +1,260 @@
...
+}
+
+// getChunkHeader - returns the
```
harshavardhana: 'returns the?'

pkg/s3signer/request-signature-streaming.go
```diff
@@ -0,0 +1,260 @@
...
+	// Calculate signature.
+	s.seedSignature = getSignature(signingKey, stringToSign)
+}
+
+// StreamingReader implements chunked upload signature as a reader on
+// top of req.Body's ReaderCloser chunk header;data;... repeat
+type StreamingReader struct {
+	accessKeyID     string
+	secretAccessKey string
+	location        string
```
harshavardhana: you can keep location as `region` - S3 has blurred the differences but for signature lets call it region for readability.

pkg/s3signer/request-signature-streaming.go
```diff
@@ -0,0 +1,260 @@
...
+	prevSignature   string
+	seedSignature   string
+	contentLen      int64
```
harshavardhana: s/contentLen/origContentLength

pkg/s3signer/request-signature-streaming.go
```diff
@@ -0,0 +1,260 @@
...
+		previousSig,
+		emptySHA256,
+		hex.EncodeToString(sum256(chunkData)),
+	}
+
+	return strings.Join(stringToSignParts, "\n")
+}
+
+// PrepareStreamingRequest - prepares a request with appropriate headers before computing the seed signature.
+func PrepareStreamingRequest(req http.Request, dataLen int64, timestamp time.Time) (*http.Request, int64, error) {
```
harshavardhana: why have this function? just take all the inputs directly in NewStreamingRequest? avoids doing `strconv` again back for dataLen from `Content-Length` header.

pkg/s3signer/request-signature-streaming.go
krisis: OK

pkg/s3signer/request-signature-streaming.go
krisis: OK

pkg/s3signer/request-signature-streaming.go
krisis: It was useful at some point in time. It is no longer useful. I will remove this in the next update.

pkg/s3signer/request-signature-streaming.go
krisis: `StreamingReader` type doesn't involve itself with the request headers. It implements a `io.ReadCloser` to build signed payload as required. This is why `NewStreamingReader` stays clear of setting request headers.

 `PrepareStreamingRequest` is meant to be a helper function for users of streaming signature for signing payload.

pkg/s3signer/request-signature-streaming.go
```diff
@@ -0,0 +1,267 @@
...
+		hex.EncodeToString(sum256(chunkData)),
+	}
+
+	return strings.Join(stringToSignParts, "\n")
+}
+
+// PrepareStreamingRequest - prepares a request with appropriate
+// headers before computing the seed signature.
+func PrepareStreamingRequest(req *http.Request, dataLen int64, timestamp time.Time) {
+	// Set x-amz-content-sha256 header.
```
harshavardhana: Since we don't modify request we can just return the set headers instead and set the header outside where the request is useful, this function can avoid messing with request header itself directly.

harshavardhana: But i am not able to understand why do you need a separate function for this?  - from what i can see from tests - PrepareStreamingRequest can be merged into NewStreamReader.

harshavardhana:
Any updates on my comments @krisis ?

pkg/s3signer/request-signature-streaming.go
krisis: StreamingReader implements chunk upload signing as a Reader wrapping req.Body. Adding those headers is a pre-requisite for chunk upload signing process but not part of the signing itself. So, preparing the request for chunk upload signing is intentionally separated from computing the signature (as a Reader). It still made sense to provide PrepareStreamingRequest as a helper.

harshavardhana:
is this ready for review again? @krisis

krisis:
@harshavardhana no, this implementation is not memory-efficient. It has been a long time since I have looked into it. I shall update with specifics of the problem.

harshavardhana:
> @harshavardhana no, this implementation is not memory-efficient. It has been a long time since I have looked into it. I shall update with specifics of the problem.

Can you update this PR what kind of problems you are facing with this? @krisis - are you planning to work on this further?

krisis:
>Can you update this PR what kind of problems you are facing with this? @krisis - are you planning to work on this further?
@harshavardhana I have rebased the PR and added comments explaining the space complexity. If the space complexity seems fine, we could merge it after review and improve it in subsequent releases.

harshavardhana:
There is a conflict in this PR can you fix @krisis

krisis:
@harshavardhana @vadmeste this PR is ready to be reviewed.

api-put-object-progress.go
```diff
@@ -121,3 +121,51 @@ func (c Client) PutObjectWithMetadata(bucketName, objectName string, reader io.R
...
+		errResp := ToErrorResponse(err)
+		// Verify if multipart functionality is not available, if not
+		// fall back to single PutObject operation.
+		if errResp.Code == "AccessDenied" && strings.Contains(errResp.Message, "Access Denied") {
+			// Verify if size of reader is greater than '5GiB'.
+			if size > maxSinglePutObjectSize {
+				return 0, ErrEntityTooLarge(size, maxSinglePutObjectSize, bucketName, objectName)
+			}
+			// Fall back to uploading as single PutObject operation.
+			return c.putObjectSingle(bucketName, objectName, reader, size, nil, nil)
```
vadmeste: Do you mean c.putObjectNoChecksum ?

pkg/s3signer/request-signature-streaming.go
```diff
@@ -0,0 +1,285 @@
...
+	streamContentLen := getStreamLength(dataLen, int64(payloadChunkSize))
+	// Set content length with streaming signature for each chunk included.
+	req.Header.Set("Content-Length", strconv.FormatInt(streamContentLen, 10))
+	req.Header.Set("x-amz-decoded-content-length", strconv.FormatInt(dataLen, 10))
+	req.ContentLength = streamContentLen
+}
+
+// getChunkHeader - returns the chunk header.
+// e.g string(IntHexBase(chunk-size)) + ";chunk-signature=" + signature + \r\n + chunk-data + \r\n
+func getChunkHeader(chunkLen int64, signature string) []byte {
```
vadmeste: I prefer if we use build instead of get each time which constructs something.

pkg/s3signer/request-signature-streaming.go
```diff
@@ -0,0 +1,285 @@
...
+	authParts := []string{
+		signV4Algorithm + " Credential=" + credential,
+		"SignedHeaders=" + getSignedHeaders(*req, ignoredStreamingHeaders),
+		"Signature=" + s.seedSignature,
+	}
+
+	// Set authorization header.
+	auth := strings.Join(authParts, ",")
+	req.Header.Set("Authorization", auth)
+	req.Header.Set("x-amz-content-sha256", "STREAMING-AWS4-HMAC-SHA256-PAYLOAD")
```
vadmeste: This header is already set in prepareStreamingRequest

api-put-object-progress.go
```diff
@@ -121,3 +121,51 @@ func (c Client) PutObjectWithMetadata(bucketName, objectName string, reader io.R
 	}
 	return n, nil
 }
+
+// PutObjectStreaming using AWS streaming signature V4
+func (c Client) PutObjectStreaming(bucketName, objectName string, reader io.Reader, size int64) (n int64, err error) {
```
vadmeste: Can you add functional test for this function? and documentation.

api-put-object-progress.go
krisis: Yes, fixed now.

pkg/s3signer/request-signature-streaming.go
krisis: OK, renamed similar functions to start wiith `build`.

pkg/s3signer/request-signature-streaming.go
krisis: Good catch! I have fixed it now.

api-put-object-progress.go
krisis: I have added an example under `examples/s3/putobject-streaming.go` and updated `docs/API.md`

api_functional_v4_test.go
```diff
@@ -354,6 +354,66 @@ func TestPutObjectWithMetadata(t *testing.T) {
...
+
+	// Make a new bucket.
+	err = c.MakeBucket(bucketName, "us-east-1")
+	if err != nil {
+		t.Fatal("Error:", err, bucketName)
+	}
+
+	// Upload an object.
+	objectName := "test-object"
+	size := int64(64 * 1024)
```
vadmeste: One last comment, can you test with different file sizes? at least add empty file case.

krisis: @vadmeste thanks for suggesting the zero byte test case, I caught a bug in the streaming signature implementation. Now that is fixed too.

PR #610, opened 2017-02-10
trie: new package

balamurugana:
This implements a simple trie tree for minio server/tools.
This package borrows idea from
https://godoc.org/golang.org/x/text/internal/triegen.

### DIFF
--- a/pkg/trie/trie.go
+++ b/pkg/trie/trie.go
@@ -1,0 +1,113 @@
+/*
+ * Minio Go SDK for Amazon S3 Compatible Cloud Storage (C) 2014, 2015, 2016, 2017 Minio, Inc.
+ *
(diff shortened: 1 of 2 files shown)


harshavardhana:
why is this needed @balamurugana ? and why in minio-go

harshavardhana:
We shouldn't move / dump packages in minio-go if they are not going to be used directly. If its used somewhere then you can create a `pkg/trie`

Also this code looks like comes from - https://github.com/minio/minio/blob/master/cmd/trie.go we can have "pkg/trie" on the server

balamurugana:
@harshavardhana mc and minio have same copy of trie hence making it as pkg

balamurugana:
@harshavardhana got it

balamurugana:
moved to https://github.com/minio/minio/pull/3729

PR #614, opened 2017-02-17
Make sure to pass the contentBody to NewRequest appropriately.

harshavardhana:
This is needed since in go1.8 wrapping the body can cause
content-length to be not calculated properly, which would
lead to failures on the server side when uploading a zero
byte object.

### DIFF
--- a/api-put-object.go
+++ b/api-put-object.go
@@ -209,6 +209,9 @@ func (c Client) putObjectSingle(bucketName, objectName string, reader io.Reader,
 		// Initialize a new temporary buffer.
 		tmpBuffer := new(bytes.Buffer)
 		size, err = hashCopyN(hashAlgos, hashSums, tmpBuffer, reader, size)
+		if err != nil {
+			return 0, err
+		}
 		reader = bytes.NewReader(tmpBuffer.Bytes())
 		tmpBuffer.Reset()
 	} else {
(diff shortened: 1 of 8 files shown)


api-put-object.go
```diff
@@ -205,10 +205,13 @@ func (c Client) putObjectSingle(bucketName, objectName string, reader io.Reader,
 		hashAlgos["sha256"] = sha256.New()
 	}
 
-	if size <= minPartSize {
+	if size <= minPutBufferSize {
 		// Initialize a new temporary buffer.
 		tmpBuffer := new(bytes.Buffer)
 		size, err = hashCopyN(hashAlgos, hashSums, tmpBuffer, reader, size)
+		if err != nil {
```
vadmeste: Actually, hashCopyN can return io.EOF in normal conditions (not sure why), besides, there is a check of err later.

api-put-object.go
```diff
@@ -205,10 +205,13 @@ func (c Client) putObjectSingle(bucketName, objectName string, reader io.Reader,
 		hashAlgos["sha256"] = sha256.New()
 	}
 
-	if size <= minPartSize {
+	if size <= minPutBufferSize {
```
vadmeste: This was minPartSize, it means else was never entered, it could have buggy code.

harshavardhana: Actually let me revert this back its not needed with this patch.

api-put-object.go
harshavardhana: hashCopy should be io.Copy equivalent so should never return io.EOF

harshavardhana: This is fixed.

PR #615, opened 2017-02-17
doc: Rewrite GetObject API description

vadmeste:


vadmeste:
@harshavardhana, done.

PR #626, opened 2017-03-17
api: Add support to override region in PostPolicy.

harshavardhana:
This is needed for many applications which are generating
post policy form without having any access to the remote
storage and also wish to avoid making a network call.

This PR is in line with the change we allowed in minio-java
to override region when generating `formData()`

### DIFF
--- a/api-presigned.go
+++ b/api-presigned.go
@@ -17,7 +17,6 @@
 package minio
 
 import (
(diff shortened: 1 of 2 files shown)


harshavardhana:
This is incorrect.. implementation decided with @abperiasamy to make it generic.

PR #630, opened 2017-03-22
allow bucket location responses with empty Message (fixes #629)

### LINKED ISSUE #629: access to buckets owned by other users not possible with ceph radosgw
Using minio-go, I cannot access buckets owned by other users (with appropriate ACLs set) on a Ceph radosgw. It is expected that they cannot be listed, but testing for existence or accessing them by name should work. 

I have identified the underlying issue and will submit a PR shortly. 

The problem is that Ceph radosgw is not returning a `<Message>` in the `AccessDenied` error response, but `processBucketLocationResponse` is checking that an `Access Denied` string is contained in the message (https://github.com/minio/minio-go/blob/master/bucket-cache.go#L128). 

An example error response I get during a `BucketExists` call on an accessible bucket owned by another user is: 
```
<?xml version="1.0" encoding="UTF-8"?><Error><Code>AccessDenied</Code><BucketName>npg-cloud-realign-wip</BucketName><RequestId>tx0000000000000002b1dad-0058d25cfa-6ccee7-default</RequestId><HostId>6ccee7-default-default</HostId></Error>
```

Because `<Message>` is not included at all in the XML, it will be left at the nil value of `""` in the `errResp` ErrorResponse structure. My PR will test for an empty `Message` and treat that the same way as a message containing the string "Access Denied".

jrandall:
When Ceph radosgw responds to a bucket location request for a bucket which is not owned by the user, it issues an "AccessDenied" XML response that does not contain any `<Message>` element. Because of this, the code that handles "anonymous" buckets in `processBucketLocationResponse` is not triggered, as it is looking for both an `errResp.Code` of "AccessDenied" and also an `errResp.Message` containing the substring "Access Denied". 

This PR changes the behaviour of `processBucketLocationResponse` to allow an empty `errResp.Message` to be treated the same as one that contains the substring "Access Denied". 

Fixes #629

### DIFF
--- a/bucket-cache.go
+++ b/bucket-cache.go
@@ -125,7 +125,7 @@ func processBucketLocationResponse(resp *http.Response, bucketName string) (buck
 			// For access denied error, it could be an anonymous
 			// request. Move forward and let the top level callers
 			// succeed if possible based on their policy.
-			if errResp.Code == "AccessDenied" && (errResp.Message == "" || strings.Contains(errResp.Message, "Access Denied")) {
+			if errResp.Code == "AccessDenied" {
 				return "us-east-1", nil
 			}
 			return "", err
--- a/bucket-cache.go
+++ b/bucket-cache.go
@@ -125,7 +125,7 @@ func processBucketLocationResponse(resp *http.Response, bucketName string) (buck
 			// For access denied error, it could be an anonymous
 			// request. Move forward and let the top level callers
 			// succeed if possible based on their policy.
-			if errResp.Code == "AccessDenied" && strings.Contains(errResp.Message, "Access Denied") {
+			if errResp.Code == "AccessDenied" && (errResp.Message == "" || strings.Contains(errResp.Message, "Access Denied")) {
 				return "us-east-1", nil
 			}
 			return "", err
--- a/bucket-cache.go
+++ b/bucket-cache.go
@@ -21,7 +21,6 @@ import (
 	"net/http"
 	"net/url"
 	"path"
-	"strings"
 	"sync"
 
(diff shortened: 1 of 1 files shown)


bucket-cache.go
```diff
@@ -125,7 +125,7 @@ func processBucketLocationResponse(resp *http.Response, bucketName string) (buck
 			// For access denied error, it could be an anonymous
 			// request. Move forward and let the top level callers
 			// succeed if possible based on their policy.
-			if errResp.Code == "AccessDenied" && strings.Contains(errResp.Message, "Access Denied") {
+			if errResp.Code == "AccessDenied" && (errResp.Message == "" || strings.Contains(errResp.Message, "Access Denied")) {
```
harshavardhana: This logic looks pretty ugly, we can just look for `errResp.Code == "AccessDenied"` only.

jrandall: @harshavardhana I would have no problem with removing the check on `errResp.message` entirely, I just was not aware of why that check existed and did not want to risk a regression against another system. 

The conditions date back to when this code was first introduced in `api-error-response.go` in commit a856d4312629f7edd58dcd0cbaa973b1e7672493 but there does not appear to be any note there as to why the `Message` was a part of the condition. As you were the committer, if you have reason to think that `errResp.code == "AccessDenied"` is sufficient, I'll change to that.

harshavardhana: This change was introduced because some users with S3 have IAM policy where they disable certain operations i.e returning `AccessDenied` only.

harshavardhana:
Can you fix the build @jrandall ?

harshavardhana:
Can you fix the build error @jrandall ?

PR #636, opened 2017-04-01
[need-review] sigv2: Do not encode canonical sub-resource value

vadmeste:
Do not encode sub-resource value (such is uploadID) when building
canonicalized request string.

### DIFF
--- a/pkg/s3signer/request-signature-v2.go
+++ b/pkg/s3signer/request-signature-v2.go
@@ -316,7 +316,7 @@ func writeCanonicalizedResource(buf *bytes.Buffer, req http.Request, isPreSign b
 				// Request parameters
 				if len(vv[0]) > 0 {
 					buf.WriteByte('=')
-					buf.WriteString(strings.Replace(url.QueryEscape(vv[0]), "+", "%20", -1))
+					buf.WriteString(vv[0])
 				}
 			}
 		}


pkg/s3signer/request-signature-v2.go
```diff
@@ -316,7 +316,7 @@ func writeCanonicalizedResource(buf *bytes.Buffer, req http.Request, isPreSign b
 				// Request parameters
 				if len(vv[0]) > 0 {
 					buf.WriteByte('=')
-					buf.WriteString(strings.Replace(url.QueryEscape(vv[0]), "+", "%20", -1))
```
harshavardhana: @vadmeste can we test this with `ListObjects()` with some UTF-8 files and see how it works.

vadmeste: > @vadmeste can we test this with ListObjects() with some UTF-8 files and see how it works.

* I already tested that case with more than 1000 objects which has unicode characters, and that to trigger token query params in web requests
* I also tested v2 tests function of minio-go with AWS S3
* I also tested manually many use cases

Besides, the change line affects only values of Resource parameters: 

```
260 var resourceList = []string{
261     "acl",
262     "delete",
263     "location",
264     "logging",
265     "notification",
266     "partNumber",
267     "policy",
268     "requestPayment",
269     "torrent",
270     "uploadId",
271     "uploads",
272     "versionId",
273     "versioning",
274     "versions",
275     "website",
276 }
```

By the way, S3 spec is not clear about the behavior of this case. 

How did I find the bug?

I made a manual change in minio-go to send uploadID="QsbxgdvySmWkYZ2X84hQPg==" instead of the one that it receives from AWS in list multipart uploads. In that case, AWS returns signature mismatch.

PR #638, opened 2017-04-04
Export ListObjectV2 to core client

nl5887:
* needed for gateway

### DIFF
--- a/api-s3-datatypes.go
+++ b/api-s3-datatypes.go
@@ -42,7 +42,7 @@ type commonPrefix struct {
 }
 
 // listBucketResult container for listObjects V2 response.
-type listBucketV2Result struct {
+type ListBucketV2Result struct {
 	// A response can contain CommonPrefixes only if you have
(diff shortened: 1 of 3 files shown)


harshavardhana:
ping @nl5887

api-s3-datatypes.go
```diff
@@ -42,7 +42,7 @@ type commonPrefix struct {
 }
 
 // listBucketResult container for listObjects V2 response.
-type listBucketV2Result struct {
+type ListBucketV2Result struct {
 	// A response can contain CommonPrefixes only if you have
 	// specified a delimiter.
 	CommonPrefixes []commonPrefix
```
harshavardhana: @nl5887 do you think this is a problem ?

nl5887: I don't think this is a problem. Users won't declare commonPrefix themselves (as they could do with ListBucketV2Result) and Go won't complain about:

```prefixes := result.CommonPrefixes```

PR #640, opened 2017-04-04
api: Do not need Expect header.

harshavardhana:
Expect header is not support and implemented by GCS
and some S3 endpoints. Remove it as it doesn't add
much purpose.

### DIFF
--- a/api.go
+++ b/api.go
@@ -625,14 +625,6 @@ func (c Client) newRequest(method string, metadata requestMetadata) (req *http.R
 		return req, nil
 	}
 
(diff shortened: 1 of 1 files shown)


vadmeste:
Hard to see the benefit in practice, let's merge this and let's the community to decide.

krishnasrinivas:
what was the reason we brought in Expect header support on client? One obvious advantage is we dont read from the stream if the auth fails. But there was something else too?

harshavardhana:
That was the only thing.

harshavardhana:
But our server doesn't do true expect continue I think.  The control is inside Go http server. By the time request comes it already yes to the client.

PR #641, opened 2017-04-07
api: Create new pkg/encrypt package and use it.

harshavardhana:
This is implemented to avoid references of any
HTTP or aws based constants. Simply to show that
encryption is implemented for generic purposes.

### DIFF
--- a/api-get-object.go
+++ b/api-get-object.go
@@ -24,10 +24,12 @@ import (
 	"strings"
 	"sync"
 	"time"
+
+	"github.com/minio/minio-go/pkg/encrypt"
 )
 
 // GetEncryptedObject deciphers and streams data stored in the server after applying a specifed encryption materiels
-func (c Client) GetEncryptedObject(bucketName, objectName string, encryptMaterials EncryptionMaterials) (io.Reader, error) {
+func (c Client) GetEncryptedObject(bucketName, objectName string, encryptMaterials encrypt.Materials) (io.Reader, error) {
 
 	if encryptMaterials == nil {
 		return nil, ErrInvalidArgument("Unable to recognize empty encryption properties")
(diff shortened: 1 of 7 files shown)


pkg/encrypt/interface.go
```diff
@@ -14,31 +14,37 @@
...
-	"net/http"
-)
+import "io"
 
-// EncryptionMaterials - provides generic interface to encrypt
-// any stream of data. Some crypt information can be
-// save in the object metadata
-type EncryptionMaterials interface {
+// Materials - provides generic interface to encrypt any stream of data.
+type Materials interface {
```
vadmeste: 'Materials' could be more generic without GetIV(), GetKey() and GetMatDesc().

Think about it this way: any encryption module can save any information it wants in headers, that information will be retrieved from headers later to decrypt data.


As this becomes a separate package, I think we can replace http.Header with map[string]string.

```
type Materials interface {
   Read(b []byte) (int, error)
   SetupEncryptMode(stream io.Reader) (map[string]string, error)
   SetupDecryptMode(stream io.Reader, decryptInfo map[string]string) error
}
```

api-get-object.go
```diff
@@ -44,8 +46,10 @@ func (c Client) GetEncryptedObject(bucketName, objectName string, encryptMateria
 		return nil, err
 	}
 
-	encryptMaterials.SetupDecryptMode(encReader, st.Metadata)
+	// Start decrypting the object.
```
vadmeste: Decryption doesn't start here actually, this is just a setup. Data are decrypted when Read() is called.

pkg/encrypt/interface.go
harshavardhana: An IV and Key are much to do with Crypto rather than to do with AWS, AWS SPEC just names them properly and uses them. 

So it is correct the way it is right now, perhaps i can change it.. further.

Issue #617, opened 2017-02-23
api: Support CopyObject for all sizes


donatello:
High-level CopyObject requirements:

- [x] Support copying objects of all sizes.
- [x] Support source objects with an arbitrary range header (i.e. any valid start and end offset of a source object) via multipart copy object.
- [x] Support all copy conditions (this is already supported).

Currently, the library supports only copying objects <= 5GiB in size. Larger objects can also be copied via a multipart-copy-object strategy.

The multipart copy object operation consists of starting a new multipart upload, followed by 1 or more copy-object-part requests, and finally a complete-multipart request.

Note that copy object via a single PUT request does not support range headers, but copy-object-part does support this.

This feature was recently implemented in the [Haskell SDK](https://github.com/minio/minio-hs/commit/e4e2576c7488878661eaa2d7fa68ab12e177d094).

harshavardhana:
> Support source objects with an arbitrary range header (i.e. any valid start and end offset of a source object) via multipart copy object.

Why should this be supported? - what benefit does this provide a user? also at what junction a user really knows that they need to copy only a certain range of the object. 

Since we cannot append on the destination i don't see how this API benefits anyone.

hashbackup:
HashBackup could put a multipart ranged copy to good use.  HB packs files into arc files during the backup.  These default to 100MB but can be larger, like 4GB.  Over time, arc files get "holes" poked in them as files are deleted due to retention policies.  

For example, you backup a 75MB file and a 25MB file into 1 arc file and store it on S3.  The first file is marked deleted.  To actually recover space, the 100MB arc file has to be downloaded, packed, and uploaded.  The download is where high costs are incurred.

By using a series of mulipart copy requests, this packing operation could be done remotely _without requiring a download_.  I think the only cost would be the request cost: I couldn't see where Amazon charges fees for copy based on the size of the data.

harshavardhana:
Yes but minio libraries are not meant for exposing lower level multipart operations. For that you should use AWS SDKs or copy minio library sounce into your repo. 

I don't see why we should explore range APIs while not exposing miultipart APIs underneath.

hashbackup:
A range list of start-end offsets could be added to copy_object without exposing multipart.

donatello:
> Why should this be supported? - what benefit does this provide a user? also at what junction a user really knows that they need to copy only a certain range of the object.
> 
> Since we cannot append on the destination i don't see how this API behavior benefits anyone.

[...]

PR #644, opened 2017-04-10
Add large-object support for CopyObject API

donatello:
This patch is up for early review.

Context of the change is at https://github.com/minio/minio-go/issues/617

The byte-range for the source object has been added into CopyConditions type - ideally the `CopyConditions` type should be called `CopySourceInfo` and it should have the copy-conditions, copy-source and the optional copy-byte-range. However, the changes in this patch aim to keep backwards compatibility.

The one wrinkle is the `NewCopyConditions` method - it is now the preferred method to create a `CopyConditions{}` instance as it initializes the byte-range to -1 (indicating that the byte range is not specified by the user).

This is as yet untested - hoping to get some input about the interface improvement to be done.

### DIFF
--- a/api-put-object-copy.go
+++ b/api-put-object-copy.go
@@ -17,12 +17,16 @@
 package minio
 
 import (
+	"fmt"
 	"net/http"
+	"net/url"
+	"strings"
 
 	"github.com/minio/minio-go/pkg/s3utils"
 )
 
-// CopyObject - copy a source object into a new object with the provided name in the provided bucket
+// CopyObject - copy a source object into a new object with the
+// provided name in the provided bucket
 func (c Client) CopyObject(bucketName string, objectName string, objectSource string, cpCond CopyConditions) error {
 	// Input validation.
 	if err := isValidBucketName(bucketName); err != nil {
@@ -31,18 +35,46 @@ func (c Client) CopyObject(bucketName string, objectName string, objectSource st
 	if err := isValidObjectName(objectName); err != nil {
 		return err
 	}
-	if objectSource == "" {
-		return ErrInvalidArgument("Object source cannot be empty.")
+	srcBucket, srcObject, err := getObjectSource(objectSource)
+	if err != nil {
+		return err
+	}
+
+	// Get info about the source object
+	srcInfo, err := c.StatObject(srcBucket, srcObject)
+	if err != nil {
+		return err
+	}
+	srcByteRangeSize := cpCond.getRangeSize()
+	if srcByteRangeSize > srcInfo.Size ||
+		(srcByteRangeSize > 0 && cpCond.byteRangeEnd >= srcInfo.Size) {
+		return ErrInvalidArgument(fmt.Sprintf(
+			"Specified byte range (%d, %d) does not fit within source object (size = %d)",
+			cpCond.byteRangeStart, cpCond.byteRangeEnd, srcInfo.Size))
+	}
+
+	copySize := srcByteRangeSize
+	if copySize == 0 {
+		copySize = srcInfo.Size
 	}
 
 	// customHeaders apply headers.
 	customHeaders := make(http.Header)
-	for _, cond := range cpCond.conditions {
-		customHeaders.Set(cond.key, cond.value)
+	for key, value := range cpCond.conditions {
+		customHeaders.Set(key, value)
 	}
 
 	// Set copy source.
 	customHeaders.Set("x-amz-copy-source", s3utils.EncodePath(objectSource))
+
+	// Check if single part copy suffices. Multipart is required when:
+	// 1. source-range-offset does not refer to full source object, or
+	// 2. size of copied object > 5gb
+	if copySize > maxPartSize ||
+		(srcByteRangeSize > 0 && srcByteRangeSize != srcInfo.Size) {
+		return c.multipartCopyObject(bucketName, objectName,
+			objectSource, cpCond, customHeaders, copySize)
+	}
 
 	// Execute PUT on objectName.
 	resp, err := c.executeMethod("PUT", requestMetadata{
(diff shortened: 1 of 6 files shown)


donatello:
API change proposal:

`func CopyObject(bucketName, objectName, srcBucket, srcObject string, csi copySourceInfo) error`

Providing explicit source bucket and object makes it easier for user to not have to format it as "/bucket/object".

Data type and methods:

```go
// Not exported so that user has to use an initialization function
type copySourceInfo struct {
    copyConditions map[string]string
    byteRangeStart int64
    byteRangeEnd int64
}

// To allocate map and initialize range
func NewCopySourceInfo() copySourceInfo {}

// funcs to set copy conditions.
func (c *copySourceInfo) SetMatchETag(etag string) error {}
func (c *copySourceInfo) SetMatchETagExcept(etag string) error {}
func (c *copySourceInfo) SetUnmodified(modTime time.Time) error {}
func (c *copySourceInfo) SetModified(modTime time.Time) error {}
func (c *copySourceInfo) SetByteRange(start, end int64) error {}

// internal utils
func (c *copySourceInfo) getRangeSize() int64 {}
func (c *copySourceInfo) duplicate() *copySourceInfo {}
```

harshavardhana:
To keep backward compatibility we can introduce a new API, we now have projects dependent on `minio-go`. So we need to be careful with API changes. 

Will discuss this with @abperiasamy and see where we go from there.

copy-conditions.go
```diff
@@ -91,9 +75,37 @@ func (c *CopyConditions) SetModified(modTime time.Time) error {
...
+
+func (c *CopyConditions) getRangeSize() int64 {
+	if c.byteRangeStart < 0 {
+		// only happens if byte-range was not set by user
+		return 0
+	}
+	return c.byteRangeEnd - c.byteRangeStart + 1
+}
+
+func (c *CopyConditions) duplicate() *CopyConditions {
```
harshavardhana: call this `clone()` or `cloneCopyConditions` in Go style.

api-put-object-copy.go
```diff
@@ -70,3 +102,91 @@ func (c Client) CopyObject(bucketName string, objectName string, objectSource st
...
 }
+
+func getObjectSource(src string) (bucket string, object string, err error) {
+	parts := strings.Split(src, "/")
+	if len(parts) != 3 || parts[0] != "" || parts[1] == "" || parts[2] == "" {
+		return "", "", ErrInvalidArgument("Object source should be formatted as '/bucketName/objectName'")
+	}
+	return parts[1], parts[2], nil
+}
+
```
harshavardhana: Comment necessary..

api-put-object-copy.go
```diff
@@ -70,3 +102,91 @@ func (c Client) CopyObject(bucketName string, objectName string, objectSource st
 	// Return nil on success.
 	return nil
 }
+
+func getObjectSource(src string) (bucket string, object string, err error) {
```
harshavardhana: Comment necessary.. and a test case..

api-put-object-copy.go
```diff
@@ -31,19 +35,47 @@ func (c Client) CopyObject(bucketName string, objectName string, objectSource st
 	if err := isValidObjectName(objectName); err != nil {
 		return err
 	}
-	if objectSource == "" {
-		return ErrInvalidArgument("Object source cannot be empty.")
+	srcBucket, srcObject, err := getObjectSource(objectSource)
+	if err != nil {
+		return err
+	}
+
```
krisis: Stray newline?

api-put-object-copy.go
```diff
@@ -31,19 +35,47 @@ func (c Client) CopyObject(bucketName string, objectName string, objectSource st
...
+	}
+	srcByteRangeSize := cpCond.getRangeSize()
+	if srcByteRangeSize > srcInfo.Size ||
+		(srcByteRangeSize > 0 && cpCond.byteRangeEnd >= srcInfo.Size) {
+		return ErrInvalidArgument(fmt.Sprintf(
+			"Specified byte range (%d, %d) does not fit within source object (size = %d)",
+			cpCond.byteRangeStart, cpCond.byteRangeEnd, srcInfo.Size))
+	}
+
+	copySize := srcByteRangeSize
```
krisis: A comment explaining how an empty range is interpreted as the entire object would provide context for future.

copy-conditions.go
```diff
@@ -91,9 +75,37 @@ func (c *CopyConditions) SetModified(modTime time.Time) error {
...
+	}
+	c.byteRangeEnd = end
+	c.byteRangeStart = start
+	return nil
+}
+
+func (c *CopyConditions) getRangeSize() int64 {
+	if c.byteRangeStart < 0 {
+		// only happens if byte-range was not set by user
+		return 0
```
vadmeste: Probably it is better to set this to -1 to say unknown. Probably zero is a valid value.

copy-conditions.go
```diff
@@ -91,9 +75,37 @@ func (c *CopyConditions) SetModified(modTime time.Time) error {
...
-		value: modTime.Format(http.TimeFormat),
-	})
+	c.conditions["x-amz-copy-source-if-modified-since"] = modTime.Format(http.TimeFormat)
 	return nil
 }
+
+// SetByteRange - set the start and end of the source object to be
+// copied.
+func (c *CopyConditions) SetByteRange(start, end int64) error {
+	if start < 0 || end < start {
```
vadmeste: I think `end < start` is wrong, we can have `start == end` and range size will be 1.

vadmeste: Like when we want to get the first byte of an object

copy-conditions.go
```diff
@@ -91,9 +75,37 @@ func (c *CopyConditions) SetModified(modTime time.Time) error {
...
+		return ErrInvalidArgument("Range start less than 0 or range end less than range start.")
+	}
+	if end-start+1 < 1 {
```
vadmeste: I think this condition is useless, 

Here we are already sure that `start <= end`:

`start <= end` **<=>** `0 <= end - start` **<=>** `1 <= end - start + 1`, 

so we will never have `end-start+1 < 1`

api-put-object-copy.go
```diff
@@ -70,3 +102,91 @@ func (c Client) CopyObject(bucketName string, objectName string, objectSource st
...
+		pCond.byteRangeStart = partSize * (int64(partNumber) - 1)
+		if partNumber < partsCount {
+			pCond.byteRangeEnd = pCond.byteRangeStart + partSize - 1
+		} else {
+			pCond.byteRangeEnd = pCond.byteRangeStart + lastPartSize - 1
+		}
+
+		// Update the source range header value.
+		headers.Set("x-amz-copy-source-range",
+			fmt.Sprintf("bytes:%d-%d", pCond.byteRangeStart,
```
harshavardhana: This is wrong btw - it should be `bytes=%d-%d`

deekoder:
Is this still blocked?

harshavardhana:
> Is this still blocked?

Yes not discussed yet.

harshavardhana:
@donatello can you update this and incorporate usage of requestHeaders instead of copyConditions . Which would expand the scope of this.

harshavardhana:
> @donatello can you update this and incorporate usage of requestHeaders instead of copyConditions . Which would expand the scope of this.

@donatello can you incorporate the request-headers.go changes to CopyObject() just like how we did with GetObject() .This would be a breaking change we will bump up the tag so that the users who have referenced using our tags atleast can understand why this change.

tejaycar:
@donatello, @harshavardhana is there anything I can do to help?

donatello:
@tejaycar We plan to get this resolved by next week. We are busy with the Minio server release at this time.

tejaycar:
Awesome. Thanks guys.

tejaycar:
@donatello I don't mean to bug, but do you have any update to offer on this?  I'm not trying to push, just trying to plan accordingly.

donatello:
> @donatello I don't mean to bug, but do you have any update to offer on this? I'm not trying to push, just trying to plan accordingly.

@tejaycar I am actively working on this, and hope to have a PR today. I would venture to say that it might get merged by the end of the week after reviews and testing.

tejaycar:
Once again, thank you.  I appreciate the update.

donatello:
We are bringing large object copy support via a more flexible `ComposeObject` API that allows server-side copying of objects to create new objects.

This PR is thus superseded by https://github.com/minio/minio-go/pull/715. Closing as obsolete.

PR #646, opened 2017-04-13
api: Add NewWithCredentials()

### LINKED ISSUE #643: api: Support IAM roles for AWS S3 endpoint

harshavardhana:
This PR adds a new API

  - NewWithCredentials()

Internally NewWithCredentials is now used with
all APIs such as New(), NewV4(), NewV2() and NewWithRegion.

Also brings a new package called `credentials` to manage
various credentials type, currently the credentials
package supports

  - Reading file from `.aws/credentials`, `.mc/config.json`
  - Reading env variables for AWS*, MINIO*
  - Fetching from IAM roles assigned to an EC2 instance.
  - Static credentials which is the current default behavior.

Example code using IAM.
```go
        iam := credentials.NewIAM("")
        s3Client, err := minio.NewWithCredentials("s3.amazonaws.com", iam, true, "")
        if err != nil {
                log.Fatalln(err)
        }

        buckets, err := s3Client.ListBuckets()
        if err != nil {
                log.Fatalln(err)
        }
        for _, bucket := range buckets {
                log.Println(bucket)
        }
```

Fixes #643


pkg/credentials/file_minio.go
```diff
@@ -0,0 +1,113 @@
...
+	if filename == "" {
+		filename = os.Getenv("MINIO_SHARED_CREDENTIALS_FILE")
+		if filename == "" {
+			homeDir, err := homedir.Dir()
+			if err != nil {
+				return nil, err
+			}
+			filename = filepath.Join(homeDir, ".mc", "config.json")
+			if runtime.GOOS == "windows" {
+				filepath.Join(homeDir, "mc", "config.json")
```
vadmeste: This looks like a typo

pkg/credentials/iam_aws_s3.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+// NewIAM returns a pointer to a new Credentials object wrapping
+// the IAM. Takes a ConfigProvider to create a EC2Metadata client.
+// The ConfigProvider is satisfied by the session.Session type.
+func NewIAM(endpoint string) *Info {
+	if endpoint == "" {
+		// IAM Roles for Amazon EC2
+		// http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html
+		endpoint = "http://169.254.169.254"
```
vadmeste: Shouldn't this be https ?

pkg/credentials/iam_aws_s3.go
```diff
@@ -0,0 +1,189 @@
...
+	u, err := url.Parse(endpoint)
+	if err != nil {
+		return nil, err
+	}
+	u.Path = iamSecurityCredsPath
+	req, err := http.NewRequest("GET", u.String(), nil)
+	if err != nil {
+		return nil, err
+	}
+	resp, err := client.Do(req)
```
vadmeste: Close response body.

pkg/credentials/iam_aws_s3.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+
+	credsList := []string{}
+	s := bufio.NewScanner(resp.Body)
+	for s.Scan() {
+		credsList = append(credsList, s.Text())
+	}
+
+	if err := s.Err(); err != nil {
+		return nil, errors.New("Failed to read from IAM metadata service")
```
vadmeste: I think it is bettter to append s.Err() to the returned error here.

pkg/credentials/iam_aws_s3.go
```diff
@@ -0,0 +1,189 @@
...
+// requestCred requests the credentials for a specific credentials from the EC2 service.
+//
+// If the credentials cannot be found, or there is an error reading the response
+// and error will be returned.
+func requestCred(client *http.Client, endpoint string, credsName string) (ec2RoleCredRespBody, error) {
+	u, err := url.Parse(endpoint)
+	if err != nil {
+		return ec2RoleCredRespBody{}, err
+	}
+	u.Path = path.Join(iamSecurityCredsPath, credsName)
```
vadmeste: Shouldn't we encode credsName here ?

pkg/credentials/iam_aws_s3.go
```diff
@@ -0,0 +1,189 @@
...
+	req, err := http.NewRequest("GET", u.String(), nil)
+	if err != nil {
+		return ec2RoleCredRespBody{}, err
+	}
+
+	resp, err := client.Do(req)
```
vadmeste: Close resp body.

pkg/credentials/file_minio.go
harshavardhana: oops yes.

pkg/credentials/iam_aws_s3.go
harshavardhana: Nope it is only accessibly from inside EC2 instance.

pkg/credentials/iam_aws_s3.go
harshavardhana: You are right.

pkg/credentials/lib.go
```diff
@@ -0,0 +1,207 @@
...
+		e.expiration = e.expiration.Add(-window)
+	}
+}
+
+// IsExpired returns if the credentials are expired.
+func (e *Expiry) IsExpired() bool {
+	if e.CurrentTime == nil {
+		e.CurrentTime = time.Now
+	}
+	return e.expiration.Before(e.CurrentTime())
```
vadmeste: You mean 'After' here ?

pkg/credentials/lib.go
```diff
@@ -0,0 +1,207 @@
...
+// reflect that override.
+func (c *Info) IsExpired() bool {
+	c.Lock()
+	defer c.Unlock()
+
+	return c.isExpired()
+}
+
+// isExpired helper method wrapping the definition of expired credentials.
+func (c *Info) isExpired() bool {
```
vadmeste: This function doesn't seem to be useful, you can merge it in IsExpired()

harshavardhana: isExpired() is used at two places one in Get() and another is IsExpired()

vadmeste: > isExpired() is used at two places one in Get() and another is IsExpired()

yes, right.

vadmeste:
Some missing LICENCES in some files

pkg/credentials/lib.go
harshavardhana: No Before is correct.  Read it as expiration time is before current time.

vadmeste: oh yes sorry, probably it is easier to read `e.CurrentTime().After(e.expiration)`

harshavardhana: But usually it is better to use on the left side what you are validating against.  Otherwise it would look like `currentTime()` is what we are validating not expiration.

pkg/credentials/iam_aws_s3.go
harshavardhana: Done

pkg/credentials/iam_aws_s3.go
harshavardhana: Done

pkg/credentials/iam_aws_s3.go
harshavardhana: Done

pkg/credentials/iam_aws_s3.go
harshavardhana: actually its not needed.. u.String() encodes it.

pkg/credentials/iam_aws_s3.go
harshavardhana: done

harshavardhana:
> Some missing LICENCES in some files

Yeah will add license at the end.

deekoder:
@harshavardhana - is this ready for merge?

harshavardhana:
> @harshavardhana - is this ready for merge?

Not yet waiting on Mattermost.

pkg/credentials/iam_aws_s3.go
```diff
@@ -0,0 +1,192 @@
...
+	for key, val := range via[0].Header {
+		req.Header[key] = val
+	}
+	return nil
+}
+
+// NewIAM returns a pointer to a new Credentials object wrapping
+// the IAM. Takes a ConfigProvider to create a EC2Metadata client.
+// The ConfigProvider is satisfied by the session.Session type.
+func NewIAM(endpoint string) *Info {
```
vadmeste: I think this API needs to accept a role name as a parameter, not only endpoint. The application needs to get credentials from the right role name before performing the needed action which is only authorized by that role.

harshavardhana: RoleName is not needed since the IAM RoleName is attached to the EC2 instance where the GET originates. So you are bound to be under the RoleName already. 

Why do you think it is necessary?

vadmeste: Yes, you are right. I thought we can attach multiple IAM roles to one instance but we can only set one. (which is weird by the way and looks painful for the users)

I found this here, http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html

`You *cannot* attach multiple IAM roles to a single instance, but you can attach a single IAM role to multiple instances. For more information about creating and using IAM roles,`

harshavardhana:
Unblocking and ready for review @vadmeste

api-presigned.go
```diff
@@ -122,21 +122,41 @@ func (c Client) PresignedPostPolicy(p *PostPolicy) (u *url.URL, formData map[str
...
+	signerType := c.signature
+	// Check if the signature type is default, it means caller didn't
+	// ask for anything specific signature style.. Check if the
+	// credentials set returned the type of signature that should
+	// be used, if not proceed to use default.
+	if signerType == SignatureDefault && value.SignatureType != "" {
+		signerType = parseSignatureType(value.SignatureType)
+	}
+
+	// FIXME: Session token should be supported.
```
vadmeste: This can't be fixed in this PR ?

api-presigned.go
```diff
@@ -122,21 +122,41 @@ func (c Client) PresignedPostPolicy(p *PostPolicy) (u *url.URL, formData map[str
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	signerType := c.signature
+	// Check if the signature type is default, it means caller didn't
+	// ask for anything specific signature style.. Check if the
+	// credentials set returned the type of signature that should
+	// be used, if not proceed to use default.
+	if signerType == SignatureDefault && value.SignatureType != "" {
```
vadmeste: I think that the signature type that comes from config, if it exists, should override the one specified in the application code. That way, we can give users more power to configure.

api.go
```diff
@@ -650,17 +679,18 @@ func (c Client) newRequest(method string, metadata requestMetadata) (req *http.R
...
-	case c.signature.isV2():
+	case signerType.isV2():
 		// Add signature version '2' authorization header.
-		req = s3signer.SignV2(*req, c.accessKeyID, c.secretAccessKey)
-	case c.signature.isStreamingV4() && method == "PUT":
-		req = s3signer.StreamingSignV4(req, c.accessKeyID,
-			c.secretAccessKey, location, metadata.contentLength, time.Now().UTC())
+		req = s3signer.SignV2(*req, accessKeyID, secretAccessKey)
+	case signerType.isStreamingV4() && method == "PUT":
+		req = s3signer.StreamingSignV4(req, accessKeyID,
```
vadmeste: No session token here ?

pkg/credentials/env_aws.go
```diff
@@ -0,0 +1,47 @@
...
+	return Value{
+		AccessKeyID:     id,
+		SecretAccessKey: secret,
+		SessionToken:    os.Getenv("AWS_SESSION_TOKEN"),
+	}, nil
+}
+
+// IsExpired returns if the credentials have been retrieved.
+func (e *EnvAWS) IsExpired() bool {
+	return !e.retrieved
```
vadmeste: This probably should always return true. A user can change his environment variables inside his application (who knows), and also it is no harm if we fetch from the environment each time Retrieve() is called.

pkg/credentials/chain.go
```diff
@@ -0,0 +1,72 @@
+package credentials
+
+import "fmt"
+
+// A Chain will search for a provider which returns credentials
```
vadmeste: This is well thought, I wasn't able to criticize this.

api-presigned.go
harshavardhana: Actually no. We have methods to override that see NewV4 and NewV2. We need to honor that. Since they don't take dynamic credentials.

api.go
harshavardhana: Need to figure it out .. not sure right now.

pkg/credentials/env_aws.go
harshavardhana: That we already do.. it is to indicate if envs can ever be expired. They cannot be. It is not correct to call it not expired.. basically envs are similar to static creds in their behavior.

api-presigned.go
harshavardhana: Spec is little vague on post policy about this. Will need to explore more.

api-presigned.go
```diff
@@ -168,13 +187,27 @@ func (c Client) PresignedPostPolicy(p *PostPolicy) (u *url.URL, formData map[str
 		return nil, nil, err
 	}
 
+	if sessionToken != "" {
```
vadmeste: :+1:

pkg/credentials/static.go
```diff
@@ -0,0 +1,53 @@
...
+
+// A Static is a set of credentials which are set programmatically,
+// and will never expire.
+type Static struct {
+	Value
+}
+
+// NewStatic returns a pointer to a new Credentials object
+// wrapping a static credentials value provider.
+func NewStatic(id, secret, token, signerType string) *Credentials {
```
vadmeste: Isn't better to create a new exported signature data type and use it instead of string ? that way the user would be less prone to error since any typo will make this module uses anonymous.

harshavardhana: Actually it can be simplified..

pkg/credentials/file_aws_credentials.go
```diff
@@ -0,0 +1,120 @@
...
+	// environment variable is also not set.
+	profile string
+
+	// retrieved states if the credentials have been successfully retrieved.
+	retrieved bool
+}
+
+// NewFileAWSCredentials returns a pointer to a new Credentials object
+// wrapping the Profile file provider.
+func NewFileAWSCredentials(filename string, profile string) (*Credentials, error) {
```
vadmeste: NewFileAWSCredentials and NewFileMinioClient are always returnig nil, shall we remove the returned error ? if we do, it will be like other New() primitive for static, IAM and envs.

harshavardhana: Yes looks like older problem.  Will fix

PR #650, opened 2017-04-15
Example of suggested wording for ceph s3

davecb:
I just raised this in slack, as ceph cyrrently assumes you will always use us-east-1, and v4 authentication fails loudly and unhelpfully if you select anything else.  We used to use ca-east-0, but when we switched to v4 authentication, it broke most _wonderfully_. (And I switched over to minio!)

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -101,7 +101,7 @@ __Parameters__
 | Param  | Type  | Description  |
 |---|---|---|
 |`bucketName`  | _string_  | Name of the bucket |
-| `location`  |  _string_ | Region where the bucket is to be created. Default value is us-east-1. Other valid values are listed below. Note: When used with minio server, use the region specified in its config file (defaults to us-east-1).|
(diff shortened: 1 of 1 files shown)


docs/API.md
```diff
@@ -101,7 +101,7 @@ __Parameters__
 | Param  | Type  | Description  |
 |---|---|---|
 |`bucketName`  | _string_  | Name of the bucket |
-| `location`  |  _string_ | Region where the bucket is to be created. Default value is us-east-1. Other valid values are listed below. Note: When used with minio server, use the region specified in its config file (defaults to us-east-1).|
+| `location`  |  _string_ | Region where the bucket is to be created. Default value is us-east-1. Other valid values are listed below. Note: When used with minio server, use the region specified in its config file (defaults to us-east-1).When used with ceph's s3 API, always use use-east-1|
 | | |us-east-1 |
```
harshavardhana: It should `us-east-1` not `use-east-1`

harshavardhana:
Fundamentally i don't think so we should mention `Ceph` specifically in our docs `minio-go` is a generic library for all S3 compatible implementations. It is Ceph problem and they should document this properly.  Even the errors generated without region not set should be fixed as well on the server side. 

So i don't think we should explicitly mention anything like this in our SDKs.

Issue #607, opened 2017-02-07
signer: [feature] Implement streaming signature support in minio-go


xxorde:
If you want to write a data stream to an S3 backend with PutObject() the data is stored in a memory buffer until the stream is finish. After the stream is finished Minio will start to write it to the backend.

This has two main disadvantages
1. The whole process takes long because Minio does not start to write to the backend before it has put the full stream to memory
2. Minio consumes as much memory as the data you want to store, it is not possible to store streams bigger than the available memory

@harshavardhana told me that a fix is already work in progress, but I like to open the issue to track the progress and to know when I can test it :)

hexadecy:
For 2. I was thinking that when the size is bigger than 5MiB the data is splitted in parts.

xxorde:
I believe its 64MB, but Minio needs to know the full size of the data before it starts to upload.
So if it can not determine the full size it writes everything in a buffer til EOF and than starts.

xxorde:
Hi,

are there any news here, is it still work in progress?
Or will it be done later because it was tagged  "enhancement"?

Kind regards.

krisis:
@xxorde I am working on this. https://github.com/minio/minio-go/pull/609 is work-in-progress. The implementation in the PR can be improved to consume lesser memory while computing chunk signature. I will be getting back to this next.

xxorde:
@krisis  thank you for your fast response. I am really looking forward to use minio for handling my backup streams. :)

xxorde:
@krisis I do not want to annoy you but do you have any idea when you are able to fix this?
If you are not able do get back to this issue please let me know, so I can try to find another solution.

krisis:
@xxorde Apolgoies for not sharing updates on where this PR is. I have not had time to commit fully on this feature, I am making changes when I find time. I shall update this PR by when we expect to complete this feature in a day or two. Thanks for being patient.

krisis:
@xxorde I have made some more changes to the PR. The memory consumption we discussed on slack might be due to md5 computations performed for multipart upload. Could you share a snippet of your code which uses this streaming signature implementation? It would help me understand where the memory usage pattern.

xxorde:
@krisis  yes of cause, here is an example.

```go
package main

import (
	"flag"
	"log"
	"os"

	minio "github.com/minio/minio-go"
)

func main() {
	var object, bucket, accessKey, secretKey string

	location := "us-east-1"

[...]

PR #657, opened 2017-04-23
Make some changes to PutObjectStreaming() API

harshavardhana:
- Detect size automatically like other PutObject() operations.
- Allow progress bar to be passed into PutObjectStreaming().
- Allow also metadata to be passed into PutObjectStreaming().
- Rename NewStreamingV4 to just StreamingV4(). Keeping it
  consistent with other signature methods.

### DIFF
--- a/api-put-object-progress.go
+++ b/api-put-object-progress.go
@@ -123,17 +123,18 @@ func (c Client) PutObjectWithMetadata(bucketName, objectName string, reader io.R
 }
 
 // PutObjectStreaming using AWS streaming signature V4
-func (c Client) PutObjectStreaming(bucketName, objectName string, reader io.Reader, size int64) (n int64, err error) {
-	if size < 0 {
-		return 0, ErrInvalidArgument("Size can't be negative.")
-	}
+func (c Client) PutObjectStreaming(bucketName, objectName string, reader io.Reader) (n int64, err error) {
+	return c.PutObjectStreamingWithProgress(bucketName, objectName, reader, nil, nil)
+}
 
-	// Check for largest object size allowed.
-	if size > int64(maxMultipartPutObjectSize) {
-		return 0, ErrEntityTooLarge(size, maxMultipartPutObjectSize, bucketName, objectName)
-	}
+// PutObjectStreamingWithMetadata using AWS streaming signature V4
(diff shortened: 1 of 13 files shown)


harshavardhana:
This PR also fixes the bug #658

api-put-object-progress.go
```diff
@@ -142,16 +143,45 @@ func (c Client) PutObjectStreaming(bucketName, objectName string, reader io.Read
...
+	}
+
+	// Check for largest object size allowed.
+	if size > int64(maxMultipartPutObjectSize) {
+		return 0, ErrEntityTooLarge(size, maxMultipartPutObjectSize, bucketName, objectName)
+	}
+
+	// If size cannot be found on a stream, it is not possible
+	// to upload using streaming signature.
+	if size < 0 {
```
vadmeste: Can't we use multipart upload when we don't know the full size of the object?
- miniog-go calls NewMultipartUpload which doesn't require to know the size of the object
- minio-go tries to read at least 64MiB from io.Reader before it uploads the part.
- then CompleteMultipartUpload

We need to disable parallel uploads here, also resuming upload should be avoided which means we need to create a new upload id each time we call PutObjectStreaming.

We can do this in a separate PR since there is API breakage.

harshavardhana: Good idea will do, perhaps in a separate PR.

krisis: @harshavardhana we should close https://github.com/minio/minio-go/issues/607 once @vadmeste's comments are addressed. This PR fixes #658.

harshavardhana: Done updated @krisis @vadmeste

api-put-object-progress.go
```diff
@@ -142,16 +143,45 @@ func (c Client) PutObjectStreaming(bucketName, objectName string, reader io.Read
...
+
+	// Check for largest object size allowed.
+	if size > int64(maxMultipartPutObjectSize) {
+		return 0, ErrEntityTooLarge(size, maxMultipartPutObjectSize, bucketName, objectName)
+	}
+
+	// If size cannot be found on a stream, it is not possible
+	// to upload using streaming signature, fall back to multipart.
+	if size < 0 {
+		return c.putObjectMultipartStream(bucketName, objectName, reader, size, metadata, progress)
```
krisis: @harshavardhana shouldn't we be using `c.putObjectMultipartStreamNoChecksum`?

PR #662, opened 2017-04-24
Validate minio.Object ETag between read requests

### LINKED ISSUE #656: minioClient.Object should validate ETag for new HTTP requests
obj := minioClient.GetObject()

obj.Read(...)

Now `obj.readAt()` at a different offset will cause new HTTP request, we have now ensure that the referred object on the server side has not been replaced by verifying that the ETag is same.

krisis:
On each new request, we compare `minio.Object`'s cached ETag with the current response ETag to error out if the object has changed since the last read.

Fixes #656

### DIFF
--- a/api-get-object.go
+++ b/api-get-object.go
@@ -67,6 +67,7 @@ func (c Client) GetObject(bucketName, objectName string) (*Object, error) {
 	var httpReader io.ReadCloser
 	var objectInfo ObjectInfo
 	var err error
+
 	// Create request channel.
 	reqCh := make(chan getRequest)
 	// Create response channel.
@@ -78,6 +79,9 @@ func (c Client) GetObject(bucketName, objectName string) (*Object, error) {
 	go func() {
 		defer close(reqCh)
 		defer close(resCh)
+
+		// Used to verify if etag of object has changed since last read.
+		var etag string
 
 		// Loop through the incoming control messages and read data.
 		for {
@@ -105,7 +109,7 @@ func (c Client) GetObject(bucketName, objectName string) (*Object, error) {
 							// Do not set objectInfo from the first readAt request because it will not get
 							// the whole object.
 							reqHeaders.SetRange(req.Offset, req.Offset+int64(len(req.Buffer))-1)
-							httpReader, _, err = c.getObject(bucketName, objectName, reqHeaders)
+							httpReader, objectInfo, err = c.getObject(bucketName, objectName, reqHeaders)
 						} else {
 							reqHeaders.SetRange(req.Offset, 0)
 							// First request is a Read request.
@@ -117,6 +121,7 @@ func (c Client) GetObject(bucketName, objectName string) (*Object, error) {
 							}
 							return
 						}
+						etag = objectInfo.ETag
 						// Read at least firstReq.Buffer bytes, if not we have
 						// reached our EOF.
 						size, err := io.ReadFull(httpReader, req.Buffer)
@@ -143,13 +148,18 @@ func (c Client) GetObject(bucketName, objectName string) (*Object, error) {
 							// Exit the go-routine.
 							return
 						}
+						etag = objectInfo.ETag
 						// Send back the first response.
 						resCh <- getResponse{
 							objectInfo: objectInfo,
 						}
 					}
 				} else if req.settingObjectInfo { // Request is just to get objectInfo.
-					objectInfo, err := c.StatObject(bucketName, objectName)
+					reqHeaders := NewGetReqHeaders()
+					if etag != "" {
+						reqHeaders.SetMatchETag(etag)
+					}
+					objectInfo, err := c.statObject(bucketName, objectName, reqHeaders)
 					if err != nil {
 						resCh <- getResponse{
 							Error: err,
@@ -170,6 +180,9 @@ func (c Client) GetObject(bucketName, objectName string) (*Object, error) {
 					// All readAt requests are new requests.
 					if req.DidOffsetChange || !req.beenRead {
 						reqHeaders := NewGetReqHeaders()
+						if etag != "" {
+							reqHeaders.SetMatchETag(etag)
+						}
 						if httpReader != nil {
 							// Close previously opened http reader.
 							httpReader.Close()
@@ -276,6 +289,12 @@ type Object struct {
 func (o *Object) doGetRequest(request getRequest) (getResponse, error) {
 	o.reqCh <- request
 	response := <-o.resCh
+
+	// Return any error to the top level.
+	if response.Error != nil {
+		return response, response.Error
+	}
+
 	// This was the first request.
 	if !o.isStarted {
 		// The object has been operated on.
@@ -291,11 +310,6 @@ func (o *Object) doGetRequest(request getRequest) (getResponse, error) {
 	if !o.beenRead {
 		o.beenRead = response.didRead
 	}
-	// Return any error to the top level.
-	if response.Error != nil {
-		return response, response.Error
-	}
-
 	// Data are ready on the wire, no need to reinitiate connection in lower level
 	o.seekData = false
 
--- a/api_functional_v4_test.go
+++ b/api_functional_v4_test.go
@@ -29,6 +29,7 @@ import (
 	"net/url"
 	"os"
 	"strconv"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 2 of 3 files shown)


api-get-object.go
```diff
@@ -289,6 +291,21 @@ func (o *Object) doGetRequest(request getRequest) (getResponse, error) {
...
+	// If response doesn't carry a valid ETag, ignore.
+	case response.objectInfo.ETag == "":
+
+	// Set Object's etag on first request.
+	case o.etag == "":
+		o.etag = response.objectInfo.ETag
+
+	// Check if etag from response matches Object's etag.
+	case o.etag != response.objectInfo.ETag:
+		return response, errors.New("object has been modified since last read")
```
harshavardhana: You can use the ErrorResponse  thingy here and return BadDigest instead.

krisis: @harshavardhana Since this error is returned by `minio.Object`'s `Stat`, `Read` and `ReadAt`, I was thinking if `syscall.EBADFD` would be more appropriate?

krisis: > You can use the ErrorResponse thingy here and return BadDigest instead.

@harshavardhana, `The Content-Md5 you specified did not match what we received.` can be misleading to someone receiving this error for `Read` or `ReadAt` on `minio.Object`. Further, the request headers sent during a `GetObject` were correct, at that point in time.Thoughts?

harshavardhana: syscall.EBADFD is good would work with *os.File behavior.

api-get-object.go
```diff
@@ -289,6 +292,21 @@ func (o *Object) doGetRequest(request getRequest) (getResponse, error) {
...
+	// If response doesn't carry a valid ETag, ignore.
+	case response.objectInfo.ETag == "":
+
+	// Set Object's etag on first request.
+	case o.etag == "":
+		o.etag = response.objectInfo.ETag
+
+	// Check if etag from response matches Object's etag.
+	case o.etag != response.objectInfo.ETag:
+		return response, syscall.EBADF
```
krishnasrinivas: since we always return type `ErrorResponse` for error it's better to:
`return ErrInvalidObjectName("ETag has changed")`

instead of returning syscall.EBADF

But this sounds better:
`return ErrInvalidObject("ETag has changed")`

we could rename `ErrInvalidObjectName()` to `ErrInvalidObject()`

krishnasrinivas: @krisis @harshavardhana mentioned that he discussed this with you. The reason `ErrInvalidObject("ETag has changed")` is better than `syscall.EBADF` is if application logs the error it will know exactly why error was returned (instead of a generic bad FD)

krisis: @krishnasrinivas, I agree that `syscall.EBADF` doesn't provide context to the user.  How about `"object has been modified since last read"`? I find "ETag has changed" not entirely helpful. Thoughts?

harshavardhana: It should be encapsulated inside ErrInvalidObjectNamet() @krisis which provides ErrorResponse compatible error.  This way caller can look at the `errResp.Code == NoSuchKey` and can decide to re-open the connection to discard and start reading again.

krisis: @harshavardhana It doesn't make sense to return `NoSuchKey`. IMO `InvalidObjectState`[1] is a more apt error response code. thoughts? 

[1] http://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html#ErrorCodeList

harshavardhana: NoSuchKey meaning we don't see the same object anymore i.e based on the ETag.

harshavardhana:
Actually now that i think about it.  This is not the right fix - we should have been using `If-Match` tag in GetObject header and let server return proper error. 

http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectGET.html

deekoder:
is this PR still valid? Please advise if it needs to be closed then?

harshavardhana:
> is this PR still valid? Please advise if it needs to be closed then?

It is valid it needs to be re-done since we merged #670

krisis:
@harshavardhana @krishnasrinivas I have updated the PR to use recent changes in `c.getObject` method to include request headers.

harshavardhana:
> @harshavardhana @krishnasrinivas I have updated the PR to use recent changes in c.getObject method to include request headers.

Will test thanks @krisis

api-get-object.go
```diff
@@ -105,7 +105,7 @@ func (c Client) GetObject(bucketName, objectName string) (*Object, error) {
 							// Do not set objectInfo from the first readAt request because it will not get
 							// the whole object.
 							reqHeaders.SetRange(req.Offset, req.Offset+int64(len(req.Buffer))-1)
-							httpReader, _, err = c.getObject(bucketName, objectName, reqHeaders)
+							httpReader, objectInfo, err = c.getObject(bucketName, objectName, reqHeaders)
```
harshavardhana: Replacing `objectInfo` for each ReadAt changes the Size() of the object.. Is the Size() of the object expected to change . Stat() method is supposed to return the actual object size.

harshavardhana: Can you test this with all other conditions such as a combination of ReadAt() and then Read()/Seek() ?

krisis: >Replacing objectInfo for each ReadAt changes the Size() of the object.. Is the Size() of the object expected to change . Stat() method is supposed to return the actual object size

objectInfo (i.e, `minio.Object.objectInfo`) is not being replaced on each ReadAt. `if !o.objectInfoSet && !req.isReadAt` check in `doGetRequest` ensures that objectInfo is set only the first time and never from a ReadAt call.

harshavardhana: The problem is that Size returned is lesser than the original Size of the object. So it means that if the Read() is ensued after ReadAt() - would result in an EOF without reading the whole object.

krisis: >Can you test this with all other conditions such as a combination of ReadAt() and then Read()/Seek() ?

I have added a functional test that makes a ReadAt, Stat and ReadAt call on `minio.Object` in that order, in api_functional_v4_test.go. Between the two ReadAt calls the object is modified in the store. I shall see how I can add tests that will cover other sequences of `minio.Object` methods, especially Read, ReadAt and Stat.

harshavardhana: You need something like a user did following sequence of events tested.

- ReadAt() to read some bytes at an offset.
- Then proceeds to start Read() instead of ReadAt(). 

Or 

- ReadAt() to read some bytes at offset.
- Seek() try to seek past the offset+length used in ReadAt(). 

The reason previously `objectInfo` was not updated in ReadAt() was to ensure that it is updated only when a Read() or a Seek() is called. I am pretty sure setting objectInfo would cause some unknown situations.

harshavardhana: I see now there is an additional check.. to ignore objectInfo sent from ReadAt..

api_functional_v4_test.go
```diff
@@ -2422,3 +2423,75 @@ func TestFunctional(t *testing.T) {
...
+
+	// Confirm that a Stat() call in between doesn't change the Object's cached etag.
+	_, err = reader.Stat()
+	if err != nil {
+		t.Fatalf("Failed to stat object %s/%s: %v", bucketName, objectName, err)
+	}
+
+	// Read again only to find object contents have been modified since last read.
+	_, err = reader.ReadAt(b, int64(n))
+	if err.Error() != s3ErrorResponseMap["PreconditionFailed"] {
```
harshavardhana: Make sure that err is not nil. Since you need to fail for that as well.

api-get-object.go
krisis: >The reason previously objectInfo was not updated in ReadAt() was to ensure that it is updated only when a Read() or a Seek() is called. I am pretty sure setting objectInfo would cause some unknown situations during ReadAt().

There is a check in `doGetRequest` which ensures `objectInfo` is not updated. See https://github.com/minio/minio-go/pull/662/files#diff-0afcc28c2029fa14fe9a221e90b524e5R307.

api_functional_v4_test.go
```diff
@@ -2422,3 +2423,75 @@ func TestFunctional(t *testing.T) {
...
+
+	// Upload different contents to the same object while object is being read.
+	newContent := "goodbyeworld"
+	_, err = c.PutObject(bucketName, objectName, strings.NewReader(newContent), "application/text")
+	if err != nil {
+		t.Fatalf("Failed to upload %s/%s: %v", bucketName, objectName, err)
+	}
+
+	// Confirm that a Stat() call in between doesn't change the Object's cached etag.
+	_, err = reader.Stat()
```
harshavardhana: Shouldn't you be reading the cached value and validating it as well?

api_functional_v4_test.go
```diff
@@ -2422,3 +2423,75 @@ func TestFunctional(t *testing.T) {
...
+	objectName := "myobject"
+	content := "helloworld"
+	_, err = c.PutObject(bucketName, objectName, strings.NewReader(content), "application/text")
+	if err != nil {
+		t.Fatalf("Failed to upload %s/%s: %v", bucketName, objectName, err)
+	}
+
+	defer c.RemoveObject(bucketName, objectName)
+
+	reader, err := c.GetObject(bucketName, objectName)
```
harshavardhana: Close the reader for posterity.

api_functional_v4_test.go
```diff
@@ -2422,3 +2423,75 @@ func TestFunctional(t *testing.T) {
 		t.Fatal("Error: ", err)
 	}
 }
+
```
harshavardhana: A comment..

api_functional_v4_test.go
krisis: In this case ReadAt call should fail. If it returned `nil` it implies that this test failed. I can add `err != nil` for readability sake.

api_functional_v4_test.go
krisis: In this sequence of method calls, namely PutObject, ReadAt, PutObject, Stat ..., we can check for object size to be `len(newContent)`. If Read call replaced the (first) ReadAt, then the objectInfo.Size would be `len(content)`. In summary, Stat method may return stale objectInfo while Read and ReadAt methods will return error if the object is modified in the object store.

This brings us to if we should return error on Stat if the object is modified since the call to `GetObject`.

krisis:
@krishnasrinivas, you are right, the way we use objectInfo is not straightforward. If you see how doGetRequest uses objectInfo you will see why we need ETag saved separately. We can simplify the code structure in a different PR.

krishnasrinivas:
@krisis no, see this is all we need: https://github.com/krishnasrinivas/minio-go/commit/fa0e6407a7eccbfdaa8fbd46a5d10d13d801baee

krisis:
@krishnasrinivas, could you test your changes with Get object followed by ReadAt on the reader, putobject modifying the object and another ReadAt? Meanwhile I shall check how it is different from what is present in this PR.

krisis:
> @krisis no, see this is all we need: krishnasrinivas@fa0e640

This fix looks good to me. I will make a minor change where `var etag string` will be moved into the go-routine. I shall revert my changes and include this.

PR #665, opened 2017-04-24
MakeBucket: Retry with correct region

vadmeste:
The main purpose if this PR is to be able to find x-amz-bucket-region
when S3 server returns InvalidRegion error. That would help clients
to retry creating bucket with the correct region.


vadmeste:
Can this PR be reviewed again ?

api.go
```diff
@@ -544,10 +551,15 @@ func (c Client) executeMethod(method string, metadata requestMetadata) (res *htt
...
+
+		// Save the body.
+		errBodySeeker = bytes.NewReader(errBodyBytes)
+		res.Body = ioutil.NopCloser(errBodySeeker)
+
 		// Bucket region if set in error response and the error code dictates invalid region,
 		// we can retry the request with the new region.
-		if errResponse.Code == "InvalidRegion" && errResponse.Region != "" {
-			c.bucketLocCache.Set(metadata.bucketName, errResponse.Region)
+		if errResponse.Code == "InvalidRegion" && errResponse.Headers.Get("x-amz-bucket-region") != "" {
```
harshavardhana: Don't do this, set the Region properly and remove Headers from errorResponse structure. Let me send a PR to this PR.

harshavardhana: There is a chance for this to go into an infinite loop, we just want to do this once.

vadmeste: > Don't do this, set the Region properly and remove Headers from errorResponse structure.

My aim was to make the code reflects exactly the spec even if it looks more longer, the reason is that the code will be easier to read and be compared with the S3 spec. Also we are not the creator of the spec and any optimization could bring new problems.

harshavardhana: Region is there because it does exist in some operations on AWS S3, so we can re-purpose making it more cleaner and use it as part of region retry. I sent a PR which retains Headers but extracts the value and sets it properly for errorResponse instead of spilling the code.

api_functional_v4_test.go
```diff
@@ -801,7 +801,7 @@ func TestRemovePartiallyUploaded(t *testing.T) {
 	if err == nil {
 		t.Fatal("Error: PutObject should fail.")
 	}
-	if err.Error() != "Proactively closed to be verified later." {
+	if err.Error() != "proactively closed to be verified later" {
```
harshavardhana: Where did this change @vadmeste ?

api.go
```diff
@@ -538,14 +538,21 @@ func (c Client) executeMethod(method string, metadata requestMetadata) (res *htt
...
 		res.Body = ioutil.NopCloser(errBodySeeker)
 
 		// For errors verify if its retryable otherwise fail quickly.
 		errResponse := ToErrorResponse(httpRespToErrorResponse(res, metadata.bucketName, metadata.objectName))
-		// Bucket region if set in error response and the error code dictates invalid region,
-		// we can retry the request with the new region.
+
+		// Save the body.
+		errBodySeeker = bytes.NewReader(errBodyBytes)
+		res.Body = ioutil.NopCloser(errBodySeeker)
```
harshavardhana: Is this needed? there seems like something similar is done above.

api.go
```diff
@@ -578,26 +585,28 @@ func (c Client) newRequest(method string, metadata requestMetadata) (req *http.R
...
-	// Default all requests to "us-east-1" or "cn-north-1" (china region)
-	location := "us-east-1"
-	if s3utils.IsAmazonChinaEndpoint(c.endpointURL) {
-		// For china specifically we need to set everything to
-		// cn-north-1 for now, there is no easier way until AWS S3
-		// provides a cleaner compatible API across "us-east-1" and
-		// China region.
-		location = "cn-north-1"
-	}
+	if metadata.bucketLocation == "" {
```
harshavardhana: Is this necessary anymore?

api.go
vadmeste: > Is this needed? there seems like something similar is done above.

Yes, this is needed. You can see just above that httpRespToErrorResponse() is reading from res.Body so this change puts back res.Body

api_functional_v4_test.go
vadmeste: Test was failing without this change, 'proactively closed to be verified later' message should be the same as returned by the go routine created just above.

api.go
vadmeste: > Is this necessary anymore?

No, we don't need this anymore since you brought back makeBucketRequest(), I'll update

api.go
harshavardhana: Yes but this is wrong see below 

```
               // Save the body back again.
                errBodySeeker.Seek(0, 0) // Seek back to starting point.
                res.Body = ioutil.NopCloser(errBodySeeker)
``` 

We seek back.

PR #666, opened 2017-04-25
core: GetBucketPolicy API should return all errors.

### LINKED ISSUE #664: getBucketPolicy returns policy without statements instead of error NoSuchBucketPolicy
api-get-policy.go#81

Currently Mint fails on this, but probably this has been done on purpose. I see the following solutions:

* don't use Client::getBucketPolicy in Core, but instead, use different implementation in Core
* fix this code, and return an error code when error occurs
* fix in gateway router when we have a policy without statements, return NoSuchBucketPolicy
* fix mint and check for statements is null

harshavardhana:
GetBucketPolicy should return exact error received
from the server and let the caller manage it.

Fixes #664

### DIFF
--- a/api-get-policy.go
+++ b/api-get-policy.go
@@ -36,6 +36,10 @@ func (c Client) GetBucketPolicy(bucketName, objectPrefix string) (bucketPolicy p
 	}
 	policyInfo, err := c.getBucketPolicy(bucketName)
 	if err != nil {
+		errResponse := ToErrorResponse(err)
+		if errResponse.Code == "NoSuchBucketPolicy" {
+			return policy.BucketPolicyNone, nil
+		}
 		return policy.BucketPolicyNone, err
 	}
 	return policy.GetPolicy(policyInfo.Statements, bucketName, objectPrefix), nil
@@ -52,9 +56,18 @@ func (c Client) ListBucketPolicies(bucketName, objectPrefix string) (bucketPolic
 	}
 	policyInfo, err := c.getBucketPolicy(bucketName)
 	if err != nil {
+		errResponse := ToErrorResponse(err)
+		if errResponse.Code == "NoSuchBucketPolicy" {
+			return map[string]policy.BucketPolicy{}, nil
+		}
 		return map[string]policy.BucketPolicy{}, err
 	}
 	return policy.GetPolicies(policyInfo.Statements, bucketName), nil
+}
+
+// Default empty bucket access policy.
+var emptyBucketAccessPolicy = policy.BucketAccessPolicy{
+	Version: "2012-10-17",
 }
 
 // Request server for current bucket policy.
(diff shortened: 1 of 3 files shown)


api-get-policy.go
```diff
@@ -51,12 +52,18 @@ func (c Client) ListBucketPolicies(bucketName, objectPrefix string) (bucketPolic
...
 	policyInfo, err := c.getBucketPolicy(bucketName)
-	if err != nil {
+	errResponse := ToErrorResponse(err)
+	if err != nil && errResponse.Code != "NoSuchBucketPolicy" {
 		return map[string]policy.BucketPolicy{}, err
 	}
 	return policy.GetPolicies(policyInfo.Statements, bucketName), nil
 }
 
+// Default empty bucket access policy.
```
krisis: This can be defined as a const.
```go
const  emptyBucketPolicy = policy.BucketAccessPolicy {
  Version: "2012-10-17",
}
```

api-put-bucket.go
```diff
@@ -163,7 +163,8 @@ func (c Client) SetBucketPolicy(bucketName string, objectPrefix string, bucketPo
 	}
 
 	policyInfo, err := c.getBucketPolicy(bucketName)
```
krisis: if all callers to `getBucketPolicy` should check for "NoSuchBucketPolicy", can't we check for the corresponding `errResponse.Code` inside the function and return `sentinelBucketAccessPolicy, nil`?

harshavardhana: That is exactly what we don't want. All callers need not check for example Core API shouldn't check.

It is a change to reflect a core API should do lower implementation and return all errors. Special interpretation is left to the caller.

api-get-policy.go
harshavardhana: Sure

harshavardhana: Struct literals cannot be constants, keep it as var.

harshavardhana:
> Is it possible to add a functional test that simulates the issue being fixed here?

Functional test added as well.

api-get-policy.go
```diff
@@ -35,7 +35,8 @@ func (c Client) GetBucketPolicy(bucketName, objectPrefix string) (bucketPolicy p
 		return policy.BucketPolicyNone, err
 	}
 	policyInfo, err := c.getBucketPolicy(bucketName)
-	if err != nil {
+	errResponse := ToErrorResponse(err)
+	if err != nil && errResponse.Code != "NoSuchBucketPolicy" {
```
vadmeste: I am not sure how this change fixes #664 ? we are not returning any error when error is NoSuchBucketPolicy

harshavardhana: It fixes #664 because Gateway uses GetBucketPolicy() Core API directly, which would return NoSuchBucketPolicy i.e error when no bucket policy exists. This makes it easier for Gateway to have a pass-through error back to the client.

vadmeste: yes, but notice that with this change, GetBucketPolicy() will return err = nil when getBucketPolicy() returns NoSuchBucketPolicy.

harshavardhana: GetBucketPolicy is overridden Core API @vadmeste, what you are talking about is GetBucketPolicy(bucket, prefix) which is not changed and should be working as expected.

harshavardhana: Look at the functional test in core_test.go to get the better idea. @vadmeste

vadmeste: ok, understood.

vadmeste: I think it is better if we return policy.BucketPolicyNone when we have NoSuchBucketPolicy without going to policy.GetPolicy() later.

api-get-policy.go
```diff
@@ -51,12 +52,18 @@ func (c Client) ListBucketPolicies(bucketName, objectPrefix string) (bucketPolic
 		return map[string]policy.BucketPolicy{}, err
 	}
 	policyInfo, err := c.getBucketPolicy(bucketName)
-	if err != nil {
+	errResponse := ToErrorResponse(err)
+	if err != nil && errResponse.Code != "NoSuchBucketPolicy" {
```
vadmeste: I think it is better if we return map[string]policy.BucketPolicy{} when we have NoSuchBucketPolicy without going to policy.GetPolicies() later.

harshavardhana: Sure

api-get-policy.go
harshavardhana: Sure

PR #668, opened 2017-04-25
Use UNSIGNED-PAYLOAD if sha256 is not calculated

krishnasrinivas:
The issue was found when testing S3 gateway. In gateway setup, if client does
presignedPUT the sha256 value is not valid in which case we should use
UNSIGNED-PAYLOAD instead of calculating sha256sum() on empty byte array

Reference: https://github.com/minio/minio-go/pull/661

Tested api_functional_v4_test.go on localhost:9000 minio
Tested with `mc --debug` on aws-s3

### DIFF
--- a/api.go
+++ b/api.go
@@ -661,12 +661,8 @@ func (c Client) newRequest(method string, metadata requestMetadata) (req *http.R
 	case c.signature.isV4():
 		// Set sha256 sum for signature calculation only with signature version '4'.
 		shaHeader := unsignedPayload
-		if !c.secure {
-			if metadata.contentSHA256Bytes == nil {
-				shaHeader = hex.EncodeToString(sum256([]byte{}))
-			} else {
-				shaHeader = hex.EncodeToString(metadata.contentSHA256Bytes)
-			}
+		if len(metadata.contentSHA256Bytes) > 0 {
+			shaHeader = hex.EncodeToString(metadata.contentSHA256Bytes)
 		}
 		req.Header.Set("X-Amz-Content-Sha256", shaHeader)
 
--- a/api.go
+++ b/api.go
@@ -658,7 +658,10 @@ func (c Client) newRequest(method string, metadata requestMetadata) (req *http.R
 	case c.signature.isV2():
 		// Add signature version '2' authorization header.
 		req = s3signer.SignV2(*req, c.accessKeyID, c.secretAccessKey)
-	case c.signature.isV4():
+	case c.signature.isStreamingV4() && method == "PUT":
+		req = s3signer.StreamingSignV4(req, c.accessKeyID,
+			c.secretAccessKey, location, metadata.contentLength, time.Now().UTC())
+	default:
 		// Set sha256 sum for signature calculation only with signature version '4'.
 		shaHeader := unsignedPayload
 		if len(metadata.contentSHA256Bytes) > 0 {
@@ -668,9 +671,6 @@ func (c Client) newRequest(method string, metadata requestMetadata) (req *http.R
 
 		// Add signature version '4' authorization header.
 		req = s3signer.SignV4(*req, c.accessKeyID, c.secretAccessKey, location)
-	case c.signature.isStreamingV4():
-		req = s3signer.StreamingSignV4(req, c.accessKeyID,
-			c.secretAccessKey, location, metadata.contentLength, time.Now().UTC())
 	}
 
 	// Return request.
--- a/api.go
+++ b/api.go
@@ -654,11 +654,11 @@ func (c Client) newRequest(method string, metadata requestMetadata) (req *http.R
 		return req, nil
 	} // Sign the request for all authenticated requests.
 
-	if c.signature.isV2() {
+	switch {
+	case c.signature.isV2():
 		// Add signature version '2' authorization header.
 		req = s3signer.SignV2(*req, c.accessKeyID, c.secretAccessKey)
-	} else if c.signature.isV4() || c.signature.isStreamingV4() &&
-		method != "PUT" {
+	case c.signature.isV4():
 		// Set sha256 sum for signature calculation only with signature version '4'.
 		shaHeader := unsignedPayload
 		if !c.secure {
@@ -672,7 +672,7 @@ func (c Client) newRequest(method string, metadata requestMetadata) (req *http.R
 
 		// Add signature version '4' authorization header.
 		req = s3signer.SignV4(*req, c.accessKeyID, c.secretAccessKey, location)
-	} else if c.signature.isStreamingV4() {
+	case c.signature.isStreamingV4():
 		req = s3signer.StreamingSignV4(req, c.accessKeyID,
 			c.secretAccessKey, location, metadata.contentLength, time.Now().UTC())
 	}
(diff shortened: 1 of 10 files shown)


signature-type.go
```diff
@@ -27,6 +27,8 @@ const (
 	SignatureV4Streaming
 )
 
+var emptySHA256 = sum256([]byte{})
```
krisis: why is `[]byte{}` used in favour of `nil`? The default value for a slice is `nil` so we can avoid setting `metadata.contentSHA256Bytes` like we are doing in this PR.

krisis:
@krishnasrinivas Could you add a functional test case capturing [this](https://github.com/minio/minio-go/pull/661#discussion_r112855629)?

signature-type.go
krishnasrinivas: I copy/pasted it from the previous code:
`shaHeader = hex.EncodeToString(sum256([]byte{}))`

let me fix that.

krishnasrinivas:
> @krishnasrinivas Could you add a functional test case capturing this?

@krisis actually I see that we don't have functional tests for the Core functions. I could bring it as a separate commit as it will take time to write all the tests.

krisis:
@krishnasrinivas see core_test.go. @harshavardhana added one yesterday.

api.go
```diff
@@ -658,12 +658,8 @@ func (c Client) newRequest(method string, metadata requestMetadata) (req *http.R
 	case c.signature.isV4():
 		// Set sha256 sum for signature calculation only with signature version '4'.
 		shaHeader := unsignedPayload
-		if !c.secure {
-			if metadata.contentSHA256Bytes == nil {
-				shaHeader = hex.EncodeToString(sum256([]byte{}))
-			} else {
-				shaHeader = hex.EncodeToString(metadata.contentSHA256Bytes)
-			}
+		if len(metadata.contentSHA256Bytes) > 0 {
```
krisis: Why can't we check for `nil` like before? I.e,
```go
if metadata.contentSHA256Bytes != nil {
     // ....
}
```

krishnasrinivas: So that we do the right thing if the caller passes `[]byte{}` like how remco's s3 gateway patch does.

krisis: To quote you from https://github.com/minio/minio-go/pull/661#discussion_r112876148
> It can be fixed in a cleaner way if we want to fix it. i.e metadata.contentSHA256Bytes being nil should not indicate whether the body itself is empty. If the body is empty the caller should set metadata.contentSHA256Bytes to Hex(SHA256Hash("")) and If metadata.contentSHA256Bytes is set to nil then we simple use UNSIGNED-PAYLOAD

So, I thought `metadata.contentSHA256Bytes == nil`  meant the client wants to send unsigned payload and signed payload otherwise.

krisis: I was coming from the point that `nil` is a sentinel value for `[]T` and makes it easier to recognise as a pattern while reading code. `if len(contentSHA256Bytes) > 0` allows users to send `[]byte{}` and `nil` interchangeably, which is convenient, so this check is fine by me.

krishnasrinivas: I had it as nil, but changed it after it failed with remco's patch

harshavardhana: The story of 

```
slice := nil 

or 

var slice []byte
```

```
slice := []byte{}
```

is that  `if slice != nil` would work for the former but not the latter but functionally both are of same `len` and `cap`

Quite surprisingly this happens with encoding JSON objects (a nil slice encodes to null, while []string{} encodes to the JSON array []).

core_test.go
```diff
@@ -86,3 +91,108 @@ func TestGetBucketPolicy(t *testing.T) {
...
+	}
+
+	buf := make([]byte, minPartSize)
+
+	size, err := io.ReadFull(crand.Reader, buf)
+	if err != nil {
+		t.Fatal("Error:", err)
+	}
+	if size != minPartSize {
+		t.Fatalf("Error: number of bytes does not match, want %v, got %v\n", minPartSize*4, size)
```
krisis: Is the error message as expected? Shouldn't it be as below?
```go
t.Fatalf(..., minPartSize, size)
```

core_test.go
```diff
@@ -86,3 +91,108 @@ func TestGetBucketPolicy(t *testing.T) {
...
+	// Read the data back
+	r, err := c.GetObject(bucketName, objectName)
+	if err != nil {
+		t.Fatal("Error:", err, bucketName, objectName)
+	}
+
+	st, err := r.Stat()
+	if err != nil {
+		t.Fatal("Error:", err, bucketName, objectName)
+	}
```
krisis: for uniformity sake, shall we have one newline between any two code-blocks. Could you fix that in this function?

api.go
```diff
@@ -651,11 +651,11 @@ func (c Client) newRequest(method string, metadata requestMetadata) (req *http.R
...
 	} // Sign the request for all authenticated requests.
 
-	if c.signature.isV2() {
+	switch {
+	case c.signature.isV2():
 		// Add signature version '2' authorization header.
 		req = s3signer.SignV2(*req, c.accessKeyID, c.secretAccessKey)
-	} else if c.signature.isV4() || c.signature.isStreamingV4() &&
-		method != "PUT" {
+	case c.signature.isV4():
```
vadmeste: Doesn't this change break something ?
I think it should be:
```
case c.signature.isV4():
   fallthrough
case c.signature.isStreamingV4() && method != "PUT":
   ..
```

krishnasrinivas: @vadmeste the previous code was not looking correct.
```
	} else if c.signature.isV4() || c.signature.isStreamingV4() &&
		method != "PUT" {
```

I reviewed the code and saw that StreamingSignature is enabled only for PUT request. Hence this code was not necessary. Let me know if I am reading anything incorrectly.

vadmeste: > I reviewed the code and saw that StreamingSignature is enabled only for PUT request. Hence this code was not necessary. Let me know if I am reading anything incorrectly.

yes you are right, a request is stream v4 only when we are going to upload an object or a part, thanks.

krisis: @krishnasrinivas the check is still required since the client's signature type is set before we know if the upload is going to be a single PUT object or a multipart upload. Streaming signature applies only to `http.MethodPUT`. API calls corresponding to get incomplete uploads, initiate new multipart, etc should use signature v4. Having said that, you could see if it can be refactored to be obvious to the reader and maintainable in the long run.

The current PR fails in `go test -v -run=TestPutObjectStreaming` when we make the following change. I have made the size of object to be uploaded to `64MB + 1` bytes.
```
diff --git a/api_functional_v4_test.go b/api_functional_v4_test.go
index dacbba8..c7c4a00 100644
--- a/api_functional_v4_test.go
+++ b/api_functional_v4_test.go
@@ -392,7 +392,7 @@ func TestPutObjectStreaming(t *testing.T) {
        }
 
        // Upload an object.
-       sizes := []int64{0, 64*1024 - 1, 64 * 1024}
+       sizes := []int64{0, 64*1024 - 1, 64*1024*1024 + 1}
        objectName := "test-object"
        for i, size := range sizes {
                data := bytes.Repeat([]byte("a"), int(size))
```

krishnasrinivas: @krisis you are right, let me fix that

PR #670, opened 2017-04-26
api, core: Add new Core API {Get,Stat}Object with pre-conditions.

### LINKED ISSUE #669: Support new API GetObjectPartial with preconditions
Reference: https://github.com/minio/minio/issues/3521

We need the following API in minio-go to fix the race explained in the issue above:
```
GetObject(bucket, object string,  startRange, length int, preconditions map[string]string, writer io.Writer) (err error) {
}
```

harshavardhana:
Implements a new API to provide a way to set headers
for GetObject(), StatObject() request such as to

 - read partial data starting at offsets.
 - read only if etag matches.
 - read only if modtime matches.
 - read only if etag doesn't match.
 - read only if modtime doesn't match.

Fixes #669

### DIFF
--- a/api-get-object-file.go
+++ b/api-get-object-file.go
@@ -78,8 +78,13 @@ func (c Client) FGetObject(bucketName, objectName, filePath string) error {
 		return err
 	}
 
+	// Initialize get object request headers to set the
+	// appropriate range offsets to read from.
+	reqHeaders := NewGetReqHeaders()
+	reqHeaders.SetRange(st.Size(), 0)
+
 	// Seek to current position for incoming reader.
-	objectReader, objectStat, err := c.getObject(bucketName, objectName, st.Size(), 0)
+	objectReader, objectStat, err := c.getObject(bucketName, objectName, reqHeaders)
 	if err != nil {
 		return err
 	}
--- a/core_test.go
+++ b/core_test.go
@@ -29,6 +29,165 @@ import (
 	"time"
 )
 
+// Tests for Core GetObject() function.
+func TestGetObjectCore(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping functional tests for the short runs")
+	}
+
+	// Seed random based on current time.
+	rand.Seed(time.Now().Unix())
+
+	// Instantiate new minio core client object.
+	c, err := NewCore(
+		os.Getenv("S3_ADDRESS"),
+		os.Getenv("ACCESS_KEY"),
+		os.Getenv("SECRET_KEY"),
+		mustParseBool(os.Getenv("S3_SECURE")),
+	)
+	if err != nil {
+		t.Fatal("Error:", err)
+	}
+
+	// Enable tracing, write to stderr.
+	// c.TraceOn(os.Stderr)
+
+	// Set user agent.
+	c.SetAppInfo("Minio-go-FunctionalTest", "0.1.0")
+
+	// Generate a new random bucket name.
+	bucketName := randString(60, rand.NewSource(time.Now().UnixNano()), "minio-go-test")
+
+	// Make a new bucket.
+	err = c.MakeBucket(bucketName, "us-east-1")
+	if err != nil {
+		t.Fatal("Error:", err, bucketName)
+	}
+
+	// Generate data more than 32K
+	buf := bytes.Repeat([]byte("3"), rand.Intn(1<<20)+32*1024)
+
+	// Save the data
+	objectName := randString(60, rand.NewSource(time.Now().UnixNano()), "")
+	n, err := c.Client.PutObject(bucketName, objectName, bytes.NewReader(buf), "binary/octet-stream")
+	if err != nil {
+		t.Fatal("Error:", err, bucketName, objectName)
+	}
+
+	if n != int64(len(buf)) {
+		t.Fatalf("Error: number of bytes does not match, want %v, got %v\n", len(buf), n)
+	}
+
+	reqHeaders := NewGetReqHeaders()
+
+	offset := int64(2048)
+
+	// read directly
+	buf1 := make([]byte, 512)
+	buf2 := make([]byte, 512)
+	buf3 := make([]byte, n)
+
+	reqHeaders.SetRange(offset, offset+int64(len(buf1))-1)
+	reader, objectInfo, err := c.GetObject(bucketName, objectName, reqHeaders)
+	if err != nil {
+		t.Fatal(err)
+	}
+	m, err := io.ReadFull(reader, buf1)
+	if err != nil {
+		reader.Close()
+		t.Fatal(err)
+	}
+	reader.Close()
+
+	if objectInfo.Size != int64(m) {
+		t.Fatalf("Error: GetObject read shorter bytes before reaching EOF, want %v, got %v\n", objectInfo.Size, m)
+	}
+	if !bytes.Equal(buf1, buf[offset:offset+512]) {
+		t.Fatal("Error: Incorrect read between two GetObject from same offset.")
+	}
+	offset += 512
+
+	reqHeaders.SetRange(offset, offset+int64(len(buf2))-1)
+	reader, objectInfo, err = c.GetObject(bucketName, objectName, reqHeaders)
+	if err != nil {
+		t.Fatal(err)
+	}
+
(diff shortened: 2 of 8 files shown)


api-get-conditions.go
```diff
@@ -0,0 +1,89 @@
...
+	"net/http"
+	"time"
+)
+
+// GetConditions - get conditions implement methods for setting
+// conditions for a GetObject request.
+// http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectGET.html
+type GetConditions struct {
+	http.Header
+}
```
krishnasrinivas: why not:
`type PreConditions http.Header`

api-get-object-file.go
```diff
@@ -78,8 +78,13 @@ func (c Client) FGetObject(bucketName, objectName, filePath string) error {
 		return err
 	}
 
+	// Initialize get object conditions and set the appropriate
+	// range offsets to read from.
+	conds := GetConditions{}
+	conds.SetRange(st.Size(), 0)
```
krishnasrinivas: will panic as the map is nil

api-get-conditions.go
harshavardhana: Too generic we already have CopyConditions.

api-get-object-file.go
harshavardhana: Ah will change this.

api-get-conditions.go
krishnasrinivas: But PreConditions are generic header information ... i.e it can apply to HEAD too if required.

`SetRange` does not sound right for a "Conditions" type.

You could name it just `Header`

harshavardhana: Ah you are right.

harshavardhana: SetRange is appropriate as it is part of the conditions here http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectGET.html - Header is not conveying any information.

pre-conditions.go
```diff
@@ -0,0 +1,89 @@
...
+		return ErrInvalidArgument("Modified since cannot be empty.")
+	}
+	c.Set("If-Modified-Since", modTime.Format(http.TimeFormat))
+	return nil
+}
+
+// SetRange - set the start and end offset of the object to be read.
+// See https://tools.ietf.org/html/rfc7233#section-3.1 for reference.
+func (c PreConditions) SetRange(start, end int64) error {
+	if start < 0 || end > 0 && end < start {
```
vadmeste: Can you reject when end is negative, it looks it can pass when end = -1 and start = 1 for example.

api-get-object-partial.go
```diff
@@ -0,0 +1,17 @@
+package minio
```
vadmeste: Looks like this file is not needed.

api-get-object-file.go
```diff
@@ -78,8 +78,13 @@ func (c Client) FGetObject(bucketName, objectName, filePath string) error {
 		return err
 	}
 
+	// Initialize get object conditions and set the appropriate
+	// range offsets to read from.
+	conds := NewPreConditions()
+	conds.SetRange(st.Size(), 0)
```
vadmeste: This should be `conds.SetRange(0, st.Size()-1)`

pre-conditions.go
```diff
@@ -0,0 +1,89 @@
...
+	if modTime.IsZero() {
+		return ErrInvalidArgument("Modified since cannot be empty.")
+	}
+	c.Set("If-Modified-Since", modTime.Format(http.TimeFormat))
+	return nil
+}
+
+// SetRange - set the start and end offset of the object to be read.
+// See https://tools.ietf.org/html/rfc7233#section-3.1 for reference.
+func (c PreConditions) SetRange(start, end int64) error {
```
vadmeste: We can't download only the first byte with this primitive.

SetRange(0, 0) should be able to return the first byte.

I think `end` should be -1 when we need to indicate that we want to download from start until the end of the target object.

api-get-conditions.go
krishnasrinivas: SetRange is under `Request Headers` section not conditions. i.e Preconditions belongs to `Request Headers` but Range does not belong to `Preconditions`

harshavardhana: Range also is part of a pre-condition such that you need to validate the requested range first. So i don't agree.

pre-conditions.go
harshavardhana: End negative is a valid case.. though.

api-get-object-partial.go
harshavardhana: will remove.

pre-conditions.go
harshavardhana: Yes there are more cases in pure HTTP, But all cases are not supported by AWS S3.

api-get-conditions.go
harshavardhana: From a HTTP point of view you are right Range is not part of Pre-Conditions - https://tools.ietf.org/html/rfc7232#section-3 one needs to set If-Range. But AWS S3 doesn't support that yet.

harshavardhana:
Can you check again @krishnasrinivas @vadmeste

request-headers.go
```diff
@@ -0,0 +1,105 @@
...
+	case start <= 0 && end < 0:
+		// Read everything until the 'end'. `bytes=-N`
+		c.Set("Range", fmt.Sprintf("bytes=%d", end))
+	case start > 0 && end == 0:
+		// Read everything starting from offset 'start'. `bytes=N-`
+		c.Set("Range", fmt.Sprintf("bytes=%d-", start))
+	case start > 0 && end > 0 && end > start:
+		// Read everything starting at 'start' till the 'end'. `bytes=N-M`
+		c.Set("Range", fmt.Sprintf("bytes=%d-%d", start, end))
+	case start == 0 && end == 0:
```
vadmeste: Only the first byte?

I guess S3 can support reading one byte in other positions. You can merge this case with the previous one, so it would be: `case start > 0 && end > 0 && end >= start:`

harshavardhana: Done

core_test.go
```diff
@@ -17,13 +17,155 @@
...
+	reqHeaders = NewGetReqHeaders()
+	reqHeaders.SetMatchETagExcept("etag")
+	reader, objectInfo, err = c.GetObject(bucketName, objectName, reqHeaders)
+	if err != nil {
+		t.Fatal(err)
+	}
+	io.ReadFull(reader, buf3)
+	reader.Close()
+
+	if objectInfo.Size != int64(len(buf3)) {
```
krisis: The above check doesn't guarantee that `len(buf3)` bytes were read from `reader`. The check should be,
```go
n, err = io.ReadFull(reader, buf3)
if err != nil {
  t.Fatal(err)
}

if objectInfo.Size != n {
  t.Fatal(err)
}
```

api-get-object-file.go
```diff
@@ -78,8 +78,13 @@ func (c Client) FGetObject(bucketName, objectName, filePath string) error {
 		return err
 	}
 
+	// Initialize get object conditions and set the appropriate
+	// range offsets to read from.
+	reqHeaders := NewGetReqHeaders()
+	reqHeaders.SetRange(st.Size(), 0)
```
krisis: As per, `case start > 0 && end == 0:` in `SetRange(...)` method  this `reqHeaders` will request bytes of object starting from `st.Size()` to EOF, which is not intended. Shouldn't it be `SetRange(0, -st.Size())`?

core_test.go
harshavardhana: Why do you think that ?

api-get-object-file.go
harshavardhana: I don't think so..

core_test.go
krisis: Number of bytes read would be different than `len(buf3)` if `io.ReadFull` returned with `io.UnexpectedEOF`.

api-get-object-file.go
krisis: @harshavardhana which case statement would match `SetRange(st.Size(), 0)`, where `st.Size() > 0`?

krishnasrinivas: @krisis it looks confusing but st.Size() is actually size of the partially downloaded file, not the size of the object

PR #674, opened 2017-05-04
Fix range header bounds check in SetRange()

donatello:
Should fix https://github.com/minio/minio/issues/4243

### DIFF
--- a/request-headers.go
+++ b/request-headers.go
@@ -1,5 +1,5 @@
 /*
- * Minio Go Library for Amazon S3 Compatible Cloud Storage (C) 2016 Minio, Inc.
+ * Minio Go Library for Amazon S3 Compatible Cloud Storage (C) 2016-17 Minio, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


request-headers.go
```diff
@@ -89,17 +89,18 @@ func (c RequestHeaders) SetRange(start, end int64) error {
 	case start > 0 && end == 0:
 		// Read everything starting from offset 'start'. `bytes=N-`
 		c.Set("Range", fmt.Sprintf("bytes=%d-", start))
-	case start > 0 && end > 0 && end >= start:
+	case start > 0 && end >= start:
 		// Read everything starting at 'start' till the 'end'. `bytes=N-M`
```
harshavardhana: Won't this still retun an error if start is 0 and end > 0 ?

request-headers.go
```diff
@@ -84,22 +84,21 @@ func (c RequestHeaders) SetModified(modTime time.Time) error {
...
-		// Read everything until the 'end'. `bytes=-N`
+		// Read last '-end' bytes. `bytes=-N`
 		c.Set("Range", fmt.Sprintf("bytes=%d", end))
 	case start > 0 && end == 0:
 		// Read everything starting from offset 'start'. `bytes=N-`
 		c.Set("Range", fmt.Sprintf("bytes=%d-", start))
-	case start > 0 && end > 0 && end >= start:
+	case start >= 0 && end >= start:
 		// Read everything starting at 'start' till the 'end'. `bytes=N-M`
 		c.Set("Range", fmt.Sprintf("bytes=%d-%d", start, end))
```
harshavardhana: Can you test if 0-0 indeed reads the whole object?

donatello: It does not.

donatello:
There are two conflicting cases here:

1. Read 1 byte starting at the first byte - in the current implementation this is supported. The interpretation of `start==0 && end==0` is `bytes=0-0` (this is correct according to RFC). 
2. Read whole file starting from byte 0 - in the current implementation this is not supported. This is the same as not setting the header at all - so the `SetHeader` method should not be called.

donatello:
If we implement in the opposite way, then:

`start == 0 && end == 0` would generate header as `bytes=0-` meaning read whole file from byte 0.

However, we lose the ability to express "read 1 byte starting from byte 0". So I chose to go with the latter. Still a better `SetHeader` API could be constructed.

request-headers.go
```diff
@@ -83,23 +83,28 @@ func (c RequestHeaders) SetModified(modTime time.Time) error {
...
 		c.Set("Range", fmt.Sprintf("bytes=%d", end))
-	case start > 0 && end == 0:
-		// Read everything starting from offset 'start'. `bytes=N-`
+	case 0 < start && end == 0:
+		// Read everything starting from offset
+		// 'start'. `bytes=N-`.
 		c.Set("Range", fmt.Sprintf("bytes=%d-", start))
-	case start > 0 && end > 0 && end >= start:
-		// Read everything starting at 'start' till the 'end'. `bytes=N-M`
+	case 0 <= start && start <= end:
```
harshavardhana: This looks good, can you add some tests for this as well? @donatello

harshavardhana:
> If we implement in the opposite way, then:
> 
> start == 0 && end == 0 would generate header as bytes=0- meaning read whole file from byte 0.
> 
> However, we lose the ability to express "read 1 byte starting from byte 0". So I chose to go with the approach in the previous comment. Still a better SetHeader API could be constructed.

This is good serves the purpose here.

donatello:
@harshavardhana Tests added.

request-headers_test.go
```diff
@@ -0,0 +1,40 @@
+package minio
```
harshavardhana: Add license header..

donatello: Done

PR #676, opened 2017-05-10
Adding a timeout setter for minio http.Client

BarthV:
Hello :)

First, thanks for this useful project !

For my project integration, I have the need to set a specific timeout when retrieving a S3 object.
It seems that there's not currently setter for underneath http.Client used by minio.

This public func should allow external code to set this timeout :).

Thanks for this review.

### DIFF
--- a/api.go
+++ b/api.go
@@ -34,6 +34,7 @@ import (
 	"strings"
 	"sync"
 	"time"
+	"errors"
 
 	"github.com/minio/minio-go/pkg/s3signer"
 	"github.com/minio/minio-go/pkg/s3utils"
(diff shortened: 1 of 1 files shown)


harshavardhana:
Thanks for the PR @BarthV  you can already do this in the following manner.  It is also not advisable to set timeouts on http.Client directly which affects all your upload and download operations.

```go
 	var transport http.RoundTripper = &http.Transport{
		Proxy: http.ProxyFromEnvironment,
		DialContext: (&net.Dialer{
			Timeout:   30 * time.Second,
			KeepAlive: 30 * time.Second,
		}).DialContext,
		MaxIdleConns:          100,
		IdleConnTimeout:       90 * time.Second,
		TLSHandshakeTimeout:   10 * time.Second,
		ExpectContinueTimeout: 1 * time.Second,
		TLSClientConfig:       tlsConfig,
	}
	api, err := minio.New(hostName, config.AccessKey, config.SecretKey, useTLS)
	if err != nil {
		fmt.Println(err)
		return
	}
	api.SetCustomTransport(transport)
```

BarthV:
ok thanks for this tips ! :)

PR #681, opened 2017-05-14
api: getReaderSize() should honor seeked file descriptors.

### LINKED ISSUE #680: Uploading file with offset uploads the whole file
When passing an open file that has been seeked in to `PutObject()`, an error occurs because it thinks it needs to upload the whole file. The following program shows it (maybe needs the code from #678):

```go
package main

import (
	crand "crypto/rand"
	"fmt"
	"io"
	"io/ioutil"
	"math/rand"
	"os"
	"time"

)

func main() {
	rand.Seed(time.Now().UnixNano())

	var length = 120000
	data := make([]byte, length)
	_, err := io.ReadFull(crand.Reader, data)
	if err != nil {
		panic(err)
	}

	tempfile, err := ioutil.TempFile("", "minio-go-upload-test-")
	if err != nil {
		panic(err)
	}

	if err = tempfile.Close(); err != nil {
		panic(err)
	}

	if err = ioutil.WriteFile(tempfile.Name(), data, 0600); err != nil {
		panic(err)
	}

	tempfile, err = os.Open(tempfile.Name())
	if err != nil {
		panic(err)
	}

	client, err := minio.New("s3.amazonaws.com", os.Getenv("AWS_ACCESS_KEY_ID"), os.Getenv("AWS_SECRET_ACCESS_KEY"), false)
	if err != nil {
		panic(err)
	}

	// client.TraceOn(os.Stderr)

	name := fmt.Sprintf("test-file-%v", rand.Uint32())

	offset := 123
	if _, err := tempfile.Seek(int64(offset), io.SeekStart); err != nil {
		panic(err)
	}

	n, err := client.PutObject(os.Getenv("BUCKET"), name, tempfile, "binary/octet-stream")
	fmt.Printf("wrote %d bytes to %v err %v\n", n, name, err)

	if n != int64(length-offset) {
		panic(fmt.Sprintf("invalid length returned, want %v, got %v", int64(length-offset), n))
	}
}
```

```
$ BUCKET=restic-test-travis go run main.go
wrote 0 bytes to test-file-1876261662 err Data read ‘119877’ is not equal to the size ‘120000’ of the input Reader.
panic: invalid length returned, want 119877, got 0

goroutine 1 [running]:
main.main()
	/home/fd0/tmp/main.go:61 +0x775
exit status 2
```

harshavardhana:
In situations where a file has been Seeked, we need to
start reading from the offset which it indeed happens.
But our reader size calculation needs to honor this to
be accurate.

Fixes #680

### DIFF
--- a/api_functional_v4_test.go
+++ b/api_functional_v4_test.go
@@ -2477,3 +2477,97 @@ func TestGetObjectObjectModified(t *testing.T) {
 		t.Errorf("Expected ReadAt to fail with error %s but received %s", s3ErrorResponseMap["PreconditionFailed"], err.Error())
 	}
 }
+
+// Test validates putObject to upload a file seeked at a given offset.
+func TestPutObjectUploadSeekedObject(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping functional tests for the short runs")
+	}
+
+	// Instantiate new minio client object.
+	c, err := NewV4(
+		os.Getenv("S3_ADDRESS"),
+		os.Getenv("ACCESS_KEY"),
+		os.Getenv("SECRET_KEY"),
+		mustParseBool(os.Getenv("S3_SECURE")),
+	)
+	if err != nil {
+		t.Fatal("Error:", err)
+	}
+
+	// Enable tracing, write to stderr.
+	// c.TraceOn(os.Stderr)
+
+	// Set user agent.
+	c.SetAppInfo("Minio-go-FunctionalTest", "0.1.0")
+
+	// Make a new bucket.
+	bucketName := randString(60, rand.NewSource(time.Now().UnixNano()), "minio-go-test")
+	err = c.MakeBucket(bucketName, "us-east-1")
+	if err != nil {
+		t.Fatal("Error:", err, bucketName)
+	}
+	defer c.RemoveBucket(bucketName)
+
+	tempfile, err := ioutil.TempFile("", "minio-go-upload-test-")
+	if err != nil {
+		t.Fatal("Error:", err)
+	}
+
+	var length = 120000
+	data := bytes.Repeat([]byte("1"), length)
+
+	if _, err = tempfile.Write(data); err != nil {
+		t.Fatal("Error:", err)
+	}
+
+	objectName := fmt.Sprintf("test-file-%v", rand.Uint32())
+
+	offset := length / 2
+	if _, err := tempfile.Seek(int64(offset), 0); err != nil {
+		t.Fatal("Error:", err)
+	}
+
+	n, err := c.PutObject(bucketName, objectName, tempfile, "binary/octet-stream")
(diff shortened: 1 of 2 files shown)


harshavardhana:
PR updated, tests fixed ready for review @vadmeste @krisis

harshavardhana:
PR is ready to review @vadmeste @krisis

api_functional_v4_test.go
```diff
@@ -2498,3 +2498,102 @@ func TestGetObjectObjectModified(t *testing.T) {
...
+	}
+
+	tempfile, err = os.Open(tempfile.Name())
+	if err != nil {
+		t.Fatal("Error:", err)
+	}
+
+	objectName := fmt.Sprintf("test-file-%v", rand.Uint32())
+
+	offset := 123
```
krisis: Suggestion: we could have `offset` as `length/2` (say).  This ensures that offset would be valid independent of the `length` we choose. Having `offset` as an independent constant allows for situations in the future where `offset` may be greater than `length` etc. We have seen similar cases with our unit tests in the past, e.g index `13` was used to access a variably sized buffer.

harshavardhana: When would that happen in this case? this is a self contained test. But it is do-able to just do length/2.

api_functional_v4_test.go
```diff
@@ -2498,3 +2498,102 @@ func TestGetObjectObjectModified(t *testing.T) {
...
+	if err = ioutil.WriteFile(tempfile.Name(), data, 0600); err != nil {
+		t.Fatal("Error:", err)
+	}
+
+	tempfile, err = os.Open(tempfile.Name())
+	if err != nil {
+		t.Fatal("Error:", err)
+	}
+
+	objectName := fmt.Sprintf("test-file-%v", rand.Uint32())
```
harshavardhana: Just for uniqueness.. Each test needs to be self contained and not overwrite each other with generic names.

api_functional_v4_test.go
```diff
@@ -2498,3 +2498,102 @@ func TestGetObjectObjectModified(t *testing.T) {
...
+	if err != nil {
+		t.Fatal("Error:", err, bucketName)
+	}
+	defer c.RemoveBucket(bucketName)
+
+	tempfile, err := ioutil.TempFile("", "minio-go-upload-test-")
+	if err != nil {
+		t.Fatal("Error:", err)
+	}
+
```
harshavardhana: We can.. let me add that.

api_functional_v4_test.go
krisis: There is no issue with this PR. It is possible that constants for `length` and `offset` may be changed independently leading to an invalid test case accidentally. So, this is only a suggestion.

harshavardhana: Fixed.

PR #683, opened 2017-05-14
tests: Use bytes.Repeat() when generating big data

### LINKED ISSUE #682: Avoid using rand source in tests as they are slow
Use bytes.Repeat instead of generating random data to accelerate a bit go test time.

vadmeste:
Avoid using random source when generating large data. This will
heavily reduces functional tests time.

Fixes  #682 

In my machine, tests execution time pass from 280 seconds to 37 seconds.

### DIFF
--- a/api_functional_v4_test.go
+++ b/api_functional_v4_test.go
@@ -18,7 +18,6 @@ package minio
 
 import (
 	"bytes"
-	crand "crypto/rand"
 	"encoding/hex"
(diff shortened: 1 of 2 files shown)


harshavardhana:
Much needed thanks @vadmeste

hackintoshrao:
This fix was introduced, but some regression reverted it. Thanks.

PR #691, opened 2017-05-23
Fill ObjectPart.LastModified time in PutObjectPart

krisis:
Fixes https://github.com/minio/minio/issues/4400

### DIFF
--- a/api-put-object-multipart.go
+++ b/api-put-object-multipart.go
@@ -32,6 +32,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"time"
 )
 
 // Comprehensive put object operation involving multipart resumable uploads.
@@ -413,6 +414,11 @@ func (c Client) uploadPart(bucketName, objectName, uploadID string, reader io.Re
 	// Trim off the odd double quotes from ETag in the beginning and end.
 	objPart.ETag = strings.TrimPrefix(resp.Header.Get("ETag"), "\"")
 	objPart.ETag = strings.TrimSuffix(objPart.ETag, "\"")
+	date, err := time.Parse(http.TimeFormat, resp.Header.Get("Date"))
+	if err != nil {
+		return ObjectPart{}, nil
+	}
+	objPart.LastModified = date
 	return objPart, nil
 }
 


vadmeste:
LGTM but I wonder if last-modified is used at all.

harshavardhana:
What is the S3 behavior here ?  @krisis. AFAIK Date header is not the representation of last-modified header.

krisis:
@harshavardhana Hmm, as per http://docs.aws.amazon.com/AmazonS3/latest/API/RESTCommonResponseHeaders.html,

> The date and time Amazon S3 responded, for example, Wed, 01 Mar 2006 12:00:00 GMT.
> Type: String

for a UploadPart API request, it is correct to assume that `Date` header is the last modified time.

harshavardhana:
Do you have a sample code with AWS SDK?

krisis:
@harshavardhana, do you mean something like this? This snippet is initialized to work with `minio gateway s3`. I am not sure I understood what you wish to see.
```go
package main

import (
	"log"
	"strings"
	"time"

)

func main() {
	clnt, err := minio.NewCore("localhost:9000", "minio", "minio123", false)
	if err != nil {
		log.Fatalln(err)
	}

	bucket := "krisis"
	object := "multipart-obj"
	uploadID, err := clnt.NewMultipartUpload(bucket, object, nil)
	if err != nil {
		log.Fatalln(err)
	}
	part, err := clnt.PutObjectPart(bucket, object, uploadID, 1, 5, strings.NewReader("aaaaa"), nil, nil)
	if err != nil {
		log.Fatalln(err)
	}

	zeroTime := time.Time{}
	if part.LastModified == zeroTime {
		log.Fatalln("ObjectPart last modified time is not filled")
	}
}
```

krisis:
@harshavardhana If we are not comfortable returning the `Date` header value as `ObjectPart.LastModifiedTime`, do you have an alternative we can use instead? If there is none, we should remove `LastModifiedTime` from `ObjectPart` type. Thoughts?

harshavardhana:
We shouldn't return LastModified at all for both cases 

- http://docs.aws.amazon.com/sdk-for-go/api/service/s3/#UploadPartOutput 
- http://docs.aws.amazon.com/sdk-for-go/api/service/s3/#PutObjectOutput

So treating Date header as Last-Modified won't be correct. For GET and HEAD we treat Last-Modified correctly.  So we should remove LastModified from ObjectPart struct to avoid the confusion,  it is safe since its never used and interpreted. 

> I am not sure I understood what you wish to see.

What i wanted to see was an aws-sdk-go example with UploadPart() output indicating if they do print Last-Modified, if they don't then we shouldn't.

krisis:
@harshavardhana makes sense. I shall close this PR.

PR #694, opened 2017-05-28
get: Fix reading an object if its size is unknown

### LINKED ISSUE #693: Google Cloud Storage giving incorrect size and prematurely closing input reader
I'm having the following error when using the command ```mc cat gcs/bucket/myfile.dat```

```
mc: <ERROR> Unable to read from `gcs/bucket/myfile.dat`. Input reader closed pre-maturely. Expected `528046` bytes, but only received `32768` bytes.
```

Where ```gcs``` is google cloud storage. I'm having a very related issue in my code, which attempts to access the same object & bucket, but no error is produced (or I have mistakenly thrown it away? I produced all code below so please do let me know if I've made a mistake in handling minio errors). Instead, the data is simply a small truncated portion of the top of my original file. 

Strangest of all, calling ```.Stat()``` on the object gives me a struct that looks like this:

```
(minio.ObjectInfo) {
 ETag: (string) (len=32) "6b76194ff4b0861488c71a86ba7cc6fb",
 Key: (string) (len=23) "snip",
 LastModified: (time.Time) 2017-05-28 00:26:29 +0000 UTC,
 Size: (int64) -1,
 ContentType: (string) (len=24) "application/octet-stream",
 Metadata: (http.Header) (len=5) {
  (string) (len=20) "X-Goog-Storage-Class": ([]string) (len=1 cap=1) {
   (string) (len=8) "STANDARD"
  },
  (string) (len=11) "X-Goog-Hash": ([]string) (len=2 cap=2) {
   (string) (len=15) "crc32c=/viWNw==",
   (string) (len=28) "md5=a3YZT/SwhhSIxxqGunzG+w=="
  },
  (string) (len=20) "X-Guploader-Uploadid": ([]string) (len=1 cap=1) {
   (string) (len=98) "snip (giant base64 value)"
  },
  (string) (len=13) "Cache-Control": ([]string) (len=1 cap=1) {
   (string) (len=18) "private, max-age=0"
  },
  (string) (len=7) "Expires": ([]string) (len=1 cap=1) {
   (string) (len=29) "Sun, 28 May 2017 00:26:34 GMT"
  }
 },
 Owner: (struct { DisplayName string "json:\"name\""; ID string "json:\"id\"" }) {
  DisplayName: (string) "",
  ID: (string) ""
 },
 StorageClass: (string) "",
 Err: (error) <nil>
}
```

Note the size of ```-1```

All of this code works with an actual instance of minio as the backend, but since I've switched to GCS I'm running into the issue described above.

I create the client using this code

```go
mClient, err = minio.New(os.Getenv("MINIO_HOST"), 
                         os.Getenv("MINIO_KEY"), 
                         os.Getenv("MINIO_SECRET"), 
                         os.Getenv("MINIO_SSL") == "true")
```

I'm using this code to create the object when:

```go
exists, err := mClient.BucketExists(to)
if err != nil {
    mErr, ok := err.(minio.ErrorResponse)
    if ok && mErr.Code == "NoSuchBucket" {
        err = nil //we still need to create the bucket
    }
}

if !exists && err == nil {
    err = mClient.MakeBucket(to, os.Getenv("MINIO_REGION"))
}

....

const contentType = "application/octet-stream"
data := bytes.NewReader(buf.Bytes())
_, err = mClient.PutObject(to, name, data, contentType)
if err != nil {
    return
}
```

In case it's relevant, the file is just a series of 1 byte numbers, gzipped. As noted above I use the content MIME type of ```application/octet-stream```, and I do not report the gzip encoding of the file to google cloud storage via a ```Content-Encoding``` header.

Finally, I access the object using this code:
```go
mObj, err := mClient.GetObject(objectBucket, getObjectName(model))
if err != nil {
    return
}
stat, err = mObj.Stat()
if err != nil {
    //this returns without error because the exists bool (also a return value) is implicitly false
    //and "not existing" isn't an error in checking for the object
    if mErr, ok := err.(minio.ErrorResponse); ok {
        switch mErr.Code {
        case "NoSuchKey", "NoSuchBucket":
            err = nil
            return
        }
    }
    return
}
rawData, err := ioutil.ReadAll(mObj)
if err != nil {
    return
}
````

This is the version of minio-go from my glide.lock
```yaml
- name: github.com/minio/minio-go
  version: 2f03abaa07d8bc57faef16cda7655ea62a7e0bed
```

vadmeste:
Getting an object and reading it doesn't work as expected when S3 server
doesn't return the object content length. (though very rare)

Fixes #693

### DIFF
--- a/api-get-object.go
+++ b/api-get-object.go
@@ -328,14 +328,14 @@ func (o *Object) setOffset(bytesRead int64) error {
 	// Update the currentOffset.
 	o.currOffset += bytesRead
 
-	if o.currOffset >= o.objectInfo.Size {
+	if o.objectInfo.Size > -1 && o.currOffset >= o.objectInfo.Size {
 		return io.EOF
 	}
 	return nil
 }
 
 // Read reads up to len(b) bytes into b. It returns the number of
-// bytes read (0 <= n <= len(p)) and any error encountered. Returns
+// bytes read (0 <= n <= len(b)) and any error encountered. Returns
 // io.EOF upon end of file.
 func (o *Object) Read(b []byte) (n int, err error) {
 	if o == nil {
@@ -442,7 +442,7 @@ func (o *Object) ReadAt(b []byte, offset int64) (n int, err error) {
 	if o.objectInfoSet {
 		// If offset is negative than we return io.EOF.
 		// If offset is greater than or equal to object size we return io.EOF.
-		if offset >= o.objectInfo.Size || offset < 0 {
+		if (o.objectInfo.Size > -1 && offset >= o.objectInfo.Size) || offset < 0 {
 			return 0, io.EOF
 		}
 	}
@@ -542,16 +542,20 @@ func (o *Object) Seek(offset int64, whence int) (n int64, err error) {
 	default:
 		return 0, ErrInvalidArgument(fmt.Sprintf("Invalid whence %d", whence))
 	case 0:
-		if offset > o.objectInfo.Size {
+		if o.objectInfo.Size > -1 && offset > o.objectInfo.Size {
 			return 0, io.EOF
 		}
 		o.currOffset = offset
 	case 1:
-		if o.currOffset+offset > o.objectInfo.Size {
+		if o.objectInfo.Size > -1 && o.currOffset+offset > o.objectInfo.Size {
 			return 0, io.EOF
 		}
 		o.currOffset += offset
 	case 2:
+		// If we don't know the object size return an error for io.SeekEnd
+		if o.objectInfo.Size < 0 {
+			return 0, ErrInvalidArgument("Whence END is not supported when the object size is unknown")
+		}
 		// Seeking to positive offset is valid for whence '2', but
 		// since we are backing a Reader we have reached 'EOF' if
 		// offset is positive.
--- a/api-stat.go
+++ b/api-stat.go
@@ -21,8 +21,6 @@ import (
 	"strconv"
 	"strings"
 	"time"
-
-	"github.com/minio/minio-go/pkg/s3utils"
 )
 
 // BucketExists verify if bucket exists and you have permission to access it.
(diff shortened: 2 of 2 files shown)


api-get-object.go
```diff
@@ -542,16 +542,20 @@ func (o *Object) Seek(offset int64, whence int) (n int64, err error) {
...
 		}
 		o.currOffset = offset
 	case 1:
-		if o.currOffset+offset > o.objectInfo.Size {
+		if o.objectInfo.Size > -1 && o.currOffset+offset > o.objectInfo.Size {
 			return 0, io.EOF
 		}
 		o.currOffset += offset
 	case 2:
+		// If we don't the object size, immediately return an error
```
harshavardhana: "If we don't know the object size return an error for io.SeekEnd"

vadmeste:
All comments addressed.

api-stat.go
```diff
@@ -126,22 +124,11 @@ func (c Client) statObject(bucketName, objectName string, reqHeaders RequestHead
...
-				RequestID:  resp.Header.Get("x-amz-request-id"),
-				HostID:     resp.Header.Get("x-amz-id-2"),
-				Region:     resp.Header.Get("x-amz-bucket-region"),
-			}
-		}
+	// Parse content length.
+	size, err := strconv.ParseInt(resp.Header.Get("Content-Length"), 10, 64)
+	if err != nil {
+		// Content-Length is not valid, do not verify.
+		size = -1
```
krishnasrinivas: The question "Why are we ignoring ParseInt error" arises in mind, previous code was better. Or you can do:

```
@@ -126,11 +124,12 @@ func (c Client) statObject(bucketName, objectName string, reqHeaders RequestHead
        md5sum := strings.TrimPrefix(resp.Header.Get("ETag"), "\"")
        md5sum = strings.TrimSuffix(md5sum, "\"")
 
-       // Content-Length is not valid for Google Cloud Storage, do not verify.
+       // If server does not send Content-Length then we set size to -1.
        var size int64 = -1
-       if !s3utils.IsGoogleEndpoint(c.endpointURL) {
+       lenStr := resp.Header.Get("Content-Length")
+       if lenStr != "" {
                // Parse content length.
-               size, err = strconv.ParseInt(resp.Header.Get("Content-Length"), 10, 64)
+               size, err = strconv.ParseInt(lenStr, 10, 64)
                if err != nil {
                        return ObjectInfo{}, ErrorResponse{
                                Code:       "InternalError",
```

vadmeste: Fixed.

PR #699, opened 2017-06-01
api/encrypt: Get() on encrypted object should be a readCloser

harshavardhana:
We should implement `io.Closer` for the encryption materials
to close the underlying reader whenever a caller requests.

GetEncryptedObject() should have returned `io.ReadCloser`
in the first place.


vadmeste:
LGTM and tested.

tejaycar:
@harshavardhana anything else I can do to help this along?  I assume we're just waiting on that final review?

PR #707, opened 2017-06-14
api: deprecate PutObject() resuming capabilities.

harshavardhana:
This is done to ensure our functionality to be
similar to AWS SDKs behavior. Since resuming
as a functionality itself is not expected or
supported by AWS S3.

This is an ongoing simplification with
the new changes coming in to our minio server
gateway and fs backend.


vadmeste:
@harshavardhana, there is some conflicts, can you rebase again ?

harshavardhana:
> @harshavardhana, there are some conflicts, can you rebase again ?

Done @vadmeste

harshavardhana:
@vadmeste can you take a look again? - let me know if any changes are needed..

api.go
```diff
@@ -575,8 +608,17 @@ func (c Client) newRequest(method string, metadata requestMetadata) (req *http.R
 		return nil, err
 	}
 
+	// Go net/http notoriously closes the request body.
+	// - The request Body, if non-nil, will be closed by the underlying Transport, even on errors.
```
vadmeste: If golang's Transport module closes contentBody, why do we close it in executeMethod() ?

harshavardhana: We shouldn't let it close it.. only once we have exhausted retries.

vadmeste: > We shouldn't let it close it.. only once we have exhausted retries.

yes, right, now I understand.

PR #708, opened 2017-06-14
Relax isValidBucketName to allow reading existing buckets.

harshavardhana:
Add a stricter check for MakeBucket alone to avoid
creating newer buckets with capitalized letters.

Fixes https://github.com/minio/mc/issues/2157

### DIFF
--- a/api-put-bucket.go
+++ b/api-put-bucket.go
@@ -26,6 +26,7 @@ import (
 	"net/url"
 
 	"github.com/minio/minio-go/pkg/policy"
+	"github.com/minio/minio-go/pkg/s3utils"
 )
 
 /// Bucket operations
@@ -46,7 +47,7 @@ func (c Client) MakeBucket(bucketName string, location string) (err error) {
 	}()
 
 	// Validate the input arguments.
-	if err := isValidBucketName(bucketName); err != nil {
+	if err := s3utils.CheckValidBucketNameStrict(bucketName); err != nil {
 		return err
 	}
 
@@ -59,17 +60,6 @@ func (c Client) MakeBucket(bucketName string, location string) (err error) {
 			location = c.region
 		}
 	}
-
-	// Try creating bucket with the provided region, in case of
-	// invalid region error let's guess the appropriate region
-	// from S3 API headers
-
-	// Create a done channel to control 'newRetryTimer' go routine.
-	doneCh := make(chan struct{}, 1)
-
-	// Indicate to our routine to exit cleanly upon return.
-	defer close(doneCh)
-
 	// PUT bucket request metadata.
 	reqMetadata := requestMetadata{
 		bucketName:     bucketName,
@@ -118,10 +108,10 @@ func (c Client) MakeBucket(bucketName string, location string) (err error) {
 //  writeonly - anonymous put/delete access to a given object prefix.
 func (c Client) SetBucketPolicy(bucketName string, objectPrefix string, bucketPolicy policy.BucketPolicy) error {
 	// Input validation.
-	if err := isValidBucketName(bucketName); err != nil {
+	if err := s3utils.CheckValidBucketName(bucketName); err != nil {
 		return err
 	}
-	if err := isValidObjectPrefix(objectPrefix); err != nil {
+	if err := s3utils.CheckValidObjectNamePrefix(objectPrefix); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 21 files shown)


harshavardhana:
> isValidObjectName()
> isValidObjectNameStrict()
> 
> isValidBucketName()
> isValidBucketNameStrict()
> isValidBucketNameCommon()

Yes this is how it was in one of the versions, added an interface to see if its easier to bring a generic validator which can be initialized and used.

Why do you think interface is a problem? @krishnasrinivas

krishnasrinivas:
> Why do you think interface is a problem? @krishnasrinivas

It is compromising code simplicity here.

Notice that this commit has introduced:
3 new types: Validator, BucketNameValidator, ObjectNameValidator
2 global vars: bucketNameValidator, objectNameValidator
5 methods: checkCommon, 2 x Check(), 2 x CheckStrict()
4 functions: isValidBucketName, isValidBucketNameStrict, isValidObjectPrefix, isValidObjectName

Instead of this we could just have:
5 functions: isValidBucketName, isValidBucketNameStrict, isValidBucketNameCommon, isValidObjectPrefix, isValidObjectName

pkg/s3utils/utils.go
```diff
@@ -181,3 +182,126 @@ func EncodePath(pathName string) string {
...
+			default:
+				return errors.New("Bucket name contains invalid characters")
+			}
+		}
+		allNumbers = allNumbers && !isNotNumber
+	}
+	// Does the bucket name look like an IP address?
+	if len(pieces) == 4 && allNumbers {
+		return errors.New("Bucket name cannot be an ip address")
+	}
```
krishnasrinivas: Why not use regex and let it do all the heavy lifting?

harshavardhana: Well this is copied from server code @krishnasrinivas . We can keep the regex no big deal i think.

harshavardhana:
> Notice that this commit has introduced:
> 3 new types: Validator, BucketNameValidator, ObjectNameValidator
> 2 global vars: bucketNameValidator, objectNameValidator
> 5 methods: checkCommon, 2 x Check(), 2 x CheckStrict()
> 4 functions: isValidBucketName, isValidBucketNameStrict, isValidObjectPrefix, isValidObjectName
> 
> Instead of this we could just have:
> 5 functions: isValidBucketName, isValidBucketNameStrict, isValidBucketNameCommon, isValidObjectPrefix, isValidObjectName

Just because we added more functions doesn't mean that code is not simple.. but i will get back the other changes i had stashed before interface. :-)

krishnasrinivas:
>> It is compromising code simplicity here.

> Just because we added more functions doesn't mean that code is not simple

I meant the code in this commit is relatively more complex than if you were to just use functions. i.e if you have the two implementations side by side, which one is simpler.

utils.go
```diff
@@ -115,6 +113,22 @@ var emptySHA256 = sum256(nil)
...
+func isValidBucketNameStrict(bucketName string) error {
+	return s3utils.CheckValidBucketNameStrict(bucketName)
+}
+
+func isValidObjectPrefix(objectPrefix string) error {
+	return s3utils.CheckValidObjectName(objectPrefix)
+}
+
+func isValidObjectName(objectName string) error {
+	return s3utils.CheckValidBucketNameStrict(objectName)
```
vadmeste: This supposed to be s3utils.CheckValidObjectNameStrict(objectName)

harshavardhana: Oh yes..oops.

krishnasrinivas:
@harshavardhana what is the convention used on when CheckValidObjectNameStrict and CheckValidObjectName is used? Because GetObject uses CheckValidObjectNameStrict. But ListObjects uses CheckValidObjectName for prefix validation. Also BucketPolicy calls use CheckValidObjectName

harshavardhana:
> @harshavardhana what is the convention used on when CheckValidObjectNameStrict and CheckValidObjectName is used? Because GetObject uses CheckValidObjectNameStrict. But ListObjects uses CheckValidObjectName for prefix validation. Also BucketPolicy calls use CheckValidObjectName

The reason is one is a prefix which can be empty as well and another is a whole object name. But it is just not clear.. I think let me make it clear rather than calling it strict.

api-put-bucket.go
```diff
@@ -118,10 +108,10 @@ func (c Client) MakeBucket(bucketName string, location string) (err error) {
 //  writeonly - anonymous put/delete access to a given object prefix.
 func (c Client) SetBucketPolicy(bucketName string, objectPrefix string, bucketPolicy policy.BucketPolicy) error {
 	// Input validation.
-	if err := isValidBucketName(bucketName); err != nil {
+	if err := s3utils.CheckValidBucketName(bucketName); err != nil {
 		return err
 	}
-	if err := isValidObjectPrefix(objectPrefix); err != nil {
+	if err := s3utils.CheckValidObjectName(objectPrefix); err != nil {
```
vadmeste: This should be `s3utils.CheckValidObjectNamePrefix`

harshavardhana:
@vadmeste @krishnasrinivas the PR is updated and rebased.. Can you guys take a look?

PR #709, opened 2017-06-14
api: Always strip 80/443 port from host

### LINKED ISSUE #695: Strip 80/443 from url when calculating v4 signature
We need that because curl & browsers strip 80 and 443 ports by default when sending 'Host' header. That leads to a signature mismatch.

abipaul commented 11 hours ago:
```
"/home/minio/mc share download --expire=30m mymini/bucket/filename"

generates wrong presigned url if you use port number while configuring mymini

minio config should work with(out) port number specified in endpoint
```

vadmeste:
HTTP clients like browsers or curl automatically strip port 80 and 443
in Host header. This PR makes minio-go follow the same behavior
so the generated presigned urls can work without having signature
mismatch error.

Fixes #695

### DIFF
--- a/api_unit_test.go
+++ b/api_unit_test.go
@@ -22,6 +22,7 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+	"net/url"
 	"os"
 	"strings"
 	"testing"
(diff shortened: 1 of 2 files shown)


harshavardhana:
> LGTM. Does it make sense to write a unit test for makeTargetURL?

:+1:

vadmeste:
> LGTM. Does it make sense to write a unit test for makeTargetURL?

Tests for makeTargetURL are added.

api_unit_test.go
```diff
@@ -302,3 +303,52 @@ func TestPartSize(t *testing.T) {
...
+		{"localhost:9000", true, "mybucket", "", "", nil, url.URL{Host: "localhost:9000", Scheme: "https", Path: "/mybucket/"}, nil},
+		// Test 4, testing against google storage API
+		{"storage.googleapis.com", true, "mybucket", "", "", nil, url.URL{Host: "mybucket.storage.googleapis.com", Scheme: "https", Path: "/"}, nil},
+		// Test 5, testing against AWS S3 API
+		{"s3.amazonaws.com", true, "mybucket", "myobject", "", nil, url.URL{Host: "mybucket.s3.amazonaws.com", Scheme: "https", Path: "/myobject"}, nil},
+		// Test 6
+		{"localhost:9000", false, "mybucket", "myobject", "", nil, url.URL{Host: "localhost:9000", Scheme: "http", Path: "/mybucket/myobject"}, nil},
+		// Test 7
+		{"localhost:9000", false, "mybucket", "myobject", "", map[string][]string{"param": []string{"val"}}, url.URL{Host: "localhost:9000", Scheme: "http", Path: "/mybucket/myobject", RawQuery: "param=val"}, nil},
+	}
```
krisis: Could you add a test case each with address `https://localhost:443` and `http://localhost:80`?

vadmeste: > Could you add a test case each with address https://localhost:443 and http://localhost:80?

oups, yes

vadmeste: > Could you add a test case each with address https://localhost:443 and http://localhost:80?

Done @krisis

PR #710, opened 2017-06-15
Strip leading and trailing quotes for etag in object listing

krishnasrinivas:
list.go:
```
	ch := minioClient.ListObjects("krisbuck", "", false, nil)
	for entry := range ch {
		fmt.Println(entry)
	}
```

output without this commit:
```
{""9246fd89b86191f32e510b61f6c588e7"" my-objectname 2017-05-10 00:33:35 +0000 UTC 105  map[] {minio minio} STANDARD <nil>}
```

output with this commig:
```
{9246fd89b86191f32e510b61f6c588e7 my-objectname 2017-05-10 00:33:35 +0000 UTC 105  map[] {minio minio} STANDARD <nil>}
```

### DIFF
--- a/api-list.go
+++ b/api-list.go
@@ -231,6 +231,13 @@ func (c Client) listObjectsV2Query(bucketName, objectPrefix, continuationToken s
 	if err != nil {
 		return listBucketResult, err
 	}
+	for i, content := range listBucketResult.Contents {
+		etag := content.ETag
+		// Trim off the odd double quotes from ETag in the beginning and end.
+		etag = strings.TrimPrefix(etag, `"`)
+		etag = strings.TrimSuffix(etag, `"`)
+		listBucketResult.Contents[i].ETag = etag
+	}
 	return listBucketResult, nil
 }
 
@@ -400,6 +407,13 @@ func (c Client) listObjectsQuery(bucketName, objectPrefix, objectMarker, delimit
 	err = xmlDecoder(resp.Body, &listBucketResult)
 	if err != nil {
 		return listBucketResult, err
+	}
+	for i, content := range listBucketResult.Contents {
+		etag := content.ETag
+		// Trim off the odd double quotes from ETag in the beginning and end.
+		etag = strings.TrimPrefix(etag, `"`)
+		etag = strings.TrimSuffix(etag, `"`)
+		listBucketResult.Contents[i].ETag = etag
 	}
 	return listBucketResult, nil
 }


harshavardhana:
> Trip leading and trailing quotes for etag in object listing #710

^^ Should be `strip`

krisis:
@krishnasrinivas I tested the output of `ListObjects` and `ListObjectsV2` with and without this PR and _didn't_ see the difference in `object.ETag` representation as described https://github.com/minio/minio-go/pull/710#issue-236328694. I tested with a local Minio server and AWS S3. Am I missing something?

vadmeste:
Also this change could be backward breaking, for example, a developer who is storing ETag with double quotes can face unexpected behavior (like if he compares present etag with his database). I encourage breaking when necessary but I don't see this PR is urgent in any way.

krishnasrinivas:
@krisis I had made a mistake. I was using setup like this AWS-S3 ---> minio-gateway ---> minio-go

Since minio-gateway uses minio-go it does not strip `"` from AWS-S3, minio-gateway now adds extra `"`causing it to have `""`

Now when I test with AWS-S3 ---> minio-go I see just one `"`. Updated the commit. You can verify with and without this commit.

@vadmeste this is a bug in Listresponse which is not as critical as get/put for etag. So it is fine. Also if clients store etag from listing into their DB then clients will fail if it uses `If-ETag` precondition. So we should be fixing this bug.

api-list.go
```diff
@@ -401,6 +409,14 @@ func (c Client) listObjectsQuery(bucketName, objectPrefix, objectMarker, delimit
 	if err != nil {
 		return listBucketResult, err
 	}
+	for i, content := range listBucketResult.Contents {
+		etag := content.ETag
+		// Trim off the odd double quotes from ETag in the beginning and end.
+		// Note that XML decoder puts two leading and trailing quotes for some reason.
```
harshavardhana: This comment is not correct @krishnasrinivas - in S3 API it is expected to set `&quot;` .

krishnasrinivas: Yes, deleted that comment @harshavardhana pushed again

krisis:
@krishnasrinivas, I am not sure I understand what is the issue in minio-go. You say, 
>Since minio-gateway uses minio-go it does not strip " from AWS-S3, minio-gateway now adds extra "causing it to have ""

It makes me wonder if we should fix the ListResponse in gateway code and not in minio-go.

vadmeste:
> Also if clients store etag from listing into their DB then clients will fail if it uses If-ETag precondition.

Do you mean If-Match? I've just tested it and it seems it behaves correctly with and without double-quotes.. I really don't understand why AWS S3 send those double-quotes in the first place.

But I also know that AWS SDKs are actually returning ETag wth double quotes when listing, see examples here http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#completeMultipartUpload-property and here http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#listObjectsV2-property (I can check other examples if you want).

So overall, I think we should avoid this if there is another way to fix the problem.

krishnasrinivas:
Here's the issue (note that there is no gateway involved in the following piece of code) :

```
krishna@escape:~/tmp$ cat list.go
package main

import (
	"fmt"
	"log"

	"github.com/minio/minio-go"
)

func main() {
	endpoint := "play.minio.io:9000"
	accessKeyID := "Q3AM3UQ867SPQQA43P2F"
	secretAccessKey := "zuf+tfteSlswRu7BJ86wekitnifILbZam1KYY3TG"
	useSSL := true

	// Initialize minio client object.
	minioClient, err := minio.New(endpoint, accessKeyID, secretAccessKey, useSSL)
	if err != nil {
		log.Fatalln(err)
	}
	bucket := "krisbuck"
	ch := minioClient.ListObjects(bucket, "", true, nil)
	for entry := range ch {
		listETag := entry.ETag
		stat, err := minioClient.StatObject(bucket, entry.Key)
		if err != nil {
			log.Fatal(err)
		}
		statETag := stat.ETag
		if listETag != statETag {
			fmt.Printf("mismatch in ETags listETag=%s statETag=%s\n", listETag, statETag)
		}
	}
}
krishna@escape:~/tmp$ go run list.go
mismatch in ETags listETag="73c2ee0acb590362c0ac3740f8c39523" statETag=73c2ee0acb590362c0ac3740f8c39523
krishna@escape:~/tmp$ 
```

Basically we are inconsistent in StatObject and ListObject behavior when it comes to ETag. If we are not consistent then applications using our API have to know when to strip and when to not.

AWS SDK is consistent i.e it does not strip the quotes in any calls.

I think we can behave the same way as AWS-SDK does.

harshavardhana:
> Basically we are inconsistent in StatObject and ListObject behavior when it comes to ETag. If we are not consistent then applications using our API have to know when to strip and when to not.
> 
> AWS SDK is consistent i.e it does not strip the quotes in any calls.
> 
> I think we can behave the same way as AWS-SDK does.

We can avoid stripping in StatObject() and GetObject() - that way its easier.

krishnasrinivas:
Will close this in favor of this https://github.com/minio/minio/pull/4590

harshavardhana:
Feel free to close this @krishnasrinivas

harshavardhana:
Closing this..PR @krishnasrinivas

PR #711, opened 2017-06-16
For GCS if size cannot be found use tempfile.

harshavardhana:
This is added as a fallback option instead of returning
an error we can just add a limit reader upto 5GB and
attempt to upload the file instead.

Related to https://github.com/restic/restic/issues/996

### DIFF
--- a/api-put-object.go
+++ b/api-put-object.go
@@ -166,8 +166,11 @@ func (c Client) putObjectNoChecksum(bucketName, objectName string, reader io.Rea
 	if err := isValidObjectName(objectName); err != nil {
 		return 0, err
 	}
-	if size > maxSinglePutObjectSize {
-		return 0, ErrEntityTooLarge(size, maxSinglePutObjectSize, bucketName, objectName)
+	if size > 0 {
+		readerAt, ok := reader.(io.ReaderAt)
+		if ok {
+			reader = io.NewSectionReader(readerAt, 0, size)
+		}
 	}
 
 	// Update progress reader appropriately to the latest offset as we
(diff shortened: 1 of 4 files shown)


krishnasrinivas:
@harshavardhana let's say the file is 6GB, are we returning success after uploading 5GB? If so then it is not the right thing to do.

harshavardhana:
> @harshavardhana let's say the file is 6GB, are we returning success after uploading 5GB? If so then it is not the right thing to do.

Oh yes, i added a code for that to error out.

api-put-object.go
```diff
@@ -170,6 +170,33 @@ func (c Client) putObjectNoChecksum(bucketName, objectName string, reader io.Rea
 		return 0, ErrEntityTooLarge(size, maxSinglePutObjectSize, bucketName, objectName)
 	}
 
+	if size <= -1 {
```
krishnasrinivas: We need not do this check and use io.LimitReader() to limit the file, instead we can just wait for the server to return error. Because I think GCS allows much bigger upload in single PUT ( > 5GB ) ... But I am not sure, not able to find documentation. We can check by uploading a large object to GCS and see if it allows > 5GB in single PUT

api-put-object.go
```diff
@@ -170,6 +170,33 @@ func (c Client) putObjectNoChecksum(bucketName, objectName string, reader io.Rea
...
+		tmpFile, err = newTempFile("single$-putobject-single")
+		if err != nil {
+			return 0, err
+		}
+		defer tmpFile.Close()
+		size, err = io.Copy(tmpFile, io.LimitReader(reader, maxSinglePutObjectSize))
+		if err != nil {
+			return 0, err
+		}
+		if size > maxSinglePutObjectSize {
```
krishnasrinivas: note that size will never be > maxSinglePutObjectSize as io.LimitReader will limit it

api-put-object.go
harshavardhana: Not sure need to check.

api-put-object.go
harshavardhana: Yes this is a wrong check. In-fact we can set the size to the max value and let the underlying reader be sent to the server.

harshavardhana:
@krishnasrinivas - updated, should have fixed all the concerns and now can do native uploads - and not restrict GCS endpoint. 

@fd0 - let me know how your test goes.. Please test this change should fix the GCS/prune issue and integration tests as well.

fd0:
Works, thank you!

PR #715, opened 2017-06-19
Add API to compose objects through server-side copying

donatello:
The new ComposeObject API provides a way to create objects by
concatenating existing objects. It takes a list of source objects
along with optional start-end range specifications, and concatenates
them into a new object.

The API supports:

* Create an object from upto 10000 existing objects.
* Create object of any size upto 5TiB.
* Creates objects from even existing objects that are larger than
  5GiB (part size limit in multipart PUTs)
* Support copy-conditions on each source object separately.
* Support SSE-C (i.e. Server-Side-Encryption with Customer provided
  key) for both encryption of destination object, and decryption of
  source objects.
* Support for setting/replacing custom metadata in the destination
  object.

This API has been used to refactor the CopyObject API - that API now
supports source objects of any size, SSE-C for source and destination,
and settings custom metadata.

Finally, the type for a source object (`SourceInfo`) allows the user to set custom headers (by hand) for further extended usage (such as using AWS KMS based SSE).

### DIFF
--- a/api-compose-object.go
+++ b/api-compose-object.go
@@ -1,0 +1,516 @@
+/*
+ * Minio Go Library for Amazon S3 Compatible Cloud Storage (C) 2017 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package minio
+
+import (
+	"encoding/base64"
+	"fmt"
+	"net/http"
+	"net/url"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/minio/minio-go/pkg/s3utils"
+)
+
+// SSEInfo - represents Server-Side-Encryption parameters specified by
+// a user.
+type SSEInfo struct {
+	key  []byte
+	algo string
+}
+
+// NewSSEInfo - specifies (binary or un-encoded) encryption key and
+// algorithm name. If algo is empty, it defaults to "AES256". Ref:
+// https://docs.aws.amazon.com/AmazonS3/latest/dev/ServerSideEncryptionCustomerKeys.html
+func NewSSEInfo(key []byte, algo string) SSEInfo {
+	if algo == "" {
+		algo = "AES256"
+	}
+	return SSEInfo{key, algo}
+}
+
+// internal method to output the generated SSE headers.
+func (s *SSEInfo) getSSEHeaders(isCopySource bool) map[string]string {
+	if s == nil {
+		return nil
+	}
+
+	cs := ""
+	if isCopySource {
+		cs = "copy-source-"
+	}
+	return map[string]string{
+		"x-amz-" + cs + "server-side-encryption-customer-algorithm": s.algo,
+		"x-amz-" + cs + "server-side-encryption-customer-key":       base64.StdEncoding.EncodeToString(s.key),
+		"x-amz-" + cs + "server-side-encryption-customer-key-MD5":   base64.StdEncoding.EncodeToString(sumMD5(s.key)),
+	}
+}
+
+// DestinationInfo - type with information about the object to be
+// created via server-side copy requests, using the Compose API.
+type DestinationInfo struct {
+	bucket, object string
+
+	// key for encrypting destination
+	encryption *SSEInfo
+
+	// if no user-metadata is provided, it is copied from source
+	// (when there is only once source object in the compose
+	// request)
+	userMetadata map[string]string
+}
+
+// NewDestinationInfo - creates a compose-object/copy-source
+// destination info object.
+//
+// `encSSEC` is the key info for server-side-encryption with customer
+// provided key. If it is nil, no encryption is performed.
+//
+// `userMeta` is the user-metadata key-value pairs to be set on the
+// destination. The keys are automatically prefixed with `x-amz-meta-`
+// if needed. If nil is passed, and if only a single source (of any
+// size) is provided in the ComposeObject call, then metadata from the
+// source is copied to the destination.
+func NewDestinationInfo(bucket, object string, encryptSSEC *SSEInfo,
+	userMeta map[string]string) (d DestinationInfo, err error) {
+
+	// Input validation.
+	if err = s3utils.CheckValidBucketName(bucket); err != nil {
+		return d, err
+	}
+	if err = s3utils.CheckValidObjectName(object); err != nil {
+		return d, err
+	}
+
+	// Process custom-metadata to remove a `x-amz-meta-` prefix if
+	// present and validate that keys are distinct (after this
+	// prefix removal).
+	m := make(map[string]string)
+	for k, v := range userMeta {
+		if strings.HasPrefix(k, "x-amz-meta-") {
+			k = strings.TrimPrefix(k, "x-amz-meta-")
+		}
+		if _, ok := m[k]; ok {
+			return d, fmt.Errorf("Cannot add both %s and x-amz-meta-%s keys as custom metadata", k, k)
+		}
+		m[k] = v
+	}
+
+	return DestinationInfo{
+		bucket:       bucket,
+		object:       object,
+		encryption:   encryptSSEC,
+		userMetadata: m,
+	}, nil
+}
+
+// getUserMetaHeadersMap - construct appropriate key-value pairs to send
+// as headers from metadata map to pass into copy-object request. For
+// single part copy-object (i.e. non-multipart object), enable the
+// withCopyDirectiveHeader to set the `x-amz-metadata-directive` to
+// `REPLACE`, so that metadata headers from the source are not copied
+// over.
+func (d *DestinationInfo) getUserMetaHeadersMap(withCopyDirectiveHeader bool) map[string]string {
+	if d.userMetadata == nil {
+		return nil
+	}
+	r := make(map[string]string)
+	if withCopyDirectiveHeader {
+		r["x-amz-metadata-directive"] = "REPLACE"
+	}
+	for k, v := range d.userMetadata {
+		r["x-amz-meta-"+k] = v
+	}
+	return r
+}
+
+// SourceInfo - represents a source object to be copied, using
+// server-side copying APIs.
+type SourceInfo struct {
+	bucket, object string
+
+	start, end int64
+
+	decryptKey *SSEInfo
+	// Headers to send with the upload-part-copy request involving
+	// this source object.
+	Headers http.Header
+}
+
+// NewSourceInfo - create a compose-object/copy-object source info
+// object.
+//
+// `decryptSSEC` is the decryption key using server-side-encryption
+// with customer provided key. It may be nil if the source is not
+// encrypted.
+func NewSourceInfo(bucket, object string, decryptSSEC *SSEInfo) SourceInfo {
+	r := SourceInfo{
+		bucket:     bucket,
+		object:     object,
+		start:      -1, // range is unspecified by default
+		decryptKey: decryptSSEC,
+		Headers:    make(http.Header),
+	}
+
+	// Set the source header
+	r.Headers.Set("x-amz-copy-source", s3utils.EncodePath(bucket+"/"+object))
+
+	// Assemble decryption headers for upload-part-copy request
+	for k, v := range decryptSSEC.getSSEHeaders(true) {
+		r.Headers.Set(k, v)
+	}
+
+	return r
+}
+
+// SetRange - Set the start and end offset of the source object to be
+// copied. If this method is not called, the whole source object is
+// copied.
+func (s *SourceInfo) SetRange(start, end int64) error {
+	if start > end || start < 0 {
+		return ErrInvalidArgument("start must be non-negative, and start must be at most end.")
+	}
+	// Note that 0 <= start <= end
+	s.start, s.end = start, end
+	return nil
+}
+
+// SetMatchETagCond - Set ETag match condition. The object is copied
+// only if the etag of the source matches the value given here.
+func (s *SourceInfo) SetMatchETagCond(etag string) error {
+	if etag == "" {
+		return ErrInvalidArgument("ETag cannot be empty.")
+	}
+	s.Headers.Set("x-amz-copy-source-if-match", etag)
+	return nil
+}
+
+// SetMatchETagExceptCond - Set the ETag match exception
+// condition. The object is copied only if the etag of the source is
+// not the value given here.
+func (s *SourceInfo) SetMatchETagExceptCond(etag string) error {
+	if etag == "" {
+		return ErrInvalidArgument("ETag cannot be empty.")
+	}
+	s.Headers.Set("x-amz-copy-source-if-none-match", etag)
+	return nil
+}
+
+// SetModifiedSinceCond - Set the modified since condition.
+func (s *SourceInfo) SetModifiedSinceCond(modTime time.Time) error {
+	if modTime.IsZero() {
+		return ErrInvalidArgument("Input time cannot be 0.")
+	}
+	s.Headers.Set("x-amz-copy-source-if-modified-since", modTime.Format(http.TimeFormat))
+	return nil
+}
+
+// SetUnmodifiedSinceCond - Set the unmodified since condition.
+func (s *SourceInfo) SetUnmodifiedSinceCond(modTime time.Time) error {
+	if modTime.IsZero() {
+		return ErrInvalidArgument("Input time cannot be 0.")
+	}
+	s.Headers.Set("x-amz-copy-source-if-unmodified-since", modTime.Format(http.TimeFormat))
+	return nil
+}
+
+// Helper to fetch size and etag of an object using a StatObject call.
+func (s *SourceInfo) getProps(c Client) (size int64, etag string, userMeta map[string]string, err error) {
+	// Get object info - need size and etag here. Also, decryption
+	// headers are added to the stat request if given.
+	var objInfo ObjectInfo
+	rh := NewGetReqHeaders()
+	for k, v := range s.decryptKey.getSSEHeaders(false) {
+		rh.Set(k, v)
+	}
+	objInfo, err = c.statObject(s.bucket, s.object, rh)
+	if err != nil {
+		err = fmt.Errorf("Could not stat object - %s/%s: %v", s.bucket, s.object, err)
+	} else {
+		size = objInfo.Size
+		etag = objInfo.ETag
+		userMeta = make(map[string]string)
+		for k, v := range objInfo.Metadata {
+			if strings.HasPrefix(k, "x-amz-meta-") {
+				if len(v) > 0 {
+					userMeta[k] = v[0]
+				}
+			}
+		}
+	}
+	return
+}
+
+// uploadPartCopy - helper function to create a part in a multipart
+// upload via an upload-part-copy request
+// https://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadUploadPartCopy.html
+func (c Client) uploadPartCopy(bucket, object, uploadID string, partNumber int,
+	headers http.Header) (p CompletePart, err error) {
+
+	// Build query parameters
+	urlValues := make(url.Values)
+	urlValues.Set("partNumber", strconv.Itoa(partNumber))
+	urlValues.Set("uploadId", uploadID)
+
+	// Send upload-part-copy request
+	resp, err := c.executeMethod("PUT", requestMetadata{
+		bucketName:   bucket,
+		objectName:   object,
+		customHeader: headers,
+		queryValues:  urlValues,
+	})
+	defer closeResponse(resp)
+	if err != nil {
+		return p, err
+	}
+
+	// Check if we got an error response.
+	if resp.StatusCode != http.StatusOK {
+		return p, httpRespToErrorResponse(resp, bucket, object)
+	}
+
+	// Decode copy-part response on success.
+	cpObjRes := copyObjectResult{}
+	err = xmlDecoder(resp.Body, &cpObjRes)
+	if err != nil {
+		return p, err
+	}
+	p.PartNumber, p.ETag = partNumber, cpObjRes.ETag
+	return p, nil
+}
+
+// ComposeObject - creates an object using server-side copying of
+// existing objects. It takes a list of source objects (with optional
+// offsets) and concatenates them into a new object using only
+// server-side copying operations.
+func (c Client) ComposeObject(dst DestinationInfo, srcs []SourceInfo) error {
+	if len(srcs) < 1 || len(srcs) > maxPartsCount {
+		return ErrInvalidArgument("There must be as least one and upto 10000 source objects.")
+	}
+
+	srcSizes := make([]int64, len(srcs))
+	var totalSize, size, totalParts int64
+	var srcUserMeta map[string]string
+	var etag string
+	var err error
+	for i, src := range srcs {
+		size, etag, srcUserMeta, err = src.getProps(c)
+		if err != nil {
+			return fmt.Errorf("Could not get source props for %s/%s: %v", src.bucket, src.object, err)
+		}
+
+		// Since we did a HEAD to get size, we use the ETag
+		// value to make sure the object has not changed by
+		// the time we perform the copy. This is done, only if
+		// the user has not set their own ETag match
+		// condition.
+		if src.Headers.Get("x-amz-copy-source-if-match") == "" {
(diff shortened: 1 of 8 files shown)


donatello:
Note to reviewers: This PR breaks compatibility by changing the `CopyObject` API signature. It was deemed necessary to bring an improved version in.

docs/API.md
```diff
@@ -559,6 +560,58 @@ if err != nil {
...
+|`srcs` | _[]minio.SourceInfo_  |Slice of struct with info about source objects to be concatenated in order. |
+
+
+__Example__
+
+
+```go
+	// Prepare source decryption key (here we assume same key to
+	// decrypt all source objects.)
+	decKey := NewSSEInfo([]byte{1, 2, 3}, "")
```
krisis: Shouldn't this be minio.NewSSEInfo?

examples/s3/composeobject.go
```diff
@@ -0,0 +1,74 @@
...
+	src2.SetMatchETag("f8ef9c385918b653a31624deb84149d2")
+
+	src3 := minio.NewSourceInfo("bucket3", "object3", decKey)
+	src3.SetMatchETag("5918b653a31624deb84149d2f8ef9c38")
+
+	// Create slice of sources.
+	srcs := []minio.SourceInfo{src1, src2, src3}
+
+	// Prepare destination encryption key
+	encKey := NewSSEInfo([]byte{8, 9, 0}, "")
```
krisis: Shouldn't this be minio.NewSSEInfo?

examples/s3/copyobject.go
```diff
@@ -42,24 +42,28 @@ func main() {
 	// Enable trace.
 	// s3Client.TraceOn(os.Stderr)
 
+	// Source object
+	src := NewSourceInfo("my-sourcebucketname", "my-sourceobjectname", nil)
```
krisis: Shouldn't this be `minio.NewSourceInfo`?

examples/s3/copyobject.go
```diff
@@ -42,24 +42,28 @@ func main() {
...
 	// Set matching ETag condition, copy object which matches the following ETag.
-	// copyConds.SetMatchETag("31624deb84149d2f8ef9c385918b653a")
+	// src.SetMatchETagCond("31624deb84149d2f8ef9c385918b653a")
 
 	// Set matching ETag except condition, copy object which does not match the following ETag.
-	// copyConds.SetMatchETagExcept("31624deb84149d2f8ef9c385918b653a")
+	// src.SetMatchETagExceptCond("31624deb84149d2f8ef9c385918b653a")
+
+	// Destination object
+	dst := NewDestinationInfo("my-bucketname", "my-objectname", nil)
```
krisis: Shouldn't this be `minio.NewDestinationInfo`?

api-compose-object.go
```diff
@@ -0,0 +1,399 @@
...
+		defer closeResponse(resp)
+		if err != nil {
+			return err
+		}
+		// Check if we got an error response.
+		if resp.StatusCode != http.StatusOK {
+			return httpRespToErrorResponse(resp, dst.bucket, dst.object)
+		}
+		// Decode copy response on success.
+		cpObjRes := copyObjectResult{}
```
donatello: You're right, it is not required here.

docs/API.md
```diff
@@ -559,6 +560,58 @@ if err != nil {
 }
 ```
 
+<a name="ComposeObject"></a>
```
donatello: Done!

docs/API.md
```diff
@@ -559,6 +560,58 @@ if err != nil {
...
+### ComposeObject(dst DestinationInfo, srcs []SourceInfo) error
```
donatello: `NewSSEInfo`, `NewSourceInfo` and `NewDestinationInfo` are added in docs. The types are only documented via godoc for now, in keeping with the existing style (e.g. CopyCondition was documented only via godoc). We may add documentation for types in the API.md later but that is a matter for discussion.

donatello:
There is ready for review.

api-compose-object.go
```diff
@@ -0,0 +1,393 @@
...
+
+// SSEInfo - represents Server-Side-Encryption parameters specified by
+// a user.
+type SSEInfo struct {
+	key  []byte
+	algo string
+}
+
+// NewSSEInfo - specifies (binary or un-encoded) encryption key and
+// algorithm name. If algo is empty, it defaults to "AES256".
```
harshavardhana: Perhaps mention the `AES256` AWS S3 link.

api-compose-object.go
```diff
@@ -0,0 +1,393 @@
...
+func NewSSEInfo(key []byte, algo string) SSEInfo {
+	if algo == "" {
+		algo = "AES256"
+	}
+	return SSEInfo{key, algo}
+}
+
+// internal method to output the generated SSE headers.
+func (e *SSEInfo) getSSEHeaders(isCopySource bool) map[string]string {
+	if e == nil {
```
harshavardhana: Is this `if e == nil` necessary?  - perhaps okay to let the function fail?

api-compose-object.go
```diff
@@ -0,0 +1,393 @@
...
+}
+
+// getCustomHeadersMap - construct appropriate key-value pairs to send
+// as headers from metadata map to pass into copy-object request. For
+// single part copy-object (i.e. non-multipart object), enable the
+// withCopyDirectiveHeader to set the `x-amz-metadata-directive` to
+// `REPLACE`, so that metadata headers from the source are not copied
+// over.
+func (d *DestinationInfo) getCustomHeadersMap(withCopyDirectiveHeader bool) map[string]string {
+	if d == nil || d.customMetadata == nil {
```
harshavardhana: Same as above..

api-compose-object.go
```diff
@@ -0,0 +1,393 @@
...
+func (e *SSEInfo) getSSEHeaders(isCopySource bool) map[string]string {
+	if e == nil {
+		return nil
+	}
+
+	cs := ""
+	if isCopySource {
+		cs = "copy-source-"
+	}
+	hsum := md5.Sum(e.key)
```
harshavardhana: You could use `sumMD5()` ?

api-compose-object.go
```diff
@@ -0,0 +1,393 @@
...
+}
+
+// SourceInfo - represents a source object to be copied, using
+// server-side copying APIs.
+type SourceInfo struct {
+	bucket, object string
+
+	start, end int64
+
+	// collection of headers to send with every upload-part-copy
```
harshavardhana: ``// collect of request headers sent with every upload-part-copy``

api-compose-object.go
```diff
@@ -0,0 +1,393 @@
...
+func NewSourceInfo(bucket, object string, decryptionInfo *SSEInfo) SourceInfo {
+	r := SourceInfo{
+		bucket:  bucket,
+		object:  object,
+		start:   -1,
+		Headers: make(http.Header),
+	}
+
+	// Set the source header
+	r.Headers.Set("x-amz-copy-source", s3utils.EncodePath(fmt.Sprintf("/%s/%s", bucket, object)))
```
harshavardhana: Perhaps use `path.Join()` ? it is okay to not have preceding `/`

constants.go
```diff
@@ -18,10 +18,14 @@ package minio
...
 /// Multipart upload defaults.
 
-// miniPartSize - minimum part size 64MiB per object after which
+// minPartSize - minimum part size 64MiB per object after which
 // putObject behaves internally as multipart.
 const minPartSize = 1024 * 1024 * 64
 
+// copyPartSize - default (and maximum) part size to copy in a
+// copy-object request.
+const copyPartSize = 1024 * 1024 * 1024 * 5
```
harshavardhana: why not 64MB? no reason for this to be 5MB?

api-compose-object.go
donatello: Not is is needed here. When `getSSEHeaders` is called, if there is no key provided during initialization, this safely returns. The other option is to check if the key is nil before calling this function, but I found that to be more repetitive.

api-compose-object.go
donatello: :+1:

api-compose-object.go
donatello: Again same reasoning as before - this is less repetitive.

api-compose-object.go
donatello: Updated to:

```
	// Headers to send with the upload-part-copy request involving
	// this source object.
```

The intension is to allow users to set custom headers on their own in case they need some feature not encapsulated by this library.

api-compose-object.go
donatello: Ok, just used `+` here.

constants.go
donatello: This number is actually 5 GiB. I will update the comment to make this more apparent.

Since this is only server-side copying, the most efficient thing to do is to reduce the number of network requests - this is done by copying the largest possible object-part in each request.

donatello:
Review comments addressed!

api-compose-object.go
harshavardhana: i see.. should we ignore it outside rather than letting this call succeed?

donatello: If we checked if key is nil outside, then each place of using this function would look like (for e.g.):

```go
if src.key != nil {
    for k, v := src.key.getSSEHeaders(false) {
        // do stuff.
    }
}
```

The outer `if` block would be repeated in each usage of this function. In the current implementation is neatly tucked away inside `getSSEHeaders` and it returns a nil back which can be safely used in a for-range loop.

harshavardhana: Correct but its quite an unnatural usage, usually unallocated struct means its a bug of sort and we are sort of hiding it. That is my only concern.

constants.go
harshavardhana: Actually let's not do that.. 5GiB causes for erasure coding block to become super big on a 4 disk setup. Say for example 5GiB *2 so each disk shard will be 2.5GiB.  We should avoid this worst case scenario since this would cause easily 2-3secsec worth of delay for ReadFileWithVerify(). 

We should pick 525MB which would make it 10,000 parts to upload maximum 5TB object. In which case each shard will be at max 256MB.

api-compose-object.go
```diff
@@ -0,0 +1,392 @@
...
+		if totalSize > maxMultipartPutObjectSize {
+			return ErrInvalidArgument(fmt.Sprintf("Cannot compose an object of size %d (> 5TiB)", totalSize))
+		}
+
+		// Append copy-part requests info for the current
+		// source object.
+		var currOffset, partSize int64
+		for currOffset < size {
+			if currOffset+copyPartSize < size {
+				partSize = copyPartSize
```
harshavardhana: Just use

```
        totalPartsCount, partSize, lastPartSize, err := optimalPartInfo(fileSize)
        if err != nil {
                return 0, err
        }
```

Since you know the right file size it is better to know the optimal part size. 5GiB is not an ideal size here.

donatello: I understand your point about the bad case for erasure coding but there is another problem. If we use up more parts for the first few sources, then there may be too many parts for later sources in the input and it may exceed the max parts limit. I will think a bit and figure out a working algo.

api-put-object-copy.go
```diff
@@ -16,57 +16,7 @@
...
-	cpObjRes := copyObjectResult{}
-	err = xmlDecoder(resp.Body, &cpObjRes)
-	if err != nil {
-		return err
-	}
-
-	// Return nil on success.
-	return nil
+// CopyObject - copy a source object into a new object
+func (c Client) CopyObject(dst DestinationInfo, src SourceInfo) error {
```
balamurugana: Why is this function made to break backward compatibility?

api-compose-object.go
```diff
@@ -0,0 +1,392 @@
...
+// algorithm name. If algo is empty, it defaults to "AES256".
+func NewSSEInfo(key []byte, algo string) SSEInfo {
+	if algo == "" {
+		algo = "AES256"
+	}
+	return SSEInfo{key, algo}
+}
+
+// internal method to output the generated SSE headers.
+func (e *SSEInfo) getSSEHeaders(isCopySource bool) map[string]string {
```
balamurugana: why `e` is short form to `SSEInfo`?  I think you can name it better

api-compose-object.go
```diff
@@ -0,0 +1,392 @@
...
+		"x-amz-" + cs + "server-side-encryption-customer-algorithm": e.algo,
+		"x-amz-" + cs + "server-side-encryption-customer-key":       base64.StdEncoding.EncodeToString(e.key),
+		"x-amz-" + cs + "server-side-encryption-customer-key-MD5":   base64.StdEncoding.EncodeToString(sumMD5(e.key)),
+	}
+}
+
+// DestinationInfo - type with information about the object to be
+// created via server-side copy requests, using the Compose API.
+type DestinationInfo struct {
+	bucket, object string
```
balamurugana: I think bucket and object can be listed in two lines for better readability.

api-compose-object.go
```diff
@@ -0,0 +1,392 @@
...
+	if d == nil || d.customMetadata == nil {
+		return nil
+	}
+	r := make(map[string]string)
+	if withCopyDirectiveHeader {
+		r["x-amz-metadata-directive"] = "REPLACE"
+	}
+	for k, v := range d.customMetadata {
+		if !strings.HasPrefix(k, "x-amz-meta-") {
+			k = "x-amz-meta-" + k
```
balamurugana: is it required key should start with `x-amz-meta-`. if so, you could error out in `new()`.  If two keys named `x-amz-meta-my-key` and `my-key` one key/value is lost.

api-compose-object.go
```diff
@@ -0,0 +1,392 @@
...
+	}
+	return r
+}
+
+// SourceInfo - represents a source object to be copied, using
+// server-side copying APIs.
+type SourceInfo struct {
+	bucket, object string
+
+	start, end int64
```
balamurugana: same as above

api-compose-object.go
```diff
@@ -0,0 +1,392 @@
...
+func (c Client) ComposeObject(dst DestinationInfo, srcs []SourceInfo) error {
+	// Input validation.
+	if err := s3utils.CheckValidBucketName(dst.bucket); err != nil {
+		return err
+	}
+	if err := s3utils.CheckValidObjectName(dst.object); err != nil {
+		return err
+	}
+	if len(srcs) > maxPartsCount {
+		return ErrInvalidArgument("Cannot copy more than 10000 source objects.")
```
vadmeste: Can you replace 10000 with maxPartsCount ?

api-compose-object.go
```diff
@@ -0,0 +1,392 @@
...
+// server-side copying operations.
+func (c Client) ComposeObject(dst DestinationInfo, srcs []SourceInfo) error {
+	// Input validation.
+	if err := s3utils.CheckValidBucketName(dst.bucket); err != nil {
+		return err
+	}
+	if err := s3utils.CheckValidObjectName(dst.object); err != nil {
+		return err
+	}
+	if len(srcs) > maxPartsCount {
```
vadmeste: Can you also return an error when srcs is empty ? this function is little long, returning early error will be easier to read.

api-compose-object.go
```diff
@@ -0,0 +1,392 @@
...
+	Headers http.Header
+}
+
+// NewSourceInfo - create a new SourceInfo object. decryptionInfo may
+// be nil if the source object is not encrypted.
+func NewSourceInfo(bucket, object string, decryptionInfo *SSEInfo) SourceInfo {
+	r := SourceInfo{
+		bucket:  bucket,
+		object:  object,
+		start:   -1,
```
vadmeste: I know start = -1 means start is unspecified but we can also set it to zero by default to simplify the code. Do we really need -1 state ?

api-compose-object.go
```diff
@@ -0,0 +1,392 @@
...
+		// the user has not set their own ETag match
+		// condition.
+		if src.Headers.Get("x-amz-copy-source-if-match") == "" {
+			src.SetMatchETagCond(etag)
+		}
+
+		// Check if a segment is specified, and if so, is the
+		// segment within object bounds?
+		if src.start != -1 {
+			if src.end >= size {
```
vadmeste: I think it is okay to do more check of start and end again to simplify the code reading.

api-compose-object.go
```diff
@@ -0,0 +1,392 @@
...
+	}
+
+	// Now, handle multipart-copy cases.
+
+	// Initiate a new multipart upload.
+	metaHeaders := make(map[string][]string)
+	for k, v := range dst.getCustomHeadersMap(false) {
+		metaHeaders[k] = []string{v}
+	}
+	uid, err := c.newUploadID(dst.bucket, dst.object, metaHeaders)
```
vadmeste: uploadID is a better name than uid (unique identifier)

api-compose-object.go
```diff
@@ -0,0 +1,392 @@
...
+	if modTime.IsZero() {
+		return ErrInvalidArgument("Input time cannot be 0.")
+	}
+	s.Headers.Set("x-amz-copy-source-if-unmodified-since", modTime.Format(http.TimeFormat))
+	return nil
+}
+
+// uploadPartCopy - helper function to perform a upload-part-copy
+// request.
+func (c Client) uploadPartCopy(bucket, object, uploadID string, partNumber int,
```
vadmeste: The name is not clear, shouldn't it be copyPart() ?

examples/s3/composeobject.go
```diff
@@ -0,0 +1,74 @@
...
+	src1.SetMatchETag("31624deb84149d2f8ef9c385918b653a")
+
+	src2 := minio.NewSourceInfo("bucket2", "object2", decKey)
+	src2.SetMatchETag("f8ef9c385918b653a31624deb84149d2")
+
+	src3 := minio.NewSourceInfo("bucket3", "object3", decKey)
+	src3.SetMatchETag("5918b653a31624deb84149d2f8ef9c38")
+
+	// Create slice of sources.
+	srcs := []minio.SourceInfo{src1, src2, src3}
```
donatello: Encrypted sources must be decrypted for copying - see https://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadUploadPartCopy.html - relevant quote:

`If the source object is encrypted using server-side encryption with a customer-provided encryption key, you must use the following headers providing encryption information so that Amazon S3 can decrypt the object for copying. `

So ComposeObject also does the same decrypt and copy - each encrypted source must have keys specified for decryption. (Note that this server-side-encryption not client side)

balamurugana: I understand server side encryption.  The question is asked for client side encryption specifically.

donatello: Oops, sorry I misread. In the case of client side encryption, the encrypted content would be just copied transparently by the server (it is not aware of client-side encrypted content) - most likely concatenating (client-side-)encrypted objects arbitrarily will lead to garbled data, but server cannot detect this - it is a user error.

balamurugana: Is it possible to find and error out in this case?  I see minio-java sets below header.

```java
    headerMap.put("x-amz-meta-x-amz-key", encDataKey);
    headerMap.put("x-amz-meta-x-amz-iv", ivString);
    headerMap.put("x-amz-meta-x-amz-matdesc", "{}");
```

api-compose-object.go
donatello: I am only using the term from the S3 API doc - https://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadUploadPartCopy.html - I think it will be generally recognizable. I will improve the comment though.

api-compose-object.go
donatello: :+1:

api-compose-object.go
donatello: Before the size of the source is known, `0 <= start <= end` represents a valid range specification so we need to use `start = -1` to represent an unspecified range. I have added lots of comments to make this clearer now. Ref: see doc for `x-amz-copy-source-range` at https://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadUploadPartCopy.html

examples/s3/composeobject.go
donatello: No, I do not think if we can do this correctly - we cannot try and interpret custom metadata set by the user or other tools that the user uses.

api-compose-object.go
donatello: This is handled in NewDestinationInfo() validation now.

api-compose-object.go
donatello: Done.

examples/s3/composeobject.go
balamurugana: As per http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingClientSideEncryption.html, client side encryption uses `x-amz-meta-x-amz-key` header.  If client side encryption is used for any of source objects, we should error out than performing compose object.  The current behavior leads to data corruption if source objects are deleted after compose.

harshavardhana:
ping @krisis @balamurugana do you guys have more reviews on this PR? @donatello are you going to send an update?

constants.go
```diff
@@ -18,10 +18,14 @@ package minio
...
 /// Multipart upload defaults.
 
-// miniPartSize - minimum part size 64MiB per object after which
+// minPartSize - minimum part size 64MiB per object after which
 // putObject behaves internally as multipart.
 const minPartSize = 1024 * 1024 * 64
 
+// copyPartSize - default (and maximum) part size to copy in a
+// copy-object request (5GiB)
+const copyPartSize = 1024 * 1024 * 1024 * 5
```
donatello: Using that lib for parsing requires checking for errors. It does not add much.

donatello:
This PR is updated again.

The previous implementation had a bug.

Example case of bug: two sources of size s given for concatenating, where s = 5GiB+1. Then compose object would try to upload 4 parts of sizes 5GiB, 1, 5GiB, 1. This fails because one of the parts not in the end is of length < 5MiB. In the fix, we make every part of a particular source (almost) the same size. So for the above example, it would generate 4 parts sized: 2.5GiB+1, 2.5GiB, 2.5GiB+1, 2.5GiB.

api-compose-object.go
```diff
@@ -0,0 +1,506 @@
...
+		completeMultipartUpload{Parts: objParts})
+	if err != nil {
+		err = fmt.Errorf("Error in complete-multipart request - %v", err)
+	}
+	return err
+}
+
+// partsRequired is ceiling(size / copyPartSize)
+func partsRequired(size int64) int {
+	var r int64 = size / copyPartSize
```
harshavardhana: This is not needed you can make this `int64(size/copyPartSize)`

donatello:
That's fixed in the latest change.

constants.go
balamurugana: I was taking about using `const copyPartSize = 5 * humanize.GiByte`

examples/s3/composeobject.go
balamurugana: does this comment make sense?

api-put-object-copy.go
balamurugana: does this comment make sense?

tejaycar:
@harshavardhana Are there really outstanding changes still?  It looks like github is blocking merge, but the only "requested change" I can see is also marked as outdated...

harshavardhana:
> @harshavardhana Are there really outstanding changes still? It looks like github is blocking merge, but the only "requested change" I can see is also marked as outdated...

@tejaycar its not was waiting on the other reviewers.

harshavardhana:
@balamurugana did you get a chance to look at the latest changes?

tejaycar:
Ah, that's great news.   Are you planning to cut a new release when this is merged?  We're really hoping to get it pulled into the next release of Mattermost, but would prefer to do that from a formal release rather than master.

harshavardhana:
> Ah, that's great news. Are you planning to cut a new release when this is merged? We're really hoping to get it pulled into the next release of Mattermost, but would prefer to do that from a formal release rather than master.

Yes @tejaycar it will be a major bump since we broke existing API.

tejaycar:
Sweet!  You guys rock, I'm looking forward to it.

api-put-object-copy.go
donatello: Without breaking backward compatibility, we would have a very complex function(s) to support copy-object with support for decryption of source, encryption of destination, multiple copy-conditions, sub-range of source and setting user metadata headers. Since this is a significant improvement to the existing API, it was agreed to break compatibility and make a new release.

examples/s3/composeobject.go
donatello: I am not sure if it is appropriate to do this - as the encryption scheme is a client side scheme. Copying encrypted objects may be a valid operation in certain situations (client logic may be something complicated and it may be an intended operation - there is no need for the library to stop this). It needs further discussion - and a call can be taken after releasing this. 

If only one object will be copied (e.g. via copyobject, with no explicit user metadata headers passed in), this problem is mitigated as the headers from source will be copied to the destination.

donatello:
More tests are added. I believe the PR is complete now and ready for merge.

api-put-object-copy.go
balamurugana: ok. got it.

examples/s3/composeobject.go
balamurugana: As this issue discussed with @abperiasamy, issue https://github.com/minio/minio-go/issues/738 is opened and this needs to be addressed after this PR merged.

PR #720, opened 2017-06-20
api: ListBuckets() should have a default region for AWS S3.

### LINKED ISSUE #717: ListBuckets() always sends a request with no region set
https://github.com/minio/minio-go/commit/8c34cc49a0b000ebe755577feab802ee721535ed changed the behaviour of `newRequest()` to only use data retrieved from the `metadata` object; in the case of `ListBuckets()` this object [does not contain this data](https://github.com/minio/minio-go/blob/master/api-list.go#L40), so a request is sent with `region == ""`.

The Minio server (at least the version we're using) returns 400 Bad Request in this case:
```
<Error>
    <Code>AuthorizationQueryParametersError</Code>
    <Message>Error parsing the X-Amz-Credential parameter; the region is wrong;</Message>
    [...]
</Error>
```

Prior to the above commit, `newRequest()` defaulted to `us-east-1` or `cn-north-1`, overriding based on metadata rather than making metadata the only source - could this be restored? Perhaps something like:
```go
	var location string
	// Gather location only if bucketName is present.
 	if metadata.bucketName != "" && metadata.bucketLocation == "" {
		location, err = c.getBucketLocation(metadata.bucketName)
		if err != nil {
			return nil, err
		}
	} else if metadata.bucketLocation != "" {
		location = metadata.bucketLocation
	} else if s3utils.IsAmazonChinaEndpoint(c.endpointURL) {
		location = "cn-north-1"
	} else {
		location = "us-east-1"
	}
```

harshavardhana:
Fixes #717

### DIFF
--- a/api.go
+++ b/api.go
@@ -591,15 +591,23 @@ func (c Client) newRequest(method string, metadata requestMetadata) (req *http.R
 		method = "POST"
 	}
 
-	var location string
-	// Gather location only if bucketName is present.
-	if metadata.bucketName != "" && metadata.bucketLocation == "" {
-		location, err = c.getBucketLocation(metadata.bucketName)
-		if err != nil {
-			return nil, err
+	location := metadata.bucketLocation
+	if location == "" {
+		if metadata.bucketName != "" {
+			// Gather location only if bucketName is present.
+			location, err = c.getBucketLocation(metadata.bucketName)
+			if err != nil {
+				if ToErrorResponse(err).Code != "AccessDenied" {
+					return nil, err
+				}
+			}
+			// Upon AccessDenied error on fetching bucket location, default
+			// to possible locations based on endpoint URL. This can usually
+			// happen when GetBucketLocation() is disabled using IAM policies.
 		}
-	} else {
-		location = metadata.bucketLocation
+		if location == "" {
+			location = getDefaultLocation(c.endpointURL, c.region)
+		}
 	}
 
 	// Construct a new target URL.
--- a/utils.go
+++ b/utils.go
@@ -192,3 +192,23 @@ func redactSignature(origAuth string) string {
 	// Strip out 256-bit signature from: Signature=<256-bit signature>
 	return regSign.ReplaceAllString(newAuth, "Signature=**REDACTED**")
 }
+
+// Get default location returns the location based on the input
+// URL `u`, if region override is provided then all location
+// defaults to regionOverride.
+//
+// If no other cases match then the location is set to `us-east-1`
+// as a last resort.
+func getDefaultLocation(u url.URL, regionOverride string) (location string) {
+	if regionOverride != "" {
+		return regionOverride
+	}
+	if s3utils.IsAmazonChinaEndpoint(u) {
+		return "cn-north-1"
+	}
+	if s3utils.IsAmazonGovCloudEndpoint(u) {
+		return "us-gov-west-1"
+	}
+	// Default to location to 'us-east-1'.
+	return "us-east-1"
+}
--- a/bucket-cache.go
+++ b/bucket-cache.go
@@ -86,6 +86,11 @@ func (c Client) getBucketLocation(bucketName string) (string, error) {
 		return "", err
 	}
 
+	// Region set then no need to fetch bucket location.
+	if c.region != "" {
+		return c.region, nil
+	}
+
 	if s3utils.IsAmazonChinaEndpoint(c.endpointURL) {
 		// For china specifically we need to set everything to
 		// cn-north-1 for now, there is no easier way until AWS S3
@@ -98,11 +103,6 @@ func (c Client) getBucketLocation(bucketName string) (string, error) {
 		// provides a cleaner compatible API across "us-east-1" and
 		// Gov cloud region.
 		return "us-gov-west-1", nil
-	}
-
-	// Region set then no need to fetch bucket location.
-	if c.region != "" {
-		return c.region, nil
 	}
 
 	if location, ok := c.bucketLocCache.Get(bucketName); ok {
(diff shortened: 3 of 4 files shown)


api.go
```diff
@@ -558,15 +558,24 @@ func (c Client) newRequest(method string, metadata requestMetadata) (req *http.R
...
+		// Gather location only if bucketName is present.
 		location, err = c.getBucketLocation(metadata.bucketName)
 		if err != nil {
 			return nil, err
 		}
-	} else {
-		location = metadata.bucketLocation
+	}
+	if location == "" {
+		// Default to location to 'us-east-1'.
```
donatello: s/Default to location to/Default location to/

api.go
```diff
@@ -558,15 +558,24 @@ func (c Client) newRequest(method string, metadata requestMetadata) (req *http.R
...
+		location = "us-east-1"
+		if s3utils.IsAmazonChinaEndpoint(c.endpointURL) {
+			location = "cn-north-1"
+		} else if s3utils.IsAmazonGovCloudEndpoint(c.endpointURL) {
+			location = "us-gov-west-1"
+		} else if c.region != "" {
+			location = c.region
+		}
 	}
```
donatello: Perhaps a switch statement would be more natural here.

harshavardhana: I tried that.. but it got complex. So an if-condition was easier. In-fact will cleanup to put this under functions for each cases.

krishnasrinivas: this would look more readable?:
```
        if location == "" {                                                                                                                                                                                        
                switch {                                                                                                                                                                                           
                case s3utils.IsAmazonChinaEndpoint(c.endpointURL):                                                                                                                                                 
                        location = "cn-north-1"                                                                                                                                                                    
                case s3utils.IsAmazonGovCloudEndpoint(c.endpointURL):                                                                                                                                              
                        location = "us-gov-west-1"                                                                                                                                                                 
                case c.region != "":                                                                                                                                                                               
                        location = c.region                                                                                                                                                                        
                default:                                                                                                                                                                                           
                        location = "us-east-1"                                                                                                                                                                     
                }                                                                                                                                                                                                  
        }
```

harshavardhana:
can you look again @krishnasrinivas @donatello ?

api.go
```diff
@@ -558,15 +558,24 @@ func (c Client) newRequest(method string, metadata requestMetadata) (req *http.R
...
-	var location string
-	// Gather location only if bucketName is present.
-	if metadata.bucketName != "" && metadata.bucketLocation == "" {
-		location, err = c.getBucketLocation(metadata.bucketName)
-		if err != nil {
-			return nil, err
+	location := metadata.bucketLocation
+	if location == "" {
+		switch {
+		case metadata.bucketName != "":
```
krishnasrinivas: as this is a switch with single case you could rather use 
`if metadata.bucketName != ""`

harshavardhana: oh yes done.

utils.go
```diff
@@ -192,3 +192,24 @@ func redactSignature(origAuth string) string {
...
 	return regSign.ReplaceAllString(newAuth, "Signature=**REDACTED**")
 }
+
+// Get default location returns the location based on the input
+// URL `u`, if none of the standard URL matches then if regionOverride
+// is given location gets defaulted to that instead.
+//
+// If no other cases match then the location is set to `us-east-1`
+// as a last resort.
+func getDefaultLocation(u url.URL, regionOverride string) (location string) {
```
krishnasrinivas: If the user tries to override AmazonChinaEndpoint region then better not auto correct but let it fail naturally as the server will return error. This will make this function simpler as well.

utils.go
```diff
@@ -192,3 +192,24 @@ func redactSignature(origAuth string) string {
...
+	case s3utils.IsAmazonChinaEndpoint(u):
+		location = "cn-north-1"
+	case s3utils.IsAmazonGovCloudEndpoint(u):
+		location = "us-gov-west-1"
+	case regionOverride != "":
+		location = regionOverride
+	default:
+		location = "us-east-1"
+	}
+	return
```
krishnasrinivas: actually this is more readable as it is not a if/else situation (and hence switch need not be used):
```
        func getDefaultLocation(u url.URL, regionOverride string) (location string) {                                                                                                                              
                if s3utils.IsAmazonChinaEndpoint(u) {                                                                                                                                                              
                        return "cn-north-1"                                                                                                                                                                        
                }                                                                                                                                                                                                  
                if s3utils.IsAmazonGovCloudEndpoint(u) {                                                                                                                                                           
                        return "us-gov-west-1"                                                                                                                                                                     
                }                                                                                                                                                                                                  
                if regionOverride != "" {                                                                                                                                                                          
                        return regionOverride                                                                                                                                                                      
                }                                                                                                                                                                                                  
                return "us-east-1"                                                                                                                                                                                 
        }            
```

utils.go
harshavardhana: I don't get it.. @krishnasrinivas

utils.go
harshavardhana: Well this is how it was before @krishnasrinivas

utils.go
harshavardhana: This is how it is in getBucketLocation() let me do that as well.

utils.go
harshavardhana: Doing it without switch.

krishnasrinivas: previously it was if/else @harshavardhana :
```
        if s3utils.IsAmazonChinaEndpoint(u) {                                                           
                location = "cn-north-1"                                                                 
        } else if s3utils.IsAmazonGovCloudEndpoint(u) {                                                 
                location = "us-gov-west-1"                                                              
        } else if regionOverride != "" {                                                                
                location = regionOverride                                                               
        } else {                                                                                        
                location = "us-east-1"                                                                  
        }    
```

PR #724, opened 2017-06-23
api: objectInfo filtering should include more headers.

harshavardhana:
Fixes https://github.com/minio/mc/issues/2193


thomasf:
This patch also solved the problem for my production/preproduction set up as well.

PR #727, opened 2017-06-24
Functional tests now run as a standalone program

poornas:
Reduced the max. filesize tested to 65mb to speed up the run. This is meant to be  run via mint which will have the data directory set up.


harshavardhana:
@poornas you have to fix the `golint` issues..

utils.go
```diff
@@ -212,3 +214,28 @@ func getDefaultLocation(u url.URL, regionOverride string) (location string) {
...
+
+const letterBytes = "abcdefghijklmnopqrstuvwxyz01234569"
+const (
+	letterIdxBits = 6                    // 6 bits to represent a letter index
+	letterIdxMask = 1<<letterIdxBits - 1 // All 1-bits, as many as letterIdxBits
+	letterIdxMax  = 63 / letterIdxBits   // # of letter indices fitting in 63 bits
+)
+
+// randString generates random names and prepends them with a known prefix.
+func randString(n int, src rand.Source, prefix string) string {
```
harshavardhana: Don't create randString() in this file. This is of no use in the library. You should move this back to the `api_functional_v4_tests.go` 

Also `api_functional_v4_tests.go` should be better named as `functional_tests.go`

api_functional_v4_tests.go
```diff
@@ -27,22 +29,34 @@ import (
...
 	"time"
 
+	minio "github.com/minio/minio-go"
+	log "github.com/sirupsen/logrus"
+
 	"github.com/minio/minio-go/pkg/encrypt"
 	"github.com/minio/minio-go/pkg/policy"
 )
 
+const minPartSize = 1024 * 1024 * 64
```
harshavardhana: Expose these values so that it can be used from the library in `constants.go`

api_functional_v4_tests.go
```diff
@@ -27,22 +29,34 @@ import (
...
 const letterBytes = "abcdefghijklmnopqrstuvwxyz01234569"
 const (
 	letterIdxBits = 6                    // 6 bits to represent a letter index
 	letterIdxMask = 1<<letterIdxBits - 1 // All 1-bits, as many as letterIdxBits
 	letterIdxMax  = 63 / letterIdxBits   // # of letter indices fitting in 63 bits
 )
 
+func getFilePath(filename string) string {
+	var dir = os.Getenv("DATA_DIR")
```
harshavardhana: Use the named returns here. 

```
func getFilePath(filename string) (filePath string) {
            dataDir := os.Getenv("DATA_DIR")
            if dataDir == "" {
                     dataDir = "/mint/data"
            }
            return dataDir + "/" + filename
}
```

Also make getDataDir() another function to return the configured DATA_DIR only and then use it with `getFilePath()`

api_functional_v4_tests.go
```diff
@@ -61,27 +75,34 @@ func randString(n int, src rand.Source, prefix string) string {
...
+	return os.Getenv("MODE") == "quick"
+}
+
 // Tests bucket re-create errors.
-func TestMakeBucketError(t *testing.T) {
-	if testing.Short() {
-		t.Skip("skipping functional tests for short runs")
+func testMakeBucketError() {
+	logTrace()
+	if isQuickMode() {
```
harshavardhana: We are better off doing this in main()

```
if !isQuickMode() {
   .. long running tests.
}
```
Avoids unrelated code creeping into test body code.

utils.go
poornas: The randString() method was moved to utils.go because api_functional_v2_test.go uses this function as well.

harshavardhana: Even v2 tests should be moved to the same file @poornas

api_functional_v2_test.go
```diff
@@ -31,6 +31,31 @@ import (
 	"github.com/minio/minio-go/pkg/policy"
 )
 
+const letterBytes = "abcdefghijklmnopqrstuvwxyz01234569"
```
harshavardhana: We need to move v2 tests as well into the functional test program. @poornas

poornas: is there a reason only a few of the tests are run for v2? We could run all tests for both v2 and v4 in functional_tests.go but it would slow the test run further.

harshavardhana: only few because it is generally just to validate that if v2 is working fine with most APIs if not all. We can add you can confirm this with @abperiasamy . IMO we should include v2 if we support it on server side.

poornas: plugged in the existing v2 tests.

harshavardhana:
You might have to rebase after recent change @poornas

functional_tests.go
```diff
@@ -0,0 +1,3648 @@
...
+		log.Fatal("Error:", err)
+	}
+	err = c.RemoveBucket(bucketName)
+	if err != nil {
+		log.Fatal("Error:", err)
+	}
+}
+
+// Tests resumable put object cloud to cloud.
+func testResumablePutObject() {
```
harshavardhana: We should remove all these resumable tests.. we don't resume anymore. @poornas

poornas:
@harshavardhana, fixed

harshavardhana:
Can you fix the build @poornas ?

poornas:
@NitishT , rebased.

NitishT:
Tests introduced in commit https://github.com/minio/minio-go/commit/129fe893b2ac95d322d531b7029981f71c351e53 seem to be missing..

poornas:
@harshavardhana , @NitishT  - rebased and added tests from the last 2 commits.

NitishT:
Data directory and environment vars need to be handled.

harshavardhana:
@poornas can you enable this with `.travis.yml` and let the CI run it on play?

PR #728, opened 2017-06-26
api: Support parallel uploads for PutObjectStreaming

harshavardhana:


harshavardhana:
This is ready for review..

harshavardhana:
Early results with this change with `mc`

```
minio@minio2:~$ time /tmp/mc-master cp 10gb myminio/testbucket/
10gb:                      9.31 GB / 9.31 GB ┃▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓┃ 100.00% 159.30 MB/s 59s
real    0m59.927s
user    0m48.400s
sys     0m8.072s
minio@minio2:~$ time /tmp/mc- cp 10gb myminio/testbucket/                                                                                         
mc-master     mc-streaming  
minio@minio2:~$ time /tmp/mc-streaming cp 10gb myminio/testbucket/1
10gb:                      9.31 GB / 9.31 GB ┃▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓┃ 100.00% 267.59 MB/s 35s
real    0m35.853s
user    0m2.696s
sys     0m8.260s
```

api-put-object-multipart.go
```diff
@@ -24,143 +24,14 @@ import (
...
 
-// Comprehensive put object operation involving multipart uploads.
-//
-// Following code handles these types of readers.
-//
-//  - *os.File
-//  - *minio.Object
-//  - Any reader which has a method 'ReadAt()'
-//
 func (c Client) putObjectMultipart(bucketName, objectName string, reader io.Reader, size int64, metaData map[string][]string, progress io.Reader) (n int64, err error) {
```
fwessels: Why is the comment gone?

harshavardhana: This function doesn't do that anymore.. It is moved to a different function which handles this situation.

api-put-object-streaming.go
```diff
@@ -0,0 +1,547 @@
...
+		return 0, err
+	}
+
+	// Initiate a new multipart upload.
+	uploadID, err := c.newUploadID(bucketName, objectName, metadata)
+	if err != nil {
+		return 0, err
+	}
+
+	// Total data read and written to server. should be equal to 'size' at the end of the call.
```
fwessels: should --> Should

harshavardhana: done.

api-put-object-streaming.go
```diff
@@ -0,0 +1,547 @@
...
+// Following code handles these types of readers.
+//
+//  - *os.File
+//  - *minio.Object
+//  - Any reader which has a method 'ReadAt()'
+//
+func (c Client) putObjectMultipartStream(bucketName, objectName string,
+	reader io.Reader, size int64, metadata map[string][]string, progress io.Reader) (int64, error) {
+
+	if size > 0 && size > minPartSize {
```
vadmeste: `size > minPartSize` means also `size > 0`

api-put-object-streaming.go
```diff
@@ -0,0 +1,547 @@
...
+	Size    int64 // Size of the part uploaded.
+	Part    *ObjectPart
+}
+
+type uploadPartReq struct {
+	PartNum int         // Number of the part uploaded.
+	Part    *ObjectPart // Size of the part uploaded.
+}
+
+func (c Client) putObjectMultipartStreamFromFile(bucketName, objectName string,
```
vadmeste: @harshavardhana, can you tell why we have both putObjectMultipartStreamFromFile and putObjectMultipartStreamFromReadAt ? both of have similar body.

api-put-object.go
```diff
@@ -143,164 +141,111 @@ func (a completedParts) Less(i, j int) bool { return a[i].PartNumber < a[j].Part
...
-	}
-	if resp != nil {
-		if resp.StatusCode != http.StatusOK {
-			return ObjectInfo{}, httpRespToErrorResponse(resp, bucketName, objectName)
+		errResp := ToErrorResponse(err)
+		// Verify if multipart functionality is not available, if not
+		// fall back to single PutObject operation.
+		if errResp.Code == "AccessDenied" && strings.Contains(errResp.Message, "Access Denied") {
+			// Verify if size of reader is greater than '5GiB'.
+			if size > maxSinglePutObjectSize {
```
vadmeste: Can we move object size limitation check in a more low level functions ?
like c.putObjectNoChecksum and c.putObjectMultipartStream and c.putObjectMultipart

api-put-object-streaming.go
```diff
@@ -0,0 +1,547 @@
...
+
+	// Return final size.
+	return totalUploadedSize, nil
+}
+
+// putObjectNoChecksum special function used Google Cloud Storage. This special function
+// is used for Google Cloud Storage since Google's multipart API is not S3 compatible.
+func (c Client) putObjectNoChecksum(bucketName, objectName string, reader io.Reader, size int64, metaData map[string][]string, progress io.Reader) (n int64, err error) {
+	// Input validation.
+	if err := s3utils.CheckValidBucketName(bucketName); err != nil {
```
vadmeste: We don't have to call s3utils.CheckValidBucketName and s3utils.CheckValidObjectName since they are already called in putObjectDo()

api-put-object-streaming.go
harshavardhana: ReadAt is for GetObject, File is for *os.File .

api-put-object.go
harshavardhana: Let me see..

api-put-object-streaming.go
harshavardhana: It is called on purpose.. @vadmeste - each functions validate their own arguments.

api-put-object-streaming.go
harshavardhana: Actually you are right we don't need it. @vadmeste - removing. It must be an old code.

vadmeste: > ReadAt is for GetObject, File is for *os.File .

I know, but the code is the same, below is the diff between putObjectMultipartStreamFromFile and putObjectMultipartStreamFromFile (I removed comments before diff)

```diff
@@ -1,4 +1,4 @@
-func (c Client) putObjectMultipartStreamFromReadAt(bucketName, objectName string,
+func (c Client) putObjectMultipartStreamFromFile(bucketName, objectName string,
 	reader io.ReaderAt, size int64, metadata map[string][]string, progress io.Reader) (int64, error) {
 	if err := s3utils.CheckValidBucketName(bucketName); err != nil {
 		return 0, err
@@ -21,10 +21,10 @@
 		return 0, err
 	}
 
-	uploadPartsCh := make(chan uploadPartReq, 10000)
-
 	uploadedPartsCh := make(chan uploadedPartRes, 10000)
 
+	uploadPartsCh := make(chan uploadPartReq, 10000)
+
 	lastPartNumber := totalPartsCount
 
 	for p := 1; p <= totalPartsCount; p++ {
@@ -35,7 +35,6 @@
 	for w := 1; w <= totalWorkers; w++ {
 		go func() {
 			for uploadReq := range uploadPartsCh {
-
 				readOffset := int64(uploadReq.PartNum-1) * partSize
 
 				if uploadReq.PartNum == lastPartNumber {
@@ -46,11 +45,10 @@
 				sectionReader := io.NewSectionReader(reader, readOffset, partSize)
 
 				var objPart ObjectPart
-				objPart, err = c.uploadPart(bucketName, objectName, uploadID, sectionReader,
-					uploadReq.PartNum, nil, nil, partSize)
+				objPart, err = c.uploadPart(bucketName, objectName, uploadID, sectionReader, uploadReq.PartNum,
+					nil, nil, partSize)
 				if err != nil {
 					uploadedPartsCh <- uploadedPartRes{
-						Size:  0,
 						Error: err,
 					}
 					return
@@ -75,7 +73,7 @@
 		}
 		part := uploadRes.Part
 		if part == nil {
-			return 0, ErrInvalidArgument(fmt.Sprintf("Missing part number %d", uploadRes.PartNum))
+			return totalUploadedSize, ErrInvalidArgument(fmt.Sprintf("Missing part number %d", uploadRes.PartNum))
 		}
 		totalUploadedSize += uploadRes.Size
 		if progress != nil {
```

api-put-object.go
harshavardhana: This is done so that we can do it comprehensively at the top, at lower level this error can originate for any API operation such as initiateMultipartUpload,  PutObjectPart, CompleteMultipart etc.

harshavardhana: Moved it to a lower function.

harshavardhana:
@fwessels any more reviews?

api-put-object.go
```diff
@@ -143,164 +141,71 @@ func (a completedParts) Less(i, j int) bool { return a[i].PartNumber < a[j].Part
...
+	// NOTE: Streaming signature is not supported by GCS.
+	if s3utils.IsGoogleEndpoint(c.endpointURL) {
+		// Do not compute MD5 for Google Cloud Storage.
+		return c.putObjectNoChecksum(bucketName, objectName, reader, size, metadata, progress)
 	}
 
-	// If Content-Type is not provided, set the default application/octet-stream one
-	if v, ok := metaData["Content-Type"]; !ok || len(v) == 0 {
-		customHeader.Set("Content-Type", "application/octet-stream")
+	if c.overrideSignerType.IsV2() {
```
Liuchang0812: hi, why does V2 singer can't use streaming upload?

harshavardhana: Streaming standard is only signature v4

Liuchang0812: thank you, It seems that I could steaming upload object with signature V2. is there any problems if we remove this check?

Liuchang0812: @harshavardhana "Streaming standard is only signature v4" do you mean this [Transferring Payload in Multiple Chunks](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html).  but minio-go's [executeMethod](https://github.com/minio/minio-go/blob/46b02d3b4e8c63804a5d4d4d14a562fd71c66730/api.go#L524) do not use this chunk upload whether it's V2 or V4 signature.

harshavardhana: of course, we do @Liuchang0812 

```
   switch {
        case signerType.IsV2():
                req = s3signer.SignV2(*req, accessKeyID, secretAccessKey, isVirtualHost)
        case metadata.objectName != "" && method == "PUT" && metadata.customHeader.Get("X-Amz-Copy-Source") == "" && !c.secure:
                // Streaming signature is used by default for a PUT object request. Additionally we also
                // look if the initialized client is secure, if yes then we don't need to perform
                // streaming signature.
                req = s3signer.StreamingSignV4(req, accessKeyID,
                        secretAccessKey, sessionToken, location, metadata.contentLength, time.Now().UTC())
        default:
                shaHeader := unsignedPayload
                if metadata.contentSHA256Hex != "" {
                        shaHeader = metadata.contentSHA256Hex
                }
                req.Header.Set("X-Amz-Content-Sha256", shaHeader)

                // Add signature version '4' authorization header.
                req = s3signer.SignV4(*req, accessKeyID, secretAccessKey, sessionToken, location)
        }
```

Liuchang0812: @hackintoshrao thanks for your help. `s3signer.SignV4` does not sign http body as [transferring payload in multiple chunks](s3signer.SignV4). please let me know if i missing anything.

what I meet is minio-go consumes nearly 600MB memory when I upload a object with V2 sign. because minio-go allocate 64MB memory [here](https://github.com/minio/minio-go/blob/master/api-put-object-multipart.go#L100).

it seems that the diff between `putObjectMultipartNoStream` and `putObjectMultipartStreamNoChecksum` is non-stream api has to read file content to memory and calculate MD5/SHA.

PR #729, opened 2017-06-27
api: Allow *.amazonaws.com but not s3 endpoints

vadmeste:
New() was rejecting endpoints with this address format: *.amazonaws.com,
however, it should only reject *.s3.amazonaws.com. With this PR,
addresses such as foo.us-east-1.elb.amazonaws.com will be accepted.

Fixes https://github.com/minio/mc/issues/2198

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -122,7 +122,7 @@ func isValidEndpointURL(endpointURL url.URL) error {
 	if endpointURL.Path != "/" && endpointURL.Path != "" {
 		return ErrInvalidArgument("Endpoint url cannot have fully qualified paths.")
 	}
-	if strings.Contains(endpointURL.Host, ".amazonaws.com") {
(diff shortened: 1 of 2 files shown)


vadmeste:
I didn't see the test error here. Now it is fixed. Can you review @harshavardhana and @balamurugana  ?

utils.go
```diff
@@ -122,7 +122,7 @@ func isValidEndpointURL(endpointURL url.URL) error {
 	if endpointURL.Path != "/" && endpointURL.Path != "" {
 		return ErrInvalidArgument("Endpoint url cannot have fully qualified paths.")
 	}
-	if strings.Contains(endpointURL.Host, ".amazonaws.com") {
+	if strings.Contains(endpointURL.Host, ".s3.amazonaws.com") {
```
harshavardhana: how does this work for `s3.amazonaws.com` ?

PR #733, opened 2017-06-29
Reduce memory consumption by staging the data locally.

### LINKED ISSUE #730: optimalPartInfo() default partSize results in running out of memory

harshavardhana:
Fixes #730

### DIFF
--- a/api-put-object-multipart.go
+++ b/api-put-object-multipart.go
@@ -82,21 +82,24 @@ func (c Client) putObjectMultipartNoStream(bucketName, objectName string, reader
 	// Part number always starts with '1'.
 	partNumber := 1
 
-	// Initialize a temporary buffer.
-	tmpBuffer := new(bytes.Buffer)
-
 	// Initialize parts uploaded map.
 	partsInfo := make(map[int]ObjectPart)
 
 	for partNumber <= totalPartsCount {
-		// Choose hash algorithms to be calculated by hashCopyN,
-		// avoid sha256 with non-v4 signature request or
-		// HTTPS connection.
+		// Initialize a temporary buffer.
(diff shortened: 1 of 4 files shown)


harshavardhana:
Any reviews ?

donatello:
Could you fix the conflicts?

harshavardhana:
> Could you fix the conflicts?

dun!

harshavardhana:
> @harshavardhana during our design discussions long time back on sdks we had decided not to use tmp files at all, did that change anytime? because if that is the case we can use tmp files in other libs too.

@krishnasrinivas the only reason i think its valid is because of few reasons that i found that our SDKs don't work. 

PutObject() with size == -1 - atleast in Go results in 570MB , 570MB x 2 due to bytes.Buffer usage (2x growth rate).  The problem is Go GC never relinquishes this unless it hits the GC=100% mark. 

At the same time user is also concurrently uploading many such PutObject() operations for 1GiB - all of them with stream size of '-1' .  This incurs a memory growth to around multiple of GB which is not GC'ed properly due to the same fact as GC=100 growth.

There is no easier way to restrict the user, unless other than allocating a large block of 570MB re-usable memory. This would also have an impact on minio-go startup performance.

Using temporary files there are few advantages i.e 

 - The stored file never grows beyond 570MB unlike usage of bytes.Buffer. File itself is garbage collected upon error or success. 
 - There is no GC overahead or no idle memory to track. The memory footprint of minio-go is around 30MB~50MB. 

This is the problem which affects all our libraries.

harshavardhana:
We should moved this as blocked for now on discussion with @abperiasamy

harshavardhana:
Closing this doing bufPool approach. We have a full API refactor #747

PR #734, opened 2017-06-30
api: Update progress bar concurrently.

harshavardhana:
Currently we had a bug where progress bar
was only updated after the upload has been
completed. With streaming signature we
don't need to do that.

This PR fixes this behavior. This PR also
removes a double progress bar jump
in case of fallback buffer based multipart upload.

### DIFF
--- a/api-put-object-multipart.go
+++ b/api-put-object-multipart.go
@@ -113,13 +113,6 @@ func (c Client) putObjectMultipartNoStream(bucketName, objectName string, reader
 		// Save successfully uploaded part metadata.
 		partsInfo[partNumber] = objPart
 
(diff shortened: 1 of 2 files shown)


vadmeste:
Good change, this also fix this issue https://github.com/minio/mc/issues/2169

api-put-object-multipart.go
```diff
@@ -113,13 +113,6 @@ func (c Client) putObjectMultipartNoStream(bucketName, objectName string, reader
 		// Save successfully uploaded part metadata.
 		partsInfo[partNumber] = objPart
 
-		// Update the progress reader for the skipped part.
```
vadmeste: +1

PR #735, opened 2017-06-30
api: use simple streaming upload for GetObject() reader.

harshavardhana:
This change is needed to avoid the offset based ReadAt()
calls which can occur if using io.SectionReader().

Optimizing this part re-implementing a sectionReader()
based on Seeker() wasn't worth the complexity.

### DIFF
--- a/api-put-object-streaming.go
+++ b/api-put-object-streaming.go
@@ -49,7 +49,7 @@ func (c Client) putObjectMultipartStream(bucketName, objectName string,
 	// while it is going to be duck typed similar to io.ReaderAt.
 	// It is to indicate that *minio.Object implements io.ReaderAt.
 	// and such a functionality is used in the subsequent code path.
-	if isObject(reader) || isReadAt(reader) || isFile(reader) {
+	if isFile(reader) || !isObject(reader) && isReadAt(reader) {
 		n, err = c.putObjectMultipartStreamFromReadAt(bucketName, objectName, reader.(io.ReaderAt), size, metadata, progress)
 	} else {
 		n, err = c.putObjectMultipartStreamNoChecksum(bucketName, objectName, reader, size, metadata, progress)
@@ -270,14 +270,8 @@ func (c Client) putObjectMultipartStreamNoChecksum(bucketName, objectName string
 
 		var objPart ObjectPart
 		objPart, err = c.uploadPart(bucketName, objectName, uploadID,
-			io.LimitReader(hookReader, partSize), partNumber,
-			nil, nil, partSize, metadata)
-		// For unknown size, Read EOF we break away.
-		// We do not have to upload till totalPartsCount.
-		if err == io.EOF && size < 0 {
-			break
-		}
-
+			io.LimitReader(hookReader, partSize),
+			partNumber, nil, nil, partSize, metadata)
 		if err != nil {
 			return totalUploadedSize, err
 		}
@@ -340,9 +334,8 @@ func (c Client) putObjectNoChecksum(bucketName, objectName string, reader io.Rea
 		return 0, ErrEntityTooSmall(size, bucketName, objectName)
 	}
 	if size > 0 {
-		readerAt, ok := reader.(io.ReaderAt)
-		if ok {
-			reader = io.NewSectionReader(readerAt, 0, size)
+		if isReadAt(reader) && !isObject(reader) {
+			reader = io.NewSectionReader(reader.(io.ReaderAt), 0, size)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


harshavardhana:
PR is updated fixes a bug as well in the streaming signature implementation. Can anyone review @krishnasrinivas @vadmeste ?

pkg/s3signer/request-signature-streaming.go
```diff
@@ -254,7 +254,15 @@ func (s *StreamingReader) Read(buf []byte) (int, error) {
 		s.chunkBufLen = 0
 		for {
 			n1, err := s.baseReadCloser.Read(s.chunkBuf[s.chunkBufLen:])
-			if err == nil || err == io.ErrUnexpectedEOF {
+			// An instance of this general case is that a Reader returning
```
krisis: We could simplify this comment like,
```
1. n > 0, err is one of io.EOF, io.ErrUnexpectedEOF, nil - nearing end of stream
2. n ==0, err is io.EOF - actual end of stream
```

pkg/s3signer/request-signature-streaming.go
```diff
@@ -254,7 +254,19 @@ func (s *StreamingReader) Read(buf []byte) (int, error) {
 		s.chunkBufLen = 0
 		for {
 			n1, err := s.baseReadCloser.Read(s.chunkBuf[s.chunkBufLen:])
-			if err == nil || err == io.ErrUnexpectedEOF {
+			// Usually we validate `err` first, but in this case
+			// we are validating n > 0 for the following reasons.
+			//
+			// 1. n > 0, err is one of io.EOF, io.ErrUnexpectedEOF, nil
```
krishnasrinivas: But Read() would not return `io.ErrUnexpectedEOF`, right?

api-put-object-streaming.go
```diff
@@ -49,7 +49,7 @@ func (c Client) putObjectMultipartStream(bucketName, objectName string,
 	// while it is going to be duck typed similar to io.ReaderAt.
 	// It is to indicate that *minio.Object implements io.ReaderAt.
 	// and such a functionality is used in the subsequent code path.
-	if isObject(reader) || isReadAt(reader) || isFile(reader) {
+	if isFile(reader) || !isObject(reader) && isReadAt(reader) {
```
krishnasrinivas: The intention is to disable parallel upload of the parts if the src is minio.Object?

api-put-object-streaming.go
```diff
@@ -270,14 +270,8 @@ func (c Client) putObjectMultipartStreamNoChecksum(bucketName, objectName string
 
 		var objPart ObjectPart
 		objPart, err = c.uploadPart(bucketName, objectName, uploadID,
-			io.LimitReader(hookReader, partSize), partNumber,
-			nil, nil, partSize, metadata)
-		// For unknown size, Read EOF we break away.
-		// We do not have to upload till totalPartsCount.
-		if err == io.EOF && size < 0 {
-			break
-		}
```
krishnasrinivas: Does this mean we will not be supporting upload if size is `-1`?

harshavardhana: Streaming we never support size == -1,  for size == -1 we fall back to regular multipart upload.

api-put-object-streaming.go
harshavardhana: That is correct.. since it is not possible to support streaming parallel downloads in current structure for GetObject to PutObject().

pkg/s3signer/request-signature-streaming.go
harshavardhana: Actually you are right it either returns io.EOF or nil.  Let me change it.

api-put-object-streaming.go
harshavardhana: Instead we will add concurrent copy code on the `mc` side for in multiples of files.

harshavardhana: There is a possibility by implementing io.SectionReader() based on io.ReadSeeker instead of io.ReaderAt.  But it is not worth the complexity for now in my opinion.

pkg/s3signer/request-signature-streaming.go
```diff
@@ -265,23 +276,23 @@ func (s *StreamingReader) Read(buf []byte) (int, error) {
 					s.signChunk(s.chunkBufLen)
 					break
 				}
+			}
+			if err != nil && err != io.ErrUnexpectedEOF {
```
krishnasrinivas: @harshavardhana there is a reference to io.ErrUnexpectedEOF here as well

harshavardhana: This is fixed.. @krishnasrinivas

harshavardhana:
@krishnasrinivas any more reviews ?

PR #737, opened 2017-07-01
api: Upon failure abort the multipart upload.

harshavardhana:
Since we do not have resuming capabilities we
should not leave stale parts behind upon any
error during putObject(). This PR pro-actively
aborts the multipart upload created during
the putObject() for large files.

### DIFF
--- a/api-put-object-multipart.go
+++ b/api-put-object-multipart.go
@@ -53,10 +53,10 @@ func (c Client) putObjectMultipart(bucketName, objectName string, reader io.Read
 func (c Client) putObjectMultipartNoStream(bucketName, objectName string, reader io.Reader, size int64,
 	metadata map[string][]string, progress io.Reader) (n int64, err error) {
 	// Input validation.
-	if err := s3utils.CheckValidBucketName(bucketName); err != nil {
+	if err = s3utils.CheckValidBucketName(bucketName); err != nil {
 		return 0, err
 	}
-	if err := s3utils.CheckValidObjectName(objectName); err != nil {
+	if err = s3utils.CheckValidObjectName(objectName); err != nil {
 		return 0, err
 	}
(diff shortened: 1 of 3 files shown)


vadmeste:
LGTM though I think this PR is not so much useful in practice. I think most of the stale uploads are due to a premature exit during multipart upload.

harshavardhana:
> LGTM though I think this PR is not so much useful in practice. I think most of the stale uploads are due to a premature exit during multipart upload.

This is similar in behavior to how aws-sdk upload manager does. Lets say if one of the parts failed to upload we will never be able to clear it unless user explicitly does it on his own, most users are not aware. We should disallow that to happen and just abort the entire request similarly how aws-sdk does this.

vadmeste:
> This is similar in behavior to how aws-sdk upload manager does. Lets say if one of the parts failed to upload we will never be able to clear it unless user explicitly does it on his own, most users are not aware. We should disallow that to happen and just abort the entire request similarly how aws-sdk does this.

I know, but in practice if a part fails to upload due to a down server or internet connection (I don't see other possible errors here), I think that most likely the next call which aborts upload would fail (but I know not always).

I will approve this PR.

Can you rebase this @harshavardhana  ?

harshavardhana:
> I know, but in practice if a part fails to upload due to a down server or internet connection (I don't see other possible errors here), I think that most likely the next call which aborts upload would fail (but I know not always).

Not necessarily with AWS a part might fail just because of some slow network whereas an abort might complete. it can also happen that input stream is somewhat corrupted as well due to a local filesystem error.  So there are valid cases.. not just network alone.

harshavardhana:
Any reviews ?

harshavardhana:
@donatello any reviews?

PR #739, opened 2017-07-03
Handle source objects used client side encryption for compose object.

### LINKED ISSUE #738: Multiple sources those use client side encryption in copy object api should return error.
Current [implementation](https://github.com/minio/minio-go/pull/715) continues destination object creation if client side encryption is used in multi source object input.  As the resultant object is not useful, copy object API should not allow this case and return error properly.

balamurugana:
When multiple source objects are used for compose object, errors out
if client side encryption is used in source object.

Fixes #738

### DIFF
--- a/api-compose-object.go
+++ b/api-compose-object.go
@@ -323,6 +323,13 @@ func (c Client) ComposeObject(dst DestinationInfo, srcs []SourceInfo) error {
 			return fmt.Errorf("Could not get source props for %s/%s: %v", src.bucket, src.object, err)
 		}
 
+		// Error out if client side encryption is used in this source object when
+		// more than one source objects are given.
+		if len(srcs) > 1 && src.Headers.Get("x-amz-meta-x-amz-key") != "" {
+			return ErrInvalidArgument(
+				fmt.Sprintf("Client side encryption is used in source object %s/%s", src.bucket, src.object))
+		}
+
 		// Since we did a HEAD to get size, we use the ETag
 		// value to make sure the object has not changed by
 		// the time we perform the copy. This is done, only if


api-compose-object.go
```diff
@@ -315,6 +315,12 @@ func (c Client) ComposeObject(dst DestinationInfo, srcs []SourceInfo) error {
 			return fmt.Errorf("Could not get source props for %s/%s: %v", src.bucket, src.object, err)
 		}
 
+		// Error out if client side encryption is used in this source object.
+		if src.Headers.Get("x-amz-meta-x-amz-key") != "" {
```
harshavardhana: Did you verify what is the AWS S3 behavior in this scenario? @balamurugana

api-compose-object.go
```diff
@@ -315,6 +315,13 @@ func (c Client) ComposeObject(dst DestinationInfo, srcs []SourceInfo) error {
 			return fmt.Errorf("Could not get source props for %s/%s: %v", src.bucket, src.object, err)
 		}
 
+		// Error out if client side encryption is used in this source object when
+		// more than one source objects are given.
+		if i > 0 && src.Headers.Get("x-amz-meta-x-amz-key") != "" {
```
krisis: This check won't catch the case where the first source object is encrypted while others are not. I.e, `i == 0 && src.Headers.Get("x-amz-meta-x-amz-key") != ""` is true.

harshavardhana: Come to think of it i think this code needs to loop through an validates all sourceInfos.

balamurugana: Done.  Fixing.

PR #740, opened 2017-07-03
Export KeyInfo.GetSSEHeaders so users can more easily generate SSE-C headers

donatello:
This adds a method for users to more simply generate and set server-side-encryption headers.

### DIFF
--- a/api-compose-object.go
+++ b/api-compose-object.go
@@ -45,7 +45,7 @@ func NewSSEInfo(key []byte, algo string) SSEInfo {
 	return SSEInfo{key, algo}
 }
 
-// internal method to output the generated SSE headers.
+// internal method that computes SSE-C headers
 func (s *SSEInfo) getSSEHeaders(isCopySource bool) map[string]string {
 	if s == nil {
 		return nil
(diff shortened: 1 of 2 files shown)


api-compose-object.go
```diff
@@ -62,6 +62,14 @@ func (s *SSEInfo) getSSEHeaders(isCopySource bool) map[string]string {
 	}
 }
 
+// GetSSEHeaders - computes and returns headers for SSE-C as key-value
+// pairs. They can be set as metadata in PutObject* requests (for
+// encryption) or be set as request headers in `Core.GetObject` (for
+// decryption).
+func (s *SSEInfo) GetSSEHeaders() map[string]string {
```
harshavardhana: Lets keep it `map[string][]string`  if we are really going to export it.

donatello: No, doing it this way is the cleaner interface for users when adding these headers to another existing set of headers. Compare:

```go
// In the scenarios below, metadata is of type map[string][]string as this is the type 
// passed to put-object requests.

// 1. When header value is just a string:
key := NewSSEInfo("blahblah", "")
for k, v := range key.GetSSEHeaders() {
     metadata[k] = append(metadata[k], v)
}

// 2. When header value is a []string:
key := NewSSEInfo("blahblah", "")
for k, v := range key.GetSSEHeaders() {
     // Either we extract v[0] alone - which means implementation could have to change in future, or...
     metadata[k] = append(metadata[k], v...)
}
```

The point here is that there is no reason to making the return type a slice - there should be no option for user to append extra stuff for these sse-c headers.

harshavardhana: Hmm also Go doesn't allow immutable maps. Makes sense..

donatello: Any other reviews?

PR #744, opened 2017-07-06
api: getObject ObjectInfo should set metadata headers.

harshavardhana:


deekoder:
@vadmeste  can you approve this again.

PR #750, opened 2017-07-10
PutObject streams with unknown size use streaming V4.

harshavardhana:
Currently we used to fall back and use non-streaming
version but we can instead just use the streaming
version instead.

Does not fix any apparent bug, but it just allows
for avoiding any checksum calculation for large buffers.

### DIFF
--- a/api-put-object.go
+++ b/api-put-object.go
@@ -17,13 +17,15 @@
 package minio
 
 import (
+	"bytes"
+	"fmt"
 	"io"
 	"os"
 	"reflect"
 	"runtime"
+	"sort"
 	"strings"
 
-	"github.com/minio/minio-go/pkg/credentials"
 	"github.com/minio/minio-go/pkg/s3utils"
 )
 
@@ -200,16 +202,9 @@ func (c Client) putObjectCommon(bucketName, objectName string, reader io.Reader,
 		return c.putObjectMultipart(bucketName, objectName, reader, size, metadata, progress)
 	}
 
-	// If size cannot be found on a stream, it is not possible
-	// to upload using streaming signature, fall back to multipart.
 	if size < 0 {
-		// Set regular signature calculation.
-		c.overrideSignerType = credentials.SignatureV4
-		return c.putObjectMultipart(bucketName, objectName, reader, size, metadata, progress)
+		return c.putObjectMultipartStreamNoLength(bucketName, objectName, reader, size, metadata, progress)
 	}
-
-	// Set streaming signature.
-	c.overrideSignerType = credentials.SignatureV4Streaming
 
 	if size < minPartSize {
 		return c.putObjectNoChecksum(bucketName, objectName, reader, size, metadata, progress)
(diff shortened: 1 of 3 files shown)


api.go
```diff
@@ -700,7 +700,8 @@ func (c Client) newRequest(method string, metadata requestMetadata) (req *http.R
 	case signerType.IsV2():
 		// Add signature version '2' authorization header.
 		req = s3signer.SignV2(*req, accessKeyID, secretAccessKey)
-	case signerType.IsStreamingV4() && method == "PUT":
+	case metadata.objectName != "" && method == "PUT":
```
krisis: This change has the following user-visible impact.
1. signature type set in configuration or overridden via overrideSignerType` is not honored, for all PUT requests on object resources. This may confuse existing users.

2. removes some performance advantages that were available to users previously.
E.g, user uploading an object of _known_ size over HTTPS would like to benefit from not having to compute sha256 sum of the payload. It looks like this request will be signed using streaming signature forcing sha256 sum computation for every 'chunk'.

harshavardhana: Correct we just default to sha256 anyways.. This is faster than actually doing md5sum with https in the benchmarks.. Defaulting to streaming was already done for all PUT operations, this change just goes one step further and enables that also for situation where size of the object is not known.

Streaming signature is a subset of V4 not a different auth mechanism in itself . That is why the differentiation is removed and implemented in such a way that .. if V4 is configured and object is being uploaded they all default to streaming signature implementation.

krisis: > Correct we just default to sha256 anyways.. This is faster than actually doing md5sum with https in the benchmarks..

How does the performance of streaming signature compare with _not_ computing sha256 sum of payload when object is being uploaded over a HTTPS connection using signature v4? Is streaming signature based upload (which computes sha256 sum of every chunk) faster than UNSIGNED PAYLOAD using signature v4 over HTTPS?

harshavardhana: Even with unsigned payload we calculate content-md5 . That is what slows it down.. but you are right I can change that for secure connection can just skip this as well.

api.go
```diff
@@ -700,7 +700,11 @@ func (c Client) newRequest(method string, metadata requestMetadata) (req *http.R
 	case signerType.IsV2():
 		// Add signature version '2' authorization header.
 		req = s3signer.SignV2(*req, accessKeyID, secretAccessKey)
-	case signerType.IsStreamingV4() && method == "PUT":
+	case metadata.objectName != "" && method == "PUT" && metadata.customHeader.Get("X-Amz-Copy-Source") == "" && !c.secure:
+		// Following condition is only invoked when objectName is set,
```
krisis: This comment describes the low-level details which are available in the code. We could describe the same at a higher-level as,
`// Streaming signature is used by default for a PUT object request on a HTTP object storage server.`
In the higher-level description the reader is required to make the connection between the goal and the case expression that represents it. This seems reasonable to me. Thoughts?

harshavardhana: Yes.. done.

api.go
```diff
@@ -700,7 +700,8 @@ func (c Client) newRequest(method string, metadata requestMetadata) (req *http.R
 	case signerType.IsV2():
 		// Add signature version '2' authorization header.
 		req = s3signer.SignV2(*req, accessKeyID, secretAccessKey)
-	case signerType.IsStreamingV4() && method == "PUT":
+	case metadata.objectName != "" && method == "PUT" && metadata.customHeader.Get("X-Amz-Copy-Source") == "" && !c.secure:
```
vadmeste: Can you add a comment to explain why we check `c.secure` ?

vadmeste:
Looks like minio-go tests are not passing here:
```
--- FAIL: TestCorePutObject (5.90s)
        core_test.go:354: Error expected: nil, got:  <nil>
```

Could be not related to this PR.

harshavardhana:
> Looks like minio-go tests are not passing here:
> 
> --- FAIL: TestCorePutObject (5.90s)
>         core_test.go:354: Error expected: nil, got:  <nil>
> Could be not related to this PR.

This looks like some other problem. There shouldn't be an error here.

api.go
harshavardhana: Because we don't wna

harshavardhana:
@vadmeste can you take a look again the PR is updated.

api-put-object.go
```diff
@@ -200,21 +202,129 @@ func (c Client) putObjectCommon(bucketName, objectName string, reader io.Reader,
...
-
 	if size < minPartSize {
 		return c.putObjectNoChecksum(bucketName, objectName, reader, size, metadata, progress)
 	}
 
 	// For all sizes greater than 64MiB do multipart.
 	return c.putObjectMultipartStream(bucketName, objectName, reader, size, metadata, progress)
 }
+
+func (c Client) putObjectMultipartStreamNoLength(bucketName, objectName string, reader io.Reader, size int64,
```
vadmeste: Sorry for the delay. I don't understand one thing, why we are passing `size` parameter her though this function name is "putObjectMultipartStreamNoLength" (nolength). Besides, this function seems to be called only when size < 0.

harshavardhana: I guess we can change that.

harshavardhana: Can this be done in a later PR which refactors PutObject and get this merged ?

vadmeste: > Can this be done in a later PR which refactors PutObject and get this merged ?

Sure

PR #752, opened 2017-07-11
Add DisableCompression in http.Transport to override auto-decompress.

harshavardhana:
Fixes https://github.com/minio/mc/issues/2209

### DIFF
--- a/api.go
+++ b/api.go
@@ -211,7 +211,7 @@ func privateNew(endpoint string, creds *credentials.Credentials, secure bool, re
 
 	// Instantiate http client and bucket location cache.
 	clnt.httpClient = &http.Client{
-		Transport:     http.DefaultTransport,
+		Transport:     defaultMinioTransport,
 		CheckRedirect: redirectHeaders,
 	}
 
--- a/pkg/s3signer/request-signature-streaming.go
+++ b/pkg/s3signer/request-signature-streaming.go
@@ -99,7 +99,7 @@ func prepareStreamingRequest(req *http.Request, sessionToken string, dataLen int
 	if sessionToken != "" {
 		req.Header.Set("X-Amz-Security-Token", sessionToken)
 	}
-	req.Header.Set("Content-Encoding", streamingEncoding)
+	req.Header.Add("Content-Encoding", streamingEncoding)
 	req.Header.Set("X-Amz-Date", timestamp.Format(iso8601DateFormat))
 
 	// Set content length with streaming signature for each chunk included.
--- a/core_test.go
+++ b/core_test.go
@@ -18,15 +18,15 @@ package minio
 
 import (
 	"bytes"
-	"crypto/md5"
-	"log"
-
 	"io"
-	"math/rand"
+	"log"
 	"os"
 	"reflect"
 	"testing"
 	"time"
+
+	"crypto/md5"
+	"math/rand"
 )
 
 const (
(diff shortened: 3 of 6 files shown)


pkg/s3signer/request-signature-streaming.go
```diff
@@ -99,7 +99,12 @@ func prepareStreamingRequest(req *http.Request, sessionToken string, dataLen int
 	if sessionToken != "" {
 		req.Header.Set("X-Amz-Security-Token", sessionToken)
 	}
-	req.Header.Set("Content-Encoding", streamingEncoding)
+	contentEncoding := req.Header.Get("Content-Encoding")
+	if contentEncoding != "" {
+		req.Header.Set("Content-Encoding", strings.Join([]string{contentEncoding, streamingEncoding}, ","))
+	} else {
+		req.Header.Set("Content-Encoding", streamingEncoding)
+	}
```
krishnasrinivas: instead you can just do:
req.Header.Add("Content-Encoding", streamingEncoding)

harshavardhana: Oops yes forgot :P @krishnasrinivas

api.go
```diff
@@ -622,6 +622,20 @@ func (c Client) newRequest(method string, metadata requestMetadata) (req *http.R
...
+	//
+	// Refer:
+	//
+	//    https://golang.org/src/net/http/transport.go?h=roundTrip#L1843
+	//    "" Ask for a compressed version if the caller didn't set their
+	//    own value for Accept-Encoding. We only attempt to
+	//    uncompress the gzip stream if we were the layer that
+	//    requested it. ""
+	//
+	req.Header.Set("Accept-Encoding", "gzip")
```
vadmeste: Aren't we here literally saying: hey server, we can accept content in compressed format to gain bandwidth, now if the server supports that feature, will the client really be able to decompress content before delivering them to our users ?

Besides.. I am noticing this weird behavior, an empty header returned by AWS S3, but it seems it has nothing to do with this PR.
```
GET /presignedgetobject.go HTTP/1.1
Host: vadmeste.s3.amazonaws.com
User-Agent: Minio (linux; amd64) minio-go/3.0.0
Accept-Encoding: gzip
Authorization: AWS4-HMAC-SHA256 Credential=**REDACTED**/20170712/us-east-1/s3/aws4_request, SignedHeaders=accept-encoding;host;if-match;x-amz-content-sha256;x-amz-date, Signatu
re=**REDACTED**
If-Match: "6795a04d7e56a596c4cacafc0c148ae8"
X-Amz-Content-Sha256: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
X-Amz-Date: 20170712T151340Z

HTTP/1.1 200 OK
Content-Length: 1736
Accept-Ranges: bytes
Content-Encoding:
Date: Wed, 12 Jul 2017 15:13:41 GMT
Etag: "6795a04d7e56a596c4cacafc0c148ae8"
Last-Modified: Wed, 12 Jul 2017 15:12:55 GMT
Server: AmazonS3
X-Amz-Id-2: jt7yeuBacaGCBGdU1cJ1BVGZTAKZdWy5u3u7expyF61LyTBiARPqC/sPpSOQbHbC6+3HdTAoafw=
X-Amz-Request-Id: 9071463091587F0D
```

harshavardhana:
> Aren't we here literally saying: hey server, we can accept content in compressed format to gain bandwidth, now if the server supports that feature, will the client really be able to decompress content before delivering it to our users ?

Yes that is correct indication but we are doing that to avoid for copying and streaming.  Such that out GetObject() gets a compressed stream and the caller can decide on decompressing it or un-compressing it. minio-go client will not do any data manipulation.

harshavardhana:
> Yes that is correct indication but we are doing that to avoid auto-decompress for copying and streaming. Such that out GetObject() gets a compressed stream and the caller can decide on decompressing it or un-compressing it. minio-go client will not do any data manipulation.

Found a way to avoid setting any additional headers instead just use DisableCompression: true in the http.DefaultTranport which is more easier way to understand solves this problem.

api.go
```diff
@@ -190,6 +190,28 @@ func redirectHeaders(req *http.Request, via []*http.Request) error {
...
+	IdleConnTimeout:       90 * time.Second,
+	TLSHandshakeTimeout:   10 * time.Second,
+	ExpectContinueTimeout: 1 * time.Second,
+	// Set this value so that the underlying transport round-tripper
+	// doesn't try to auto decode the body of objects with
+	// content-encoding set to `gzip`.
+	//
+	// Refer:
+	//    https://golang.org/src/net/http/transport.go?h=roundTrip#L1843
+	DisableCompression: true,
```
vadmeste: We also need to do this in mc since mc sets its custom transport (primarily to satisfy --insecure flag). But I am OK with this.

harshavardhana: Yes we need to. @vadmeste

harshavardhana:
ping @vadmeste waiting on your review.

PR #767, opened 2017-07-20
Fix typo in the readme

fetmar:
Changed Println with formatting to a Printf

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ func main() {
 		log.Fatalln(err)
 	}
 
-	log.Printf("%v\n", minioClient) // minioClient is now setup
+	log.Printf("%#v\n", minioClient) // minioClient is now setup
 ```
 
 ## Quick Start Example - File Uploader
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ func main() {
 		log.Fatalln(err)
 	}
 
-	log.Println("%v", minioClient) // minioClient is now setup
+	log.Printf("%v\n", minioClient) // minioClient is now setup
 ```
 
 ## Quick Start Example - File Uploader


deekoder:
@fetmar why do you suggest we change Println to Printf ? Is the output looking better in the latter?

fetmar:
It's failing in travis but I'm not sure why since I didn't change anything but the readme. 

Fails here:
```
The command "SERVER_ENDPOINT=play.minio.io:9000 ACCESS_KEY=Q3AM3UQ867SPQQA43P2F SECRET_KEY=zuf+tfteSlswRu7BJ86wekitnifILbZam1KYY3TG ENABLE_HTTPS=1 go run functional_tests.go" exited with 2.
```

fetmar:
@deekoder Just because the original line there says %v, which is a formatting string. `Println` doesn't accept formatting strings, only things like `Sprintf` and `Printf`. `Println` is by default `%v`. So the way this is written, it's basically just as good as saying `Println(minioClient)`.

Actually, I like that better. I'll remove the formatting string altogether. :)

fetmar:
On second thought...

Println(minioClient)
```
&{{http  <nil> 192.168.99.100:30199   false  } 0xc420122240 0 { } false 0xc42011c7e0 0xc42010c4c0 true 0xc42000c018   0xc42011c840}
```

Printf("%+v\n", minioClient)
```
&{endpointURL:{Scheme:http Opaque: User:<nil> Host:192.168.99.100:30199 Path: RawPath: ForceQuery:false RawQuery: Fragment:} credsProvider:0xc420122240 overrideSignerType:0 appInfo:{appName: appVersion:} secure:false httpClient:0xc42011c7e0 bucketLocCache:0xc42010c4c0 isTraceEnabled:true traceOutput:0xc42000c018 s3AccelerateEndpoint: region: random:0xc42011c840}
```

Printf("%#v\n", minioClient)
```
&minio.Client{endpointURL:url.URL{Scheme:"http", Opaque:"", User:(*url.Userinfo)(nil), Host:"192.168.99.100:30199", Path:"", RawPath:"", ForceQuery:false, RawQuery:"", Fragment:""}, credsProvider:(*credentials.Credentials)(0xc420122240), overrideSignerType:0, appInfo:struct { appName string; appVersion string }{appName:"", appVersion:""}, secure:false, httpClient:(*http.Client)(0xc42011c7e0), bucketLocCache:(*minio.bucketLocationCache)(0xc42010c4a0), isTraceEnabled:true, traceOutput:(*os.File)(0xc42000c018), s3AccelerateEndpoint:"", region:"", random:(*rand.Rand)(0xc42011c840)}
```

I prefer the last option. It's the prettiest of the prints that's a part of go's standard library. 🌹

harshavardhana:
// cc @deekoder we can take this in if you don't have further questions?

harshavardhana:
// cc @deekoder if you don't have any further questions we can take this in.

deekoder:
Sorry for the delay. Merging now.

PR #770, opened 2017-07-24
Add Client.PresignedHeadObject method

dmage:
Added a method to get a presigned URL to access an object [metadata](http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectHEAD.html) without credentials.

### DIFF
--- a/api-presigned.go
+++ b/api-presigned.go
@@ -91,6 +91,12 @@ func (c Client) PresignedGetObject(bucketName string, objectName string, expires
 	return c.presignURL("GET", bucketName, objectName, expires, reqParams)
 }
 
+// PresignedHeadObject - Returns a presigned URL to access an object metadata
+// without credentials. Expires maximum is 7days - ie. 604800 and minimum is 1.
+func (c Client) PresignedHeadObject(bucketName string, objectName string, expires time.Duration, reqParams url.Values) (u *url.URL, err error) {
+	return c.presignURL("HEAD", bucketName, objectName, expires, reqParams)
+}
+
 // PresignedPutObject - Returns a presigned URL to upload an object without credentials.
 // Expires maximum is 7days - ie. 604800 and minimum is 1.
 func (c Client) PresignedPutObject(bucketName string, objectName string, expires time.Duration) (u *url.URL, err error) {


harshavardhana:
Can you explain a bit more about this ? why is this needed which is not achieved by PresignedGetObject() ?

dmage:
I'm implementing yet another S3 driver for [Docker Distribution](https://github.com/docker/distribution). And Docker uses HEAD requests to check if the layer of an image is already uploaded. If I redirect the Docker client to PresignedGetObject URL while it performs a HEAD request, it will get 403, because the URL is signed for the different method.

harshavardhana:
> I'm implementing yet another S3 driver for Docker Distribution. And Docker uses HEAD requests to check if the layer of an image is already uploaded. If I redirect the Docker client to PresignedGetObject URL while it performs a HEAD request, it will get 403, because the URL is signed for the different method.

Curious does current implementation of S3 driver of any presigned operations, why is it necessary to use presigned operations? is this URL being shared to some other entity which would validate upon the object? 

Another question is why a new S3 driver any improvements that you have in mind that you can share with us?

dmage:
> does current implementation of S3 driver use any presigned operations

[Yes, it does.](https://github.com/docker/distribution/blob/e18fe7d3f4573faac5a32a0df3322a4812969c92/registry/storage/driver/s3-aws/s3.go#L859-L874)

They are used to take off workload from the docker registry. For example, when you say `docker pull minio/minio`, Docker gets from Docker Hub presigned redirects to image layers. So it saves bandwidth on the registry servers.

It doesn't save bandwidth in case of HEAD request, though. I need to investigate it further.

> why a new S3 driver

Current implementation uses aws-sdk-go which AFAIK does not support chunked uploads. It means that you need at least a 5 MB buffer to sign an upload part. This causes OOM if a burst of uploads come to our registry.

harshavardhana:
> They are used to take off workload from the docker registry. For example, when you say docker pull minio/minio, Docker gets from Docker Hub presigned redirects to image layers. So it saves bandwidth on the registry servers.
> 
> It doesn't save bandwidth in case of HEAD request, though. I need to investigate it further.
> 
> why a new S3 driver
> Current implementation uses aws-sdk-go which AFAIK does not support chunked uploads. It means that you need at least a 5 MB buffer to sign an upload part. This causes OOM if a burst of uploads come to our registry.

Understood now. Here are few things i have in mind. 

In a scenario like this i think we can make Presign little generic? like for rest of all object operations? .. like an API like this. 

```
PresignURL(method string, bucketName, objectName string, expiry time.Duration, reqParams url.Values)

Usage: 

PresignURL(http.MethodHead, "testbucket", "myobject", time.Hour * 1,  url.Values{})
```

https://github.com/minio/minio-go/blob/master/api-presigned.go#L40  we already have an internal code which can be exposed. i.e just rename it as PresignURL. 

I am of the opinion that it kind of doesn't make sense to add new API's for HEAD - since there are hardly users for presigned HEAD, DELETE, POST etc. and they are not required to be first class APIs.

Will check with @abperiasamy and get back to you if we can make it explicit or have a single function for rest of the HTTP methods ops.

harshavardhana:
// cc @deekoder we need to discuss this with @abperiasamy on the final API for presignedHeadObject.

harshavardhana:
We have decided to implement this API @dmage thanks for sending this over. Can you rebase to master and 

  - Add functional tests
  - Update the relevant API doc in docs/API.md 
  - Add an example at `examples/s3/presignedhead-object.go` 

Thanks for the PR and sorry for the delay.

harshavardhana:
This is already merged after adding documentation and tests, closing this PR.

PR #772, opened 2017-07-26
added insecure mode

tommyvicananza:
Since it's possible to put an object to s3 endpoint and certificate fails, I added a new parameter to enable TLS Insecure Skip Verify

### DIFF
--- a/api.go
+++ b/api.go
@@ -21,6 +21,7 @@ import (
 	"bytes"
 	"crypto/md5"
 	"crypto/sha256"
+	"crypto/tls"
 	"encoding/base64"
 	"encoding/hex"
 	"errors"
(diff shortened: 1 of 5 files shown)


harshavardhana:
You don't need to do this.. look at how it is done in `mc`  https://github.com/minio/mc/blob/master/cmd/client-s3.go#L122 

Once you have a custom roundTripper you can use the SetCustomTransport() to have this behavior.

tommyvicananza:
You're totally right I didn't notice about the SetCustomTranpsort() sorry for the inconvenience.

I close the PR.

harshavardhana:
> You're totally right I didn't notice about the SetCustomTranpsort() sorry for the inconvenience.
> 
> I close the PR.

No problem feel free to send PRs if you find any issues.. really appreciate your time.

PR #776, opened 2017-07-31
api: Optimize multipart upload memory usage for unknown sized stream

### LINKED ISSUE #730: optimalPartInfo() default partSize results in running out of memory

harshavardhana:
Fixes #730

### DIFF
--- a/api-put-object-multipart.go
+++ b/api-put-object-multipart.go
@@ -27,13 +27,14 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
 
 	"github.com/minio/minio-go/pkg/s3utils"
 )
 
 func (c Client) putObjectMultipart(bucketName, objectName string, reader io.Reader, size int64,
 	metadata map[string][]string, progress io.Reader) (n int64, err error) {
-	n, err = c.putObjectMultipartNoStream(bucketName, objectName, reader, size, metadata, progress)
+	n, err = c.putObjectMultipartNoStream(bucketName, objectName, reader, metadata, progress)
 	if err != nil {
 		errResp := ToErrorResponse(err)
 		// Verify if multipart functionality is not available, if not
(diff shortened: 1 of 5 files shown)


harshavardhana:
Any reviews here @balamurugana @vadmeste ?

harshavardhana:
Changes are done @balamurugana - ready to review again.

api-put-object-multipart.go
```diff
@@ -100,53 +107,54 @@ func (c Client) putObjectMultipartNoStream(bucketName, objectName string, reader
 		// HTTPS connection.
 		hashAlgos, hashSums := c.hashMaterials()
 
-		// Calculates hash sums while copying partSize bytes into tmpBuffer.
-		prtSize, rErr := hashCopyN(hashAlgos, hashSums, tmpBuffer, reader, partSize)
-		if rErr != nil && rErr != io.EOF {
+		bufp := bufPool.Get().(*[]byte)
+		prtSize, rErr := io.ReadFull(reader, *bufp)
```
balamurugana: `length` may be better than `prtSize`

api-put-object-multipart.go
```diff
@@ -100,53 +107,54 @@ func (c Client) putObjectMultipartNoStream(bucketName, objectName string, reader
...
-
 		// Save successfully uploaded size.
-		totalUploadedSize += prtSize
+		totalUploadedSize += int64(prtSize)
 
 		// Increment part number.
 		partNumber++
 
+		// Put back data into bufpool.
+		bufPool.Put(bufp)
```
balamurugana: this statement can be in `defer`

harshavardhana: We cannot defer @balamurugana we need to get this back in the next loop.

api-put-object-multipart.go
harshavardhana: Sure.

harshavardhana:
@vadmeste need your reviews as well.

PR #787, opened 2017-08-15
fix: Extract region from s3-<region>.amazonaws.com

### LINKED ISSUE #785: Using IAM instance profile credentials to authenticate to S3
I'm trying to update [Gitlab's CI runner](https://gitlab.com/gitlab-org/gitlab-ci-multi-runner) to be able to use instance profile credentials (merged in to Minio with https://github.com/minio/minio-go/commit/5d7ee332f62e83d36beba669be671801180abb89) but I'm getting a 400 response back when actually trying to use it.

Looking at the HTTP Trace I can see the following:
```
---------START-HTTP---------
GET /my-bucket-name/?location= HTTP/1.1
Host: s3-eu-west-1.amazonaws.com
User-Agent: Minio (linux; amd64) minio-go/3.0.0
Authorization: AWS4-HMAC-SHA256 Credential=**REDACTED**/20170814/us-east-1/s3/aws4_request, SignedHeaders=host;x-amz-content-sha256;x-amz-date;x-amz-security-token, Signature=**REDACTED**
X-Amz-Content-Sha256: UNSIGNED-PAYLOAD
X-Amz-Date: 20170814T081701Z
X-Amz-Security-Token: **REDACTED**
Accept-Encoding: gzip
[1B blob data]
---------END-HTTP---------
```

The thing that jumps out to me is the authorization header containing `us-east-1` even though my host is `s3-eu-west-1.amazonaws.com`. I previously had it requesting `s3.amazonaws.com` and still got a 400 response back but presumed that's because my bucket is not in `us-east-1` so corrected the host that I was passing into Minio's client.

Do I need to be doing something different here to get the authorization header to use the correct region or should I actually be using `s3.amazonaws.com` as the Host (despite my bucket not being in us-east-1/standard) and then doing something else to fix that?

harshavardhana:
This PR fixes a behavior problem on our end when URLs
like `s3-<region>` are allowed we should use the region
name automatically otherwise calls like GetBucketLocation()
would fail by defaulting to `us-east-1`.

Fixes #785

### DIFF
--- a/api.go
+++ b/api.go
@@ -190,6 +190,31 @@ func redirectHeaders(req *http.Request, via []*http.Request) error {
 	return nil
 }
 
+// getRegionFromURL - parse region from URL if present.
+func getRegionFromURL(u url.URL) (region string) {
+	region = ""
+	if s3utils.IsGoogleEndpoint(u) {
+		return
+	} else if s3utils.IsAmazonChinaEndpoint(u) {
+		// For china specifically we need to set everything to
+		// cn-north-1 for now, there is no easier way until AWS S3
+		// provides a cleaner compatible API across "us-east-1" and
+		// China region.
+		return "cn-north-1"
+	} else if s3utils.IsAmazonGovCloudEndpoint(u) {
+		// For us-gov specifically we need to set everything to
+		// us-gov-west-1 for now, there is no easier way until AWS S3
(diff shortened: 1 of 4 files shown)


harshavardhana:
@poornas @vadmeste any reviews?

Issue #747, opened 2017-07-07
PutObject* refactor


donatello:
Plan to converge to a smaller set of PutObject functions:

```go
type PutObjectOptions struct {
	UserMetadata       map[string]string
	Progress           io.Reader
	ContentType        string
	ContentEncoding    string
	ContentDisposition string
	CacheControl       string
}

func (c Client) PutObject(bucketName, objectName string, reader io.Reader, objectSize int64,
	opts *PutObjectOptions) (n int64, err error) {
	// TODO
}

func (c Client) FPutObject(bucketName, objectName, filePath string,
	opts *PutObjectOptions) (n int64, err error) {
	// TODO
}
```

All other put object functions shall be removed.

PR #794, opened 2017-08-21
Allow number of workers used for multipart operation to be configurable

### LINKED ISSUE #759: Allow multipart upload workers to be configurable.
Currently 

```
vendor/github.com/minio/minio-go/constants.go:var totalWorkers = 3
```

totalWorkers are set to 3 and are not configurable. Perhaps make this global variable public and allow PutObject() to use this variable instead. 

Related issue please see https://github.com/minio/mc/issues/1974

poornas:
Fixes #759.

Currently, total number of parallel workers is set as constant at 3. With this PR, it is exposed as a public variable that can be configured so that user can configure number of parallel workers for streaming put operations.

### DIFF
--- a/api-put-object-streaming.go
+++ b/api-put-object-streaming.go
@@ -152,7 +152,7 @@ func (c Client) putObjectMultipartStreamFromReadAt(bucketName, objectName string
 	close(uploadPartsCh)
 
 	// Receive each part number from the channel allowing three parallel uploads.
-	for w := 1; w <= totalWorkers; w++ {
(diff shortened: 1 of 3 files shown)


krisis:
@poornas since this configuration is applicable only for multipart-upload using streaming signature, it makes sense to keep it with other PutObject related options as proposed in the PutObject API refactor in #747.

poornas:
closing this PR as #811 addresses this as part of the PutObject refactor.

PR #798, opened 2017-08-25
Add new API PresignedHeadObject method.

harshavardhana:
This PR borrows original work from @dmage in PR #770

### DIFF
--- a/api-presigned.go
+++ b/api-presigned.go
@@ -91,6 +91,12 @@ func (c Client) PresignedGetObject(bucketName string, objectName string, expires
 	return c.presignURL("GET", bucketName, objectName, expires, reqParams)
 }
 
+// PresignedHeadObject - Returns a presigned URL to access an object metadata
+// without credentials. Expires maximum is 7days - ie. 604800 and minimum is 1.
+func (c Client) PresignedHeadObject(bucketName string, objectName string, expires time.Duration, reqParams url.Values) (u *url.URL, err error) {
+	return c.presignURL("HEAD", bucketName, objectName, expires, reqParams)
+}
+
 // PresignedPutObject - Returns a presigned URL to upload an object without credentials.
 // Expires maximum is 7days - ie. 604800 and minimum is 1.
 func (c Client) PresignedPutObject(bucketName string, objectName string, expires time.Duration) (u *url.URL, err error) {
--- a/api-presigned.go
+++ b/api-presigned.go
@@ -84,23 +84,33 @@ func (c Client) presignURL(method string, bucketName string, objectName string, 
 }
 
 // PresignedGetObject - Returns a presigned URL to access an object
-// without credentials. Expires maximum is 7days - ie. 604800 and
-// minimum is 1. Additionally you can override a set of response
-// headers using the query parameters.
+// data without credentials. URL can have a maximum expiry of
+// upto 7days or a minimum of 1sec. Additionally you can override
+// a set of response headers using the query parameters.
 func (c Client) PresignedGetObject(bucketName string, objectName string, expires time.Duration, reqParams url.Values) (u *url.URL, err error) {
 	return c.presignURL("GET", bucketName, objectName, expires, reqParams)
 }
 
-// PresignedHeadObject - Returns a presigned URL to access an object metadata
-// without credentials. Expires maximum is 7days - ie. 604800 and minimum is 1.
+// PresignedHeadObject - Returns a presigned URL to access object
+// metadata without credentials. URL can have a maximum expiry of
+// upto 7days or a minimum of 1sec. Additionally you can override
+// a set of response headers using the query parameters.
 func (c Client) PresignedHeadObject(bucketName string, objectName string, expires time.Duration, reqParams url.Values) (u *url.URL, err error) {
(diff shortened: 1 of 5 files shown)


api-presigned.go
```diff
@@ -91,12 +91,24 @@ func (c Client) PresignedGetObject(bucketName string, objectName string, expires
...
+}
+
 // PresignedPutObject - Returns a presigned URL to upload an object without credentials.
 // Expires maximum is 7days - ie. 604800 and minimum is 1.
 func (c Client) PresignedPutObject(bucketName string, objectName string, expires time.Duration) (u *url.URL, err error) {
 	return c.presignURL("PUT", bucketName, objectName, expires, nil)
 }
 
+// Presign - returns a presigned URL for any http method of your choice.
+// Expires maximum in 7days - i.e 604800 and minimum in 1.
```
poornas: Expires maximum in 7days  -> Expires maximum is 7 days.

docs/API.md
```diff
@@ -1093,14 +1091,44 @@ if err != nil {
 fmt.Println(presignedURL)
 ```
 
+<a name="PresignedHeadObject"></a>
+### PresignedHeadObject(bucketName, objectName string, expiry time.Duration, reqParams url.Values) (*url.URL, error)
+
+Generates a presigned URL for HTTP HEAD operations. Browsers/Mobile clients may point to this URL to directly get metadata from objects even if the bucket is private. This presigned URL can have an associated expiration time in seconds after which it is no longer operational. The default expiry is set to 7 days.
+
```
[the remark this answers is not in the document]
harshavardhana: No default is 7 days if not set that is.

functional_tests.go
```diff
@@ -2166,14 +2166,32 @@ func testFunctional() {
 		logger().Fatal("Error: ", err)
 	}
 
+	// Generate presigned HEAD object url.
+	presignedHeadURL, err := c.PresignedHeadObject(bucketName, objectName, 3600*time.Second, nil)
```
vadmeste: Can you please use another instance of Minio client initialized without credentials ? We will be more close to real world use case

examples/s3/presignedheadobject.go
```diff
@@ -0,0 +1,53 @@
...
+func main() {
+	// Note: YOUR-ACCESSKEYID, YOUR-SECRETACCESSKEY, my-bucketname and my-objectname
+	// are dummy values, please replace them with original values.
+
+	// Requests are always secure (HTTPS) by default. Set secure=false to enable insecure (HTTP) access.
+	// This boolean value is the last argument for New().
+
+	// New returns an Amazon S3 compatible client object. API compatibility (v2 or v4) is automatically
+	// determined based on the Endpoint value.
+	s3Client, err := minio.New("s3.amazonaws.com", "YOUR-ACCESSKEYID", "YOUR-SECRETACCESSKEY", true)
```
vadmeste: This example should be updated to do not use access and secret keys

docs/API.md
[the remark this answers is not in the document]
vadmeste: But also the maximum is 7 AFAIK

harshavardhana:
Can you guys review again @vadmeste @poornas ?

api-presigned.go
```diff
@@ -84,25 +84,40 @@ func (c Client) presignURL(method string, bucketName string, objectName string,
...
+// the query parameters.
 func (c Client) PresignedHeadObject(bucketName string, objectName string, expires time.Duration, reqParams url.Values) (u *url.URL, err error) {
 	return c.presignURL("HEAD", bucketName, objectName, expires, reqParams)
 }
 
-// PresignedPutObject - Returns a presigned URL to upload an object without credentials.
-// Expires maximum is 7days - ie. 604800 and minimum is 1.
+// PresignedPutObject - Returns a presigned URL to upload an object
+// without credentials. URL can have a maximum expiry of upto 7days
+// or a minimum of 1sec, if expiry duration is set to 0 URL a
```
vadmeste: I didn't test it but after reading the code, it seems setting 0 to expires returns an error (invoked by isValidExpiry() in presignURL())

harshavardhana: I see looks like behavior has changed.. @vadmeste - let me change the comment to reflect the new changes.

api-presigned.go
```diff
@@ -84,19 +84,40 @@ func (c Client) presignURL(method string, bucketName string, objectName string,
 }
 
 // PresignedGetObject - Returns a presigned URL to access an object
-// without credentials. Expires maximum is 7days - ie. 604800 and
-// minimum is 1. Additionally you can override a set of response
-// headers using the query parameters.
+// data without credentials. URL can have a maximum expiry of
+// upto 7days or a minimum of 1sec, if expiry duration is set to 0
+// URL a maximum value of 7 days is automatically choosen. Additionally
```
poornas: +// upto 7days or a minimum of 1sec, if expiry duration is set to 0
 +// URL a maximum value of 7 days is automatically choosen.

Should be:
+// upto 7days or a minimum of 1sec, if expiry duration is set to 0
 +// ~~URL~~ a maximum value of 7 days is automatically cho~~o~~sen.

Ditto for the other 3 function header comments

harshavardhana:
@vadmeste waiting on your review.

PR #807, opened 2017-08-31
Move PutObjectStreaming() to the correct column in the ToC

### LINKED ISSUE #806: Is PutObjectStreaming() listed in the wrong spot?
Looking at the hmm... "Table of Contents" thing here:

&nbsp; &nbsp; https://docs.minio.io/docs/golang-client-api-reference

`PutObjectStreaming()` is listed underneath the "Encrypted Object operations" column.

Is that correct?  Kind of thinking it should it be listed under "Object operations" instead, but not sure.

justinclift:
Fixes #806


harshavardhana:
ping @donatello waiting on your review..

PR #810, opened 2017-09-02
Fix for functional tests using random file sizes with static files (#809)

### LINKED ISSUE #809: Functional tests use existing files but size of the files is expected to be random, causing it to fail in Mint
Code such as the following, will work, if the file datafile-33-kB doesn't exist.

```go
bufSize := rand.Intn(1<<20) + 32*1024
var reader = getDataReader("datafile-33-kB", bufSize)
```

In the case of Mint, these files do exist and because that, bufSize is set to anywhere from 32Kb to 1MB + 32KB. The file size is always set to 33KB. So, code which checks if the size of the object that was pushed to the s3 server will fail if the buf size is > 33Kb. 

Fix is to not do any random size or change the parameter that is passed to the random number generator to 10.

kannappanr:
Fixed the issue by using the size of the 33KB file instead of random size. 

Fixes #809


functional_tests.go
```diff
@@ -45,7 +45,8 @@ import (
 )
 
 const (
-	sixtyFiveMiB = 1024 * 1024 * 65 // 65MiB
+	sixtyFiveMiB  = 1024 * 1024 * 65 // 65MiB
+	thirtyThreeKB = 1024 * 33        //33KB
```
balamurugana: How about using `humanize.MiByte` and `humanize.KiByte`?

PR #811, opened 2017-09-05
Refactor PutObject API calls to add context and additional metadata.

### LINKED ISSUE #700: Support for context.Context?
Hi,

are there plans to support request cancellation via `context.Context`? I'm currently in the progress of adding proper cancellation to restic, and I'd like to be able to cancel outgoing requests to the s3 backend...

Cheers,
Alex

### LINKED ISSUE #747: PutObject* refactor

### LINKED ISSUE #759: Allow multipart upload workers to be configurable.

poornas:
- Add additional Api calls PutObjectWithContext, GetObjectWithContext, FPutObjectWithContext and FGetObjectWithContext to allow request cancellation.
- Remove PutObjectWithSize,PutObjectStreaming,PutEncryptedObject,PutObjectWithMetadata calls and fold into a single PutObject call which accepts a pointer to PutObjectOptions struct. PutObjectOptions struct allows user to enter custom metadata, Content-Encoding, Cache-Control, Content-Type and Content-Disposition headers, configure number of threads for multipart PutObject calls, and set Encryption materials module to encrypt data.
- Change FPutObject API signature to add additional parameter - a pointer to PutObjectOptions struct.
- Number of workers used in multipart Put operation is now configurable by setting the number of threads as an optional field in the PutObjectOptions struct.

Fixes #700 
Fixes #747 
Fixes #759


api-get-object-file.go
```diff
@@ -21,11 +21,18 @@ import (
...
 	"path/filepath"
 
+	"context"
+
 	"github.com/minio/minio-go/pkg/s3utils"
 )
 
 // FGetObject - download contents of an object to a local file.
 func (c Client) FGetObject(bucketName, objectName, filePath string) error {
+	return c.fGetObjectWithContext(context.Background(), bucketName, objectName, filePath)
```
vadmeste: I am not sure how it is useful to provide two APIs which has the same functionality, for example here:
- FGetObject
and
- FGetObjectWithContext

I think we need to decide if they should have different behaviors or merge them into one function

poornas: It was decided to have an extra variant of Put/GetObject when context is an additional argument, rather than force user to pass nil each time if they do not need request cancellation. We could however revisit this in the next meeting and reach a consensus.

vadmeste: > rather than force user to pass nil each time if they do not need request cancellation.

emm, in that case, we need to pass nil here (if it works) and not context.Background()

poornas: as per https://golang.org/pkg/context/
Do not pass a nil Context, even if a function permits it. Pass context.TODO if you are unsure about which Context to use. 

Since we are wrapping the functions, it makes sense to use context.Background(). Agree?

vadmeste: > Since we are wrapping the functions, it makes sense to use context.Background(). Agree?

sorry, yeah I see now how it can be useful to have two APIs

poornas:
@vadmeste  and @harshavardhana , this needs reviews.

api-put-object-context.go
```diff
@@ -0,0 +1,41 @@
...
+
+// PutObjectWithContext - Identical to PutObject call, but accepts context to facilitate request cancellation.
+func (c Client) PutObjectWithContext(ctx context.Context, bucketName, objectName string, reader io.Reader, objectSize int64,
+	opts *PutObjectOptions) (n int64, err error) {
+	if opts == nil {
+		opts = &PutObjectOptions{}
+	}
+
+	if opts.EncryptMaterials != nil {
+		if err := opts.EncryptMaterials.SetupEncryptMode(reader); err != nil {
```
harshavardhana: This `err` assignment shadows with `err` return value. Just use `err =` instead of `err :=`

api-put-object-file-context.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	// Save the file size.
+	fileSize := fileStat.Size()
+
+	if opts == nil {
+		opts = &PutObjectOptions{}
+	}
+	// Set contentType based on filepath extension if not given or default
+	// value of "binary/octet-stream" if the extension has no associated type.
```
harshavardhana: Perhaps we can name this as `application/octet-stream` instead.  Since that is what we set eventually instead of `binary/octet-stream`

api-put-object.go
```diff
@@ -18,119 +18,73 @@ package minio
...
-			if len(result) == 1 {
-				size = toInt(result[0])
-			}
-		}
+	return
+}
+
+// getMetaData - constructs the headers from metadata entered by user in
+// PutObjectOptions struct
+func (opts *PutObjectOptions) getMetadata() (headers map[string][]string) {
```
harshavardhana: We can make the return value to be more typed i.e `getMetadataHeaders() (headers http.Header)` instead.

poornas: done.

api-put-object-file-context.go
poornas: this was unintentional. fixed.

api-put-object-context.go
poornas: fixed.

harshavardhana:
One suggestion is

>Add additional Api calls PutObjectWithContext, GetObjectWithContext,
FPutObjectWithContext and FGetObjectWithContext to allow request cancellation.
> Remove PutObjectWithSize,PutObjectStreaming,PutEncryptedObject,PutObjectWithMetadata
calls and fold into a single PutObject call which accepts a pointer to PutObjectOptions
struct. PutObjectOptions struct allows user to enter custom metadata, Content-Encoding,
Cache-Control, Content-Type and Content-Disposition headers, configure number of threads
for multipart PutObject calls, and set Encryption materials module to encrypt data.

^^ This paragraph can be worded a bit better.

> Pass pointer to PutObjectOptions struct as argument to FPutObject call.
This is a not a commit body worthy message can can be fixed.

> Core API methods PutObject, NewMultipartUpload, PutObjectPart and PutObjectPartWithMetadata
signatures have changed.
This can be moved above.

> Remove go 1.5,1.6 from Travis. Add 1.9

We can say `Deprecated support for Go versions 1.5.x, 1.6.x,  Add support for 1.9.x`

poornas:
@harshavardhana, reworded commit message.

api-get-policy.go
```diff
@@ -79,7 +80,7 @@ func (c Client) getBucketPolicy(bucketName string) (policy.BucketAccessPolicy, e
 	urlValues.Set("policy", "")
 
 	// Execute GET on bucket to list objects.
-	resp, err := c.executeMethod("GET", requestMetadata{
+	resp, err := c.executeMethod(context.Background(), "GET", requestMetadata{
```
vadmeste: @poornas, is there any reason why we are not accepting letting developers their own context here?

api-put-object-common.go
```diff
@@ -77,7 +78,7 @@ func optimalPartInfo(objectSize int64) (totalPartsCount int, partSize int64, las
 
 // getUploadID - fetch upload id if already present for an object name
 // or initiate a new request to fetch a new upload id.
-func (c Client) newUploadID(bucketName, objectName string, metaData map[string][]string) (uploadID string, err error) {
+func (c Client) newUploadID(ctx context.Context, bucketName, objectName string, opts *PutObjectOptions) (uploadID string, err error) {
```
vadmeste: I like PutObjectOptions change, but couldn't we do it in a separate PR ? (let's keep it as it now since I am already late reviewing this PR)

api-get-object-context.go
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package minio
+
+import "context"
+
+// GetObjectWithContext - returns an seekable, readable object.
+func (c Client) GetObjectWithContext(ctx context.Context, bucketName, objectName string) (*Object, error) {
```
vadmeste: I think we can avoid creating new file for this function, we can put it in api-get-file.go

api-get-object-file-context.go
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package minio
+
+import "context"
+
+// FGetObjectWithContext - download contents of an object to a local file.
+func (c Client) FGetObjectWithContext(ctx context.Context, bucketName, objectName, filePath string) error {
```
vadmeste: We probably need to avoid creating a new file for this function

api-put-object.go
```diff
@@ -263,21 +185,20 @@ func (c Client) putObjectMultipartStreamNoLength(bucketName, objectName string,
 
 	for partNumber <= totalPartsCount {
 		length, rErr := io.ReadFull(reader, buf)
-		if rErr == io.EOF {
+		if rErr == io.EOF && partNumber > 1 {
 			break
 		}
-		if rErr != nil && rErr != io.ErrUnexpectedEOF {
+		if rErr != nil && rErr != io.ErrUnexpectedEOF && partNumber > 1 {
```
vadmeste: Here,  `io.ReadFull()` can return an error other than `io.ErrUnexpectedEOF` and `io.EOF` but we ignore it when `partNumber == 1`. In other terms, we will create an object even if the reader returns a serious error.

api-put-object-context.go
```diff
@@ -0,0 +1,41 @@
...
+	if opts == nil {
+		opts = &PutObjectOptions{}
+	}
+
+	if opts.EncryptMaterials != nil {
+		if err = opts.EncryptMaterials.SetupEncryptMode(reader); err != nil {
+			return 0, err
+		}
+
+		// Set the necessary encryption headers, for future decryption.
```
vadmeste: Can you remove this comment ? it doesn't make sense anymore

api-get-policy.go
poornas: @vadmeste, the scope of this PR is to limit adding context to PutObject calls because it is likely to be long running enough to warrant a request cancellation. There is not much of a use case for GetBucketPolicy to use context.

vadmeste: :+1:

api-get-object-context.go
poornas: done.

api-get-object-file-context.go
poornas: done.

api-put-object-context.go
poornas: done.

api-put-object.go
poornas: fixed.

api-put-object-common.go
poornas: @vadmeste, the PutObjectOptions change is the main reason for this PR. Context change is minor since we are just passing it through to the executeMethod everywhere so that request can get context. It made sense to combine the changes in one PR.

poornas:
@vadmeste, updated the PR with your suggestions.Could you please review?

PR #813, opened 2017-09-19
api: Retry only request body is seekable

### LINKED ISSUE #791: Functional tests not passing when Minio server has a new default region
When I set Minio region to "us-west-2" in ~/.minio/config.json, functional tests won't pass anymore:

```
$ go run functional_tests.go
...
INFO[0015]                                               file=functional_tests.go function:=main.testGetObjectClosedTwice line#=665
FATA[0041] Error:Put http://localhost:9000/minio-go-testiuzn23cfq9jrzlcnz/1yzgz903xtgj9tceycazirws0cmlla: Connection closed by foreign host http://localhost:9000/minio-go-testiuzn23cfq9jrzlcnz/1yzgz903xtgj9tceycazirws0cmlla. Retry again.minio-go-testiuzn23cfq9jrzlcnz1yzgz903xtgj9tceycazirws0cmlla  file=functional_tests.go function:=main.testGetObjectClosedTwice line#=705
exit status 1
```

vadmeste:
It doesn't make sense to retry a HTTP request when its body is not
seekable. Req Body can be delivered by a developer such as in
PutObject() API.

Partially fix - ref #791

### DIFF
--- a/api.go
+++ b/api.go
@@ -497,6 +497,8 @@ var successStatus = []int{
 func (c Client) executeMethod(method string, metadata requestMetadata) (res *http.Response, err error) {
 	var isRetryable bool     // Indicates if request can be retried.
 	var bodySeeker io.Seeker // Extracted seeker from io.Reader.
+	var reqRetry = MaxRetry  // Indicates how many times we can retry the request
+
 	if metadata.contentBody != nil {
 		// Check if body is seekable then it is retryable.
 		bodySeeker, isRetryable = metadata.contentBody.(io.Seeker)
@@ -504,6 +506,11 @@ func (c Client) executeMethod(method string, metadata requestMetadata) (res *htt
 		case os.Stdin, os.Stdout, os.Stderr:
 			isRetryable = false
 		}
+		// Retry only when reader is seekable
+		if !isRetryable {
+			reqRetry = 1
+		}
+
 		// Figure out if the body can be closed - if yes
 		// we will definitely close it upon the function
 		// return.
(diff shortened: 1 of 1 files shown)


api.go
```diff
@@ -519,15 +519,21 @@ func (c Client) executeMethod(method string, metadata requestMetadata) (res *htt
...
 	// blank identifiers is only supported since go1.4
 	// https://golang.org/doc/go1.4#forrange.
-	for range c.newRetryTimer(MaxRetry, DefaultRetryUnit, DefaultRetryCap, MaxJitter, doneCh) {
+	for range c.newRetryTimer(reqRetry, DefaultRetryUnit, DefaultRetryCap, MaxJitter, doneCh) {
 		// Retry executes the following function body if request has an
 		// error until maxRetries have been exhausted, retry attempts are
 		// performed after waiting for a given period of time in a
 		// binomial fashion.
-		if isRetryable {
+		if isSeekable {
```
harshavardhana: Why change the variable name? @vadmeste `isRetryable` is fine here. isSeekable is not providing any meaning.

api.go
```diff
@@ -519,15 +519,21 @@ func (c Client) executeMethod(method string, metadata requestMetadata) (res *htt
 	// Indicate to our routine to exit cleanly upon return.
 	defer close(doneCh)
 
+	// Retry only when reader is seekable
+	reqRetry := MaxRetry
+	if !isSeekable {
+		reqRetry = 1
```
harshavardhana: We should set this value perhaps in above `metadata.contentBody` fold itself.  Avoids having multiple ifs.

api.go
vadmeste: > Why change the variable name? @vadmeste isRetryable is fine here. isSeekable is not providing any meaning. Its also not associated with what this PR tries to fix.

Yeah, it is a simple change, I thought isSeekable is more precise than isRetryable, will revert

Issue #823, opened 2017-09-25
Support Client and Server side encryption


aead:
Currently minio-go supports only client-side encryption (and only a "insecure" AWS compatible version). The client and server-side encryption should be unified. Further the minio-go API should be provided functions like GetEncryptedObject and PutEncryptedObject (and the file-based equivalents).

aead:
#830 is marked as high priority but recent changes of SSE-C require some conceptual changes to the PR @harshavardhana proposed to split PR to get the 4.0.0 release out.
I agree that we may can define the API in advance (to avoid breaking changes) and later add additional implementations. (Of course we would not release just the API but maybe defer different implementations (AES-CBC vs. DARE))

One important API change I would like to see in the 4.0 release is:
`GetEncryptedObject(bucket, object, password string) (Object, error)` (also for PUTs)
This would make encryption **much** easier for users - they even don't need to import the `pkg/encrypt` package anymore.
Further `Get/PutObjectOptions` need two different fields (one for SSE and one for CSE) because clients can use SSE and CSE at the same time.
There is bunch of other stuff to do - so is 4.0 release that important to break issue/PR into parts @deekoder ?

harshavardhana:
> There is bunch of other stuff to do - so is 4.0 release that important to break issue/PR into parts @deekoder ?

4.0 release here is blocking mint release @aead  - that is one of the reasons why i asked you to split PRs.

aead:
I see - Okay I try to get this PR (re-viewable / ready-to-merge) this week. Will split it into API and functionality...

harshavardhana:
We are going to make 4.0.0 anyways, we can make 5.0.0 if we break the API.

aead:
Close - See #883

PR #824, opened 2017-09-25
refactor GET object operations

aead:
This change refactors get-object calls to take GetOptions.
This is required to use client and server side encryption within
minio-go.

Updates #823

### DIFF
--- a/api-get-object-context.go
+++ b/api-get-object-context.go
@@ -19,6 +19,7 @@ package minio
 import "context"
 
 // GetObjectWithContext - returns an seekable, readable object.
-func (c Client) GetObjectWithContext(ctx context.Context, bucketName, objectName string) (*Object, error) {
-	return c.getObjectWithContext(ctx, bucketName, objectName)
+// The options can be used to specify the GET request further.
+func (c Client) GetObjectWithContext(ctx context.Context, bucketName, objectName string, opts GetObjectOptions) (*Object, error) {
+	return c.getObjectWithContext(ctx, bucketName, objectName, opts)
 }
--- a/api-get-object-file.go
+++ b/api-get-object-file.go
@@ -21,23 +21,34 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/minio/minio-go/pkg/encrypt"
+
 	"context"
 
 	"github.com/minio/minio-go/pkg/s3utils"
 )
 
 // FGetObjectWithContext - download contents of an object to a local file.
-func (c Client) FGetObjectWithContext(ctx context.Context, bucketName, objectName, filePath string) error {
-	return c.fGetObjectWithContext(ctx, bucketName, objectName, filePath)
+// The options can be used to specify the GET request further.
+func (c Client) FGetObjectWithContext(ctx context.Context, bucketName, objectName, filePath string, opts GetObjectOptions) error {
+	return c.fGetObjectWithContext(ctx, bucketName, objectName, filePath, opts)
 }
 
 // FGetObject - download contents of an object to a local file.
-func (c Client) FGetObject(bucketName, objectName, filePath string) error {
-	return c.fGetObjectWithContext(context.Background(), bucketName, objectName, filePath)
+func (c Client) FGetObject(bucketName, objectName, filePath string, opts GetObjectOptions) error {
+	return c.fGetObjectWithContext(context.Background(), bucketName, objectName, filePath, opts)
+}
+
+// FGetEncryptedObject - Decrypt and store an object at filePath.
+func (c Client) FGetEncryptedObject(bucketName, objectName, filePath string, materials encrypt.Materials) error {
+	if materials == nil {
+		return ErrInvalidArgument("Unable to recognize empty encryption properties")
+	}
+	return c.FGetObject(bucketName, objectName, filePath, GetObjectOptions{Materials: materials})
 }
 
 // fGetObjectWithContext - fgetObject wrapper function with context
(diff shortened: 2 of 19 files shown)


poornas:
travis fail - your examples need to be updated.

aead:
Oh, yes, thanks @poornas :+1:

api-stat.go
```diff
@@ -81,20 +81,19 @@ func extractObjMetadata(header http.Header) http.Header {
 }
 
 // StatObject verifies if object exists and you have permission to access.
-func (c Client) StatObject(bucketName, objectName string) (ObjectInfo, error) {
+func (c Client) StatObject(bucketName, objectName string, opts GetOptions) (ObjectInfo, error) {
```
poornas: Api.md needs to document this.

api-get-object-context.go
```diff
@@ -19,6 +19,6 @@ package minio
 import "context"
 
 // GetObjectWithContext - returns an seekable, readable object.
-func (c Client) GetObjectWithContext(ctx context.Context, bucketName, objectName string) (*Object, error) {
-	return c.getObjectWithContext(ctx, bucketName, objectName)
+func (c Client) GetObjectWithContext(ctx context.Context, bucketName, objectName string, opts GetOptions) (*Object, error) {
+	return c.getObjectWithContext(ctx, bucketName, objectName, opts)
```
poornas: needs documentation update for GetObjectWithContext and FGetObjectWithContext

api-get-options.go
```diff
@@ -19,81 +19,98 @@ package minio
...
 )
 
-// RequestHeaders - implement methods for setting special
-// request headers for GET, HEAD object operations.
-// http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectGET.html
-type RequestHeaders struct {
-	http.Header
+// GetOptions are used to specify additional headers or options
+// during GET requests.
+type GetOptions struct {
```
poornas: Can we rename this GetObjectOptions to be symmetric to the PutObjectOptions

api-get-object-file.go
```diff
@@ -72,7 +72,7 @@ func (c Client) fGetObjectWithContext(ctx context.Context, bucketName, objectNam
 	}
```
poornas: needs a FGetEncryptedObject variant to match the FPutEncryptedObject.

api-get-options.go
aead: My plan was to rename `PutObjectOptions` -> `PutOptions` during encryption refactoring.
I can ask Harsha

harshavardhana: Actually it is more consistent to keep `GetObjectOptions()` instead of `GetOptions()` GetOptions is just too generic.

aead: Okay

docs/API.md
```diff
@@ -831,7 +870,7 @@ if err != nil {
 }
 ```
 <a name="StatObject"></a>
-### StatObject(bucketName, objectName string) (ObjectInfo, error)
+### StatObject(bucketName, objectName string, opts GetObjectOptions) (ObjectInfo, error)
```
poornas: It looks odd to pass GetObjectOptions to a HEAD call.

aead: The problem is a SSE-C call requires encryption keys - Otherwise there would be no way to get stats of an SSE-C object...

examples/s3/getobject-context.go
```diff
@@ -47,7 +47,7 @@ func main() {
 	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Minute)
 	defer cancel()
 
-	reader, err := s3Client.GetObjectWithContext(ctx, "my-bucketname", "my-objectname")
+	reader, err := s3Client.GetObjectWithContext(ctx, "my-bucketname", "my-objectname", minio.GetObjectOptions{})
```
poornas: would be good to have the example demonstrate use of the options

docs/API.md
harshavardhana: Perhaps `StatObject(bucketName, objectName string, opts StatObjectOptions)`  --> `GetObjectOptions --> StatObjectOptions`

harshavardhana:
Looks like there is a formatting issue @aead can you fix?

poornas:
@aead, travis fails for the example. A couple of minor additions required - FGetEncryptedObject test needed for functional tests. Also believe readme.md is missing link to the example. Are we going to have a StatObjectOptions?

harshavardhana:
> @aead, travis fails for the example. A couple of minor additions required - FGetEncryptedObject test needed for functional tests. Also believe readme.md is missing link to the example. Are we going to have a StatObjectOptions?

We can have a simple 

```
// GetObjectOptions are used to specify additional headers or options
// during GET requests.
	headers map[string]string

	UserMetadata map[string]string
	Materials    encrypt.Materials
}

type StatObjectOptions GetObjectOptions
```

aead:
@harshavardhana 
Won't work nicely. The problem is that we would use Go's type system in a wrong way.
E.g. All functions of `GetObjectOptions` are not accessible (strong type system).
A solution would be:
```
type StatObjectOptions struct {
     GetObjectOptions
}
``` 
But this will lead to weird constructions like `StatObjectOptions{GetObjectObptions{}}`

harshavardhana:
> A solution would be:
> 
> type StatObjectOptions struct {
>      GetObjectOptions
> }
> But this will lead to weird constructions like StatObjectOptions{GetObjectObptions{}}

This would be do-able 

```go
type StatObjectOptions struct {
     GetObjectOptions
}
```

`StatObjectOptions{}` would initialize properly and inherit all the methods.

PR #827, opened 2017-09-26
Fix preSign signature calculation to honor resource lists.

### LINKED ISSUE #825: pkg/s3signer: PreSignV2 is encoding all query params erroneously
The subresources that must be included when constructing the CanonicalizedResource Element are acl, lifecycle, location, logging, notification, partNumber, policy, requestPayment, torrent, uploadId, uploads, versionId, versioning, versions, and website.

http://docs.aws.amazon.com/AmazonS3/latest/dev/RESTAuthentication.html#RESTAuthenticationQueryStringAuth

harshavardhana:
Currenlty we used to sign all the query parameters, this
is not in accordance with AWS S3 spec. We should only
canonicalize query params which are present in resouce list.

Fixes #825

### DIFF
--- a/pkg/s3signer/request-signature-v2.go
+++ b/pkg/s3signer/request-signature-v2.go
@@ -76,7 +76,7 @@ func PreSignV2(req http.Request, accessKeyID, secretAccessKey string, expires in
 	}
 
 	// Get presigned string to sign.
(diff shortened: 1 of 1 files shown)


harshavardhana:
@vadmeste @krisis waiting on your reviews.

PR #829, opened 2017-09-27
Support redirect headers to sign again with new Host header.

harshavardhana:
This is a requirement if say an S3 API endpoint redirects to
a new location based on perhaps a redirect proxy between
two servers.

This is implemented to support bucket re-direction which
might be implemented in future by employing a redirect
proxy.

To this change please use redirector here -
 https://gist.github.com/harshavardhana/a135fafff0d97b31b25ee303ebf28d46

### DIFF
--- a/api.go
+++ b/api.go
@@ -48,7 +48,7 @@ type Client struct {
 	///  Standard options.
 
 	// Parsed endpoint url provided by the user.
-	endpointURL url.URL
+	endpointURL *url.URL
 
 	// Holds various credential providers.
 	credsProvider *credentials.Credentials
@@ -130,11 +130,11 @@ func New(endpoint, accessKeyID, secretAccessKey string, secure bool) (*Client, e
 		return nil, err
 	}
 	// Google cloud storage should be set to signature V2, force it if not.
-	if s3utils.IsGoogleEndpoint(clnt.endpointURL) {
+	if s3utils.IsGoogleEndpoint(*clnt.endpointURL) {
 		clnt.overrideSignerType = credentials.SignatureV2
 	}
 	// If Amazon S3 set to signature v4.
-	if s3utils.IsAmazonEndpoint(clnt.endpointURL) {
+	if s3utils.IsAmazonEndpoint(*clnt.endpointURL) {
 		clnt.overrideSignerType = credentials.SignatureV4
 	}
 	return clnt, nil
@@ -177,6 +177,68 @@ func (r *lockedRandSource) Seed(seed int64) {
 	r.lk.Unlock()
 }
 
+// Redirect requests by re signing the request.
+func (c *Client) redirectHeaders(req *http.Request, via []*http.Request) error {
+	if len(via) >= 5 {
+		return errors.New("stopped after 5 redirects")
+	}
+	if len(via) == 0 {
+		return nil
+	}
+	lastRequest := via[len(via)-1]
+	var reAuth bool
+	for attr, val := range lastRequest.Header {
+		// if hosts do not match do not copy Authorization header
+		if attr == "Authorization" && req.Host != lastRequest.Host {
+			reAuth = true
+			continue
+		}
+		if _, ok := req.Header[attr]; !ok {
+			req.Header[attr] = val
+		}
+	}
+
+	*c.endpointURL = *req.URL
+
+	value, err := c.credsProvider.Get()
+	if err != nil {
+		return err
+	}
+	var (
+		signerType      = value.SignerType
+		accessKeyID     = value.AccessKeyID
+		secretAccessKey = value.SecretAccessKey
+		sessionToken    = value.SessionToken
+		region          = c.region
+	)
+
+	// Custom signer set then override the behavior.
+	if c.overrideSignerType != credentials.SignatureDefault {
+		signerType = c.overrideSignerType
+	}
+
+	// If signerType returned by credentials helper is anonymous,
+	// then do not sign regardless of signerType override.
+	if value.SignerType == credentials.SignatureAnonymous {
+		signerType = credentials.SignatureAnonymous
+	}
+
+	if reAuth {
+		// Check if there is no region override, if not get it from the URL if possible.
+		if region == "" {
+			region = s3utils.GetRegionFromURL(*c.endpointURL)
+		}
+		switch {
+		case signerType.IsV2():
+			// Add signature version '2' authorization header.
+			req = s3signer.SignV2(*req, accessKeyID, secretAccessKey)
+		case signerType.IsV4():
+			req = s3signer.SignV4(*req, accessKeyID, secretAccessKey, sessionToken, getDefaultLocation(*c.endpointURL, region))
+		}
+	}
+	return nil
+}
+
 func privateNew(endpoint string, creds *credentials.Credentials, secure bool, region string) (*Client, error) {
 	// construct endpoint.
 	endpointURL, err := getEndpointURL(endpoint, secure)
@@ -194,16 +256,17 @@ func privateNew(endpoint string, creds *credentials.Credentials, secure bool, re
 	clnt.secure = secure
 
 	// Save endpoint URL, user agent for future uses.
-	clnt.endpointURL = *endpointURL
+	clnt.endpointURL = endpointURL
 
 	// Instantiate http client and bucket location cache.
 	clnt.httpClient = &http.Client{
-		Transport: defaultMinioTransport,
+		Transport:     defaultMinioTransport,
+		CheckRedirect: clnt.redirectHeaders,
 	}
 
 	// Sets custom region, if region is empty bucket location cache is used automatically.
 	if region == "" {
-		region = s3utils.GetRegionFromURL(clnt.endpointURL)
+		region = s3utils.GetRegionFromURL(*clnt.endpointURL)
 	}
 	clnt.region = region
 
@@ -276,7 +339,7 @@ func (c *Client) TraceOff() {
 // please vist -
 // http://docs.aws.amazon.com/AmazonS3/latest/dev/transfer-acceleration.html
 func (c *Client) SetS3TransferAccelerate(accelerateEndpoint string) {
-	if s3utils.IsAmazonEndpoint(c.endpointURL) {
+	if s3utils.IsAmazonEndpoint(*c.endpointURL) {
 		c.s3AccelerateEndpoint = accelerateEndpoint
 	}
 }
@@ -380,6 +443,7 @@ func (c Client) dumpHTTP(req *http.Request, resp *http.Response) error {
 			}
 		}
 	}
+
 	// Write response to trace output.
 	_, err = fmt.Fprint(c.traceOutput, strings.TrimSuffix(string(respTrace), "\r\n"))
 	if err != nil {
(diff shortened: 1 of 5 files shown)


harshavardhana:
This is a future work, we can review and take this in with leisure.

vadmeste:
Isn't this not secure? we sign in the first place to prevent requests modification by a man in the middle

harshavardhana:
> Isn't this not secure? we sign in the first place to prevent requests modification by a man in the middle

No it is not redirect server will need to do a bit more validation @vadmeste current example is plain and simple.  

But minio-go should also honor re-directs such that we can re-authorize the request.  It really doesn't matter though if the server redirects to another server which doesn't have the credentials the request will fail.  In essence the server and client needs to know the right credentials, it doesn't matter who re-directs to what server.

nitisht:
Can this be reviewed again, @vadmeste @harshavardhana

harshavardhana:
Not yet @nitisht . Will get confirmation if we need to add this feature in minio-go.

api.go
```diff
@@ -177,6 +177,47 @@ func (r *lockedRandSource) Seed(seed int64) {
 	r.lk.Unlock()
 }
 
+// Redirect requests by re signing the request.
+func (c *Client) redirectHeaders(req *http.Request, via []*http.Request) error {
```
fdawg4l: Should you get rid of the redirect handling on line 488 of this file?  `mc` already checks for go `1.9.1`, so I think it's likely safe to remove.

harshavardhana: Yeah but minio-go is used by apps which still run on go1.7.x etc.

harshavardhana: Anyways removed it, most apps would be using go1.8 or 1.9

fdawg4l: Cool!  Thanks!

api.go
```diff
@@ -177,6 +177,66 @@ func (r *lockedRandSource) Seed(seed int64) {
...
+		// Check if there is no region override, if not get it from the URL if possible.
+		if region == "" {
+			region = getRegionFromURL(c.endpointURL)
+		}
+		switch {
+		case signerType.IsV2():
+			// Add signature version '2' authorization header.
+			req = s3signer.SignV2(*req, accessKeyID, secretAccessKey)
+		case signerType.IsV4():
+			req = s3signer.SignV4(*req, accessKeyID, secretAccessKey, sessionToken, getDefaultLocation(c.endpointURL, region))
```
fdawg4l: This is using `c.endpointURL`.  Curious as to why not use the redirected host?

harshavardhana: c.endpointURL is replaced with the new host.

harshavardhana: So all future operations happen directly on new host rather than doing a continuous redirection for all requests.

fdawg4l: Ah, this make sense.  Thanks!  But this leads me to another question-

Line 51, `endpointURL` isn't a pointer.  And `executeMethod` on 525 doesn't have a pointer receiver.  So, for every request, `executeMethod` will get the redirect, the go client will retry and follow the `Location`, we end up here and resign the request after setting the new `endpoint`.  

Basically changes to `c.endpointURL` won't persist between requests, right?  The redirect will happen at least once per request OR the client will need to be updated.

harshavardhana: Not really. *Client is a pointer so endpointURL gets updated anyways, you can test this out yourself.

harshavardhana: That is one of the reasons why redirectHeaders() function has a pointer receiver.

fdawg4l: :+1:   yup, you're right.  Thanks for indulging me.  This patch works for my use-case.  Tested it with a redirector without an issue.

fdawg4l: For my own edification, I did some probing.

For every request, there is 1 redirection.  But subsequent calls also recieve a redirection because `endpoint` is not updated.  And this is because because `do` does not have a pointer receiver.  Infact, every call in the stack before `do` doesn't either.  Notice the lack of `(*Client)` in calls before the frame with `redirectHeaders` in the stack below.

```
goroutine 1 [running]:
github.com/minio/mc/vendor/github.com/minio/minio-go.(*Client).redirectHeaders(0xc4201f8a20, 0xc420202300, 0xc42000e228, 0x1, 0x1, 0xc42008ff10, 0xc4201e5590)
	/food/s4/src/github.com/minio/mc/vendor/github.com/minio/minio-go/api.go:182 +0x39
github.com/minio/mc/vendor/github.com/minio/minio-go.(*Client).(github.com/minio/mc/vendor/github.com/minio/minio-go.redirectHeaders)-fm(0xc420202300, 0xc42000e228, 0x1, 0x1, 0xc4201e6a50, 0x25)
	/food/s4/src/github.com/minio/mc/vendor/github.com/minio/minio-go/api.go:269 +0x52
net/http.(*Client).checkRedirect(0xc4201e5170, 0xc420202300, 0xc42000e228, 0x1, 0x1, 0x0, 0x101)
	/usr/local/go/src/net/http/client.go:411 +0x5a
net/http.(*Client).Do(0xc4201e5170, 0xc420202200, 0xc, 0x0, 0x0)
	/usr/local/go/src/net/http/client.go:568 +0x918
github.com/minio/mc/vendor/github.com/minio/minio-go.Client.do(0xc4201fa6c0, 0x4, 0x0, 0x0, 0x0, 0xc4201fa6c7, 0xf, 0x0, 0x0, 0x0, ...)
	/food/s4/src/github.com/minio/mc/vendor/github.com/minio/minio-go/api.go:474 +0x50
github.com/minio/mc/vendor/github.com/minio/minio-go.Client.getBucketLocation(0xc4201fa6c0, 0x4, 0x0, 0x0, 0x0, 0xc4201fa6c7, 0xf, 0x0, 0x0, 0x0, ...)
	/food/s4/src/github.com/minio/mc/vendor/github.com/minio/minio-go/bucket-cache.go:104 +0x1a4
<snip>
```

And my trivial redirector confirms this.  I issued an `mc ls redirectedHost/foo`
```
redirecting from /foo/?location=
redirecting from /foo/
```
If the endpoint in the client was updated correctly, we should have seen only 1 redirect.

Simplified example
https://play.golang.org/p/TTKjoFO_Wz

Anyway, again this patch does work so the above is a minor gripe.  But if you reuse the client, you'll see 1 redirect for each request.  Maybe that's desireable.

harshavardhana: Yes let me verify it again. And fix it properly making pointer receiver is not a big deal

harshavardhana:
This PR is ready for review @vadmeste @kannappanr

vadmeste:
I am just little worried about minio-go getting blamed when a man in the middle instructs minio-go to follow redirection though the user didn't have a proxy in the first place. Can you at least enable this redirection only with HTTPS ?

harshavardhana:
> I am just little worried about minio-go getting blamed when a man in the middle instructs minio-go to follow redirection though the user didn't have a proxy in the first place. Can you at least enable this redirection only with HTTPS ?

what is the benefit of only doing with HTTPS ?

harshavardhana:
@vadmeste @kannappanr i need to move further with this PR, let me know how i can help.

api.go
```diff
@@ -398,38 +462,22 @@ func (c Client) dumpHTTP(req *http.Request, resp *http.Response) error {
...
-			}
-			req.URL = newURL
-		} else {
-			break
 		}
+		return nil, err
 	}
 
-	// Response cannot be non-nil, report if its the case.
+	// Response cannot be non-nil, report error if thats the case.
```
kannappanr: Typo, should say Response cannot be nil

api.go
```diff
@@ -177,6 +177,68 @@ func (r *lockedRandSource) Seed(seed int64) {
...
+		return errors.New("stopped after 5 redirects")
+	}
+	if len(via) == 0 {
+		return nil
+	}
+	lastRequest := via[len(via)-1]
+	var reAuth bool
+	for attr, val := range lastRequest.Header {
+		// if hosts do not match do not copy Authorization header
+		if attr == "Authorization" && req.Host != lastRequest.Host {
```
vadmeste: I guess not only Host can change, even Path can do. A proxy can decide to tweak a little the url, like /user1/bucket/object to /bucket/object.

What about re-authenticating in all cases without condition, since this is a redirection.

harshavardhana: That is not expected to be handled here. Redirect is only to ensure that we are reAuth the host, changing paths is not supported. The server side redirection is supposed to redirect to a different hostname only.

vadmeste: > The server side redirection is supposed to redirect to a different hostname only.

Are you saying this is an S3 requirement ?

harshavardhana: The specific requirement here is a customer, who wishes the client to be namespace aware. Namespace is cheaply provided by servers redirecting to servers where the incoming request belongs.  There is no mangling of path expected. Such scenario we should fail and it fails properly.

I could additionally add such a check along the lines of Host , but that is not expected as a requirement.

vadmeste:
@harshavardhana can you rebase this to last master to be able to check without function tests ?

vadmeste:
I manually rebased in my machine, it seems there is a bug somewhere in function tests.. it only tests few test cases:
```
18:47 $ go run functional_tests.go
{"args":{"bucketName":"minio-go-test-has5x6fg0mzxgplh"},"duration":104,"function":"MakeBucket(bucketName, region), BucketExists(bucketName), GetBucketPolicy(bucketName, objectPrefix), SetBucketPolicy(bucketName, objectPrefix, bucketPolicy), GetBucketPolicy(bucketName, objectPrefix), SetBucketPolicy(bucketName, objectPrefix, bucketPolicy), GetBucketPolicy(bucketName, objectPrefix), SetBucketPolicy(bucketName, objectPrefix, bucketPolicy), GetBucketPolicy(bucketName, objectPrefix), ListBuckets(), PutObject(bucketName, objectName, reader, contentType), ListObjects(bucketName, objectName, isRecursive, doneCh), ListObjectsV2(bucketName, objectName, isRecursive, doneCh), ListIncompleteUploads(bucketName, objectName, isRecursive, doneCh), GetObject(bucketName, objectName), FGetObject(bucketName, objectName, fileName), PresignedHeadObject(bucketName, objectName, expires, reqParams), PresignedHeadObject(bucketName, objectName, expires, reqParams), PresignedGetObject(bucketName, objectName, expires, reqParams), PresignedGetObject(bucketName, objectName, expires, reqParams), PresignedPutObject(bucketName, objectName, expires), PresignedPutObject(bucketName, objectName, expires), RemoveObject(bucketName, objectName), RemoveBucket(bucketName)","name":"minio-go: testFunctional","status":"PASS"}
{"args":{"bucketName":"minio-go-test-pjmxamfbqvt3s5pc","objectName":"minio-go-test-pjmxamfbqvt3s5pcunique-presigned"},"duration":73,"function":"MakeBucket(bucketName, location), BucketExists(bucketName), SetBucketPolicy(bucketName, objectPrefix, bucketPolicy), ListBuckets(), ListObjects(bucketName, objectName, isRecursive, doneCh), ListIncompleteUploads(bucketName, objectName, isRecursive, doneCh), GetObject(bucketName, objectName), FGetObject(bucketName, objectName, fileName), PresignedHeadObject(bucketName, objectName, expires, reqParams), PresignedGetObject(bucketName, objectName, expires, reqParams), PresignedPutObject(bucketName, objectName, expires), GetObject(bucketName, objectName)","name":"minio-go: testFunctionalV2","status":"PASS"}
```

harshavardhana:
You need run with MINT_MODE=full

vadmeste:
> You need run with MINT_MODE=full

Oh right, now it is passing..

PR #830, opened 2017-09-28
refactor client/server side encryption

aead:
This change unifies with client and server side encryption and
makes enrcyption easier to use.


examples/s3/put-encrypted-object.go
```diff
@@ -68,16 +68,10 @@ func main() {
...
 
-	// Build encryption materials which will encrypt uploaded data
-	cbcMaterials, err := encrypt.NewCBCSecureMaterials(symmetricKey)
-	if err != nil {
-		log.Fatalln(err)
-	}
-
 	// Encrypt file content and upload to the server
-	n, err := s3Client.PutEncryptedObject("my-bucketname", "my-objectname", file, cbcMaterials)
+	n, err := s3Client.PutEncryptedObject("aead", "my-objectname", file, symmetricKey)
```
harshavardhana: Keep the bucket name more generic like `my-bucketname`

examples/s3/put-encrypted-object.go
```diff
@@ -68,16 +68,10 @@ func main() {
...
 	if err != nil {
-		log.Fatalln(err)
+		panic(err)
```
harshavardhana: use `log.Fatalln` not `panic`

pkg/encrypt/encrypt.go
```diff
@@ -0,0 +1,76 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Package encrypt implements a generic interface to encrypt any stream of data.
+// currently this package implements two types of encryption
+// - Symmetric encryption using AES.
+// - Asymmetric encrytion using RSA.
```
harshavardhana: is this still the case?

pkg/encrypt/encrypt.go
```diff
@@ -0,0 +1,76 @@
...
+const (
+	// AesCbcPkcs5 specifies the client-side-encryption algorithm AES-CBC with
+	// PKCS5 padding. This algorithm is implemented for AWS compability but is
+	// not recommended because of security issues.
+	AesCbcPkcs5 = "AES/CBC/PKCS5"
+
+	// DareHmacSha256 specifies the client-side-encryption algorithm DARE with
+	// a HMAC-SHA256 KDF scheme. This algorithm provides tamper-proof encryption
+	// and is recommended over any current AWS S3 client-side-encryption algorithm.
+	DareHmacSha256 = "DARE-HAMC-SHA256"
```
harshavardhana: s/HAMC/HMAC

pkg/encrypt/encrypt.go
```diff
@@ -0,0 +1,76 @@
...
+)
+
+const (
+	cseIV        = "X-Amz-Meta-X-Amz-Iv"
```
harshavardhana: Links to these headers with relevant docs?

pkg/encrypt/encrypt.go
```diff
@@ -0,0 +1,76 @@
...
+	cseKey       = "X-Amz-Meta-X-Amz-Key-v2"
+	cseAlgorithm = "X-Amz-Meta-X-Amz-Cek-Alg"
+)
+
+// Cipher is a generic interface for en/decrypting streams using
+// S3 client/server side encryption.
+type Cipher interface {
+	Seal(header map[string]string, src io.Reader) (io.ReadCloser, error)
```
harshavardhana: Each interface function requires a comment, also requires that we explain if its callers responsibility to call Close(), and what type of errors are allowed to be generated in what conditions.

api-get-object-file.go
```diff
@@ -40,11 +38,12 @@ func (c Client) FGetObject(bucketName, objectName, filePath string, opts GetObje
 }
 
 // FGetEncryptedObject - Decrypt and store an object at filePath.
-func (c Client) FGetEncryptedObject(bucketName, objectName, filePath string, materials encrypt.Materials) error {
-	if materials == nil {
-		return ErrInvalidArgument("Unable to recognize empty encryption properties")
+func (c Client) FGetEncryptedObject(bucketName, objectName, filePath string, key encrypt.Key) error {
+	cipher, err := encrypt.NewCipher(encrypt.DareHmacSha256, key)
```
harshavardhana: Why change the word Cipher? , we kept materials to be inline with what AWS S3 calls this.  Advantages of calling it materials is that users of aws-sdks have a fair understanding of what is providing by our library.  Cipher is more technically correct in the encryption context.. But since its S3 API we can stick to what AWS SDKs already choose.  For example we should look at aws-sdk-java implementation as the standard for such naming conventions.

examples/s3/put-encrypted-object.go
aead: Thanks, debug artifact...

examples/s3/put-encrypted-object.go
aead: Thanks, debug artifact...

pkg/encrypt/encrypt.go
aead: Currently yes, but I plan to purge it...

pkg/encrypt/encrypt.go
aead: done

pkg/encrypt/encrypt.go
aead: Done

pkg/encrypt/encrypt.go
aead: Agree. Will update.

api-get-object-file.go
aead: In general correct. The cipher is actually low-level API (for e.g. GetEncryptedObject we just require a key). So I agree to not bother users with low-level details but I also think we should do it correctly at the low-level. If somebody will use the low-level API he/she (hopefully) knows what he/she is doing. So I think it's okay to use "correct" terms.

AWS "Materials" is a java thing and don't say anything about what is required or what it does...  I don't think it's a good thing to carry bloated java idioms to Go.

harshavardhana: Okay that is a valid argument. I thought this is a standard representation if it is only aws-sdk-java then we can stick to simpler approach.

aead: Also AWS Go sdk has it but you can see there that it's "literally" copied from aws-java...
I think it's okay to do it "correctly" underneath and provide a "nice" high-level API (just the key). From my perspective the AWS materials approach is also not nice from an API point...

harshavardhana: In that sense we can reference that these are materials from aws-sdk-go . And we choose it to be better as Cipher.

aead: Okay

pkg/encrypt/encrypt.go
```diff
@@ -0,0 +1,96 @@
...
+// of EncryptionMaterials of the aws-go-sdk.
+type Cipher interface {
+
+	// Seal returns an io.ReadCloser encrypting everything it reads from
+	// the provided io.Reader. It adds HTTP headers to the provided header
+	// if neccessary. Seal returns an error if it is not able to encrypt
+	// the io.Reader
+	Seal(header map[string]string, src io.Reader) (io.ReadCloser, error)
+
+	// Seal returns an io.ReadCloser decrypting everything it reads from
```
harshavardhana: This should be 's/Seal/Open'

aead: Thanks c/p err :-/

pkg/encrypt/encrypt.go
```diff
@@ -0,0 +1,96 @@
...
+	// the io.Reader
+	Seal(header map[string]string, src io.Reader) (io.ReadCloser, error)
+
+	// Open returns an io.ReadCloser decrypting everything it reads from
+	// the provided io.Reader. It reads HTTP headers from the provided header
+	// if neccessary. Open returns an error if it is not able to decrypt
+	// the io.Reader
+	Open(header map[string]string, src io.Reader) (io.ReadCloser, error)
+
+	// Overhead retruns the size of an encrypted stream with the provided
```
vadmeste: s/retruns/returns/

pkg/encrypt/dare.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+	header[cseIV] = base64.StdEncoding.EncodeToString(iv)
+	header[cseAlgorithm] = DareHmacSha256
+
+	if closer, ok := r.(io.Closer); ok {
+		type readCloser struct {
+			io.Reader
+			io.Closer
+		}
+		return readCloser{reader, closer}, nil
```
vadmeste: :+1:

pkg/encrypt/dare.go
```diff
@@ -0,0 +1,87 @@
...
+	"crypto/rand"
+	"crypto/sha256"
+	"encoding/base64"
+	"errors"
+	"io"
+	"io/ioutil"
+
+	"github.com/minio/sio"
+)
+
```
vadmeste: Can you add a comment in this file saying that we are satisfying Cipher interface ?

pkg/encrypt/dare.go
```diff
@@ -0,0 +1,87 @@
...
+			io.Closer
+		}
+		return readCloser{reader, closer}, nil
+	}
+	return ioutil.NopCloser(reader), nil
+}
+
+func (d dareHmacSha256) Open(header map[string]string, r io.Reader) (io.ReadCloser, error) {
+	if header[cseAlgorithm] != DareHmacSha256 {
+		return nil, errors.New("invalid encryption algorithm")
```
vadmeste: s/invalid/unexpected

pkg/encrypt/dare.go
```diff
@@ -0,0 +1,87 @@
...
+			io.Reader
+			io.Closer
+		}
+		return readCloser{reader, closer}, nil
+	}
+	return ioutil.NopCloser(reader), nil
+}
+
+func (d dareHmacSha256) Overhead(size int64) int64 {
+	encSize := (size / (64 * 1024)) * (64*1024 + 32)
```
vadmeste: Can you add a brief text about how you calculate the overhead of this algorithm ?

harshavardhana: It is in the dare spec on sio project. @vadmeste

functional_tests.go
```diff
@@ -2328,52 +2327,7 @@ func testEncryptionPutGet() {
...
-	if err != nil {
-		failureLog(function, args, startTime, "", "DecodeString for symmetric Key generation failed", err).Fatal()
-	}
-
-	// Generate an asymmetric key
-	asymKey, err := encrypt.NewAsymmetricKey(privateKey, publicKey)
-	if err != nil {
-		failureLog(function, args, startTime, "", "NewAsymmetricKey for symmetric Key generation failed", err).Fatal()
-	}
+	symKey := encrypt.NewSymmetricKey(make([]byte, 32))
```
vadmeste: Can you also add a test for the password+salt key type ?

pkg/encrypt/dare.go
```diff
@@ -0,0 +1,87 @@
...
+import (
+	"crypto/hmac"
+	"crypto/rand"
+	"crypto/sha256"
+	"encoding/base64"
+	"errors"
+	"io"
+	"io/ioutil"
+
+	"github.com/minio/sio"
```
vadmeste: This is the first dependency from outside minio/minio-go.. I am not sure, shouldn't we vendorize this ?

harshavardhana: Unless we are sticking to a version we should vendorize. If we say sio will be always backward compatible we don't have to.

aead: Not really. E.g `github.com/dustin/go-humanize`
With sio the situation is better IMO because it's our repository...

functional_tests.go
aead: Will do

pkg/encrypt/dare.go
aead: I add a link to the spec

pkg/encrypt/dare.go
aead: I'm not sure if I understand your comment correctly. Are you requesting docs for the `dareHmacSha256` type?
I think something like
```
// dareHmacSha256 implements encrypt.Cipher with the DARE format
// and HMAC-SHA256 as encryption key derivation function.
type dareHmacSha256 [32]byte
```

aead: is reasonable?!

pkg/encrypt/dare.go
vadmeste: > Not really. E.g github.com/dustin/go-humanize
> With sio the situation is better IMO because it's our repository...

Well I found go-humanize only in functional_tests.go

Anyway, I don't think this is urgent and needs to be addressed right now, so we can let it until we need to open this subject again.

pkg/encrypt/dare.go
vadmeste: Actually I was talking about the whole file and what it is  for.. because if we don't have a priori knowledge about the design, we can get lost when we read this file, it doesn't clearly say this file satisfy Cipher interface.

pkg/encrypt/dare.go
aead: Oh, your right - I just looked into the appveyor.yml...
Agree we can revisit later...

pkg/encrypt/dare.go
aead: Hm - I'm not sure about this... This style (special implementations in separate files) is kind of well known in the Go world - I would expect separated implementations in more or less this way (standard library)...
On the other side a comment is no big deal. I'm just not sure whether we should start something like this...

vadmeste: > This style (special implementations in separate files) is kind of well known in the Go world

yes.. but if you don't know that this file implements Cipher, how much time it takes for you to know that ? e.g. is there an emacs shortcut that can help to do that ?

aead: That's the advantage of Go interfaces - that types can implement interfaces without knowing that the interface exists. The downside is that this relationship is not explicit anymore.
If you want to make it explicit (and enforced by the compiler) we can add:
`var _ Cipher = (*dareHmacSha256)(nil)`
That would be more explicit that a comment.

vadmeste: Sorry @aead, it looks like you already did this (not sure when)

```
// dareHmacSha256 implements encrypt.Cipher with the DARE format
// and HMAC-SHA256 as encryption key derivation function.
```

For me, this is sufficient.

Generally, in a package with few files, it is not problem, but when there are many files in the package, it will be hard to know that a type satisfies an interface when you read the code the first time. That's my point.

So the current state already looks good.

aead: `HEAD~1` 😉 
Completely agree - can be hard in large packages.

functional_tests.go
```diff
@@ -2438,7 +2441,7 @@ func testEncryptionFPut() {
 	}
 
 	// Generate a symmetric key
-	symKey := encrypt.NewSymmetricKey(make([]byte, 32))
+	symKey := encrypt.DeriveKey("my-passwod", []byte("salt-123"))
```
harshavardhana: `my-password` ?

pkg/encrypt/encrypt.go
```diff
@@ -0,0 +1,96 @@
...
+// AWS client-side-encryption headers.
+// See: https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingClientSideEncryption.html
+const (
+	cseIV        = "X-Amz-Meta-X-Amz-Iv"
+	cseKey       = "X-Amz-Meta-X-Amz-Key-v2"
+	cseAlgorithm = "X-Amz-Meta-X-Amz-Cek-Alg"
+)
+
+// Cipher is a generic interface for en/decrypting streams using
+// S3 client/server side encryption. Cipher is the functional equalient
```
poornas: s/equalient/equivalent/

harshavardhana:
The scope of the change in the PR like `encrypt.Cipher` has changed, since we are also thinking about KMS like support along with this.  SO we need to keep it more generic than just Cipher.

harshavardhana:
This also means that when we make the change such as API suggestions like `GetEncryptedObject(bucket, object, password)` will not work @aead 

Since password may be a KMS id, too it is also possible that we might not even have the password on client side at all just a bucket policy which encrypts everything. 

So the scope here has changed and i can see encrypt.Materials is a correct naming convention, this is what AWS S3 did.

aead:
>The scope of the change in the PR like encrypt.Cipher has changed, since we are also thinking about KMS like support along with this. SO we need to keep it more generic than just Cipher.

Yes and no. A CSE object can be decrypted with KMS if:
 - The CSE algorithm supports key wrapping.
 - The operator/admin added the client-side encryption key to the KMS service.
But agree that we can think about that...

>This also means that when we make the change such as API suggestions like GetEncryptedObject(bucket, object, password) will not work @aead

It will be correct because my current proposal is to use SSE-C as the encryption method for
Get/PutEncryptedObject. SSE-C cannot be converted to KMS so this would be fine. 

>Since password may be a KMS id, too it is also possible that we might not even have the password on client side at all just a bucket policy which encrypts everything.

There is a problem: AWS knows 3 types of encryption:
 - client side
 - server side (SSE-C)
 - server side (KMS)

A client side encrypted object can be transformed into a KMS encrypted object - a SSE-C encrypted object can't. So no matter how we encrypt the object with `PutEncryptedObject` - one decryption method is not able to decrypt the object. If the client uses CSE for an SSE-C object it will fail and also the other way around...
My proposal is to use SSE-C for Put/GetEncryptedObject and provide a low-level (`encrypt` package) API for CSE and KMS. The advantage is that the Put/GetEncryptedObject can be kept very simple.

harshavardhana:
While I do like the idea. There are way too many variations here and it will confuse people more doc burden. 

We should have one API for this IMO and leave it at that. So we should take a structured input .  Otherwise we will make other things like KMS or CSE second grade APIs and more supported style is SSE-C .

aead:
>While I do like the idea. There are way too many variations here and it will confuse people more doc burden.

That I agree. Currently you can do either an `PutEncryptedObject` call or an `PutObject` call and specify the cipher / encryption material as part of the `PutObjectOptions - so there are already 2 different ways for encryption.
From my understanding:
 - Any encryption should be possible by a regular `PutObject` - It is specified as part of the `PutObjectOptions`. That would be the "advanced" / low-level API.
- One (preferred) encryption can be used by a "nicer" API function which makes encryption much smoother for the clients.

>We should have one API for this IMO and leave it at that. So we should take a structured input . Otherwise we will make other things like KMS or CSE second grade APIs and more supported style is SSE-C .

I agree. The problem is that if we provide one struct for all encryption styles as argument to `PutEncryptedObject`:
 - There is no advantage for an user to use 
  `PutEncryptedObject("bucket", "name", reader, size, material)` instead of 
  `PutEncryptedObject("bucket", "name", reader, size, PutObjectOptions{Material: material})`
  The encryption API would be kind of pointless
- There are some cases which are quite complicated to express in one e.g. `Material` struct.
  For example an object can be encrypted with SSE-C **and** CSE at the same time. 
  (Further I've no clue what happens if you try to convert such an object to an KMS encr. object...)
  With `PutObjectOptions` you can express that by:
  ```
   type PutObjectOptions  struct {
       Material                        encrypt.Material
       ServerSideEncryption  encrypt.ServerSide
      ....
   }
  ```
  But if I try to combine SSE-C and CSE/KMS the API gets quite ugly... 

From my point of view the `PutEncryptedObject` functions should provide a easier API while the generic `PutObject` should provide all options. However we may should check with ab about that?!

aead:
Close - See #883

PR #832, opened 2017-09-29
Do not need to strictly validate request params in Presign API.

harshavardhana:
This is done so that we can enable the existing API to work for
all API operations under S3.

### DIFF
--- a/api-presigned.go
+++ b/api-presigned.go
@@ -18,22 +18,13 @@ package minio
 
 import (
 	"errors"
+	"net/http"
 	"net/url"
 	"time"
 
 	"github.com/minio/minio-go/pkg/s3signer"
 	"github.com/minio/minio-go/pkg/s3utils"
 )
-
-// supportedGetReqParams - supported request parameters for GET presigned request.
-var supportedGetReqParams = map[string]struct{}{
-	"response-expires":             {},
-	"response-content-type":        {},
-	"response-cache-control":       {},
-	"response-content-language":    {},
-	"response-content-encoding":    {},
-	"response-content-disposition": {},
-}
 
 // presignURL - Returns a presigned URL for an input 'method'.
 // Expires maximum is 7days - ie. 604800 and minimum is 1.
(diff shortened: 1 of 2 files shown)


api-presigned.go
```diff
@@ -45,33 +35,18 @@ func (c Client) presignURL(method string, bucketName string, objectName string,
 	if err := s3utils.CheckValidBucketName(bucketName); err != nil {
 		return nil, err
 	}
-	if err := s3utils.CheckValidObjectName(objectName); err != nil {
-		return nil, err
-	}
```
poornas: why should object name validation be relaxed?

harshavardhana: To make it generic, need to move this check to presignedGetObject()

harshavardhana:
This is ready for review again @poornas @krisis

api-presigned.go
```diff
@@ -18,66 +18,42 @@ package minio
...
 func (c Client) presignURL(method string, bucketName string, objectName string, expires time.Duration, reqParams url.Values) (u *url.URL, err error) {
 	// Input validation.
 	if method == "" {
 		return nil, ErrInvalidArgument("method cannot be empty.")
 	}
-	if err := s3utils.CheckValidBucketName(bucketName); err != nil {
-		return nil, err
-	}
-	if err := s3utils.CheckValidObjectName(objectName); err != nil {
+	if err = s3utils.CheckValidBucketName(bucketName); err != nil {
```
krisis: Which http methods' presignURL doesn't require a valid object name?

harshavardhana: listObjects, headBucket, getBucketPolicy,getBucketLocation,.. there are more APIs.

krisis: Ok, makes sense.

krisis:
@harshavardhana, is it possible to add a unit test case at least cover negative test cases like invalid bucket and object names?

harshavardhana:
> @harshavardhana, is it possible to add a unit test case at least cover negative test cases like invalid bucket and object names?

Yes will do @krisis good idea.

PR #835, opened 2017-09-30
Add cleanup function to remove all the objects from a bucket after tests

nitisht:
Fixes : https://github.com/minio/mint/issues/165
and improves logging by adding args in few test cases.

### DIFF
--- a/functional_tests.go
+++ b/functional_tests.go
@@ -121,6 +121,43 @@ func ignoredLog(function string, args map[string]interface{}, startTime time.Tim
 	return log.WithFields(fields)
 }
 
+// Delete objects in given bucket, recursively
+func cleanupBucket(bucketName string, c *minio.Client) error {
+	// Create a done channel to control 'ListObjectsV2' go routine.
+	doneCh := make(chan struct{})
+	// Exit cleanly upon return.
+	defer close(doneCh)
+	// Iterate over all objects in the bucket via listObjectsV2 and delete
+	for objCh := range c.ListObjectsV2(bucketName, "", true, doneCh) {
+		if objCh.Err != nil {
+			return objCh.Err
+		}
+		if objCh.Key != "" {
+			err := c.RemoveObject(bucketName, objCh.Key)
+			if err != nil {
+				return err
+			}
+		}
+	}
+	for objPartInfo := range c.ListIncompleteUploads(bucketName, "", true, doneCh) {
+		if objPartInfo.Err != nil {
+			return objPartInfo.Err
+		}
+		if objPartInfo.Key != "" {
+			err := c.RemoveIncompleteUpload(bucketName, objPartInfo.Key)
+			if err != nil {
+				return err
+			}
+		}
+	}
+	// objects are already deleted, clear the buckets now
+	err := c.RemoveBucket(bucketName)
+	if err != nil {
+		return err
+	}
+	return err
+}
+
 func init() {
 	// If server endpoint is not set, all tests default to
 	// using https://play.minio.io:9000
@@ -255,10 +292,10 @@ func testMakeBucketError() {
 		minio.ToErrorResponse(err).Code != "BucketAlreadyOwnedByYou" {
 		failureLog(function, args, startTime, "", "Invalid error returned by server", err).Fatal()
 	}
-	if err = c.RemoveBucket(bucketName); err != nil {
-		failureLog(function, args, startTime, "", "Remove bucket failed", err).Fatal()
+	// Delete all objects and buckets
+	if err = cleanupBucket(bucketName, c); err != nil {
+		failureLog(function, args, startTime, "", "Cleanup failed", err).Fatal()
 	}
-
 	successLogger(function, args, startTime).Info()
 }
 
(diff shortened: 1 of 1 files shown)


functional_tests.go
```diff
@@ -379,9 +411,9 @@ func testMakeBucketRegions() {
 		failureLog(function, args, startTime, "", "MakeBucket failed", err).Fatal()
 	}
 
-	// Remove the newly created bucket.
-	if err = c.RemoveBucket(bucketName + ".withperiod"); err != nil {
-		failureLog(function, args, startTime, "", "Remove bucket failed", err).Fatal()
+	// cleanup
```
harshavardhana: Add more descriptive comment rather than place holder like `cleanup`

functional_tests.go
```diff
@@ -1613,7 +1608,7 @@ func testPutObjectWithContext() {
 	args := map[string]interface{}{
 		"bucketName": "",
 		"objectName": "",
-		"opts":       "minio.PutObjectOptions{ContentType:objectContentType}",
```
harshavardhana: why this change? shouldn't we add this?

nitisht: was a typo, fixed.

functional_tests.go
nitisht: Agreed. Updated the comments

functional_tests.go
```diff
@@ -121,6 +121,31 @@ func ignoredLog(function string, args map[string]interface{}, startTime time.Tim
...
+// Delete objects and buckets recursively
+func cleanupBuckets(c *minio.Client) (err error) {
+
+	doneCh := make(chan struct{})
+	defer close(doneCh)
+
+	buckets, err := c.ListBuckets()
+
+	for _, bucket := range buckets {
+		for obj := range c.ListObjects(bucket.Name, "", true, doneCh) {
```
harshavardhana: you can put each ListObjects() on a go-routine since they are independent buckets and clean them up.

functional_tests.go
```diff
@@ -121,6 +121,37 @@ func ignoredLog(function string, args map[string]interface{}, startTime time.Tim
...
+					}
+				} else {
+					err = object.Err
+				}
+			}
+		}(bucket.Name)
+	}
+	return err
+}
+
```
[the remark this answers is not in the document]
nitisht: Yes it will delete everything on the server. Let me add the bucket prefix check

functional_tests.go
```diff
@@ -121,6 +121,37 @@ func ignoredLog(function string, args map[string]interface{}, startTime time.Tim
...
+	var err error
+	var buckets []minio.BucketInfo
+
+	// Create a done channel to control 'ListObjectsV2' go routine.
+	doneCh := make(chan struct{})
+
+	// Indicate to our routine to exit cleanly upon return.
+	defer close(doneCh)
+
+	buckets, err = c.ListBuckets()
```
harshavardhana: Return on error right here..

functional_tests.go
```diff
@@ -121,6 +121,37 @@ func ignoredLog(function string, args map[string]interface{}, startTime time.Tim
...
+	for _, bucket := range buckets {
+		go func(bucketName string) {
+			objectCh := c.ListObjectsV2(bucketName, "", true, doneCh)
+			for object := range objectCh {
+				if object.Err != nil {
+					if object.Key != "" {
+						err = c.RemoveObject(bucketName, object.Key)
+					}
+				} else {
+					err = object.Err
```
harshavardhana: Do not set the err from outside scope in a go-routine,  collect all errors.. in a errs []error and then store them based on the buckets index. like this 

```go
for i, bucket := range buckets {
     go func(i int, bucketName string) {
           ....
           errs[i] = err
     }(i, bucket)
}
```

harshavardhana: Also since there is no wait group, this won't delete all buckets all the time. Since the program might exit if this function returns after all go-routines have spawned. 

Use wait group to limit yourself and use `wg.Wait()` to wait for cleaning up of all buckets. Once done loop through 

```go
for _, err := range errs {
        if err != nil {
             return err 
        }
}
```

nitisht:
Updated the PR with a simpler approach to cleanup, since previous approach failed if there are multiple instances of `minio-go` tests running in parallel against a single server, `cleanup` deleted buckets belonging to other test instances as well - for example when travis runs multiple `minio-go` tests against play server. 

Please take a look @harshavardhana @poornas

functional_tests.go
```diff
@@ -121,6 +121,33 @@ func ignoredLog(function string, args map[string]interface{}, startTime time.Tim
...
+			}
+		}
+	}
+	// objects are already deleted, clear the buckets now
+	err := c.RemoveBucket(bucketName)
+	if err != nil {
+		return err
+	}
+	return err
+}
```
poornas: what if there are incomplete upload parts? that would need cleanup as well.

nitisht: can you explain a bit more @poornas

poornas: if there are incomplete uploads to a bucket, RemoveBucket would fail. For tests using large uploads, shouldn't the cleanup also remove incomplete uploads if any?

nitisht: done @poornas

harshavardhana: RemoveBucket works fine on all gateways even with incomplete uploads, that is the S3 behavior. But it is a good change since it will cater for B2 gateway backend as well. Which requires explicit removal of incomplete uploads.

harshavardhana:
> Partially fixes : minio/mint#165
> and improves logging by adding args in few test cases.

Is this true that it partially fixes? @nitisht

nitisht:
> Is this true that it partially fixes?

Let me change it to fixes, initially there were stale buckets left by minio-js as well

```
[2017-09-29 17:11:02 PDT] 100KiB 41b17354-ff91-478d-9a46-353830d48767/datafile-100-kB
[2017-09-29 17:11:02 PDT] 100KiB 41b17354-ff91-478d-9a46-353830d48767/datafile-100-kB.buffer
```

but that is not seen anymore.

PR #836, opened 2017-09-30
Add new CopyObject API in Core client

harshavardhana:
This PR also fixes the problem of PutObject
call not handling additional headers.

Refer original problem here https://github.com/minio/minio/pull/5000

### DIFF
--- a/api-compose-object.go
+++ b/api-compose-object.go
@@ -266,6 +266,51 @@ func (s *SourceInfo) getProps(c Client) (size int64, etag string, userMeta map[s
 	return
 }
 
+// Low level implementation of CopyObject API, supports only upto 5GiB worth of copy.
+func (c Client) copyObjectDo(ctx context.Context, srcBucket, srcObject, destBucket, destObject string,
+	metadata map[string]string) (ObjectInfo, error) {
+
+	// Build headers.
+	headers := make(http.Header)
+
+	// Set all the metadata headers.
+	for k, v := range metadata {
+		headers.Set(k, v)
+	}
+
+	// Set the source header
+	headers.Set("x-amz-copy-source", s3utils.EncodePath(srcBucket+"/"+srcObject))
+
+	// Send upload-part-copy request
+	resp, err := c.executeMethod(ctx, "PUT", requestMetadata{
+		bucketName:   destBucket,
+		objectName:   destObject,
+		customHeader: headers,
+	})
+	defer closeResponse(resp)
+	if err != nil {
+		return ObjectInfo{}, err
+	}
+
+	// Check if we got an error response.
+	if resp.StatusCode != http.StatusOK {
+		return ObjectInfo{}, httpRespToErrorResponse(resp, srcBucket, srcObject)
+	}
+
+	cpObjRes := copyObjectResult{}
+	err = xmlDecoder(resp.Body, &cpObjRes)
+	if err != nil {
+		return ObjectInfo{}, err
+	}
+
+	objInfo := ObjectInfo{
+		Key:          destObject,
+		ETag:         strings.Trim(cpObjRes.ETag, "\""),
+		LastModified: cpObjRes.LastModified,
+	}
+	return objInfo, nil
+}
+
 // uploadPartCopy - helper function to create a part in a multipart
 // upload via an upload-part-copy request
 // https://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadUploadPartCopy.html
--- a/api-s3-datatypes.go
+++ b/api-s3-datatypes.go
@@ -128,7 +128,7 @@ type initiator struct {
 // copyObjectResult container for copy object response.
 type copyObjectResult struct {
 	ETag         string
-	LastModified string // time string format "2006-01-02T15:04:05.000Z"
+	LastModified time.Time // time string format "2006-01-02T15:04:05.000Z"
 }
 
 // ObjectPart container for particular part of an object.
--- a/core.go
+++ b/core.go
@@ -19,6 +19,7 @@ package minio
 import (
 	"context"
 	"io"
+	"strings"
 
 	"github.com/minio/minio-go/pkg/policy"
 )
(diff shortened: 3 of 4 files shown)


core.go
```diff
@@ -53,9 +54,49 @@ func (c Core) ListObjectsV2(bucketName, objectPrefix, continuationToken string,
...
+func (c Core) CopyObject(sourceBucket, sourceObject, destBucket, destObject string, metadata map[string]string) (ObjectInfo, error) {
+	return c.copyObjectDo(context.Background(), sourceBucket, sourceObject, destBucket, destObject, metadata)
+}
+
 // PutObject - Upload object. Uploads using single PUT call.
 func (c Core) PutObject(bucket, object string, data io.Reader, size int64, md5Sum, sha256Sum []byte, metadata map[string]string) (ObjectInfo, error) {
-	return c.putObjectDo(context.Background(), bucket, object, data, md5Sum, sha256Sum, size, PutObjectOptions{UserMetadata: metadata})
+	delKey := func(m map[string]string, key string) {
+		delete(m, strings.ToLower(key))
+		delete(m, strings.ToUpper(key))
```
poornas: will miss key in mixed-case.

core.go
```diff
@@ -53,9 +54,49 @@ func (c Core) ListObjectsV2(bucketName, objectPrefix, continuationToken string,
...
+		delKey(metadata, "content-type")
+	}
+	if v := getKeyValue(metadata, "content-disposition"); v != "" {
+		opts.ContentDisposition = v
+		delKey(metadata, "content-disposition")
+	}
+	if v := getKeyValue(metadata, "content-encoding"); v != "" {
+		opts.ContentEncoding = v
+		delKey(metadata, "content-encoding")
+	}
```
poornas:
```
opts := PutObjectOptions{}
m := make(map[string]string)
for k, v := range metadata {
	if strings.ToLower(k) == "content-encoding" {
		opts.ContentEncoding = v
	} else if strings.ToLower(k) == "content-disposition" {
		opts.ContentDisposition = v
	} else if strings.ToLower(k) == "content-type" {
		opts.ContentType = v
	} else if strings.ToLower(k) == "cache-control" {
		opts.CacheControl = v
	} else {
		m[k] = metadata[k]
	}
}
opts.UserMetadata = m
```
this would be more concise

core.go
```diff
@@ -53,9 +54,49 @@ func (c Core) ListObjectsV2(bucketName, objectPrefix, continuationToken string,
 	return c.listObjectsV2Query(bucketName, objectPrefix, continuationToken, fetchOwner, delimiter, maxkeys)
 }
 
+// CopyObject - copies an object from source object to destination object on server side.
+func (c Core) CopyObject(sourceBucket, sourceObject, destBucket, destObject string, metadata map[string]string) (ObjectInfo, error) {
+	return c.copyObjectDo(context.Background(), sourceBucket, sourceObject, destBucket, destObject, metadata)
+}
```
poornas: need tests.

core.go
harshavardhana: can you give me an example?

core.go
harshavardhana: This makes sense..

core.go
harshavardhana: sure

donatello:
Sorry to come in late on this. The bug identified https://github.com/minio/minio/pull/5000#issuecomment-333323200 can be fixed properly in ComposeObject().

Particularly, this part https://github.com/minio/minio-go/blob/master/api-compose-object.go#L110-L122 needs to be fixed to handle the standard headers (like Content-Encoding, etc). IIUC, then this PR is not needed?

harshavardhana:
> Sorry to come in late on this. The bug identified minio/minio#5000 (comment) can be fixed properly in ComposeObject().
> 
> Particularly, this part https://github.com/minio/minio-go/blob/master/api-compose-object.go#L110-L122 needs to be fixed to handle the standard headers (like Content-Encoding, etc). IIUC, then this PR is not needed (except perhaps to fix PutObject)?

No we need to add a simpler version @donatello - we shouldn't use ComposeObject IMHO. Which uses multipart API underneath.

core.go
poornas: example - if  "Content-Type" is a key in the metadata map, the above code will not delete the entry.

harshavardhana: Oops yeah i should have used http.CanonicalHeaderKey()

donatello:
> No we need to add a simpler version @donatello - we shouldn't use ComposeObject IMHO. Which uses multipart API underneath.

@harshavardhana Got it, you're right. ComposeObject would also do a HEAD, which is not desirable in gateway mode where we do expect it to be a one-to-one API mapping.

api-s3-datatypes.go
```diff
@@ -128,7 +128,7 @@ type initiator struct {
 // copyObjectResult container for copy object response.
 type copyObjectResult struct {
 	ETag         string
-	LastModified string // time string format "2006-01-02T15:04:05.000Z"
+	LastModified time.Time // time string format "2006-01-02T15:04:05.000Z"
```
donatello: How would `xmlDecoder()` know the format (time layout) with which to parse the time? I am not sure the decoding will be correct as is.

core.go
```diff
@@ -53,9 +54,30 @@ func (c Core) ListObjectsV2(bucketName, objectPrefix, continuationToken string,
...
+	for k, v := range metadata {
+		if strings.ToLower(k) == "content-encoding" {
+			opts.ContentEncoding = v
+		} else if strings.ToLower(k) == "content-disposition" {
+			opts.ContentDisposition = v
+		} else if strings.ToLower(k) == "content-type" {
+			opts.ContentType = v
+		} else if strings.ToLower(k) == "cache-control" {
+			opts.CacheControl = v
+		} else {
```
donatello: Two headers seem to be missing from this list - `content-length` and `content-md5` according to https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPUT.html

api-compose-object.go
```diff
@@ -266,6 +266,51 @@ func (s *SourceInfo) getProps(c Client) (size int64, etag string, userMeta map[s
...
+func (c Client) copyObjectDo(ctx context.Context, srcBucket, srcObject, destBucket, destObject string,
+	metadata map[string]string) (ObjectInfo, error) {
+
+	// Build headers.
+	headers := make(http.Header)
+
+	// Set all the metadata headers.
+	for k, v := range metadata {
+		headers.Set(k, v)
+	}
```
donatello: For metadata replacement to actually happen we need to make sure the `x-amz-metadata-directive` header is set to `REPLACE`. Perhaps we should throw an error if `metadata` is non-empty and this header is not set?

api-s3-datatypes.go
harshavardhana: that is the feature of xml Marshal @donatello also we need to have typed value.

core.go
harshavardhana: that is set internally @donatello

api-compose-object.go
harshavardhana: You are right let me do that.

harshavardhana: actually we don't need to.. there is no error @donatello according to CopyObject() if REPLACE is set then the set metadata is honored, otherwise it does copy from the source object by default. So we don't need to error out.

api-s3-datatypes.go
harshavardhana: For example look all the List() APIs.

donatello: @harshavardhana - xml decoder is using a particular format within the ISO8601 standard by default to parse timestamps. Looks like by dumb luck AWS is using the same format!

api-compose-object.go
donatello: Ah yes. It is not needed, as AWS behaviour is also to silently ignore the headers without the directive header.

PR #837, opened 2017-10-02
Fix handling of metadata headers in compose-object

donatello:
- Previously common headers such as `Content-Type` were not replaced
  on the copy of the object.

- This patch addresses a problem related to https://github.com/minio/minio/pull/5000

### DIFF
--- a/functional_tests.go
+++ b/functional_tests.go
@@ -4670,6 +4670,10 @@ func testUserMetadataCopyingWrapper(c *minio.Client) {
 				}
 			}
 		}
+		// Also add content-type header
+		if objInfo.ContentType != "" {
+			h.Add("content-type", objInfo.ContentType)
+		}
 		return h
 	}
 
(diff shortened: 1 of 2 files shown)


functional_tests.go
```diff
@@ -4709,6 +4713,7 @@ func testUserMetadataCopyingWrapper(c *minio.Client) {
 
 	expectedHeaders := make(http.Header)
 	expectedHeaders.Set("x-amz-meta-notmyheader", "notmyvalue")
+	expectedHeaders.Set("content-type", "application/javascript")
```
harshavardhana: While this is a good fix.. There is a lot of simplification missing @donatello . It was significantly harder to add these parts of the code. For example if you look at the getProps() filters out x-amz-meta- only and ignores all headers..

harshavardhana: We should perhaps take a stab at simplifying the metadata handling in ComposeObject and avoid it being interwoven.. perhaps you can think about this further.

donatello: Yes, it is complicated. The difficulty seems to be due to the lack of a simple way to extract  from  `ObjectInfo` the set of values that can be set via headers. I will give it some thought.

harshavardhana:
@donatello are you planning to fix this? if not should we close this PR?

donatello:
Will resubmit this after fixing with newer changes.

PR #849, opened 2017-10-18
Add PartSize to be configurable as part of PutObjectOptions

### LINKED ISSUE #848: Uploading small files of unknown size uses too much memory
## Version
minio-go 3.0.3
FreeBSD 11.1/amd64
Go 1.9

## Background
My app accepts multipart file uploads over HTTP (thus the file size can not be known) and uses this library to stream them directly into an S3-compatible store.  I run it on a small VPS with 512 MiB RAM and a tiny hard drive.  The purpose of streaming them directly into S3 is to prevent buffering large files in RAM or on disk.

## Expected Behavior
I should be able to stream small files of unknown size into S3 without using lots of memory.

## Actual Behavior
When minio-go starts uploading a file of unknown size, `optimalPartInfo(-1)` assumes the worst case of needing to fit a 5 TiB file within 10000 chunks and allocates a 576 MiB buffer.

## The Problem
Allocating this massive buffer causes my server to grind to a halt for 60 seconds while the page daemon swaps to make room for this large allocation that can't fit in RAM.  This is silly when I know I'll never handle a file larger than 10 GiB.

## Temporary Workaround
Fork and modify `constants.go`.

## Solutions
- The constants could be configurable, and belong to the Client struct.
- The new `PutObjectOptions` struct you're adding could have a field for the maximum buffer size or maximum expected file size.

harshavardhana:
This is done to ensure that caller can control the amount of
memory being used when uploading a stream without knowing
prior length.

Fixes #848

### DIFF
--- a/api-put-object-context.go
+++ b/api-put-object-context.go
@@ -24,8 +24,7 @@ import (
 // PutObjectWithContext - Identical to PutObject call, but accepts context to facilitate request cancellation.
 func (c Client) PutObjectWithContext(ctx context.Context, bucketName, objectName string, reader io.Reader, objectSize int64,
 	opts PutObjectOptions) (n int64, err error) {
-	err = opts.validate()
-	if err != nil {
+	if err = opts.validate(); err != nil {
 		return 0, err
 	}
 	if opts.EncryptMaterials != nil {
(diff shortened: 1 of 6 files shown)


nitisht:
Blocked till further discussion, as proposed by @harshavardhana

harshavardhana:
Closing this as not needed we have a different solution for this.

PR #850, opened 2017-10-19
Fix docs and add a validator tool for our api docs

harshavardhana:


docs/validator.go
```diff
@@ -0,0 +1,164 @@
...
+			t.Execute(w, p)
+			w.Sync()
+			w.Close()
+			entryN++
+
+			cmdArgs := []string{"-s", "-w", "-l", testFilePath}
+			exec.Command("gofmt", cmdArgs...).Run()
+
+			cmdArgs = []string{"-w", testFilePath}
+			exec.Command("goimports", cmdArgs...).Run()
```
nitisht: validation fails for `FGetEncryptedObject`, looks like `encrypt` is not getting imported

```sh
Failed to validate code at docs/API.md:18247

// Generate a master symmetric key
key := encrypt.NewSymmetricKey([]byte("my-secret-key-00"))

// Build the CBC encryption material
cbcMaterials, err := encrypt.NewCBCSecureMaterials(key)
if err != nil {
    fmt.Println(err)
    return
}

err = minioClient.FGetEncryptedObject("mybucket", "myobject", "/tmp/myobject", cbcMaterials)
if err != nil {
    fmt.Println(err)
    return
}

```

docs/validator.go
```diff
@@ -0,0 +1,164 @@
...
+
+			// Go build current directory
+			cmdArgs = []string{"build", "-o", testBinPath, testFilePath}
+			if cerr := exec.Command("go", cmdArgs...).Run(); cerr != nil {
+				fmt.Printf("Failed to validate code at %s:%d\n%s", docPath, p.Pos, p.Text)
```
nitisht: p.Pos is a little confusing, not clear what position it is referring to.

docs/validator.go
```diff
@@ -0,0 +1,164 @@
...
+				os.Exit(255)
+			}
+			fmt.Printf("Successfully validated %s code at %s:%d\n", p.Lang, docPath, p.Pos)
```
nitisht: p.Pos is a little confusing, not clear what position it is referring to

```sh
Successfully validated go code at docs/API.md:7695
Successfully validated go code at docs/API.md:8261
Successfully validated go code at docs/API.md:8847
Successfully validated go code at docs/API.md:9279
Successfully validated go code at docs/API.md:10467
Successfully validated go code at docs/API.md:11698
Successfully validated go code at docs/API.md:13270
Successfully validated go code at docs/API.md:14553
Successfully validated go code at docs/API.md:15424
Successfully validated go code at docs/API.md:16389
Successfully validated go code at docs/API.md:17463
```

docs/validator.go
harshavardhana: It runs fine on CI looks like you have a local issue `goimports` has a bug btw it stops scanning imports if it reaches a directory which it cannot access.

docs/validator.go
harshavardhana: A position is the position in file with number of characters. It doesn't equate to line number.

docs/validator.go
nitisht: > goimports has a bug btw it stops scanning imports if it reaches a directory which it cannot access.

This is what I faced. 

> Go to the directory and check manually what is going on.

will do

docs/validator.go
nitisht: in that case, I dont feel it adds any value, can we remove it?

docs/validator.go
```diff
@@ -0,0 +1,164 @@
...
+	"path/filepath"
+	"strings"
+	"text/template"
+
+	"github.com/a8m/mark"
+	"github.com/minio/cli"
+)
+
+var templateMain = `
+// +build ignore
```
balamurugana: This text can be in separate file like `checker.go.template`

harshavardhana: Sure

harshavardhana:
@balamurugana can you review this again?

docs/checker.go.tpl
```diff
@@ -0,0 +1,23 @@
+// +build ignore
```
balamurugana: 1. I think this line can be removed.

2. `tpl` extension is too limited.  IMO we would have fully expanded `.template`

docs/validator.go
```diff
@@ -0,0 +1,167 @@
...
```
balamurugana: I think we should build on `make` to ensure its compiling all the time.

docs/validator.go
```diff
@@ -0,0 +1,167 @@
...
+	cli.StringFlag{
+		Name:  "t",
+		Value: "checker.go.tpl",
+		Usage: "Template used for generating the programs.",
+	},
+	cli.IntFlag{
+		Name:  "skip",
+		Value: 2,
+		Usage: "Skip entries before validating the code.",
+	},
```
balamurugana: I think we would need to have a template like `mc --help` to show how this can be used with examples.

docs/validator.go
```diff
@@ -0,0 +1,167 @@
...
+			if werr != nil {
+				panic(werr)
+			}
+			t.Execute(w, p)
+			w.Sync()
+			w.Close()
+			entryN++
+
+			cmdArgs := []string{"-s", "-w", "-l", testFilePath}
+			exec.Command("gofmt", cmdArgs...).Run()
```
balamurugana: You could have a function or closure hide this logic here.  This way you could have cmdArgs globally.  If available I prefer long option than short option for readability (if available)
```go
function runGofmt(string testFilePath) {
    ...
}
```

docs/validator.go
```diff
@@ -0,0 +1,167 @@
...
+
+			cmdArgs = []string{"-w", testFilePath}
+			exec.Command("goimports", cmdArgs...).Run()
```
balamurugana: Same as above.

docs/validator.go
```diff
@@ -0,0 +1,167 @@
...
+
+			// Go build current directory
+			cmdArgs = []string{"build", "-o", testBinPath, testFilePath}
+			if cerr := exec.Command("go", cmdArgs...).Run(); cerr != nil {
```
balamurugana: same as above.

I think we don't need have args `-o testBinPath`.  This intention here is to check whether `testFilePath` compiles are not.

Also we would need to show compilation error if any.  This way we know what exact problem in the code.

docs/checker.go.tpl
harshavardhana: We cannot remove build ignore, it is needed to avoid conflicting with other `packages` in current directory .tpl is acceptable name most projects like hugo etc support this. We don't need to call it .template

docs/validator.go
harshavardhana: There is no `make` in minio-go, We ensure by running on travis.

docs/validator.go
harshavardhana: -o builds a binary with right name. Compilation error is displayed with the cerr.

docs/validator.go
harshavardhana: Will add that.

docs/validator.go
harshavardhana: Added a make in minio-go

harshavardhana:
Can you review this again Bala?

harshavardhana:
PR is updated @balamurugana please check.

Makefile
```diff
@@ -0,0 +1,17 @@
+all: checks
```
balamurugana: :+1:

PR #851, opened 2017-10-19
Add doc for GetObjectOptions and PutObjectOptions

krisis:


docs/API.md
```diff
@@ -560,7 +566,19 @@ __Parameters__
...
-|`opts` | _minio.PutObjectOptions_  |Pointer to struct that allows user to set optional custom metadata, content-type, content-encoding,content-disposition and cache-control headers, pass encryption module for encrypting objects, and optionally configure number of threads for multipart put operation. |
+|`opts` | _minio.PutObjectOptions_  | Allows user to set optional custom metadata, content headers, encryption keys and number of threads for multipart upload operation. |
+
+__minio.PutObjectOptions__
+
+|Field | Type | Description |
+|:--- |:--- | :--- |
+| `opts.UserMetadata` | _map[string]string_ | Map of user metadata|
+| `opts.Progress` | _io.Reader_ | Reader to fetch progress of an upload |
+| `opts.ContentType` | _string_ | Content type of object, e.g "application/text" |
```
deekoder: Is there an exhaustive list of values that can be shown application/text, application/zip ....

docs/API.md
```diff
@@ -524,7 +530,7 @@ __Parameters__
 |`bucketName`  | _string_  |Name of the bucket |
 |`objectName` | _string_  |Name of the object  |
 |`filePath` | _string_  |Path to download object to |
-|`materials` | _encrypt.Materials_ | The module to decrypt the object data |
+|`materials` | _encrypt.Materials_ | Interface provided by `encrypt` package to encrypt a stream of data (For more information see https://godoc.org/github.com/minio/minio-go) |
```
deekoder: Need to update examples to show how to use with the encrypt option.

docs/API.md
deekoder: can we show all exhaustive values not just eg. What all content types can the api accept application/zip, application/octet. application/zip etc.

harshavardhana: We can add a link here https://en.wikipedia.org/wiki/Media_type we don't need to write exhaustive list here.

harshavardhana: Same as above.

docs/API.md
harshavardhana: AFAICS these are under FPutEncryptedObject()  and FGetEncryptedObject.

docs/API.md
harshavardhana: Basically meaning there is no restriction this ContentType field is HTTP Content-Type header with any value, we can mention that as well.

PR #853, opened 2017-10-23
Update functional tests to handle NotImplemented error

nitisht:
For API requests not supported by Minio Gateway, it responds with
NotImplemented error. This response needs to be interpreted in
functional tests as Not applicable (NA) instead of failure.

Fixes https://github.com/minio/mint/issues/189,
https://github.com/minio/mint/issues/194,
https://github.com/minio/mint/issues/197


functional_tests.go
```diff
@@ -96,6 +96,19 @@ func successLogger(function string, args map[string]interface{}, startTime time.
 	return log.WithFields(fields)
 }
 
+// As few of the features are not available in Gateway(s) currently, Check if err value is NotImplemented,
+// and log as NA in that case and continue execution. Otherwise log as failure and return
+func validateErrorLog(function string, args map[string]interface{}, startTime time.Time, alert string, message string, err error) {
+	// If server returns notimplmented we assume it is gateway mode and hence log it as info and move on to next tests
```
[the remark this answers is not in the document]
poornas: the function could be named as logErrors/logFailures - there is no real validation going on here except to skip logging if it is a NotImplemented error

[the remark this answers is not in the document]
nitisht: done

nitisht:
@poornas @balamurugana can you please review

functional_tests.go
```diff
@@ -96,6 +96,19 @@ func successLogger(function string, args map[string]interface{}, startTime time.
 	return log.WithFields(fields)
 }
 
+// As few of the features are not available in Gateway(s) currently, Check if err value is NotImplemented,
+// and log as NA in that case and continue execution. Otherwise log as failure and return
+func logError(function string, args map[string]interface{}, startTime time.Time, alert string, message string, err error) {
+	// If server returns NotImplemented we assume it is gateway mode and hence log it as info and move on to next tests
+	// Special case for ComposeObject API as it is implemented on client side and adds specific error details like `Error in upload-part-copy` in
+	// addition to NotImplemented error returned from server
+	if isErrNotImplemented(err) || err.Error() == "Error in upload-part-copy - A header you provided implies functionality that is not implemented" {
```
balamurugana: Unable to get why ignoring based on error string

nitisht: `Compose-Object` is not server side, so it prepends the error string from server with `Error in xyz`, (xyz being the exact server api that failed)

balamurugana: On what case we get an error from `ComposeObject()`?

nitisht:
Updated the PR with change in `err` returned by `ComposeObject` API. Now, it returns the server error transparently.

functional_tests.go
balamurugana: You would need to fix by removing err.Error()

nitisht: done

nitisht:
@balamurugana this is updated

PR #856, opened 2017-10-26
Honor file pointer set in PutObject method

### LINKED ISSUE #855: PutObject does not honor file pointer
If the file pointer is moved to the middle of the file and then PutObject is called with the actual size of the file, it is expected to fail. Here it succeeds, because internally, minio-go pushes it back to the start and then tries to do a putObject.
Here is a sample to code to reproduce the issue.
```
package main

import (
	"io"
	"log"
	"os"

	"github.com/minio/minio-go"
)

func main() {
	s3Client, err := minio.New("s3.amazonaws.com", "Access_Key", "Secret_Key", true)
	s3Client.TraceOn(os.Stdout)
	if err != nil {
		log.Fatalln(err)
	}
	filename := "testfile"
	var object *os.File
	object, err = os.Open(filename)
	if err != nil {
		log.Fatalln(err)
	}
	defer object.Close()

	objectStat, err := object.Stat()
	if err != nil {
		log.Fatalln(err)
	}	

	offset := objectStat.Size() / 2
	if _, err := object.Seek(int64(offset), 0); err != nil {
		log.Fatalln(err)
	}

	n, err := s3Client.PutObject("bucket", filename, object, int64(objectStat.Size()), minio.PutObjectOptions{ContentType: "application/octet-stream"})
	if err != nil {
		log.Fatalln(err)
	}
	log.Println("Uploaded", filename, " of size: ", n, "Successfully.")
}
```

Minio-go is expected to fail in this scenario.

kannappanr:
Create a sectionreader by honoring the file pointer already set on the file.

Fixes #855

### DIFF
--- a/api-put-object-streaming.go
+++ b/api-put-object-streaming.go
@@ -343,7 +343,12 @@ func (c Client) putObjectNoChecksum(ctx context.Context, bucketName, objectName 
 	}
 	if size > 0 {
 		if isReadAt(reader) && !isObject(reader) {
-			reader = io.NewSectionReader(reader.(io.ReaderAt), 0, size)
+			seeker, _ := reader.(io.Seeker)
+			offset, err := seeker.Seek(0, io.SeekCurrent)
+			if err != nil {
+				return 0, ErrInvalidArgument(err.Error())
+			}
+			reader = io.NewSectionReader(reader.(io.ReaderAt), offset, size)
 		}
 	}
 
--- a/functional_tests.go
+++ b/functional_tests.go
@@ -3477,7 +3477,7 @@ func testPutObjectUploadSeekedObject() {
 		failureLog(function, args, startTime, "", "TempFile seek failed", err).Fatal()
 	}
 
-	n, err := c.PutObject(bucketName, objectName, tempfile, int64(length), minio.PutObjectOptions{ContentType: "binary/octet-stream"})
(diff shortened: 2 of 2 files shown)


deekoder:
blocking release.

poornas:
needs rebase

api-put-object-streaming.go
```diff
@@ -348,7 +349,12 @@ func (c Client) putObjectNoChecksum(ctx context.Context, bucketName, objectName
 	}
 	if size > 0 {
 		if isReadAt(reader) && !isObject(reader) {
-			reader = io.NewSectionReader(reader.(io.ReaderAt), 0, size)
+			seeker, _ := reader.(io.Seeker)
+			offset, err := seeker.Seek(0, io.SeekCurrent)
+			if err != nil {
+				return 0, errors.New("error in seeker")
```
poornas: why not return err as is? The message "error in seeker" does not give actionable message to user of the api.

harshavardhana: :+1:  - we shouldn't translate errors in any-form to Go native errors, we should in-fact do `ErrInvalidArgument(err.Error())`  such that we have consistent errors being sent.

kannappanr:
Updated the PR based on review comments. Please review the PR again.

PR #859, opened 2017-10-27
Avoid setting server encryption headers with x-amz-meta

harshavardhana:


utils.go
```diff
@@ -255,3 +256,24 @@ func isCSEHeader(headerKey string) bool {
...
+var sseHeaders = []string{
+	"x-amz-server-side-encryption",
+	"x-amz-server-side-encryption-aws-kms-key-id",
+	"x-amz-server-side-encryption-context",
+	"x-amz-server-side-encryption-customer-algorithm",
+	"x-amz-server-side-encryption-customer-key",
+	"x-amz-server-side-encryption-customer-key-MD5",
+}
+
+// isCSEHeader returns true if header is a client side encryption header.
```
kannappanr: Looks like a Typo. Needs to be

isSSEHeader returns true if header is a server side encryption header.

utils.go
```diff
@@ -255,3 +256,24 @@ func isCSEHeader(headerKey string) bool {
 	}
 	return false
 }
+
+// sseHeaders is list of client side encryption headers
```
kannappanr: Comment needs to be 

sseHeaders is list of server side encryption headers

PR #861, opened 2017-11-01
Allow listen bucket notification proactively return errors

harshavardhana:
ListenBucketNotification should error out on all
conditions for invalid arguments. Network error
should be retried.


harshavardhana:
Can i get some reviews ? @krishnasrinivas @kannappanr

PR #866, opened 2017-11-08
Add status code to ErrorResponse object

jawnsy:
This change exposes the underlying HTTP status code, so that similar
classes of errors (e.g. 409 Conflict) can be handled easily.

Other minor changes:

* Error messages have been updated to consistently use directional quotes (‘ and ’)
* Used %d format parameter for int64 values, as Sprintf can handle stringifying these valuese correctly
* Added status codes for errors returned from functions like ErrInvalidBucketName
* Fixed typo: EntityTooSmall was returning code EntityTooLarge
* Changed error messages in tests to use %#v instead of %+v, to show the full struct state instead of just String() value

### DIFF
--- a/api-error-response.go
+++ b/api-error-response.go
@@ -20,7 +20,6 @@ import (
 	"encoding/xml"
 	"fmt"
 	"net/http"
-	"strconv"
 )
 
 /* **** SAMPLE ERROR RESPONSE ****
@@ -48,6 +47,9 @@ type ErrorResponse struct {
 	// Region where the bucket is located. This header is returned
 	// only in HEAD bucket and ListObjects response.
 	Region string
+
+	// Underlying HTTP status code for the returned error
+	StatusCode int `xml:"-" json:"-"`
 
 	// Headers of the returned S3 XML error
 	Headers http.Header `xml:"-" json:"-"`
@@ -100,7 +102,10 @@ func httpRespToErrorResponse(resp *http.Response, bucketName, objectName string)
 		msg := "Response is empty. " + reportIssue
 		return ErrInvalidArgument(msg)
 	}
-	var errResp ErrorResponse
+
+	errResp := ErrorResponse{
+		StatusCode: resp.StatusCode,
+	}
 
 	err := xmlDecoder(resp.Body, &errResp)
 	// Xml decoding failed with no body, fall back to HTTP headers.
@@ -109,12 +114,14 @@ func httpRespToErrorResponse(resp *http.Response, bucketName, objectName string)
 		case http.StatusNotFound:
 			if objectName == "" {
 				errResp = ErrorResponse{
+					StatusCode: resp.StatusCode,
 					Code:       "NoSuchBucket",
 					Message:    "The specified bucket does not exist.",
 					BucketName: bucketName,
 				}
 			} else {
 				errResp = ErrorResponse{
+					StatusCode: resp.StatusCode,
 					Code:       "NoSuchKey",
 					Message:    "The specified key does not exist.",
 					BucketName: bucketName,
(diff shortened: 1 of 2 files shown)


api-error-response.go
```diff
@@ -100,7 +100,10 @@ func httpRespToErrorResponse(resp *http.Response, bucketName, objectName string)
 		msg := "Response is empty. " + reportIssue
 		return ErrInvalidArgument(msg)
 	}
-	var errResp ErrorResponse
+
+	errResp := ErrorResponse{
+		StatusCode: resp.StatusCode,
+	}
```
nitisht: IMO `resp.StatusCode` should be added to the `errResp` var being assigned below as well

api-error-response_test.go
```diff
@@ -182,29 +184,31 @@ func TestHttpRespToErrorResponse(t *testing.T) {
...
+		StatusCode: http.StatusBadRequest,
 		Code:       "EntityTooLarge",
 		Message:    msg,
 		BucketName: "minio-bucket",
 		Key:        "Asia/",
 	}
 	actualResult := ErrEntityTooLarge(1000000, 99999, "minio-bucket", "Asia/")
 	if !reflect.DeepEqual(expectedResult, actualResult) {
-		t.Errorf("Expected result to be '%+v', but instead got '%+v'", expectedResult, actualResult)
+		t.Errorf("Expected result to be '%#v', but instead got '%#v'", expectedResult, actualResult)
```
nitisht: `%#v` instead of `%+v` add lot of information and feels verbose IMO. 

Here is a sample run with `%+v` 

```
$: Expected result to be 'Your proposed upload size ‘1000000’ exceeds the maximum allowed object size ‘99999’ for single PUT operation.', but instead got 'Your proposed upload size ‘1000000’ exceeds the maximum allowed object size ‘199999’ for single PUT operation.'
```

While with `%#v`

```
$: Expected result to be 'minio.ErrorResponse{XMLName:xml.Name{Space:"", Local:""}, StatusCode:400, Code:"EntityTooLarge", Message:"Your proposed upload size ‘1000000’ exceeds the maximum allowed object size ‘99999’ for single PUT operation.", BucketName:"minio-bucket", Key:"Asia/", RequestID:"", HostID:"", Region:"", Headers:http.Header(nil)}', but instead got 'minio.ErrorResponse{XMLName:xml.Name{Space:"", Local:""}, StatusCode:400, Code:"EntityTooLarge", Message:"Your proposed upload size ‘1000000’ exceeds the maximum allowed object size ‘199999’ for single PUT operation.", BucketName:"minio-bucket", Key:"Asia/", RequestID:"", HostID:"", Region:"", Headers:http.Header(nil)}'
```
Is there any specific benefit of doing this?

harshavardhana: @nitisht since the reflect.DeepEqual might not match for lot of fields other than the message itself, having #v is okay in this scenario. Since as a mismatch should be more noticeable.

nitisht: :+1:

jawnsy: Yeah, the reason I added it was that it made it a little harder to debug an error, since %+v only returns the String() form, which doesn't include the status code or other fields that might be mismatched. In my own projects I use `gomega` which does this automatically.

api-error-response.go
```diff
@@ -38,6 +37,7 @@ import (
 // ErrorResponse - Is the typed error returned by all API operations.
 type ErrorResponse struct {
 	XMLName    xml.Name `xml:"Error" json:"-"`
+	StatusCode int
```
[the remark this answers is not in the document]
jawnsy: Done! Good catch :smile:

PR #868, opened 2017-11-09
Re-sign signature and follow redirects

### LINKED ISSUE #867: Redirects are being followed with the wrong signature
`minio-go` will follow redirects, but the signature is invalid since the canonical string includes the original `Host`.  This will return mismatched signatures error by the server, failing the request.

PR to follow.

fdawg4l:
On redirections (503), the default http client will retry the request on
the new Location.  However, the signature will be invalid since it
includes the original Host in the canonical string.

Retry the request after resigning with the new host.  This change
disables the default client retry mechanism and retries the request in
the executeMethod retry loop.  Unfortunately this incurs a retry delay
penalty.

Fixes #867

### DIFF
--- a/api-presigned.go
+++ b/api-presigned.go
@@ -53,7 +53,7 @@ func (c Client) presignURL(method string, bucketName string, objectName string, 
 	// Instantiate a new request.
 	// Since expires is set newRequest will presign the request.
 	var req *http.Request
-	if req, err = c.newRequest(method, reqMetadata); err != nil {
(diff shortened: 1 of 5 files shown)


harshavardhana:
@fdawg4l this is not a correct fix a correct fix is posted here https://github.com/minio/minio-go/pull/829

harshavardhana:
I would suggest you close this PR and test #829 on your end.

fdawg4l:
@harshavardhana you're very right.  I like that patch much better than this.  Thanks!

PR #876, opened 2017-11-16
add support of x-amz-acl header for v2 acl (ceph server, aws s3 v2)

slimus:
We use CEPH server for store object. CEPH has old protocol (v2) and use special header for set ACl for object. Also AWS S3 v2 use this header
1. http://docs.ceph.com/docs/jewel/radosgw/s3/objectops/#put-object
2. http://docs.aws.amazon.com/AmazonS3/latest/dev/RESTAuthentication.html#RESTAuthenticationExamples-6

I added support of this header and add test for this method

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -283,9 +283,5 @@ func isSSEHeader(headerKey string) bool {
 func isAmzHeader(headerKey string) bool {
 	key := strings.ToLower(headerKey)
 
-	if strings.HasPrefix(key, "x-amz-meta-") || key == "x-amz-acl" {
-		return true
-	}
-
-	return false
+	return strings.HasPrefix(key, "x-amz-meta-") || key == "x-amz-acl"
 }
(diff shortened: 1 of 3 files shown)


utils.go
```diff
@@ -278,3 +278,14 @@ func isSSEHeader(headerKey string) bool {
 	}
 	return false
 }
+
+// isAmzHeader returns true if header is a x-amz-meta-* or x-amz-acl header.
+func isAmzHeader(headerKey string) bool {
+	key := strings.ToLower(headerKey)
+
+	if strings.HasPrefix(key, "x-amz-meta-") || key == "x-amz-acl" {
```
harshavardhana: this can be done in one line here 

```
return strings.HasPrefix(key, "x-amz-meta-") || key == "x-amz-acl" 
```

slimus: yep. fixed

PR #877, opened 2017-11-16
pkg/credentials: Credentials retrieval should prioritize credentials

harshavardhana:
Always prefer proper credentials over anonymous ones, if no
credentials are found default to anonymous value.

### DIFF
--- a/pkg/credentials/chain.go
+++ b/pkg/credentials/chain.go
@@ -17,8 +17,6 @@
 
 package credentials
 
-import "fmt"
-
 // A Chain will search for a provider which returns credentials
 // and cache that provider until Retrieve is called again.
 //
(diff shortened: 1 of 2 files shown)


pkg/credentials/chain.go
```diff
@@ -58,28 +56,30 @@ func NewChainCredentials(providers []Provider) *Credentials {
 	})
 }
 
-// Retrieve returns the credentials value or error if no provider returned
-// without error.
+// Retrieve returns the credentials value, or no anonymous credentials if
```
donatello: This comment is quite confusing - s/no anonymous credentials/the anonymous credential value/ perhaps.

harshavardhana: fixed

harshavardhana:
Can you review again @nitisht @donatello ?

PR #878, opened 2017-11-16
Add CopyObjectPart support for Core client

donatello:
This is towards resolving https://github.com/minio/minio/issues/4885

### DIFF
--- a/core.go
+++ b/core.go
@@ -60,6 +60,15 @@ func (c Core) CopyObject(sourceBucket, sourceObject, destBucket, destObject stri
 	return c.copyObjectDo(context.Background(), sourceBucket, sourceObject, destBucket, destObject, metadata)
 }
 
+// CopyObjectPart - creates a part in a multipart upload by copying (a
+// part of) an existing object.
+func (c Core) CopyObjectPart(srcBucket, srcObject, destBucket, destObject string, uploadID string,
+	partID int, startOffset, length int64, metadata map[string]string) (p CompletePart, err error) {
+
+	return c.copyObjectPartDo(context.Background(), srcBucket, srcObject, destBucket, destObject, uploadID,
+		partID, startOffset, length, metadata)
+}
+
 // PutObject - Upload object. Uploads using single PUT call.
 func (c Core) PutObject(bucket, object string, data io.Reader, size int64, md5Base64, sha256Hex string, metadata map[string]string) (ObjectInfo, error) {
 	opts := PutObjectOptions{}
--- a/api-compose-object.go
+++ b/api-compose-object.go
@@ -312,6 +312,56 @@ func (c Client) copyObjectDo(ctx context.Context, srcBucket, srcObject, destBuck
 	return objInfo, nil
 }
 
+func (c Client) copyObjectPartDo(ctx context.Context, srcBucket, srcObject, destBucket, destObject string, uploadID string,
+	partID int, startOffset int64, length int64, metadata map[string]string) (p CompletePart, err error) {
+
+	headers := make(http.Header)
+
+	// Set source
+	headers.Set("x-amz-copy-source", s3utils.EncodePath(srcBucket+"/"+srcObject))
+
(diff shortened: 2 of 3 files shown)


core.go
```diff
@@ -60,6 +60,15 @@ func (c Core) CopyObject(sourceBucket, sourceObject, destBucket, destObject stri
 	return c.copyObjectDo(context.Background(), sourceBucket, sourceObject, destBucket, destObject, metadata)
 }
 
+// CopyObjectPart - creates a part in a multipart upload by copying (a
+// part of) an existing object.
+func (c Core) CopyObjectPart(srcBucket, srcObject, destBucket, destObject string, uploadID string,
```
harshavardhana: One problem with this API is that right now we are not ensuring how to do encryption yet. We need to fix our CopyObjectPart() API on server side first to take in additional metadata and add it here. Like how we have. 

```
// PutObjectPartWithMetadata - upload an object part with additional request metadata.
func (c Core) PutObjectPartWithMetadata(bucket, object, uploadID string, partID int, data io.Reader,
        size int64, md5Base64, sha256Hex string, metadata map[string]string) (ObjectPart, error) {
        return c.uploadPart(context.Background(), bucket, object, uploadID, data, partID, md5Base64, sha256Hex, size, metadata)
}
```

harshavardhana:
> Can we add CopyObjectOptions? Than we can extend the API. For users it should feel even more convenient:
> 
> GetObjectOptions
> PutObjectOptions
> StatObjectOptions
> CopyObjectOptions?
> EDIT: Especially because CopyObjectPart is the only way S3 provides to change object metadata for objects > 5GB...

CopyObject already has ComposeObject - SourceInfo, DestinationInfo - CopyObjectOptions doesn't cater for both source and destination though.

aead:
>CopyObject already has ComposeObject - SourceInfo, DestinationInfo - CopyObjectOptions doesn't cater for both source and destination though.

Okay, than why do we need CopyObjectPart at all? For completeness?
As far as I can see ComposeObject works like multipart copy...

harshavardhana:
> Okay, than why do we need CopyObjectPart at all? For completeness?
> As far as I can see ComposeObject works like multipart copy...

This is for Core API for S3 gateway, this is not documented API.

core.go
donatello: @harshavardhana Currently the server does not accept any extra headers for PutObjectPart either - see in object API interface:

```
	CopyObjectPart(srcBucket, srcObject, destBucket, destObject string, uploadID string, partID int, startOffset int64, length int64) (info PartInfo, err error)
	PutObjectPart(bucket, object, uploadID string, partID int, data *hash.Reader) (info PartInfo, err error)
```

This means that PutObjectPart is also not able to accept encryption headers though they are accepted by the [S3 API](https://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadUploadPart.html).

We can go two ways with this:
1. Add server side support for extra headers for both putobjectpart and copyobjectpart, and then add the appropriate support for minio-go
2. Take this change in that adds copyobjectpart for minio-go (without extra headers), and then add support for the extra headers in the server and then circle back and update the client side APIs.

Thoughts?

harshavardhana: Yes we need to do that @donatello that is why i said we need to do on server side as well. This was waiting on @aead to finish CopyObject encryption support on server side.

We also need to do metadata translation for each such incoming requests to the equivalent encryption features for the backend.

donatello:
@harshavardhana @aead Added extra metadata parameter to copy-object-part API. I think this facilitates adding encryption for copying APIs too.

core.go
aead: For SSE-C copy I opened minio/minio#5193. However, current decision is that encryption support for minio-go has higher priority than minio server features. 

IMO we should avoid unnecessary API changes - so I would prefer the first option. However I don't no how important the CopyObjectPart support is. If it's required we have to do 2. approach...

PR #879, opened 2017-11-17
Fix functional tests to honor MINT_DATA_DIR properly

harshavardhana:
getDataReader() now doesn't rely on multiple arguments
for different behavior, instead getDataReader() always
returns a reader depending on the input file name.

This map of file name maps to size which correspond with
MINT_DATA_DIR requirements.

This fixes a whole bunch of bugs in interpretation, repetition
in tests.

### DIFF
--- a/functional_tests.go
+++ b/functional_tests.go
@@ -33,6 +33,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"path/filepath"
 	"reflect"
 	"strconv"
 	"strings"
@@ -44,12 +45,6 @@ import (
 
 	"github.com/minio/minio-go/pkg/encrypt"
 	"github.com/minio/minio-go/pkg/policy"
-)
-
-const (
-	sixtyFiveMiB   = 65 * humanize.MiByte // 65MiB
-	thirtyThreeKiB = 33 * humanize.KiByte // 33KiB
-	oneMiB         = 1 * humanize.MiByte  // 1MiB
 )
 
 const letterBytes = "abcdefghijklmnopqrstuvwxyz01234569"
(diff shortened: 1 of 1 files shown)


functional_tests.go
```diff
@@ -6244,7 +6255,7 @@ func testPutObjectWithContextV2() {
 	}
 	defer c.RemoveBucket(bucketName)
 	bufSize := thirtyThreeKiB
```
poornas: bufSize can also be picked from the map based on filename

harshavardhana: done

functional_tests.go
```diff
@@ -188,19 +183,13 @@ func init() {
...
-	}
-	return
-}
+var mintDataDir = os.Getenv("MINT_DATA_DIR")
 
-func getFilePath(filename string) (filepath string) {
-	if getDataDir() != "" {
-		filepath = getDataDir() + "/" + filename
+func getFilePath(filename string) (fp string) {
+	if mintDataDir == "" {
```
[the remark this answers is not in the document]
harshavardhana: No we are not returning the file name we are only returning if the filepath can be constructed. because this function is meaningless..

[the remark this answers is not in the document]
harshavardhana: Changed the function name to avoid confusion.

harshavardhana:
@ebozduman please take a look again.

functional_tests.go
```diff
@@ -617,9 +623,10 @@ func testPutObjectWithMetadata() {
 		return
 	}
 
-	// Generate data using 2 parts
+	// Generate data for using 2 parts
```
kannappanr: Comment can be more clear.

harshavardhana: sure

harshavardhana:
Removed all unnecessary comments @kannappanr @ebozduman can you review again?

functional_tests.go
```diff
@@ -3803,32 +3796,38 @@ func testPutObjectUploadSeekedObject() {
...
+		// Seek back to the beginning of the file.
+		tempfile.Seek(0, 0)
+	}
+	var length = 100 * humanize.KiByte
 	objectName := fmt.Sprintf("test-file-%v", rand.Uint32())
 	args["objectName"] = objectName
 
 	offset := length / 2
-	if _, err := tempfile.Seek(int64(offset), 0); err != nil {
+	if _, err = tempfile.Seek(int64(offset), 0); err != nil {
```
kannappanr: You should not do os.Remove at line 3845

functional_tests.go
```diff
@@ -1525,11 +1528,6 @@ func testFPutObject() {
 		return
 	}
 
-	err = os.Remove(fName + ".gtar")
```
kannappanr: This file should be removed. This could is needed, otherwise it will leave an extra file in the /mint/data directory

PR #883, opened 2017-11-19
add SSE-C support for GET, PUT, HEAD

### LINKED ISSUE #882: Support SSE-C GET, PUT and HEAD
This is a proposal for adding SSE-C support. 

## Proposal
In general there are two different ways to use encryption:
 - SSE-C through `Get/Put/Stat EncryptedObject`
 - SSE-C, CSE, and KMS through `Get/Put/Stat Options` (will be addressed in a separate issue / PR)

Why use SSE-C for `Get/Put/Stat EncryptedObject`:
 - SSE-C works for all S3 compatible object storage solutions (Minio, AWS, ....)
 - SSE-C works across all S3 compatible object storage solutions (Minio, AWS, ....)
 - SSE-C does not require encryption library support for the client-sdk

### API
This proposal defines the following API:
 - `GetEncryptedObject(bucket, object, password string) (io.ReadCloser, error)`
 - `FGetEncryptedObject(bucket, object, filePath, password string) error`
 - `PutEncryptedObject(bucket, object string, reader io.Reader, size int64, password string) (int64, error)`
 - `FPutEncryptedObject(bucket, object, filePath, password string) (int64, error)`
 - `StatEncryptedObject(bucket, object, password string) (ObjectInfo, error)`
 - `ComposeObject(dst DestinationInfo, src ...SourceInfo) error`

The general idea is that the `password` and the `bucket + object` is used to derive an encryption key:
`key = scrypt.Key(password, bucket + object, 32768, 8, 1, 32) // scrypt 2017 params`
The `key` is unique per object identifier (`bucket + object`). It's the responsibility of the server to ensure that the encryption key is unique (minio server does that - see DARE integration). Even if one `key` is compromised the `password` is not compromised (as long as scrypt is secure) - this may happen e.g. during TLS MitM.

Additionally it's very convenient for clients to use (one) passwords instead of dealing with symmetric keys and  the security of algorithms.

aead:
This change adds SSE-C support. The encryption API uses SSE-C as
encryption method:
 - GetEncryptedObject
 - PutEncryptedObject
 - FGetEncryptedObject
 - FPutEncryptedObject
 - StatEncryptedObject

Further this change unifies the SSE usage (ComposeObject).
Additionally:
 - Adjust mint tests
 - Update API.md doc

Fixes #882


aead:
Will do - The 2. change for encryption (CSE, KMS) may also cause API breaking changes - am I supposed to:
 - combine both changes into one PR
 - change version number in second PR (than we may block this after review until both PRs are ready to merge)
 - change version number twice

api-get-object-encrypted.go
```diff
@@ -0,0 +1,56 @@
...
+	"golang.org/x/crypto/scrypt"
+)
+
+// GetEncryptedObject tries to get an server-side-encrypted object.
+// It returns an error if the key - derived from the provided password - does not
+// match the encryption key of the object. GetEncryptedObject requires a TLS connection.
+func (c Client) GetEncryptedObject(bucketName, objectName, password string) (io.ReadCloser, error) {
+	key, err := scrypt.Key([]byte(password), []byte(bucketName+objectName), 32768, 8, 1, 32) // recommended scrypt parameter for 2017
+	if err != nil {
+		return nil, err // TODO(aead): may panic - this error can only occur if scrypt parameters are invalid
```
poornas: @aead , are the TODO comments to be left in codebase?

api-get-object-encrypted.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+	sse, err := encrypt.NewServerSide(key)
+	if err != nil {
+		return nil, err // TODO(aead): may panic - this error can only occur if derivied key != 256 bits
```
poornas: ditto as above.

functional_tests.go
```diff
@@ -6551,6 +6425,8 @@ func main() {
 
 		// SSE-C tests will only work over TLS connection.
 		if tls {
+			testEncryptionPutGet()
+			testEncryptionFPut()
```
poornas: can we add a test for FGetEncryptedObject?

api-get-object-encrypted.go
aead: Oh, thanks for the hint - forgot to replace with panic.

functional_tests.go
aead: Will add test 👍

aead:
Incremented to 5.0 - But I cannot guarantee that the CSE/KMS change breaks API compatibility.

harshavardhana:
> Incremented to 5.0 - But I cannot guarantee that the CSE/KMS change breaks API compatibility.

It did FGetEncryptedObject() proposed here uses password instead of encrypt.Materials so it is breaking.

aead:
Yes, but this is not the CSE/KMS change. This is "just" the SSE-C change. CSE/KMS will follow...

harshavardhana:
> Yes, but this is not the CSE/KMS change. This is "just" the SSE-C change. CSE/KMS will follow...

Then perhaps we should now design this part such that we don't break future APIs unless really needed.   CSE/KMS change should be incorporated into the a similar looking style IMHO. 

AWS sdks did it with encrypt.Materials which as you suggested is named incorrectly, but encrypt.Materials in their SDKs is serving multiple purpose.

aead:
This was my question 😉 
> Will do - The 2. change for encryption (CSE, KMS) may also cause API breaking changes - am I >supposed to:
> -  combine both changes into one PR
>   - change version number in second PR (than we may block this after review until both PRs are ready to merge)
>    - change version number twice

So I guess block after review until second PR is ready to merge?

kannappanr:
@aead Have marked this PR blocked based on your comments. Can you please update the status?

kannappanr:
@aead Also Travis fails

aead:
@kannappanr Yes, will update status
Travis was failing because of storage class feature - rebased on master - should pass now.

kannappanr:
PR is blocked on CSE and KMS support (needs investigation).

harshavardhana:
> PR is blocked on CSE and KMS support (needs investigation).

Shall we slate this for discussion soonish?

aead:
>Shall we slate this for discussion soonish?

I can investigate requirements for KMS and make proposal, than we can discuss solution/correct approach if needed?

deekoder:
@aead - Is this PR completely blocked by KMS? Can you not independently send separate PRs for the SSE for get/put calls and do key management seperately?

aead:
@deekoder It's not blocked by KMS directly - The problem is that this PR (SSE-C) changes API, so we have to increment version number. The next PR (CSE) changes API again - so we have to change version number again. That's the reason why we didn't review/merged yet. But I can close and open again as soon as other PRs are also ready for review?

PR #886, opened 2017-11-21
PutObjectOptions Header: Don't include invalid header

### LINKED ISSUE #885: Adding X-Amz-Meta headers for macOS xattrs results in invalid headers
When attempting to use this code with `mc` from macOS, it can result in an invalid http header which kills the request.

For example, this xattr on a mac: `com.apple.metadata:kMDItemWhereFroms`. This is included as part of the file metadata, but when you convert the metadata to headers it results in a header with an invalid character `:`.

I'll fork and submit a pull request to fix.

isaachess:
If the provided metadata key contains a character
which is not valid for an http header, do not include
it in the generated header map.

Fixes #885.

### DIFF
--- a/api-put-object.go
+++ b/api-put-object.go
@@ -23,12 +23,15 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"regexp"
 	"runtime/debug"
 	"sort"
 
 	"github.com/minio/minio-go/pkg/encrypt"
 	"github.com/minio/minio-go/pkg/s3utils"
 )
+
+var isValidHeader = regexp.MustCompile("^[a-zA-Z0-9!#$%&'*+-.^_`|~]+$").MatchString
 
 // PutObjectOptions represents options specified by user for PutObject call
 type PutObjectOptions struct {
@@ -78,7 +81,7 @@ func (opts PutObjectOptions) Header() (header http.Header) {
 		header[amzHeaderMatDesc] = []string{opts.EncryptMaterials.GetDesc()}
 	}
 	for k, v := range opts.UserMetadata {
-		if !isAmzHeader(k) && !isStandardHeader(k) && !isSSEHeader(k) {
+		if !isAmzHeader(k) && !isStandardHeader(k) && !isSSEHeader(k) && isValidHeader(k) {
 			header["X-Amz-Meta-"+k] = []string{v}
 		} else {
 			header[k] = []string{v}
--- a/api-put-object_test.go
+++ b/api-put-object_test.go
@@ -51,3 +51,36 @@ func TestPutObjectOptionsValidate(t *testing.T) {
 		}
 	}
 }
+
+// test that metadata is properly converted to headers
+func TestPutObjectOptionsHeaderMetadata(t *testing.T) {
+
+	// test that invalid header keys are not included
+	testCases := []struct {
+		metakey    string
+		shouldPass bool
+	}{
+		{"It has spaces", false},
+		{"It:has@illegal=characters", false},
+		{"It-Is-Fine", true},
+		{"Numbers-098987987-Should-Work", true},
+		{"Crazy-!#$%&'*+-.^_`|~-Should-193832-Be-Fine", true},
+	}
+
+	// construct the meta map from test cases
+	meta := make(map[string]string)
+	for _, testCase := range testCases {
+		meta[testCase.metakey] = "somevalue"
+	}
+
+	opt := PutObjectOptions{UserMetadata: meta}
+	header := opt.Header()
+
+	// iterate through test cases, ensure header is correct
+	for _, testCase := range testCases {
+		_, ok := header["X-Amz-Meta-"+testCase.metakey]
+		if ok != testCase.shouldPass {
+			t.Errorf("Test case %s should be in headers %t, but was %t", testCase.metakey, testCase.shouldPass, ok)
+		}
+	}
+}


harshavardhana:
This is not the right fix @isaachess we should fix it correctly in `mc`  and there is a fix for it too at https://github.com/minio/mc/pull/2310

We can additionally add input validation, but go http client already fails properly.

isaachess:
Fixing it in `mc` makes sense. The reason I made the change in `minio-go` is because it's possible the metadata will be used for something _other_ than a header, in which case you'd want it in the metadata map but throw it out when the metadata is converted to headers (which is the fix I made).

That said, if we fix it in `mc` there are two possible ways to fix it, please let me know which is better to you:

1) We can only load the xattrs into the metadata map if they are valid headers. (This is weird to me simply because, at this point in the code, it is unclear that metadata and headers have a 1:1 relationship. Isn't it possible to want to see metadata that _isn't_ a valid http header?)

2) We can replace any invalid http header character with a valid one, like `-`, so `com.apple.metadata:kMDItemWhereFroms` becomes `com.apple.metadata-kMDItemWhereFroms`.

Thoughts? Preferences?

harshavardhana:
> Fixing it in mc could make sense. The reason I made the change in minio-go is because it's possible the metadata will be used for something other than a header, in which case you'd want it in the metadata map but throw it out when the metadata is converted to headers (which is the fix I made).

Whoever adds an incompatible header simply errors out before making the http request, but we can do a pro-active check as well and return ErrInvalidArgument()

> We only load xattrs into the metadata map if the xattr key is a valid http header name. (This is weird to me simply because, at this point in the code, it is unclear that metadata and headers have a 1:1 relationship. Isn't it possible to want to use metadata somewhere that isn't a valid http header?)

No it is not possible metadata is just headers you can set it to whatever you want, we will fail letting you know that if it was a compatible HTTP header.  We could simply do what http client already does i.e just use this https://godoc.org/golang.org/x/net/lex/httplex#ValidHeaderFieldName to validate the UserMetadata keys and return appropriate error. This can be done in `options.validate()` function. 

> We can replace any invalid http header character with a valid one, like -, so com.apple.metadata:kMDItemWhereFroms becomes com.apple.metadata-kMDItemWhereFroms.

Even this is invalid after the translation as well, there is no real way to translate it back. You cannot differentiate these correctly and there is no cleaner encoding mechanism which can be used.  So there is no reason to go about this route.  Users who want to preserve this will preserve the extended attributes (key, value) in the Header value itself.  Basically http request will have something like this

```
x-amz-meta-xattr: "com.apple.metadata:kMDItemWhereFroms: value"
```

api-put-object.go
```diff
@@ -78,7 +81,7 @@ func (opts PutObjectOptions) Header() (header http.Header) {
 		header[amzHeaderMatDesc] = []string{opts.EncryptMaterials.GetDesc()}
 	}
 	for k, v := range opts.UserMetadata {
-		if !isAmzHeader(k) && !isStandardHeader(k) && !isSSEHeader(k) {
+		if !isAmzHeader(k) && !isStandardHeader(k) && !isSSEHeader(k) && isValidHeader(k) {
```
harshavardhana: Not a good idea to skip the invalid inputs, it can be unexpected from an API point of view - we should return error during validate() for options and return proper error.

isaachess:
Thank you for your comments. I still would like to get this fixed, and want to understand the approach you would prefer.

The issue is:

- The `mc` client loads all xattrs into the metadata map (this makes sense)
- The `mc` client puts that metadata (which includes _all_ xattrs) into the `UserMetadata` field of the `PutObjectOptions` struct.
- When you call `Header()` on `PutObjectOptions`, it attempts to create an http header with every key in the `UserMetadata` struct.

Right now the user has no control over this process, which makes it break when the xattr has a name that is not a valid header.

You said above that you want to give an error if there is an invalid header when you call `Header()`, which is fine, but that means we need to ensure there are on invalid header names in the map when we give it to `PutObjectOptions`.

So I'm still confused about where you prefer the fix to be. It seems like from your comment you want a change in `mc` to fix it. This is the fix I imagine in `mc`:

- When `mc` initializes the `PutObjectOptions` in `cmd/client-s3.go`, it first removes anything that would be an invalid header name. (I consider this the best option, since `minio-go` can give you an error if you include an invalid header, but `mc` is responsible for ensuring invalid header names never go in there.)

If there is a different place / way you'd like to see it fixed, please help me understand. Thanks for your patience on this!

- Isaac

harshavardhana:
Okay lets backup a bit only two things are needed

- Taking in the `mc` PR which would fix the problem
- Adding a check in minio-go using httplex package to validate the header names and values in PutObjectOptions{}.validate() 

That's it.

isaachess:
Are you referring to minio/mc#2310? That appears to allow skipping loading the xattrs, but is not quite the behavior I was hoping for. I would like to include xattrs if they can be converted to headers, I don't want to skip them completely.

harshavardhana:
> Are you referring to minio/mc#2310? That appears to allow skipping loading the xattrs, but is not quite the behavior I was hoping for. I would like to include xattrs if they can be converted to headers, I don't want to skip them completely.

If you don't want them to be skipped its a feature we should add them as part of `mc-xattr` and interpret it at the `mc` level.  It should be like this 

```
x-amz-meta-mc-xattr: base64(xattr)
```

Then apply it back by specially interpreting it.  `mc` fix is correct and the change you want should be done subsequently as a  feature. 

You shouldn't think xattrs on operating systems to be a direct representation as part of UserMetadata the reasons i have explained before.

harshavardhana:
Were you able to do this change @isaachess ?

isaachess:
I apologize, we had the Thanksgiving holiday and then I got quite sick. Let me take a look and get back to you soon.

isaachess:
I am happy to make this change, but want to be sure I understand what you mean. You mean that in `mc`, if have `isFetchMeta` set to true, instead of adding the keys/values directly to `context.Metadata` we should change the key name to `x-amz-meta-mc-xattr` as you mentioned above?

Or do you mean that we should have a separate option somewhere that triggers this behavior?

harshavardhana:
We should preserve the xattr as a special value. Since they cannot be directly encoded into http headers. But that is a separate change . In this PR we should simply validate and fail for incorrect headers passed.

isaachess:
Yes sorry, I will make that change for this repository.

My questions was for how you would like to make the change in the `mc` repository.

harshavardhana:
> Yes sorry, I will make that change for this repository.
> 
> My questions was for how you would like to make the change in the mc repository.

are you still going to work on this PR @isaachess ?

harshavardhana:
Closing this in favor of #907

PR #896, opened 2017-12-12
Add storage class support

nitisht:
This enables `x-amz-storage-class` support in `minio-go`. This is
required as Minio server is adding storage class support.

See https://github.com/minio/minio/issues/4997

### DIFF
--- a/api-put-object.go
+++ b/api-put-object.go
@@ -40,6 +40,7 @@ type PutObjectOptions struct {
 	CacheControl       string
 	EncryptMaterials   encrypt.Materials
 	NumThreads         uint
+	StorageClass       string
 }
 
 // getNumThreads - gets the number of threads to be used in the multipart
@@ -77,8 +78,11 @@ func (opts PutObjectOptions) Header() (header http.Header) {
 		header[amzHeaderKey] = []string{opts.EncryptMaterials.GetKey()}
 		header[amzHeaderMatDesc] = []string{opts.EncryptMaterials.GetDesc()}
 	}
+	if opts.StorageClass != "" {
+		header[amzStorageClass] = []string{opts.StorageClass}
+	}
 	for k, v := range opts.UserMetadata {
-		if !isAmzHeader(k) && !isStandardHeader(k) && !isSSEHeader(k) {
+		if !isAmzHeader(k) && !isStandardHeader(k) && !isSSEHeader(k) && !isStorageClassHeader(k) {
 			header["X-Amz-Meta-"+k] = []string{v}
 		} else {
 			header[k] = []string{v}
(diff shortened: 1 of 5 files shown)


nitisht:
`{"args":{},"duration":1108,"function":"testStorageClassMetadata()","message":"Metadata match failed","name":"minio-go: testStorageClassMetadata","status":"FAIL"}`

This test is expected to fail since play server doesn't support `x-amz-storage-class` yet.

api-put-object.go
```diff
@@ -77,8 +78,11 @@ func (opts PutObjectOptions) Header() (header http.Header) {
 		header[amzHeaderKey] = []string{opts.EncryptMaterials.GetKey()}
 		header[amzHeaderMatDesc] = []string{opts.EncryptMaterials.GetDesc()}
 	}
+	if opts.StorageClass != "" {
+		header["X-Amz-Storage-Class"] = []string{opts.StorageClass}
```
harshavardhana: You might have to add a const for this string.

nitisht: sure will do

nitisht:
PR is updated @harshavardhana @poornas PTAL

constants.go
```diff
@@ -65,3 +65,5 @@ const (
 	amzHeaderKey     = "X-Amz-Meta-X-Amz-Key"
 	amzHeaderMatDesc = "X-Amz-Meta-X-Amz-Matdesc"
 )
+
```
harshavardhana: Add a comment..

utils.go
```diff
@@ -227,13 +227,28 @@ var supportedHeaders = []string{
 	// Add more supported headers here.
 }
 
+var storageClassHeader = []string{
```
harshavardhana: You don't need this to be a slice, this can be simply kept single string and use amzStorageClass?

api-put-object.go
```diff
@@ -40,6 +40,7 @@ type PutObjectOptions struct {
 	CacheControl       string
 	EncryptMaterials   encrypt.Materials
 	NumThreads         uint
+	StorageClass       string
```
harshavardhana: Need to update docs/API.md with this option.

constants.go
nitisht: done

api-put-object.go
nitisht: done

nitisht:
This is updated @harshavardhana PTAL

nitisht:
@harshavardhana could you please take a look on this as well

harshavardhana:
We need to update minio on play.minio.io and also what is the error that is returned when x-amz-storage-class is used with FS backend ?

nitisht:
> what is the error that is returned when x-amz-storage-class is used with FS backend ?

`x-amz-storage-class` value if present in the header is not validated. Only if it matches specific values - data is erasure coded accordingly else it is ignored.

harshavardhana:
Strange then why the test fails ?

nitisht:
> Strange then why the test fails ?

Tests failed as Play server doesn't store and respond with expected `x-amz-storage-class` values.

PR #898, opened 2017-12-22
remove functional tests for context timeout

poornas:
Fixes https://github.com/minio/mint/issues/241

Tests for checking context timeouts for Put|Get object are removed in this PR. This is a sdk functionality test, and hard to set meaningful timeout that works for both server running locally and across network. As request context cancellation is happening at http handler level, there seems to be no value adding these tests to Mint.

### DIFF
--- a/functional_tests.go
+++ b/functional_tests.go
@@ -1642,7 +1642,7 @@ func testFPutObjectWithContext() {
 	// Set base object name
 	objectName := bucketName + "FPutObjectWithContext"
 	args["objectName"] = objectName
-	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Millisecond)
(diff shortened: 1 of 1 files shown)


functional_tests.go
```diff
@@ -1569,309 +1568,6 @@ func testFPutObject() {
...
-		}
-		defer os.Remove(file.Name())
-		fName = file.Name()
-	}
-	totalSize := dataFileMap["datafile-1-MB"]
-
-	// Set base object name
-	objectName := bucketName + "FPutObjectWithContext"
-	args["objectName"] = objectName
-	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Millisecond)
```
nitisht: Instead of removing tests, can we test other contexts e.g. `context.WithValue`

poornas:
closing as this is is a duplicate of #900

PR #899, opened 2017-12-22
fix s3signer to use req.Host header

bnogas:
s3signer module should use req.Host and only if it's not set it should fall back to req.URL.Host. The documentation specifies that HTTP Host header(req.Host) should be used to create canonical request (http://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html). However in aws-sdk there is a fallback to a host from the URL req.URL.Host.

This change fix support for signing requests without absolute path in URI, for example:
GET /resource HTTP/1.1
Host: example.com

### DIFF
--- a/pkg/s3signer/request-signature-streaming_test.go
+++ b/pkg/s3signer/request-signature-streaming_test.go
@@ -33,7 +33,7 @@ func TestGetSeedSignature(t *testing.T) {
 
 	req := NewRequest("PUT", "/examplebucket/chunkObject.txt", body)
 	req.Header.Set("x-amz-storage-class", "REDUCED_REDUNDANCY")
-	req.URL.Host = "s3.amazonaws.com"
+	req.Host = "s3.amazonaws.com"
 
 	reqTime, err := time.Parse("20060102T150405Z", "20130524T000000Z")
 	if err != nil {
@@ -69,6 +69,7 @@ func TestSetStreamingAuthorization(t *testing.T) {
 
 	req := NewRequest("PUT", "/examplebucket/chunkObject.txt", nil)
 	req.Header.Set("x-amz-storage-class", "REDUCED_REDUNDANCY")
+	req.Host = ""
 	req.URL.Host = "s3.amazonaws.com"
 
 	dataLen := int64(65 * 1024)
@@ -93,6 +94,7 @@ func TestStreamingReader(t *testing.T) {
 	req := NewRequest("PUT", "/examplebucket/chunkObject.txt", nil)
 	req.Header.Set("x-amz-storage-class", "REDUCED_REDUNDANCY")
 	req.ContentLength = 65 * 1024
+	req.Host = ""
 	req.URL.Host = "s3.amazonaws.com"
 
 	baseReader := ioutil.NopCloser(bytes.NewReader(bytes.Repeat([]byte("a"), 65*1024)))
--- a/pkg/s3signer/request-signature-v2.go
+++ b/pkg/s3signer/request-signature-v2.go
@@ -40,22 +40,23 @@ const (
 )
 
 // Encode input URL path to URL encoded path.
-func encodeURL2Path(u *url.URL) (path string) {
+func encodeURL2Path(req *http.Request) (path string) {
+	reqHost := getHostAddr(req)
 	// Encode URL path.
-	if isS3, _ := filepath.Match("*.s3*.amazonaws.com", u.Host); isS3 {
-		bucketName := u.Host[:strings.LastIndex(u.Host, ".s3")]
+	if isS3, _ := filepath.Match("*.s3*.amazonaws.com", reqHost); isS3 {
+		bucketName := reqHost[:strings.LastIndex(reqHost, ".s3")]
 		path = "/" + bucketName
-		path += u.Path
+		path += req.URL.Path
 		path = s3utils.EncodePath(path)
 		return
 	}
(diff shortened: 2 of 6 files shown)


vadmeste:
@bnogas, thanks for this PR.

> This change add support for signing requests without absolute path in URI, for example:
> GET /resource HTTP/1.1
> Host: example.com

Are you sure this description is consistent with this PR ?

bnogas:
I've changed from add to fix. In the original code this requested would be signed with empty host for v4, v2 there would be only a problem with encodeURL2Path

vadmeste:
@bnogas, I am not able to find any S3 documentation regarding absolute URI in requests. Are you fixing any bug? Are you using a http proxy in your setup ?

bnogas:
This is a bug fix. S3Signer uses host address extracted from URL (Absolute URI) instead of using http Host Header.
This pull request changes this behavior to use http Host Header by default, and Host extracted from absolute URI if the Host header is not set.

Longer explanation:
**url.URL.Host(req.URL.Host)** - is the host extracted from URL(or Absolute URI) (https://github.com/golang/go/blob/acce8268b611613bef4c422a861bd4863e85b0f1/src/net/url/url.go#L316-L321)
**http.Request.Host(req.Host)** - is the HTTP host header (https://github.com/golang/go/blob/acce8268b611613bef4c422a861bd4863e85b0f1/src/net/http/request.go#L145-L146)
 
**Problem**:
For now the SignV2 and SignV4 methods are using req.URL.Host(Absolute URI)
https://github.com/minio/minio-go/blob/17b9efe2ee358a550ff2d414160b75fc85c86f2e/pkg/s3signer/request-signature-v4.go#L191

But the AWS documentation states that HTTP Host header should be used for signing.
http://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html
"Add the canonical headers, followed by a newline character. The canonical headers consist of a list of all the HTTP headers that you are including with the signed request.

At a minimum, you must include the host header. Standard headers like content-type are optional. Different services might require other headers."

**Why does it work now?**
The signing in api works because of this workaround golang http library:  https://github.com/golang/go/blob/acce8268b611613bef4c422a861bd4863e85b0f1/src/net/http/request.go#L214-L218
If req.Host is empty it is replaced by req.URL.Host when sending request

**What doesn't work correctly?**
If you try to use s3signer package outside minio-go api it works correctly only for requests of this form:
GET http://example.com/resource HTTP/1.1
Host: example.com
Otherwise - if there is no Host specified in URI or if req.Host != req.URL.Host it won't work correctly.

Also because of this workaround in golang http library: https://github.com/golang/go/blob/acce8268b611613bef4c422a861bd4863e85b0f1/src/net/http/request.go#L214-L218 
I suggest to use req.Host as default and req.URL.Host as fallback. Just as Amazon does in it's SDK:
https://github.com/aws/aws-sdk-go/blob/793c7bad9fd33bf3cf2095f32f8024bf47f3cd0a/aws/signer/v4/v4.go#L611-L616

vadmeste:
@bnogas, if I understood correctly, s3signer should not use u.Host is the first place, it should always use u.URL.Host, since s3signer is a http client, not a server. Do you think this way will fix the bug ?

vadmeste:
Happy new year @bnogas

vadmeste:
Functional tests passing if we rebase this PR against master. (Some old bug related to context with short timeout in local machine)

bnogas:
@vadmeste Happy new year!

I think that using req.Host if it's set and req.URL.Host otherwise (like [here](https://github.com/bnogas/minio-go/blob/06d57cab9d121c5083af0362ef9fdb097908c6a6/pkg/s3signer/utils.go#L44-L49)) will fix the problem for all cases. [Canonical Request](https://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html) should reflect the actual _HTTP Host Header_ that will  be sent with the request. And according to [this](https://github.com/golang/go/blob/acce8268b611613bef4c422a861bd4863e85b0f1/src/net/http/request.go#L214-L218) URL.Host will be sent as _HTTP Host Header_ only if req.Host is missing

krisis:
@bnogas  Happy New Year and thank you for sending this PR. Do you have a test program or application that fails due to the current implementation of s3signer?

krisis:
The changes look good to me. While you rebase to resolve conflicts, could you add the above test too? The test needed a small change for it to compile. 
```
diff --git a/pkg/s3signer/request-signature_test.go b/pkg/s3signer/request-signature_test.go
index d53483e..5423435 100644
--- a/pkg/s3signer/request-signature_test.go
+++ b/pkg/s3signer/request-signature_test.go
@@ -69,3 +69,17 @@ func TestSignatureCalculation(t *testing.T) {
                t.Fatal("Error: normal credentials should not have Signature query resource.")
        }
 }
+
+func TestRequestHost(t *testing.T) {
+       req, err := http.NewRequest("GET", "https://s3.amazonaws.com", nil)
+       if err != nil {
+               t.Fatal("Error:", err)
+       }
+       req.URL.RawQuery = "Foo=z&Foo=o&Foo=m&Foo=a"
+       req.Host = "myhost"
+       canonicalHeaders := getCanonicalHeaders(*req, v4IgnoredHeaders)
+
+       if !strings.Contains(canonicalHeaders, "host:"+req.Host) {
+               t.Errorf("canonical host header invalid")
+       }
+}

```

bnogas:
Sure, I've added the test

harshavardhana:
@vadmeste @krisis PTAL

PR #901, opened 2017-12-22
fix: use virtual hosted style if server supports it.

poornas:
This partially fixes https://github.com/minio/mc/issues/2335

Auto probe if virtual hosted style urls are supported
by server. If not, default to path style urls.  Will also be submitting a PR for mc

### DIFF
--- a/api.go
+++ b/api.go
@@ -81,6 +81,9 @@ type Client struct {
 
 	// Random seed.
 	random *rand.Rand
+
+	// whether server supports virtual hosted style urls
+	isVirtual bool
 }
 
 // Global constants.
@@ -238,6 +241,8 @@ func privateNew(endpoint string, creds *credentials.Credentials, secure bool, re
 	// Introduce a new locked random seed.
 	clnt.random = rand.New(&lockedRandSource{src: rand.NewSource(time.Now().UTC().UnixNano())})
 
+	// auto probe if server supports virtual hosted style urls
+	clnt.probeVirtualHostStyle()
 	// Return.
 	return clnt, nil
 }
@@ -799,12 +804,11 @@ func (c Client) makeTargetURL(bucketName, objectName, bucketLocation string, que
 	// endpoint URL.
 	if bucketName != "" {
 		// Save if target url will have buckets which suppport virtual host.
-		isVirtualHostStyle := s3utils.IsVirtualHostSupported(c.endpointURL, bucketName)
-
+		isVirtual := c.isVirtualHostStyle(bucketName)
 		// If endpoint supports virtual host style use that always.
 		// Currently only S3 and Google Cloud Storage would support
 		// virtual host style.
-		if isVirtualHostStyle {
+		if isVirtual {
 			urlStr = scheme + "://" + bucketName + "." + host + "/"
 			if objectName != "" {
 				urlStr = urlStr + s3utils.EncodePath(objectName)
@@ -829,4 +833,113 @@ func (c Client) makeTargetURL(bucketName, objectName, bucketLocation string, que
 	}
 
 	return u, nil
+}
+
+// isVirtualHostStyle - returns true if server supports
+// virtual hosted-style urls
+func (c *Client) isVirtualHostStyle(bucketName string) bool {
+	if c.endpointURL == sentinelURL {
+		return false
+	}
+	// bucketName can be valid but '.' in the hostname will fail SSL
+	// certificate validation. So do not use host-style for such buckets.
+	if c.endpointURL.Scheme == "https" && strings.Contains(bucketName, ".") {
+		return false
+	}
+	return c.isVirtual
+}
+
+// probeVirtualHostStyle - sets the isVirtualHostStyle bool
+// to true if server can accept virtual hosted style urls.
+// Send a HEAD bucket request for a random bucket name
+// with request url in virtual hosted style format,
+// and see if server responds with NotFound status.
+func (c *Client) probeVirtualHostStyle() {
+	// set default to false
+	c.isVirtual = false
+
+	host := c.endpointURL.Host
+	scheme := c.endpointURL.Scheme
+	method := "HEAD"
+	// Strip port 80 and 443 so we won't send these ports in Host header.
+	if h, p, err := net.SplitHostPort(host); err == nil {
+		if scheme == "http" && p == "80" || scheme == "https" && p == "443" {
+			host = h
+		}
+	}
+	// Generate a new random bucket name.
+	bucketName := randString(60, rand.NewSource(time.Now().UnixNano()), "auto-probe")
+
+	// Construct HEAD bucket request with virtual style
+	urlStr := scheme + "://" + bucketName + "." + host + "/"
+
+	u, err := url.Parse(urlStr)
+	if err != nil {
+		return
+	}
+	// Initialize a new HTTP request for the method.
+	req, err := http.NewRequest(method, u.String(), nil)
+	if err != nil {
+		return
+	}
+
+	// Get credentials from the configured credentials provider.
+	value, err := c.credsProvider.Get()
+	if err != nil {
+		return
+	}
+
+	var (
+		signerType      = value.SignerType
+		accessKeyID     = value.AccessKeyID
+		secretAccessKey = value.SecretAccessKey
+		sessionToken    = value.SessionToken
+	)
+
+	// Custom signer set then override the behavior.
+	if c.overrideSignerType != credentials.SignatureDefault {
+		signerType = c.overrideSignerType
+	}
+
+	// If signerType returned by credentials helper is anonymous,
+	// then do not sign regardless of signerType override.
+	if value.SignerType == credentials.SignatureAnonymous {
+		signerType = credentials.SignatureAnonymous
+	}
+	// Set 'User-Agent' header for the request.
+	c.setUserAgent(req)
+
+	metadata := requestMetadata{
+		bucketName:       bucketName,
+		contentSHA256Hex: emptySHA256Hex,
+	}
+	// Set all headers.
+	for k, v := range metadata.customHeader {
+		req.Header.Set(k, v[0])
+	}
+
+	req.Body = nil
+	// Set incoming content-length.
+	req.ContentLength = metadata.contentLength
+	switch {
+	case signerType.IsV2():
+		// Add signature version '2' authorization header.
+		req = s3signer.SignV2(*req, accessKeyID, secretAccessKey)
+	default:
+		// Set sha256 sum for signature calculation only with signature version '4'.
+		shaHeader := unsignedPayload
+		if metadata.contentSHA256Hex != "" {
+			shaHeader = metadata.contentSHA256Hex
+		}
+		req.Header.Set("X-Amz-Content-Sha256", shaHeader)
+		// Add signature version '4' authorization header.
+		req = s3signer.SignV4(*req, accessKeyID, secretAccessKey, sessionToken, "us-east-1")
+	}
+	// Initiate the request.
+	r, _ := c.do(req)
+	if r != nil && (r.StatusCode == http.StatusNotFound) {
+		// Bucket doesn't exist, means server supports virtual hosted style urls.
+		c.isVirtual = true
+	}
+	return
 }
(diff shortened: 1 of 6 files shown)


poornas:
@harshavardhana, if this method is marked private - mc will need to also have similar call to check if virtual hosted style is permitted.

harshavardhana:
mc already has it, also what is the purpose of this function ?

poornas:
mc has a virtualStyle bool that currently checks against hardcoded urls for amazon and google storage. If the function in minio-go is public it is accessible to mc s3 client as well. If we do not want to expose this method, will mark it private and need to have a similar function in mc to detect if virtual hosted style can be used for the storage provider.

harshavardhana:
I know but why it is used. Like why does mc need to know if it's virtual or path style ?

poornas:
to filter out bucket for virtual style requests to get hostname, url parsing of bucket & object.

harshavardhana:
There might be some problem in implementation here and need for improvement.. because mc shouldn't have to look for if it's virtual style.

harshavardhana:
So exposing an API here is not needed IMO.

api.go
```diff
@@ -830,3 +834,105 @@ func (c Client) makeTargetURL(bucketName, objectName, bucketLocation string, que
 
 	return u, nil
 }
+
+// isVirtualHostedStyle - returns true if server supports
+// virtual hosted-style urls
+func (c *Client) isVirtualHostedStyle(bucketName string) bool {
```
vadmeste: Isn't 'probeVirtualHostStyle' name better ?

api.go
```diff
@@ -830,3 +834,105 @@ func (c Client) makeTargetURL(bucketName, objectName, bucketLocation string, que
...
+	// certificate validation. So do not use host-style for such buckets.
+	if c.endpointURL.Scheme == "https" && strings.Contains(bucketName, ".") {
+		return false
+	}
+	return c.isVirtualHostStyle
+}
+
+// setVirtualHostStyle - auto probes if server supports virtual
+// hosted-style urls
+func (c *Client) setVirtualHostStyle() {
```
vadmeste: Can you add more comments about the logic of this function ?

api.go
```diff
@@ -830,3 +834,105 @@ func (c Client) makeTargetURL(bucketName, objectName, bucketLocation string, que
...
+	host := c.endpointURL.Host
+	scheme := c.endpointURL.Scheme
+	method := "HEAD"
+	// Strip port 80 and 443 so we won't send these ports in Host header.
+	if h, p, err := net.SplitHostPort(host); err == nil {
+		if scheme == "http" && p == "80" || scheme == "https" && p == "443" {
+			host = h
+		}
+	}
+	bucketName := "auto-probe-bucketname"
```
vadmeste: What is this `auto-probe-bucketname` ? (I am being lazy but it also means this code needs more comments)

api.go
poornas: @vadmeste, you are right. Changed this.

api.go
poornas: updated function with comments

api.go
poornas: this is just a non existent bucket - updated comments to make it self explanatory.

api.go
```diff
@@ -830,3 +834,111 @@ func (c Client) makeTargetURL(bucketName, objectName, bucketLocation string, que
...
+		if metadata.contentSHA256Hex != "" {
+			shaHeader = metadata.contentSHA256Hex
+		}
+		req.Header.Set("X-Amz-Content-Sha256", shaHeader)
+		// Add signature version '4' authorization header.
+		req = s3signer.SignV4(*req, accessKeyID, secretAccessKey, sessionToken, "us-east-1")
+	}
+	// Initiate the request.
+	r, _ := c.do(req)
+	if r != nil && r.StatusCode == http.StatusNotFound {
```
krisis: what happens if "auto-probe-bucketname" is
a) exists and owned by a different user in AWS S3 (or GCS etc)
b) exists and owned by this user

Shouldn't we accommodate errors due to above possibilities too?

api.go
vadmeste: > @vadmeste, you are right. Changed this.

Sorry @poornas, I didn't mean this to be changed to probeVirtualHostStyle, I wanted to say setVirtualHostStyle() to be renamed instead. I think this way is still better, what do you think ?

api.go
vadmeste: yeah. In case of AWS S3, any user kidding and creating a auto-probe-bucketname bucket in AWS S3 will make all software depending on minio-go switch to use path style. (I am not sure why minio-go is always selecting virtual host style with AWS S3 though)

poornas: I agree that there is a possibility of preexisting bucket with that name - generating a random bucket name could mitigate possibility of a bucket existing with same name on AWS. Do you think that is better approach?

vadmeste: > I agree that there is a possibility of preexisting bucket with that name - generating a random bucket name could mitigate possibility of a bucket existing with same name on AWS. Do you think that is better approach?

I think so.. but the bucket name should be highly random, otherwise it will be dangerous as I explained above.

For example, using 'autoprobe-bucketname' + fmt.Sprintf("%d\", rand.Intn(10000)) won't work because someone can create all the possibility. Then all software using minio-go will get Access denied instead of Bucket Not Found, hence path style will be always activated in case of AWS S3.

poornas: @vadmeste, updated PR to also check if server responds with StatusOK in the event bucket already exists. StatusOK indicates that server could resolve the url path, and hence can support virtual hosted style url.

api.go
```diff
@@ -830,3 +834,112 @@ func (c Client) makeTargetURL(bucketName, objectName, bucketLocation string, que
...
+		if metadata.contentSHA256Hex != "" {
+			shaHeader = metadata.contentSHA256Hex
+		}
+		req.Header.Set("X-Amz-Content-Sha256", shaHeader)
+		// Add signature version '4' authorization header.
+		req = s3signer.SignV4(*req, accessKeyID, secretAccessKey, sessionToken, "us-east-1")
+	}
+	// Initiate the request.
+	r, _ := c.do(req)
+	if r != nil && (r.StatusCode == http.StatusNotFound || r.StatusCode == http.StatusOK) {
```
krisis: A more succinct way of expressing this would be,
```go
switch r.StatusCode {
case http.StatusNotFound, http.StatusOK:
  c.isVirtual = true
}
```

vadmeste: There is a still a problem here:
If a random user creates 'auto-probe-bucketname' in AWS S3, all other users will get access denied after a HEAD, which means that they will never have virtual host enabled.

https://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketHEAD.html

```
13:14 $ aws s3api head-bucket --bucket aws

An error occurred (403) when calling the HeadBucket operation: Forbidden
```

(aws is a bucket created by another user)

api.go
```diff
@@ -830,3 +834,112 @@ func (c Client) makeTargetURL(bucketName, objectName, bucketLocation string, que
...
+	// certificate validation. So do not use host-style for such buckets.
+	if c.endpointURL.Scheme == "https" && strings.Contains(bucketName, ".") {
+		return false
+	}
+	return c.isVirtual
+}
+
+// probeVirtualHostStyle - sets the isVirtualHostStyle bool
+// to true if server can accept virtual hosted style urls.
+// Send a HEAD bucket request for a non existent bucket to
```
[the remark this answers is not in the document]
poornas: done

api.go
```diff
@@ -830,3 +834,112 @@ func (c Client) makeTargetURL(bucketName, objectName, bucketLocation string, que
...
+	host := c.endpointURL.Host
+	scheme := c.endpointURL.Scheme
+	method := "HEAD"
+	// Strip port 80 and 443 so we won't send these ports in Host header.
+	if h, p, err := net.SplitHostPort(host); err == nil {
+		if scheme == "http" && p == "80" || scheme == "https" && p == "443" {
+			host = h
+		}
+	}
+	// bucket doesn't exist.
```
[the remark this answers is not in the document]
poornas: done

api.go
poornas: yes, @vadmeste - this is a little tricky. I have changed the check to using a random bucket name - this has a high probability of being non-existent. And even if it is, the client will switch to path style only for one invocation.

poornas:
@krisis, can't add a unit test for the newly introduced function as it makes an api call.

krisis:
@poornas, I know, which is why I was suggesting a functional test.

poornas:
@krisis, this is an internal method, so there wouldn't be an exposed function to test. fwiw, all existing functional tests will hit this method when target url is constructed.

api.go
```diff
@@ -830,3 +834,112 @@ func (c Client) makeTargetURL(bucketName, objectName, bucketLocation string, que
...
+	host := c.endpointURL.Host
+	scheme := c.endpointURL.Scheme
+	method := "HEAD"
+	// Strip port 80 and 443 so we won't send these ports in Host header.
+	if h, p, err := net.SplitHostPort(host); err == nil {
+		if scheme == "http" && p == "80" || scheme == "https" && p == "443" {
+			host = h
+		}
+	}
+	// Generate a new random bucket name.
```
krisis: ### Motivation
How about we simplify the heuristic to determine if virtual host style access is supported by the object storage service? Apologies for making this recommendation this late in the review.
My aim is,
1. to ensure that we *never* incorrectly detect virtual host support but can incorrectly fall back to path style access. 
2. to avoid deciding which signature version to use in `privateNew`

### Suggested approach
One way would be to simply make an anonymous [GET bucket location](https://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketGETlocation.html) request to a constant bucket name. There are 3 outcomes which indicate virtual host style support.

1. Bucket Found 200 OK (valid location XML response) - Bucket owned by user
(Only possible if there is a [bucket policy](https://docs.aws.amazon.com/AmazonS3/latest/dev/using-with-s3-actions.html#using-with-s3-actions-related-to-buckets) allowing GetBucketLocation operation on the bucket)
2. Bucket Not Found XML error response
3. Access Denied XML error response - Bucket owned by someone else  

For all other responses, we can safely fall back to path-style.

### Remarks
N B in the current approach, using HEAD bucket request we can't differentiate between an S3 bucket and a plain HTTP asset hosted at `bucketname.endpoint`, from a 200 OK response. This is why I recommend GET bucket location, which returns an XML response on 200 OK as well.

poornas: @krisis, how would we guarantee a constant bucket name , short of creating one while probing virtual hosted style support? The other option going by your approach would be to listbuckets and then pick the first bucket and do a HEAD bucket or GET location request - that way, the bucket is guaranteed to belong to user and we would revert to path style only if the server has no buckets created so far / bucket got deleted before HEAD operation is attempted. Thoughts?

krisis: > how would we guarantee a constant bucket name , short of creating one while probing virtual hosted style support? 

We need __not__ create a bucket with a fixed name. We can use any arbitrarily chosen (fixed) bucket name and perform an anonymous GET bucket location request. The heuristic in my previous comment describes how we handle all possible responses.

kannappanr:
@poornas Can you please update the status of the PR?

poornas:
@kannappanr , this PR needs to be withdrawn in current state. Both the approach I used to do a HEAD on bucket vs @krisis suggestion of using GET bucket location have some cases where it would not detect with certainty if the server supports virtual hosted style. Will have to let the user specify virtual hosted style in client constructor in minio-go and in mc config rather than auto detecting.

api.go
poornas: @krisis, user may not have GET Bucket Location permission - in which case this has same pitfalls as   the HEAD bucket probing.

krisis:
@poornas, could you give me an example case where the heuristic I suggested doesn't work?

api.go
krisis: @poornas, for the S3 like service to determine that GET bucket location or HEAD request is not allowed requires the service to resolve the bucket from the host or request resource. In virtual host style access, it is determined from subdomain name of hostname. So, we can infer that the service should have determined the bucket from hostname.

poornas: @krisis, tested your heuristic locally - the AccessDenied error can be raised when the bucket is owned by someone else as well as if the server doesn't support virtual host style (for e.g. while running against localhost). On the other hand, if we excluded AccessDenied in the heuristic we would miss out cases where the server  supports virtual hosted style but the bucket policy doesnt allow anonymous access.

krisis:
@poornas, you are right. I tested the heuristic I suggested and it can't conclusively say if virtual host style bucket access is supported.

poornas:
@krisis, agree - there isn't a fool proof way of detecting whether server supports virtual hosted style request. Will discuss with @abperiasamy  about introducing a config option in minio-go constructor that will allow user to specify virtual hosted as part of config. Closing this PR.

PR #904, opened 2017-12-31
Update functional tests to not expect STANDARD storage class

nitisht:
Minio server doesn't respond with storage class as object metadata
response, if the storage class is set to STANDARD. This PR updates
the tests to fail if STANDARD is sent as storage class in response
metadata.
    
Additionally, Minio gateway (except S3 Gateway) currently ignore storage 
class requests and hence don't respond with storage class in response
metadata for GetObjectInfo calls. Updated the functional tests to
ignore if storage class is not present in metadata.

### DIFF
--- a/functional_tests.go
+++ b/functional_tests.go
@@ -5531,6 +5531,8 @@ func testStorageClassMetadataPutObject() {
 	metadata := make(http.Header)
 	metadata.Set("x-amz-storage-class", "REDUCED_REDUNDANCY")
 
+	emptyMetadata := make(http.Header)
+
 	const srcSize = 1024 * 1024
 	buf := bytes.Repeat([]byte("abcde"), srcSize) // gives a buffer of 1MiB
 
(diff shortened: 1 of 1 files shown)


poornas:
@nitisht , travis failure needs to be fixed.

functional_tests.go
```diff
@@ -5554,8 +5554,8 @@ func testStorageClassMetadataPutObject() {
 		logError(testName, function, args, startTime, "", "PutObject failed", err)
 		return
 	}
-	if !reflect.DeepEqual(metadata, fetchMeta("srcObjectSSClass")) {
-		logError(testName, function, args, startTime, "", "Metadata match failed", err)
+	if reflect.DeepEqual(metadata, fetchMeta("srcObjectSSClass")) {
+		logError(testName, function, args, startTime, "", "Metadata match failed, STANDARD should not be a part of response metadata", err)
```
poornas: Suggest wording this as ""Metadata match failed, STANDARD storage class should not be a part of response metadata".

poornas: "Metadata match failed, STANDARD storage class should not be a part of response metadata" would be clearer. Suggest changing this message and on line 5694

nitisht: done, at both the locations

nitisht:
> travis failure needs to be fixed. 

The storage class related test is expected to fail till https://github.com/minio/minio/pull/5338 is merged. Initially we returned `STANDARD` in response metadata, but this is being changed to comply with S3 behaviour.

kannappanr:
@nitisht Can you reword,  `"Metadata match failed, STANDARD storage class should not be a part of response metadata` ? There is no Metadata mismatch here.

nitisht:
PR is updated @harshavardhana @poornas @kannappanr PTAL

PR #905, opened 2018-01-02
Add new regions for Paris and China west.

harshavardhana:


poornas:
@harshavardhana , think getRegionFromUrl will also need a change - it returns cn-north-1 for all china endpoints.

pkg/s3utils/utils.go
```diff
@@ -122,7 +122,7 @@ func IsAmazonChinaEndpoint(endpointURL url.URL) bool {
 	if endpointURL == sentinelURL {
 		return false
 	}
-	return endpointURL.Host == "s3.cn-north-1.amazonaws.com.cn"
+	return strings.HasSuffix(endpointURL.Host, ".amazonaws.com.cn")
```
ebozduman: The comment only talks about `Beijing` region. How about the newly added `China (Ningxia)` region?
Don't we need to explain covered regions and hence this function returns `true` either the region is `Beijing` or `Ningxia` ?

s3-endpoints.go
```diff
@@ -36,6 +37,7 @@ var awsS3EndpointMap = map[string]string{
 	"sa-east-1":      "s3-sa-east-1.amazonaws.com",
 	"us-gov-west-1":  "s3-us-gov-west-1.amazonaws.com",
 	"cn-north-1":     "s3.cn-north-1.amazonaws.com.cn",
+	"cn-northwest-1": "s3.cn-northwest-1.amazonaws.com.cn",
```
ebozduman: The above comment,
```
// "cn-north-1" adds support for AWS China.
```
is obsolete.

ebozduman: It looks like the following codes in `api.go` and `utilities.go` files also need to be touched to accommodate the newly added region.

api.go
```
func getRegionFromURL(u url.URL) (region string) {
	region = ""
	if s3utils.IsGoogleEndpoint(u) {
		return
	} else if s3utils.IsAmazonChinaEndpoint(u) {
		// For china specifically we need to set everything to
		// cn-north-1 for now, there is no easier way until AWS S3
		// provides a cleaner compatible API across "us-east-1" and
		// China region.
		return "cn-north-1"
	} else if s3utils.IsAmazonGovCloudEndpoint(u) {
        :
        :
```

utilities.go
```
func getDefaultLocation(u url.URL, regionOverride string) (location string) {
	if regionOverride != "" {
		return regionOverride
	}
	if s3utils.IsAmazonChinaEndpoint(u) {
		return "cn-north-1"
	}
        :
        :
```

ebozduman: I'd either update the following test in `utilities_test.go` file to cover the new region, `s3.cn-northwest-1.amazonaws.com.cn`, or add a new test for the new region to test the new code change.
```
		// China region should be honored, region override not provided. - Test 4.
		{
			endpointURL:      url.URL{Host: "s3.cn-north-1.amazonaws.com.cn"},
			regionOverride:   "",
			expectedLocation: "cn-north-1",
		},
```

harshavardhana:
changes are done @poornas @ebozduman

pkg/s3utils/utils.go
```diff
@@ -84,6 +84,9 @@ func IsVirtualHostSupported(endpointURL url.URL, bucketName string) bool {
 // AmazonS3Host - regular expression used to determine if an arg is s3 host.
 var AmazonS3Host = regexp.MustCompile("^s3[.-]?(.*?)\\.amazonaws\\.com$")
 
+// AmazonS3ChinaHost - regular expression used to determin if the arg is s3 china host.
+var AmazonS3ChinaHost = regexp.MustCompile("^s3[.-]?(.*?)\\.amazonaws\\.com.cn$")
```
poornas: doesnt the . before cn$ need to be escaped as well?

poornas: typo -> determin

harshavardhana: > doesnt the . before cn$ need to be escaped as well?

oh yes @poornas

poornas:
LGTM, travis seems to have timed out though

pkg/s3utils/utils.go
```diff
@@ -84,6 +84,9 @@ func IsVirtualHostSupported(endpointURL url.URL, bucketName string) bool {
 // AmazonS3Host - regular expression used to determine if an arg is s3 host.
 var AmazonS3Host = regexp.MustCompile("^s3[.-]?(.*?)\\.amazonaws\\.com$")
 
+// AmazonS3ChinaHost - regular expression used to determine if the arg is s3 china host.
+var AmazonS3ChinaHost = regexp.MustCompile("^s3[.-]?(.*?)\\.amazonaws\\.com\\.cn$")
```
ebozduman: I think not only the period in `[.-]?` needs to be escaped, but the `?` also needs to be removed. We must have a `.` or `-` in the match. However, `?` matches zero or one of the preceding character. So, not having either one of `[.-]`, passes the match check and even if there is an offending character in between `s3` and the region name becomes ok and captured by the `(.*?)`.
Here is an example of this kind of a failure:
Try it with `"s3xcn-north-1.amazonaws.com.cn"`. With `?`, it'll return `parts: [s3xcn-north-1.amazonaws.com.cn xcn-north-1]` because of the `?`.

harshavardhana: Oh yes you are right .

api.go
```diff
@@ -183,11 +183,10 @@ func getRegionFromURL(u url.URL) (region string) {
 	if s3utils.IsGoogleEndpoint(u) {
 		return
 	} else if s3utils.IsAmazonChinaEndpoint(u) {
-		// For china specifically we need to set everything to
-		// cn-north-1 for now, there is no easier way until AWS S3
-		// provides a cleaner compatible API across "us-east-1" and
-		// China region.
-		return "cn-north-1"
+		parts := s3utils.AmazonS3ChinaHost.FindStringSubmatch(u.Host)
+		if len(parts) > 1 {
```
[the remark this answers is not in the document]
harshavardhana: There is no error. The error is 'us-east-1'

pkg/s3utils/utils.go
poornas: @harshavardhana , AmazonS3Host regex also needs to be corrected. Would be good to add negative test cases such as the one above.

harshavardhana:
fixed everything by using different regexes, AFAICS we cannot use a single regex @ebozduman @poornas  - if possible do let me know.  But the code is complete now, i also refactored code to handle region as well part of s3utils.

pkg/s3utils/utils.go
```diff
@@ -81,18 +81,53 @@ func IsVirtualHostSupported(endpointURL url.URL, bucketName string) bool {
...
+var amazonS3HostHyphen = regexp.MustCompile(`^s3-(.*?)\.amazonaws\.com$`)
+
+// amazonS3HostDualStack - regular expression used to determine if an arg is s3 host dualstack.
+var amazonS3HostDualStack = regexp.MustCompile(`^s3\.dualstack\.(.*?)\.amazonaws\.com$`)
+
+// amazonS3HostDot - regular expression used to determine if an arg is s3 host in . style.
+var amazonS3HostDot = regexp.MustCompile(`^s3\.(.*?)\.amazonaws\.com$`)
+
+// amazonS3ChinaHost - regular expression used to determine if the arg is s3 china host.
+var amazonS3ChinaHost = regexp.MustCompile(`^s3\.(cn.*?)\.amazonaws\.com\.cn?$`)
```
ebozduman: Why `?` in `\.cn?$` at the end of the regular expression.
I don't see the need for it.

harshavardhana: Yes not needed..

pkg/s3utils/utils.go
```diff
@@ -81,18 +81,53 @@ func IsVirtualHostSupported(endpointURL url.URL, bucketName string) bool {
...
+		return ""
 	}
 	if IsAmazonGovCloudEndpoint(endpointURL) {
+		return "us-gov-west-1"
+	}
+	parts := amazonS3HostDualStack.FindStringSubmatch(endpointURL.Host)
+	if len(parts) > 1 {
+		return parts[1]
+	}
+	parts = amazonS3HostHyphen.FindStringSubmatch(endpointURL.Host)
```
poornas: the amazonS3HostHyphen match will get region as 'external-1' for s3-external-1.amazonaws.com [us-east-1 region]. This should also be treated specially.

harshavardhana: Done fixed.

harshavardhana:
can i get some reviews here? @poornas @ebozduman

fd0:
It'd be awesome if we could get a new release with this PR included, people are already asking for support of `eu-west-3` in restic: https://forum.restic.net/t/aws-s3-eu-paris/384

harshavardhana:
> It'd be awesome if we could get a new release with this PR included, people are already asking for support of eu-west-3 in restic: https://forum.restic.net/t/aws-s3-eu-paris/384

sure will do soon @fd0

PR #907, opened 2018-01-08
Fail in PutObject if invalid user metadata is passed 

harshavardhana:


kannappanr:
@harshavardhana Travis and Appveyor failed.

krisis:
@harshavardhana is this PR fixing a known issue? Could you describe a use-case where an invalid header name or field value affects minio server's functioning?

harshavardhana:
It is not affecting minio servers functioning. It is just to guard PutObject avoiding invalid user metadata.

krisis:
Why should we guard this at the SDK side while the server is expected to return `InvalidArgument`?

harshavardhana:
> Why should we guard this at the SDK side while the server is expected to return InvalidArgument?

Just like we do BucketName validation, ObjectName validation etc @krisis

kannappanr:
ping @nitisht

PR #910, opened 2018-01-09
Fix bucket region estimation process

nitisht:
When creating a new bucket with a client that doesn't 
have a default region, minio-go doesn't retry if the initial
request failed with `InvalidRegion`. 

This PR fixes this behaviour. Now, minio-go retries make
bucket operation with the correct region as returned by
server in `InvalidRegion` error response.

fixes https://github.com/minio/mc/issues/2352

### DIFF
--- a/api.go
+++ b/api.go
@@ -557,16 +557,17 @@ func (c Client) executeMethod(ctx context.Context, method string, metadata reque
 		// code dictates invalid region, we can retry the request
 		// with the new region.
 		//
-		// Additionally we should only retry if bucketLocation and custom
-		// region is empty.
-		if metadata.bucketLocation == "" && c.region == "" {
+		// we should only retry if custom region is empty.
+		if c.region == "" {
 			if errResponse.Code == "AuthorizationHeaderMalformed" || errResponse.Code == "InvalidRegion" {
+				// Gather Cached location only if bucketName is present.
 				if metadata.bucketName != "" && errResponse.Region != "" {
-					// Gather Cached location only if bucketName is present.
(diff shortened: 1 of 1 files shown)


krisis:
@harshavardhana is it in scope for our request retry logic to detect the correct location of the bucket from an `InvalidRegion` error response for any operation? If yes, for which operations is this meant to be used? If no, we should remove the region detection logic from our request retry code.

harshavardhana:
> @harshavardhana is it in scope for our request retry logic to detect the correct location of the bucket from an InvalidRegion error response for any operation? If yes, for which operations is this meant to be used? If no, we should remove the region detection logic from our request retry code.

Region detection logic is good for all operations such as GET, PUT, HEAD when bucket already exists it is not a good idea to use region logic for MakeBucket() - since region is tied to a bucket it is expected that user provides the right region. If not don't configure Minio server with a specific region just leave it empty and then you can use any region of your choice.

harshavardhana:
> Region detection logic is good for all operations such as GET, PUT, HEAD when bucket already exists it is not a good idea to use region logic for MakeBucket() - since region is tied to a bucket it is expected that user provides the right region. If not don't configure Minio server with a specific region just leave it empty and then you can use any region of your choice.

It is incidental that `x-amz-bucket-region` header exists for MakeBucket() for Minio server since we are not differentiating calls on which this header is actually set, in case of AWS S3 that's not the case only GET and HEAD return this header. Neither would be case for any other endpoints. This in-fact makes me think that we should avoid setting this header altogether to comply with AWS S3 and avoid clients relying on it.

nitisht:
closing as per https://github.com/minio/mc/issues/2352#issuecomment-356872536

PR #913, opened 2018-01-10
increase max retry 

poornas:
Fixes https://github.com/minio/mc/issues/2304

Increase max number of binomial retries to 10 so that long running uploads have a chance of completing in the event of connection timeout.

### DIFF
--- a/retry.go
+++ b/retry.go
@@ -26,7 +26,7 @@ import (
 )
 
 // MaxRetry is the maximum number of retries before stopping.
-var MaxRetry = 5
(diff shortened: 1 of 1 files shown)


retry.go
```diff
@@ -26,7 +26,7 @@ import (
 )
 
 // MaxRetry is the maximum number of retries before stopping.
-var MaxRetry = 5
+var MaxRetry = 10
```
harshavardhana: Interestingly you can do `minio.MaxRetry = 15` in `mc`

PR #915, opened 2018-01-26
Add new constructor NewWithOptions to SDK

poornas:
This is apropos : https://github.com/minio/mc/issues/2335.

When this constructor is used, allows user to specify type of bucket lookup - DNS, Path or Auto. The default option is Auto - meaning SDK defaults to DNS lookup for Google/Amazon endpoints and Path lookups for all other endpoints.

### DIFF
--- a/api.go
+++ b/api.go
@@ -81,6 +81,19 @@ type Client struct {
 
 	// Random seed.
 	random *rand.Rand
+
+	// lookup indicates type of url lookup supported by server. If not specified,
+	// default to Auto.
+	lookup BucketLookupType
+}
+
+// Options for New method
+type Options struct {
+	Creds        *credentials.Credentials
+	Secure       bool
+	Region       string
+	BucketLookup BucketLookupType
+	// Add future fields here
 }
 
 // Global constants.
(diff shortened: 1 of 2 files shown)


docs/API.md
```diff
@@ -96,6 +96,19 @@ __Parameters__
...
+
+__Parameters__
+
+|Param   |Type   |Description   |
+|:---|:---| :---|
+|`endpoint`   | _string_  |S3 compatible object storage endpoint |
+|`accessKeyID`  |_string_   |Access key for the object storage |
+|`secretAccessKey`  | _string_  |Secret key for the object storage |
+|`ssl` | _bool_  | If 'true' API requests will be secure (HTTPS), and insecure (HTTP) otherwise |
+|`region`| _string_ | Region for the object storage |
```
harshavardhana: Missing virtual..

poornas: deliberately so - since the default is path style in any case, using this constructor indicates user wants to explicitly use virtual hosted style. If a bool virtual flag was permitted, the  SDK would need to maintain 3 states internally to be able to identify whether user specified virtual=false for endpoints such as Amazon where we default to virtual style. It wouldn't add much value to provide an additional parameter in the constructor, when user could instantiate with New/NewWithRegion when path style is desired.

harshavardhana: I think this is time to make our New a much more future compatible and break the version compatiblity.

poornas: meaning, we provide a virtual flag in the constructor and remove default behavior for aws/google cloud?

harshavardhana: My proposal is we bring PutObjectOptions like behavior for our New()

```go
NewWithOptions(endpoint, opts *Options)

type Options struct {
       Creds *credentials.Credentials
       Secure bool
       Region string
       VirtualStyle bool
       // Add future fields here
}
```

poornas: yes, it looks cleaner. will check with @abperiasamy

api.go
```diff
@@ -81,6 +81,19 @@ type Client struct {
 
 	// Random seed.
 	random *rand.Rand
+
+	// virtual indicates whether server supports virtual hosted style requests. If not,
+	// default to path style unless it is an Amazon/Google cloud storage endpoint.
+	virtual urlStyle
```
ebozduman: The variable name `virtual` is misleading, since it can have all 3 different possible values.:
1. `UknownURLStyle`
2. `PathStyle`
3. `VirtualHostedStylev`

I guess, it is the left-over name when it was a `boolean` in the previous commit. The comment also sounds like the old comment when `virtual` variable was a boolean.

How about `urlStyle` for the variable name?
However, since `urlStyle` is used for the type name, maybe, the type name can be modified as `urlStyleType`... just a suggestion.

As part  of this change, line#274 will also need to be changed, possibly like:
```
func privateNew(endpoint string, creds *credentials.Credentials, secure bool, region string, urlStyle urlStyleType) (*Client, error) {
:
:
	clnt.urlStyle = urlStyle
:
```

poornas: ok, done.

PR #919, opened 2018-02-08
bucket policy: Support json string in Principal field

vadmeste:
In bucket policy json document, Principal field can take a raw string
value `"*"` which is equivalent to this json `struct{ "AWS": ["*"] }`.
This PR adds support of both json type.

Fixes https://github.com/minio/mc/issues/2170

### DIFF
--- a/pkg/policy/bucket-policy.go
+++ b/pkg/policy/bucket-policy.go
@@ -18,6 +18,8 @@
 package policy
 
 import (
+	"encoding/json"
+	"errors"
 	"reflect"
 	"strings"
 
@@ -80,6 +82,33 @@ var startsWithFunc = func(resource string, resourcePrefix string) bool {
 type User struct {
 	AWS           set.StringSet `json:"AWS,omitempty"`
 	CanonicalUser set.StringSet `json:"CanonicalUser,omitempty"`
+}
+
+// UnmarshalJSON is a custom json unmarshaler for Principal field,
+// the reason is that Principal can take a json struct represented by
+// User string but it can also take a string.
+func (u *User) UnmarshalJSON(data []byte) error {
+	// Try to unmarshal data in a struct equal to User, we need it
+	// to avoid infinite recursive call of this function
+	type AliasUser User
+	var au AliasUser
+	err := json.Unmarshal(data, &au)
+	if err == nil {
+		*u = User(au)
+		return nil
+	}
+	// Data type is not known, check if it is a json string
+	// which contains a star, which is permitted in the spec
+	var str string
+	err = json.Unmarshal(data, &str)
+	if err == nil {
+		if str != "*" {
+			return errors.New("unrecognized Principal field")
+		}
+		*u = User{AWS: set.CreateStringSet("*")}
+		return nil
+	}
+	return err
 }
 
 // Statement - minio policy statement
--- a/pkg/policy/bucket-policy_test.go
+++ b/pkg/policy/bucket-policy_test.go
@@ -26,6 +26,205 @@ import (
 	"github.com/minio/minio-go/pkg/set"
 )
 
+// TestUnmarshalBucketPolicy tests unmarsheling various examples
+// of bucket policies, to verify the correctness of BucketAccessPolicy
+// struct defined in this package.
+func TestUnmarshalBucketPolicy(t *testing.T) {
+	var testCases = []struct {
+		policyData    string
+		shouldSucceed bool
+	}{
+		// Test 1
+		{policyData: `{
+  "Version":"2012-10-17",
+  "Statement":[
+    {
+      "Sid":"AddCannedAcl",
+      "Effect":"Allow",
+      "Principal": {"AWS": ["arn:aws:iam::111122223333:root","arn:aws:iam::444455556666:root"]},
+      "Action":["s3:PutObject","s3:PutObjectAcl"],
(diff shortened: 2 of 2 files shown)


pkg/policy/bucket-policy.go
```diff
@@ -76,12 +78,39 @@ var startsWithFunc = func(resource string, resourcePrefix string) bool {
...
+	err := json.Unmarshal(data, &star)
+	if err == nil {
+		if star != "*" {
+			return errors.New("unrecognized Principal field")
+		}
+		*u = User{AWS: set.CreateStringSet("*")}
+		return nil
+	}
+	// Try to do regular json unmarsheling
+	var _u _User
```
harshavardhana: You can create a struct inside UnmarshalJSON than exposing it outside.. @vadmeste

vadmeste: > You can create a struct inside UnmarshalJSON than exposing it outside.. @vadmeste

@harshavardhana, if I add the struct inside UnmarshalJSON, wouldn't I be obliged to duplicate that struct outside ?

harshavardhana: No you will be doing it a bit differently. Will show you an example.

vadmeste: > No you will be doing it a bit differently. Will show you an example.

ping @harshavardhana :)

vadmeste:
@harshavardhana I changed the code using your suggestion, please review again.

pkg/policy/bucket-policy.go
```diff
@@ -82,6 +84,33 @@ type User struct {
 	CanonicalUser set.StringSet `json:"CanonicalUser,omitempty"`
 }
 
+// UnmarshalJSON is a custom json unmarshaler for Principal field,
+// the reason is that Principal can take a json struct represented by
+// User string but it can also take a string.
+func (u *User) UnmarshalJSON(data []byte) error {
```
balamurugana: As fields in `User` are `set.StringSet`, `set.StringSet` supports single string or string array as JSON value.   Could you tell me what exactly is not working?  As per commit log, I see this method behaves like in `set.StringSet`.

harshavardhana: Perhaps adding a test would be better here @vadmeste to demonstrated.

vadmeste: Added tests @harshavardhana 

@balamurugana, if you can see test n°2, you can see `"Principal": "*"`, it wasn't supported before this PR. Only `"Principal": { "AWS": "xx", "CanonicalUser": "xx"}` is supported.

set.StringSet is not enough in this case, AFAIK, it can only support a string and an array of strings but not a struct (struct User in this case)

balamurugana: @vadmeste understood now.

It would be better to have below approach which makes things clean.

1. Have a type `user` exactly same as `User`
2. Extend `user` as `User`
3. Have `UnmarshalJSON()` to `User`.

pkg/policy/bucket-policy.go
```diff
@@ -82,6 +84,33 @@ type User struct {
...
+		*u = User(au)
+		return nil
+	}
+	// Data type is not known, check if it is a json string
+	// which contains a star, which is permitted in the spec
+	var str string
+	err = json.Unmarshal(data, &str)
+	if err == nil {
+		if str != "*" {
+			return errors.New("unrecognized Principal field")
```
balamurugana: Is it only `*` allowed by AWS?  If not, we have to support it to preserve policy rules for updates.

pkg/policy/bucket-policy.go
vadmeste: > It would be better to have below approach which makes things clean.
> 
> Have a type user exactly same as User
> Extend user as User
> Have UnmarshalJSON() to User.

This is practically what is done in the current approach, anything wrong with it ?

pkg/policy/bucket-policy.go
vadmeste: > Is it only * allowed by AWS?

I guess so, nothing indicates it in this link https://docs.aws.amazon.com/AmazonS3/latest/dev/s3-bucket-user-policy-specifying-principal-intro.html

pkg/policy/bucket-policy.go
balamurugana: you are doing in reverse

pkg/policy/bucket-policy.go
balamurugana: then we have to support all values not limiting to `*`

pkg/policy/bucket-policy.go
vadmeste: > you are doing in reverse

oh yes, understood, but if I do so, it means I need to declare "user" outside UnmarshalJSON(), however no code except UnmarshalJSON() needs it, that's why I did it this way.

pkg/policy/bucket-policy.go
vadmeste: I tested with `aws s3api put-bucket-policy`, AWS responds with "Invalid bucket policy syntax" for whatever value we put, it accepts only `*`.  So I guess it is fine.

pkg/policy/bucket-policy.go
balamurugana: The fact is a type is buried inside `UnmarshalJSON()`.  Hope it won't confuse new comers.

PR #921, opened 2018-02-10
SignV2: Fix signature calculation in virtual host style

vadmeste:
In signature V2, the Resource Path that will be signed should have
the form of /bucket/path/.. even in the case of vhost requests. The
commit fixes the issue. The downside is to forbid automatic http
redirection for V2 requests.

### DIFF
--- a/pkg/s3signer/request-signature-v2.go
+++ b/pkg/s3signer/request-signature-v2.go
@@ -25,7 +25,6 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
-	"path/filepath"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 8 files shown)


pkg/s3signer/request-signature-v2.go
```diff
@@ -40,29 +39,25 @@ const (
...
+		}
 	}
 	path = s3utils.EncodePath(req.URL.Path)
 	return
 }
 
 // PreSignV2 - presign the request in following style.
 // https://${S3_BUCKET}.s3.amazonaws.com/${S3_OBJECT}?AWSAccessKeyId=${S3_ACCESS_KEY}&Expires=${TIMESTAMP}&Signature=${SIGNATURE}.
-func PreSignV2(req http.Request, accessKeyID, secretAccessKey string, expires int64) *http.Request {
+func PreSignV2(req http.Request, accessKeyID, secretAccessKey string, expires int64, vhost bool) *http.Request {
```
poornas: it might be better to rename vhost as  isVirtualHostStyle - vhost is cryptic

pkg/s3signer/request-signature-v2.go
```diff
@@ -132,7 +127,7 @@ func PostPresignSignatureV2(policyBase64, secretAccessKey string) string {
 // CanonicalizedProtocolHeaders = <described below>
 
 // SignV2 sign the request before Do() (AWS Signature Version 2).
-func SignV2(req http.Request, accessKeyID, secretAccessKey string) *http.Request {
+func SignV2(req http.Request, accessKeyID, secretAccessKey string, vhost bool) *http.Request {
```
poornas: rename vhost -> isVirtualHostStyle in this & preStringToSignV2() stringToSignV2() and writeCanonicalizedResource()

pkg/s3signer/request-signature-v2.go
vadmeste: > it might be better to rename vhost as isVirtualHostStyle - vhost is cryptic

Fixed.

pkg/s3signer/request-signature-v2.go
vadmeste: > rename vhost -> isVirtualHostStyle in this & preStringToSignV2() stringToSignV2() and writeCanonicalizedResource()

Fixed.

pkg/s3signer/utils_test.go
```diff
@@ -66,9 +66,9 @@ func TestEncodeURL2Path(t *testing.T) {
 		if err != nil {
 			t.Fatal("Error:", err)
 		}
-		urlPath := "/" + bucketName + "/" + o.encodedObjName
-		if urlPath != encodeURL2Path(&http.Request{URL: u}) {
-			t.Fatal("Error")
+		expectedPath := "/" + bucketName + "/" + o.encodedObjName
+		if foundPath := encodeURL2Path(&http.Request{URL: u}, true); foundPath != expectedPath {
```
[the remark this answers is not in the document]
vadmeste: > we should have tests for vhost being both true and false.

yeah, done.

vadmeste:
ping @krisis

PR #925, opened 2018-02-20
allow OS to cleanup sockets in TIME_WAIT

### LINKED ISSUE #924: Client wasting connections
Under load, the client is currently wasting TCP connection and not properly re-using them. This can very quickly lead to exhaustion and errors like:
```
Put http://127.0.0.1:9000/store/e29c518982e2bbf93efc7743e7ef3abff8f403cf811cdbc9e27b7701a42d92a5: dial tcp 127.0.0.1:9000: connect: can't assign requested address
```

Reproduce with this code, while watching the number of connections on the system.
```golang
	minioClient, err := minio.New(endpoint, accessKeyID, secretAccessKey, useSSL)
	if err != nil {
		log.Fatalln(err)
	}

	var wg sync.WaitGroup
	for i := 0; i < 10; i++ {
		wg.Add(1)
		go func() {
			for i := 0; i < 10000; i++ {
				minioClient.BucketExists(bucketName)
			}
			wg.Done()
		}()
	}
	wg.Wait()
```

The issue appears to be in this line https://github.com/minio/minio-go/blob/368f6da0bdf6d5bb6fbcbf6afdb626e711bab6df/transport.go#L37 
It actually needs to read:
```golang
MaxIdleConnsPerHost:   100,
```
If that is not set, it defaults to http.DefaultMaxIdleConnsPerHost (2) and connections are wasted.

harshavardhana:
fixes #924

### DIFF
--- a/api.go
+++ b/api.go
@@ -1,6 +1,6 @@
 /*
  * Minio Go Library for Amazon S3 Compatible Cloud Storage
- * Copyright 2015-2017 Minio, Inc.
+ * Copyright 2015-2018 Minio, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -287,7 +287,7 @@ func privateNew(endpoint string, creds *credentials.Credentials, secure bool, re
 
 	// Instantiate http client and bucket location cache.
 	clnt.httpClient = &http.Client{
-		Transport:     defaultMinioTransport,
+		Transport:     DefaultTransport,
 		CheckRedirect: clnt.redirectHeaders,
 	}
 
@@ -337,7 +337,7 @@ func (c *Client) SetCustomTransport(customHTTPTransport http.RoundTripper) {
 	//           TLSClientConfig:    &tls.Config{RootCAs: pool},
 	//           DisableCompression: true,
 	//   }
-	//   api.SetTransport(tr)
+	//   api.SetCustomTransport(tr)
 	//
 	if c.httpClient != nil {
 		c.httpClient.Transport = customHTTPTransport
--- a/transport.go
+++ b/transport.go
@@ -2,7 +2,7 @@
 
 /*
  * Minio Go Library for Amazon S3 Compatible Cloud Storage
- * Copyright 2017 Minio, Inc.
+ * Copyright 2017-2018 Minio, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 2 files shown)


transport.go
```diff
@@ -25,16 +25,18 @@ import (
 	"time"
 )
 
-// This default transport is similar to http.DefaultTransport
-// but with additional DisableCompression:
-var defaultMinioTransport http.RoundTripper = &http.Transport{
+// DefaultTransport - this default transport is similar to
+// http.DefaultTransport but with additional param  DisableCompression
+// is set to true toavoid decompressing content with 'gzip' encoding.
```
krisis: s/toavoid/to avoid/

krisis: We should add 2018 to copyright for this file.

api.go
```diff
@@ -338,7 +338,7 @@ func (c *Client) SetCustomTransport(customHTTPTransport http.RoundTripper) {
 	//           TLSClientConfig:    &tls.Config{RootCAs: pool},
 	//           DisableCompression: true,
 	//   }
-	//   api.SetTransport(tr)
+	//   api.SetCustomTransport(tr)
```
krisis: Since we are here, shouldn't `SetCustomTransport` set `DisableCompression` to `true` unconditionally?

krisis: Add 2018 to copyright header of this file.

harshavardhana: > Since we are here, shouldn't SetCustomTransport set DisableCompression to true unconditionally?

Not really since its custom we are not governing what a user wants to do.

transport.go
```diff
@@ -25,16 +25,18 @@ import (
...
+// is set to true to avoid decompressing content with 'gzip' encoding.
+var DefaultTransport http.RoundTripper = &http.Transport{
 	Proxy: http.ProxyFromEnvironment,
 	DialContext: (&net.Dialer{
 		Timeout:   30 * time.Second,
 		KeepAlive: 30 * time.Second,
 		DualStack: true,
 	}).DialContext,
 	MaxIdleConns:          100,
+	MaxIdleConnsPerHost:   100,
```
ebozduman: When you have `MaxIdleConnsPerHost`, do you still need `MaxIdleConns`?

harshavardhana: yes we need @ebozduman see the original issue for more explanation. http://tleyden.github.io/blog/2016/11/21/tuning-the-go-http-client-library-for-load-testing/

PR #928, opened 2018-02-23
Add support for Content-Language metadata header

### LINKED ISSUE #922: support for Content-Language
Amazon supports a `Content-Language` field in metadata, but minio does not have an entry in its `PutObjectOptions` struct for it. Trying to put it in the `UserMetadata` map will prepend `X-Amz-Meta-`, so that doesn't work.

kannappanr:
Fixes #922

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -584,6 +584,7 @@ __minio.PutObjectOptions__
 | `opts.ContentType` | _string_ | Content type of object, e.g "application/text" |
 | `opts.ContentEncoding` | _string_ | Content encoding of object, e.g "gzip" |
 | `opts.ContentDisposition` | _string_ | Content disposition of object, "inline" |
+| `opts.ContentLanguage` | _string_ | Content language of object, e.g "French" |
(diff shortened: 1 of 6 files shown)


docs/API.md
```diff
@@ -911,7 +912,7 @@ __Parameters__
 |`bucketName`  | _string_  |Name of the bucket  |
 |`objectName` | _string_  |Name of the object |
 |`filePath` | _string_  |Path to file to be uploaded |
-|`opts` | _minio.PutObjectOptions_  |Pointer to struct that allows user to set optional custom metadata, content-type, content-encoding,content-disposition and cache-control headers, pass encryption module for encrypting objects, and optionally configure number of threads for multipart put operation.  |
+|`opts` | _minio.PutObjectOptions_  |Pointer to struct that allows user to set optional custom metadata, content-type, content-encoding, content-language and cache-control headers, pass encryption module for encrypting objects, and optionally configure number of threads for multipart put operation.  |
```
ebozduman: Why is `content-disposition` over written?

poornas:
perhaps content-language should be filtered from the object metadata and added as a separate field in ObjectInfo so that it doesn't show under user defined metadata.

PR #929, opened 2018-02-23
Honor prefix parameter in ListBucketPolicies API

### LINKED ISSUE #926: ListBucketPolicies does not filter policy rules using objectPrefix parameter
When reading the code of ListBucketPolicies() (https://github.com/minio/minio-go/blob/master/api-get-policy.go#L52), we can see that objectPrefix is not being used to filter policy rules although ListBucketPolicies() claims to do so.

```
// ListBucketPolicies - list all policies for a given prefix and all its children.
func (c Client) ListBucketPolicies(bucketName, objectPrefix string) (bucketPolicies map[string]policy.BucketPolicy, err error) {
        // Input validation.
                return map[string]policy.BucketPolicy{}, err
        }
        if err := s3utils.CheckValidObjectNamePrefix(objectPrefix); err != nil {
                return map[string]policy.BucketPolicy{}, err
        }
        policyInfo, err := c.getBucketPolicy(bucketName)
        if err != nil {
                errResponse := ToErrorResponse(err)
                if errResponse.Code == "NoSuchBucketPolicy" {
                        return map[string]policy.BucketPolicy{}, nil
                }       
                return map[string]policy.BucketPolicy{}, err
        }
        return policy.GetPolicies(policyInfo.Statements, bucketName), nil
}
```
This needs to be fixed.

kannappanr:
Fixes #926

### DIFF
--- a/pkg/policy/bucket-policy.go
+++ b/pkg/policy/bucket-policy.go
@@ -593,14 +593,14 @@ func GetPolicy(statements []Statement, bucketName string, prefix string) BucketP
 	return policy
 }
 
-// GetPolicies - returns a map of policies rules of given bucket name, prefix in given statements.
-func GetPolicies(statements []Statement, bucketName string) map[string]BucketPolicy {
+// GetPolicies - returns a map of policies of given bucket name, prefix in given statements.
+func GetPolicies(statements []Statement, bucketName, prefix string) map[string]BucketPolicy {
 	policyRules := map[string]BucketPolicy{}
 	objResources := set.NewStringSet()
 	// Search all resources related to objects policy
 	for _, s := range statements {
 		for r := range s.Resources {
-			if strings.HasPrefix(r, awsResourcePrefix+bucketName+"/") {
+			if strings.HasPrefix(r, awsResourcePrefix+bucketName+"/"+prefix) {
 				objResources.Add(r)
 			}
 		}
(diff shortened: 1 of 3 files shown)


pkg/policy/bucket-policy.go
```diff
@@ -620,6 +620,33 @@ func GetPolicies(statements []Statement, bucketName string) map[string]BucketPol
 	return policyRules
 }
 
+// GetPrefixPolicies - returns a map of policies of given bucket name, prefix in given statements.
+func GetPrefixPolicies(statements []Statement, bucketName, prefix string) map[string]BucketPolicy {
```
harshavardhana: actually a better function name for go style is `GetPoliciesWithPrefix` helps in editor auto completion and also conveys the purpose.

kannappanr: done

kannappanr:
@harshavardhana Changes made. PTAL

pkg/policy/bucket-policy.go
```diff
@@ -620,6 +620,33 @@ func GetPolicies(statements []Statement, bucketName string) map[string]BucketPol
 	return policyRules
 }
 
+// GetPoliciesWithPrefix - returns a map of policies of given bucket name, prefix in given statements.
+func GetPoliciesWithPrefix(statements []Statement, bucketName, prefix string) map[string]BucketPolicy {
```
vadmeste: It looks like there is little difference between GetPolicies & GetPoliciesWithPrefix, they look practically the same, can you do with this style ?

```
func GetPolicies(statements []Statement, bucketName string) map[string]BucketPolicy {
   return GetPoliciesWithPrefix(statements, bucketName, "")
}
```

kannappanr: Handling both the cases in GetPolicies function itself.

PR #930, opened 2018-02-24
Encryption headers should also be set during initMultipart

harshavardhana:
This was reproduced while testing encryption features.

### DIFF
--- a/api-compose-object.go
+++ b/api-compose-object.go
@@ -527,6 +527,12 @@ func (c Client) ComposeObject(dst DestinationInfo, srcs []SourceInfo) error {
 	for k, v := range metaMap {
 		metaHeaders[k] = v
 	}
+
+	// Add destination encryption headers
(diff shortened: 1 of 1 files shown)


api-compose-object.go
```diff
@@ -527,6 +527,12 @@ func (c Client) ComposeObject(dst DestinationInfo, srcs []SourceInfo) error {
 	for k, v := range metaMap {
 		metaHeaders[k] = v
 	}
+
+	// Add destination encryption headers
+	for k, v := range dst.encryption.getSSEHeaders(false) {
```
poornas: more readable to just use GetSSEHeaders().LGTM otherwise.

harshavardhana: We don't need to stick to requirements of a public function. Since this is an internal functionality it makes sense to use the internal function.

poornas: ok

PR #933, opened 2018-02-26
Fix errors in file when using the Asymmetrical ver

wilyarti:
The t.Fatal(err) will not compile, changed to log.Fatalln. Both symmetrical and asymmetrical versions now compile fine.

### DIFF
--- a/examples/s3/fputencrypted-object.go
+++ b/examples/s3/fputencrypted-object.go
@@ -20,10 +20,10 @@
 package main
 
 import (
-	"log"
-
+	//"io/ioutil"
 	"github.com/minio/minio-go"
 	"github.com/minio/minio-go/pkg/encrypt"
+	"log"
 )
 
 func main() {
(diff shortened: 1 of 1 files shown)


examples/s3/fputencrypted-object.go
```diff
@@ -20,10 +20,10 @@
 package main
 
 import (
-	"log"
-
+	//"io/ioutil"
```
harshavardhana: Remove this comment..

examples/s3/fputencrypted-object.go
```diff
@@ -20,10 +20,10 @@
...
 	"github.com/minio/minio-go"
 	"github.com/minio/minio-go/pkg/encrypt"
+	"log"
```
harshavardhana: standard imports "log" should be placed above the public imports https://github.com/golang/go/wiki/CodeReviewComments#imports

PR #938, opened 2018-03-01
Moving doneCh handling outside of conditional check for events received

magaldima:


api-notification.go
```diff
@@ -207,11 +207,14 @@ func (c Client) ListenBucketNotification(bucketName, prefix, suffix string, even
 				}
 				// Send notifications on channel only if there are events received.
 				if len(notificationInfo.Records) > 0 {
-					select {
-					case notificationInfoCh <- notificationInfo:
-					case <-doneCh:
-						return
-					}
+					notificationInfoCh <- notificationInfo
```
harshavardhana: If reading notification is blocked then doneCh will never get called. We should do this differently, we don't need to do `if len(notificationInfo.Records) > 0` , we should simply remove the `if` condition and keep the rest of the code as is as below. 

```
select {
case notificationInfoCh <- notificationInfo:
case <-doneCh:
              return
}
```

magaldima: Okay. I will update the PR. Thanks!

Issue #906, opened 2018-01-05
Implement WithContext methods for remove APIs


mahtuag:
There's a PutObjectWithContext to modernize PutObject but no RemoveObjectWithContext to modernize RemoveObject. Instead the current implementation of RemoveObject passes context.Background() to requests. This is true for other methods in `api-remove.go` as well.

https://github.com/minio/minio-go/blob/368f6da0bdf6d5bb6fbcbf6afdb626e711bab6df/api-remove.go#L41

harshavardhana:
> There's a PutObjectWithContext to modernize PutObject but no RemoveObjectWithContext to modernize RemoveObject. Instead the current implementation of RemoveObject passes context.Background() to requests. This is true for other methods in api-remove.go as well.

what would be the purpose of this API ? @mahtuag 

PutObject/GetObject was implemented because usually uploading/downloading an object takes time and easier way to cancel ongoing requests is by using contexts.  RemoveObject is quick has literally no ongoing data transfer adding contexts for cancellation doesn't really make much sense.

mahtuag:
All new Go APIs that involve network operations pass a context along so this should help make the RemoveObject API consistent with the other APIs. I agree realistically nobody would be cancelling a RemoveObject request so this would definitely be a low-ish priority issue.

kannappanr:
@mahtuag Will discuss within the team and see if we need to take this up.

kannappanr:
@mahtuag We have decided to add an API called RemoveObjectswithContext, to modernize RemoveObjects . We have also decided not to implement the same for RemoveObject call.

deekoder:
We have not had the cycles to fix this issue. We will close it and reopen when we actually work on it. This is also a great item for community contribution, if anyone wants to take it up.

PR #939, opened 2018-03-03
Add RemoveObjectsWithContext call

awwalker:
Fix for: #906


nitisht:
Thanks for the PR @awwalker will review in a bit.

PR #941, opened 2018-03-06
Add SSE-C tests for multipart, copy, get range operations

harshavardhana:


functional_tests.go
```diff
@@ -2715,6 +2715,385 @@ func testCopyObject() {
...
+		os.Getenv(secretKey),
+		mustParseBool(os.Getenv(enableHTTPS)),
+	)
+	if err != nil {
+		logError(testName, function, args, startTime, "", "Minio client object creation failed", err)
+		return
+	}
+
+	// Enable tracing, write to stderr.
+	// c.TraceOn(os.Stderr)
```
aead: Maybe we can remove disabled code?

functional_tests.go
```diff
@@ -2715,6 +2715,385 @@ func testCopyObject() {
...
+
+	// Make a new bucket.
+	err = c.MakeBucket(bucketName, "us-east-1")
+	if err != nil {
+		logError(testName, function, args, startTime, "", "MakeBucket failed", err)
+		return
+	}
+
+	// Generate 33K of data.
+	bufSize := dataFileMap["datafile-33-kB"]
```
aead: Should we not use an object which is at least large enough to fit into multiple encrypted packages?

aead: Should we not use an object which is at least large enough to fit into multiple encrypted packages?

functional_tests.go
aead: Maybe we can remove disabled code?

functional_tests.go
aead: Should we not use an object which is at least large enough to fit into multiple encrypted packages?

functional_tests.go
aead: Maybe we can remove disabled code?

harshavardhana: No, it is kept on purpose for debugging.

aead: Yes, but if someone wants to debug he/she can simple add. However very minior - if you think we should keep, than we can...

functional_tests.go
harshavardhana: Sure adding that.

functional_tests.go
harshavardhana: We have kept it in all SDKs

aead: Okay

functional_tests.go
```diff
@@ -2715,6 +2715,385 @@ func testCopyObject() {
...
+		{-1024, 2, int64(bufSize) - 1024, nil, true, bufSize - 1024, bufSize},
+		// Test with whence SEEK_END and with large negative offset
+		{-int64(bufSize) * 2, 2, 0, seekErr, true, 0, 0},
+	}
+
+	for i, testCase := range testCases {
+		// Perform seek operation
+		n, err := r.Seek(testCase.offset, testCase.whence)
+		// We expect an error
+		if testCase.err == seekErr && err == nil {
```
kannappanr: Why do we need this check? The next one seems to capture this also.

harshavardhana: This is a copy paste from the original code, my hope is that i don't need to fix those things in this PR.

functional_tests.go
```diff
@@ -5421,6 +5802,43 @@ func testEncryptedCopyObjectWrapper(c *minio.Client) {
 		logError(testName, function, args, startTime, "", "Downloaded object mismatched for encrypted object", err)
 		return
 	}
+
+	// Test key rotation technique.
+	dst, err = minio.NewDestinationInfo(bucketName, "srcObject", &key1, nil)
```
kannappanr: Destination and source seems to use the same key

harshavardhana: fixing

functional_tests.go
harshavardhana: Fixed this as well.

functional_tests.go
```diff
@@ -2715,6 +2717,380 @@ func testCopyObject() {
...
+	args["bucketName"] = bucketName
+
+	// Make a new bucket.
+	err = c.MakeBucket(bucketName, "us-east-1")
+	if err != nil {
+		logError(testName, function, args, startTime, "", "MakeBucket failed", err)
+		return
+	}
+
+	// Generate 33K of data.
```
kannappanr: Comment should say 65 MB of data

functional_tests.go
```diff
@@ -2715,6 +2717,380 @@ func testCopyObject() {
...
+		logError(testName, function, args, startTime, "", "PutObject failed", err)
+		return
+	}
+
+	if n != int64(bufSize) {
+		logError(testName, function, args, startTime, "", "Number of bytes does not match, expected "+string(int64(bufSize))+", got "+string(n), err)
+		return
+	}
+
+	defer func() {
```
kannappanr: should we move this code after successful makebucket call? That way it is guaranteed to cleanup in this function

functional_tests.go
```diff
@@ -2715,6 +2717,380 @@ func testCopyObject() {
...
+	args["bucketName"] = bucketName
+
+	// Make a new bucket.
+	err = c.MakeBucket(bucketName, "us-east-1")
+	if err != nil {
+		logError(testName, function, args, startTime, "", "MakeBucket failed", err)
+		return
+	}
+
+	// Generate 33K of data.
```
kannappanr: Comment should say 65MB of dat

functional_tests.go
```diff
@@ -2715,6 +2717,380 @@ func testCopyObject() {
...
+	}
+	offset := int64(2048)
+
+	// read directly
+	buf1 := make([]byte, 512)
+	buf2 := make([]byte, 512)
+	buf3 := make([]byte, 512)
+	buf4 := make([]byte, 512)
+
+	// Test readAt before stat is called.
```
kannappanr: why is calling readAt before stat significant?

harshavardhana: because we need to see if stat() returns the correct size not the range size which is obtained with ReadAt()

functional_tests.go
```diff
@@ -5421,6 +5799,43 @@ func testEncryptedCopyObjectWrapper(c *minio.Client) {
 		logError(testName, function, args, startTime, "", "Downloaded object mismatched for encrypted object", err)
 		return
 	}
+
+	// Test key rotation technique for multipart files.
```
kannappanr: The file size is 1MB, maybe the comment should just say `Testkey rotation`

functional_tests.go
kannappanr: But we don't seem to make the stat() call

functional_tests.go
harshavardhana: yes good idea.

functional_tests.go
harshavardhana: We should I removed it while I was testing. There was a bug.

functional_tests.go
```diff
@@ -5421,6 +5810,79 @@ func testEncryptedCopyObjectWrapper(c *minio.Client) {
...
+	}
+
+	// Test in-place decryption.
+	dst, err = minio.NewDestinationInfo(bucketName, "srcObject", nil, nil)
+	if err != nil {
+		logError(testName, function, args, startTime, "", "NewDestinationInfo failed", err)
+		return
+	}
+	args["destination"] = dst
+
```
kannappanr: `src` is using `key1`, while theSSE-C key for `srcobject` has changed to `key2`. We need a new sourceinfo variable here that uses `key2`

harshavardhana: actually, we don't the source is not unencrypted.

harshavardhana: Oh, i see you meant CopyObject, there was a problem in GetObject() as well.

PR #942, opened 2018-03-08
Add SSE and refactor encryption API

### LINKED ISSUE #882: Support SSE-C GET, PUT and HEAD

aead:
This commit does the following changes:

 - Remove all client-side-encryption code
 - Add unified server-side-encryption interface
 - Implement SSE-C and SSE-S3
   (SSE-KMS needs a bit more investigation to finally decide how to implement)
 - Remove special encryption API calls like Get/PutEncryptedObject
   (They are not really usefull after the encryption refactoring)
 - Update functional tests. The functional tests use the new API now.
 - Update API doc and increment version number to 5.0.0
 - Update S3 encryption examples

The client-side-encryption code is removed because it was hardly used,
poorly designed/specified by AWS and has no advantages compared to SSE
for allmost all uses-cases we encounter.

The SSE implementation is located at the `encrypt` package. It exposes
and abstract and easy to use API. Currently the `encrypt` package
implements SSE-C and SSE-S3. SSE-KMS is not yet implemented because some
more investigation is needed to be aware of all details/requirements/implications
of the encryption context which can be passed to SSE-KMS.

All encryption-specific client APIs - like Get/PutEncryptedObject - are
removed because there is no real benefit in keeping them. SSE can simply
be specified as part of Get/Put/Stat ObjectOptions.

The functional tests are updated.

The API documentations is updated.

The S3 encryption examples are updated.
The new examples use the new encryption API.

Fixes #882


pkg/encrypt/server-side.go
```diff
@@ -0,0 +1,146 @@
...
+	"encoding/base64"
+	"errors"
+	"net/http"
+
+	"golang.org/x/crypto/argon2"
+)
+
+const (
+	// sse is the AWS SSE header used for SSE-S3 and SSE-KMS.
+	sse = "X-Amz-Server-Side-Encryption"
```
vadmeste: I think it could be better if we rename these constats to little more meaningful like sseHdr, sseHdrCustomerAlgorithm, etc..

What do you think @aead ?

pkg/encrypt/server-side.go
```diff
@@ -0,0 +1,146 @@
...
+	sseCopyCustomerKey = "X-Amz-Copy-Source-Server-Side-Encryption-Customer-Key"
+	// sseCopyCustomerKeyMD5 is the AWS SSE-C encryption key MD5 HTTP header key for CopyObject API.
+	sseCopyCustomerKeyMD5 = "X-Amz-Copy-Source-Server-Side-Encryption-Customer-Key-MD5"
+)
+
+// PBKDF creates a SSE-C key from the provided password and salt.
+// PBKDF is a password-based key derivation function
+// which can be used to derive a high-entropy cryptographic
+// key from a low-entropy password and a salt.
+type PBKDF func(password, salt []byte) ServerSide
```
vadmeste: What about renaming from PBKDF to KeySalter ? I guess it would be easier for users to understand

pkg/encrypt/server-side.go
aead: The constants are taken from the server code (but unreported). `sse` is the only one I added directly - but I agree that we can use a more expressive name for `sse`. Maybe `sseMethod` / `sseType`.

pkg/encrypt/server-side.go
harshavardhana: PBKDF is more technically correct here @vadmeste - KeySalter is a very unintuitive name. For Crypto, we can be a bit more expressive.

aead: Hm - that is kind of difficult because for example I would not know what a "KeySalter" does and it does not tell me that I deals with passwords two. I see that `PBKDF` is kind of cryptic to an average user - but `PasswordBasedKeyDerivationFunction` is also a WTF and very annoying to write in code.
However the `PBKDF` type is only for advanced user how want to customize password -> key derivation. For example use the `PBKDF2` or `Scrypt`. I guess does users may know what it means...

aead: It's only there to provide an upgrade mechanism in the future...

harshavardhana: Crypto I think we should assume that users know, don't use crypto otherwise you might shoot yourself.

api-compose-object.go
```diff
@@ -169,21 +121,21 @@ type SourceInfo struct {
 // `decryptSSEC` is the decryption key using server-side-encryption
 // with customer provided key. It may be nil if the source is not
 // encrypted.
-func NewSourceInfo(bucket, object string, decryptSSEC *SSEInfo) SourceInfo {
+func NewSourceInfo(bucket, object string, sse encrypt.ServerSide) SourceInfo {
```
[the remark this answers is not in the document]
aead: Yes, so for example an user can specify SSE-C for source but S3 for destination.
If the encryption method does not match the previously uploaded objects the server will return an error.

api-compose-object.go
```diff
@@ -169,21 +121,21 @@ type SourceInfo struct {
...
 	}
 
 	// Set the source header
 	r.Headers.Set("x-amz-copy-source", s3utils.EncodePath(bucket+"/"+object))
 
 	// Assemble decryption headers for upload-part-copy request
-	for k, v := range decryptSSEC.getSSEHeaders(true) {
-		r.Headers.Set(k, v)
+	if r.encryption != nil {
+		encrypt.ServerSideCopy(r.encryption).Marshal(r.Headers)
```
vadmeste: I think it is little easier if we write MarshalCopy() function instead of doing like this

aead: I thought about this first too - but there are multiple issue (IMO) with the approach:
 1. `*Copy` sounds like we are copying something. But we don't do any copy - bit misleading...
 2. SSE-C is the only form of server-side-encryption which distinguish copy / non-copy.
     So we would pollute the interface only for that...
 3. This is (at the moment) the only place we use the copy construction...

vadmeste:
@aead can you update README.md, there are still references to the removed encryption APIs.

aead:
@vadmeste Oh yes - forgot that, thanks.

pkg/encrypt/server-side.go
deekoder: The comment helps me understand the acronym PBKDF.

api-put-object.go
```diff
@@ -78,16 +78,14 @@ func (opts PutObjectOptions) Header() (header http.Header) {
...
-		header[amzHeaderMatDesc] = []string{opts.EncryptMaterials.GetDesc()}
+	if opts.ServerSide != nil {
+		opts.ServerSide.Marshal(header)
 	}
 	if opts.StorageClass != "" {
 		header[amzStorageClass] = []string{opts.StorageClass}
 	}
 	for k, v := range opts.UserMetadata {
-		if !isAmzHeader(k) && !isStandardHeader(k) && !isSSEHeader(k) && !isStorageClassHeader(k) {
+		if !isAmzHeader(k) && !isStandardHeader(k) && !isStorageClassHeader(k) {
```
harshavardhana: we should error out if users set conflicting sse-c keys in metadata such that we can ask them to use ServerSide{}

examples/s3/fputencrypted-object.go
```diff
@@ -40,38 +41,16 @@ func main() {
...
-	//	log.Fatalln(err)
-	// }
-	//
-	// asymmetricKey, err := encrypt.NewAsymmetricKey(privateKey, publicKey)
-	// if err != nil {
-	//	log.Fatalln(err)
-	// }
-	////
+	filePath := "my-testfile"                  // Specify a local file that we will upload
+	bucketname := "my-bucektname"              // Specify a bucket name - the bucket must already exist
```
harshavardhana: s/my-bucektname/my-bucketname

examples/s3/get-encrypted-object.go
```diff
@@ -42,35 +42,15 @@ func main() {
...
-	// if err != nil {
-	//	log.Fatalln(err)
-	// }
-	//
-	// asymmetricKey, err := encrypt.NewAsymmetricKey(privateKey, publicKey)
-	// if err != nil {
-	//	log.Fatalln(err)
-	// }
-	////
+	bucketname := "my-bucektname"              // Specify a bucket name - the bucket must already exist
```
harshavardhana: same as above

api-put-object.go
aead: Okay - yes, agree.

examples/s3/fputencrypted-object.go
aead: Done

examples/s3/get-encrypted-object.go
aead: Done

api-put-object.go
aead: Done - but added this to the `validate()` function - so that the we error out.

api-get-options.go
```diff
@@ -28,9 +28,8 @@ import (
 // GetObjectOptions are used to specify additional headers or options
 // during GET requests.
 type GetObjectOptions struct {
-	headers map[string]string
-
-	Materials encrypt.Materials
+	headers    map[string]string
+	ServerSide encrypt.ServerSide
```
harshavardhana: @abperiasamy suggested we should make this more explicit i.e not call it a generic name "ServerSide" instead call it "ServerSideEncryption"

pkg/encrypt/server-side.go
```diff
@@ -0,0 +1,146 @@
...
+
+	// Marshal adds encryption headers to the provided HTTP headers.
+	// It marks an HTTP request as server-side-encryption request
+	// and inserts the required data into the headers.
+	Marshal(h http.Header)
+}
+
+// NewS3 returns a server-side-encryption using S3 storage encryption.
+// Using SSE-S3 the server will encrypt the object with server-managed keys.
+func NewS3() ServerSide { return s3{} }
```
harshavardhana: @abperiasamy suggested that we should keep this `encrypt.NewSSE()` such that it complies with the S3 API docs.  Since NewS3() would mean like a client constructor.

pkg/encrypt/server-side.go
```diff
@@ -0,0 +1,146 @@
...
+	return sse, nil
+}
+
+// ServerSideCopy transforms a SSE-C encryption into a SSE-C copy
+// encryption. This is required for SSE-C key rotation or a SSE-C
+// copy where the source and the destination should be encrypted.
+//
+// If the provided sse is no SSE-C encryption ServerSideCopy returns
+// sse unmodified.
+func ServerSideCopy(sse ServerSide) ServerSide {
```
harshavardhana: perhaps we should bring encryption keyword here. ```encrypt.SSECopy()```

api-get-options.go
aead: Also fine - will change

pkg/encrypt/server-side.go
aead: Are you suggesting `s/ServerSideCopy/SSECopy` ? If so I think it is okay. However currently
a user would call it: `encrypt.ServerSideCopy(...)` So I code would explicitly mention "encrypted copy"

harshavardhana: but we can stick to "SSE" because it mentions the protocol if we bring in client-side encryption if at all in future.

aead: Okay

PR #947, opened 2018-03-13
Add travis-ci badge

halfcrazy:


deekoder:
@halfcrazy  I am not for this change at the moment because we load these md files on our docs site and the badges are slowing down load times on our docs site. We are working on selectively filtering out some things we stream directly to our docs site so that this is not an issue.

harshavardhana:
> @halfcrazy I am not for this change at the moment because we load these md files on our docs site and the badges are slowing download times on our docs site. We are working on selectively filtering out some things we stream directly to our docs site so that this is not an issue.

I concur with @deekoder here. NACK from my end.

PR #948, opened 2018-03-13
Add putObject/getObject() client side encryption examples

### LINKED ISSUE #945: Issues with high memory consuption when running concurrent go routines and minio s3Client.PutEncryptedObject
I have a program that runs (5) concurrent go routines that implement the s3Client.PutEncryptedObject function.

I have over 2000 files to upload and have noticed a large memory consumption of around 2GB for the length of the programs life.

I have analyzed the program while running using pprof and the output is in this folder: https://github.com/wilyarti/hashtree/tree/master/debug

The problem seems to be: github.com/minio/minio-go.Client.putObjectMultipartStreamNoLength consuming over 1200MB of RAM.

There are obviously inefficiencies elsewhere in my program but this part is consuming %96 of the ram.

The other interesting detail is high cpu usage:
 0 0% 96.41% 4.16s 87.95% github.com/minio/minio-go.Client.putObjectMultipartStreamNoLength

Please refer to https://github.com/wilyarti/hashtree/blob/master/debug/analysis.txt for CPU usage.

Please let me know if I can provide more details.

harshavardhana:
Fixes #945

### DIFF
--- a/examples/s3/getobject-client-encryption.go
+++ b/examples/s3/getobject-client-encryption.go
@@ -1,0 +1,106 @@
+// +build ignore
+
+/*
+ * Minio Go Library for Amazon S3 Compatible Cloud Storage
+ * Copyright 2018 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package main
+
+import (
+	"errors"
+	"io"
+	"log"
+	"os"
+	"path"
+
+	"github.com/minio/minio-go"
+	"github.com/minio/sio"
+	"golang.org/x/crypto/argon2"
+)
+
+const (
+	// SSE DARE package block size.
+	sseDAREPackageBlockSize = 64 * 1024 // 64KiB bytes
+
+	// SSE DARE package meta padding bytes.
+	sseDAREPackageMetaSize = 32 // 32 bytes
+)
+
+func decryptedSize(encryptedSize int64) (int64, error) {
+	if encryptedSize == 0 {
+		return encryptedSize, nil
+	}
+	size := (encryptedSize / (sseDAREPackageBlockSize + sseDAREPackageMetaSize)) * sseDAREPackageBlockSize
+	if mod := encryptedSize % (sseDAREPackageBlockSize + sseDAREPackageMetaSize); mod > 0 {
+		if mod < sseDAREPackageMetaSize+1 {
+			return -1, errors.New("object is tampered")
+		}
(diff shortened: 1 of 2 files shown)


deekoder:
but encryption is still server side here?

examples/s3/putobject-client-encryption.go
```diff
@@ -0,0 +1,116 @@
...
+	"log"
+	"os"
+
+	"github.com/minio/minio-go"
+	"github.com/minio/sio"
+	"golang.org/x/crypto/hkdf"
+)
+
+const (
+	// SSE dare package block size.
```
poornas: dare -> DARE

examples/s3/putobject-client-encryption.go
```diff
@@ -0,0 +1,116 @@
...
+	sseDAREPackageBlockSize = 64 * 1024 // 64KiB bytes
+
+	// SSE dare package meta padding bytes.
```
poornas: dare -> DARE

aead:
@deekoder No, the encrypted reader returned from `sio.Encrypt(....)` encrypts everything. So encryption happens on client side. 

@harshavardhana Should we use a password-based key derivation (maybe easier for users and if the use secret keys it would still be perfectly fine)?

examples/s3/getobject-client-encryption.go
```diff
@@ -0,0 +1,105 @@
...
+	localFile, err := os.Create("my-testfile")
+	if err != nil {
+		log.Fatalln(err)
+	}
+	defer localFile.Close()
+
+	password := "myfavoritepassword"                    // Change as per your needs.
+	salt := path.Join("my-bucketname", "my-objectname") // Change as per your needs.
+	decrypted, err := sio.DecryptReader(obj, sio.Config{
+		Key: argon2.IDKey(password, salt, 1, sseDAREPackageBlockSize, 4, sseDAREPackageMetaSize),
```
aead: You can use `encrypt.DefaultPBKDF` here.

poornas: in both examples the argon2.IDKey method needs to have password and salt converted to byte array
```
encrypted, err := sio.EncryptReader(object, sio.Config{
		Key: argon2.IDKey([]byte(password), []byte(salt), 1, sseDAREPackageBlockSize, 4, sseDAREPackageMetaSize),
	})
decrypted, err := sio.DecryptReader(obj, sio.Config{
		Key: argon2.IDKey([]byte(password), []byte(salt), 1, sseDAREPackageBlockSize, 4, sseDAREPackageMetaSize),
	})
```

harshavardhana: > You can use encrypt.DefaultPBKDF here.

I don't want to use code from server-side.go which has special methods attached to it - which will confuse the user @aead - here PBKDF is all about client side and how they can leverage it using argon2.

harshavardhana: updated @poornas

aead: @harshavardhana Okay - than one last thing. The parameters for argon2 (`sseDAREPackageBlockSize`, `sseDAREPackageMetaSize`) look like they are somehow tied to DARE / sio - but they are not. 64 MB is a (recommended) value from the RFC draft and the 32 is the key size. I see you are using constants here instead of literals but IMO the constants give a wrong impression of the meaning of these values. Maybe we can use literals or properly named constants. (I didn't mentioned it before because of `PBKDF` suggestion)

PR #949, opened 2018-03-14
Fix xml parsing error for RemoveObjects API

poornas:
Fixes https://github.com/minio/mc/issues/2409

Also fixed the RemoveObjects example.


api-remove.go
```diff
@@ -195,6 +195,13 @@ func (c Client) RemoveObjectsWithContext(ctx context.Context, bucketName string,
 				contentMD5Base64: sumMD5Base64(removeBytes),
 				contentSHA256Hex: sum256Hex(removeBytes),
 			})
+			if resp != nil {
```
krisis: Delete Multiple Objects REST API returns 200 OK even when request parameters (headers and query params included) are invalid. See `Example 3: Malformed XML in the Request` in https://docs.aws.amazon.com/AmazonS3/latest/API/multiobjectdeleteapi.html.

We have to fix minio server too to send 200 OK with error message as XML body, but that will be a different PR. So, in this change we need to handle server's response as documented above.
(N B this is something I found (only) when I was reviewing this change.)

poornas: @krisis , server already sends error message as xml body along with 200 OK. The httpRespToErrorResponse method decodes the error response from the xml body.

krisis:
@poornas see below the HTTP trace gotten by running a modified version of `examples/s3/removeobjects.go`
on minio server built from https://github.com/minio/minio/pull/5602/ (WORM enabled). We can see that the status code for a failed DeleteMultipleObjects API call is 405 instead of 200. This PR also interprets status code other than 200 as error, which is different from what the S3 documentation describes. 
```
---------START-HTTP---------
POST /test/?delete= HTTP/1.1
Host: localhost:9000
User-Agent: Minio (linux; amd64) minio-go/5.0.1
Content-Length: 68
Authorization: AWS4-HMAC-SHA256 Credential=minio/20180315/us-east-1/s3/aws4_request, SignedHeaders=content-md5;host;x-amz-content-sha256;x-amz-date, Signature=**REDACTED**
Content-Md5: OqWHHC/iJd38l8IRNjCBjg==
X-Amz-Content-Sha256: 9bb418c901fe5b56db2c1d01f23d6aeda4a3fd46d295092dea26d0c4190c5cc1
X-Amz-Date: 20180315T025817Z
Accept-Encoding: gzip

HTTP/1.1 405 Method Not Allowed
Transfer-Encoding: chunked
Accept-Ranges: bytes
Content-Type: application/xml
Date: Thu, 15 Mar 2018 02:58:17 GMT
Server: Minio/DEVELOPMENT.GOGET (linux; amd64)
Vary: Origin
X-Amz-Request-Id: 151BF8A77C1C5AA0

111
<Error><Code>MethodNotAllowed</Code><Message>The specified method is not allowed against this resource.</Message><Key></Key><BucketName></BucketName><Resource>/test/</Resource><RequestId>3L137</RequestId><HostId>3L137</HostId></Error>
0
---------END-HTTP---------
2018/03/15 08:28:17 Failed to remove , error: The specified method is not allowed against this resource.

```

api-remove.go
balamurugana: It is wrong to blindly decoding response body without checking HTTP status code.  I believe multiple delete response XML is sent only on status code = 200.

poornas: @krisis, my bad - server actually sends a 404 The specified key does not exist.

api-remove.go
```diff
@@ -195,6 +195,13 @@ func (c Client) RemoveObjectsWithContext(ctx context.Context, bucketName string,
...
+				if resp.StatusCode != http.StatusOK {
+					e := httpRespToErrorResponse(resp, bucketName, "")
+					errorCh <- RemoveObjectError{ObjectName: "", Err: e}
```
vadmeste: I think we should not send one error and break here.. We need to send the same error multiple times and also continue to the next batch as done in the next code block (if err != nil)

Why ? because some of our users can have their code blocked since here we decided to stop reading from objectsCh channel.

poornas: ok, good idea. Updated PR.

poornas:
rebased PR to latest master and removed fix to removeobjects.go example as it has already been fixed by another PR.

PR #955, opened 2018-03-19
Add missing doneCh in the example for removeobjects

oivoodoo:


harshavardhana:
We do not need a doneCh this code to meant to drain out all objects. Once there are no more objects the objectsCh is closed automatically.

oivoodoo:
Its producing the issue in the code because of the usage of doneCh var.

harshavardhana:
Also this same code is being fixed by another PR #949

harshavardhana:
Sorry this PR accidentally got closed...

PR #959, opened 2018-03-23
Update get/setBucketPolicy methods to use files instead of pkg/policy

### LINKED ISSUE #943: set/getBucketPolicy should take/return policy JSON file
Currently the set/getBucketPolicy APIs are high level providing fine grained access. It was decided to make set/getBucketPolicy a low level API directly dealing with the policy json file to make the client libs simpler.

Changes similar to minio/minio-js#676 is needed.

policy package needs to be moved to `mc` so that `mc` can provide high level interface to the user.

nitisht:
Also removed `ListBucketPolicies` API as we move to json file based
policy handling.

Fixes #943


harshavardhana:
Update `api.go` to update the API version we need to move to 6.0.0 here.

poornas:
@nitisht, second @krishnasrinivas suggestion on using policy as a string or passed in as a file.

nitisht:
updated the PR @harshavardhana @poornas @krishnasrinivas PTAL

nitisht:
@krishnasrinivas could you pls take a look again

PR #960, opened 2018-03-27
Fix stream SSE uploads with S3 encrypt type

jespino:
When you try to upload files stream based with encryption activated, minio-go
try to upload the parts using the ServerSideEncryption header, but it isn't
supported for uploadPart with encrypt type S3, only is needed for customer
keys. For S3 encryption we simply add the header at the creation, and in the
following uploadPart we don't pass the header.

I'm not sure, but maybe we need to exclude the encrypt KMS type too (I'm not
sure).

Here is an example of it failing before the patch, and working after it.

```go
package main

import (
	"log"
	"os"

)

func main() {
	s3Client, err := minio.New("s3.amazonaws.com", "API_KEY", "API_SECRET", true)
	if err != nil {
		log.Fatalln(err)
	}

	object, err := os.Open("test.txt")
	if err != nil {
		log.Fatalln(err)
	}
	defer object.Close()

	options := minio.PutObjectOptions{ServerSideEncryption: encrypt.NewSSE()}

	n, err := s3Client.PutObject("my-bucket", "my-objectname", object, -1, options)
	if err != nil {
		log.Fatalln(err)
	}

	log.Println("Uploaded", "my-objectname", " of size: ", n, "Successfully.")
}
```

### DIFF
--- a/pkg/encrypt/server-side.go
+++ b/pkg/encrypt/server-side.go
@@ -30,6 +30,11 @@ const (
 	// sseGenericHeader is the AWS SSE header used for SSE-S3 and SSE-KMS.
 	sseGenericHeader = "X-Amz-Server-Side-Encryption"
 
+	// sseKmsKeyID is the AWS SSE-KMS key id.
+	sseKmsKeyID = sseGenericHeader + "-Aws-Kms-Key-Id"
+	// sseEncryptionContext is the AWS SSE-KMS Encryption Context data.
+	sseEncryptionContext = sseGenericHeader + "-Encryption-Context"
+
 	// sseCustomerAlgorithm is the AWS SSE-C algorithm HTTP header key.
 	sseCustomerAlgorithm = sseGenericHeader + "-Customer-Algorithm"
 	// sseCustomerKey is the AWS SSE-C encryption key HTTP header key.
@@ -90,6 +95,14 @@ type ServerSide interface {
 // Using SSE-S3 the server will encrypt the object with server-managed keys.
 func NewSSE() ServerSide { return s3{} }
 
+// NewSSEKMS returns a new server-side-encryption using SSE-KMS and the provided Key Id.
+func NewSSEKMS(keyID string) ServerSide { return kms{key: keyID, context: nil} }
+
+// NewSSEKMSWithContext returns a new server-side-encryption using SSE-KMS and the provided Key Id and context.
+func NewSSEKMSWithContext(keyID string, context *string) ServerSide {
+	return kms{key: keyID, context: context}
+}
+
 // NewSSEC returns a new server-side-encryption using SSE-C and the provided key.
 // The key must be 32 bytes long.
 func NewSSEC(key []byte) (ServerSide, error) {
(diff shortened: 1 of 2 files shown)


aead:
@jespino I think you're also right about [KMS](https://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadUploadPart.html) - but we should check to make sure...

jespino:
You are welcome, I'm implementing the encrypt.NewSSEKMS to try it ;) If I realize that is needed for KMS too, I will send another PR, if you want :)

aead:
That would be awesome :) @jespino

jespino:
I have created what I think is needed, but I don't know too much about KMS, so I don't know if the context as an optional string is enough.

pkg/encrypt/server-side.go
```diff
@@ -90,6 +95,14 @@ type ServerSide interface {
 // Using SSE-S3 the server will encrypt the object with server-managed keys.
 func NewSSE() ServerSide { return s3{} }
 
+// NewSSEKMS returns a new server-side-encryption using SSE-KMS and the provided Key Id.
+func NewSSEKMS(keyID string) ServerSide { return kms{key: keyID, context: nil} }
+
+// NewSSEKMSWithContext returns a new server-side-encryption using SSE-KMS and the provided Key Id and context.
+func NewSSEKMSWithContext(keyID string, context *string) ServerSide {
```
aead: Since::
>The encryption context can be any value that you want, provided that the header adheres to the Base64-encoded JSON format. 

I think we can follow this approach:
```
NewSSEKMS(keyID string, context interface{}) (ServerSide, error) {
      // if context is nil -> no context
      // if context implements json.Marshaler -> marshal
      // json.Marshal(context) 
}
```

pkg/encrypt/server-side.go
```diff
@@ -90,6 +95,14 @@ type ServerSide interface {
 // Using SSE-S3 the server will encrypt the object with server-managed keys.
 func NewSSE() ServerSide { return s3{} }
 
+// NewSSEKMS returns a new server-side-encryption using SSE-KMS and the provided Key Id.
+func NewSSEKMS(keyID string) ServerSide { return kms{key: keyID, context: nil} }
```
aead: Maybe we don't need two functions right now - removing a function breaks API while we can always add if requested.

pkg/encrypt/server-side.go
```diff
@@ -144,3 +157,18 @@ type s3 struct{}
 func (s s3) Type() Type { return S3 }
 
 func (s s3) Marshal(h http.Header) { h.Set(sseGenericHeader, "AES256") }
+
+type kms struct {
+	key     string
+	context *string
```
aead: Can we use a bool var - like `contextPresent` instead of a `*string` and nil checks?

pkg/encrypt/server-side.go
jespino: I like this approach and then we don't need the `contextPresent` proposed in the comment below. I'll do this changes tomorrow.

pkg/encrypt/server-side.go
jespino: Ok, perfect, I'll leave one only.

PR #962, opened 2018-03-30
avoid unnecessary stat call during single copy

### LINKED ISSUE #961: [SSE]: CopyObject fails for encrypted-to-unencrypted object
## What I did

```
package main

import (
	"crypto/tls"
	"log"
	"net/http"

	"github.com/minio/minio-go"
)

func main() {
	accessKey, secretKey := "", ""
	minioClient, err := minio.New("s3.amazonaws.com", accessKey, secretKey, true)
	if err != nil {
		log.Fatalln(err)
	}

	minioClient.SetCustomTransport(&http.Transport{
		TLSClientConfig: &tls.Config{
		  InsecureSkipVerify: true,
		},
	})

	bucket, object, file := "bucket", "object", "your-file"
	options := minio.PutObjectOptions{
		ServerSideEncryption: encrypt.DefaultPBKDF([]byte("123456"), []byte(bucket+object)),
	}
	if _, err = minioClient.FPutObject(bucket, object, file, options); err != nil {
		log.Fatalln(err)
	}

	src := minio.NewSourceInfo(bucket, object, encrypt.SSECopy(encrypt.DefaultPBKDF([]byte("123456"), []byte(bucket+object))))
	dst, _ := minio.NewDestinationInfo(bucket, object+"-copy", nil, nil)
	if err = minioClient.CopyObject(dst, src); err != nil {
		log.Fatalln(err)
	}
	log.Println("OK")
}
```

### What did you expect to see

Expected `OK`

### What did you see instead

Could not stat object - bucket/object: 400 Bad Request

### Solution

The problem is that `ComposeObject` (which is called by `CopyObject`) tries a Stat per source-info to do some checks and get the size of all source objects. The Stat fails because only the SSE-C copy headers are provided. The solution is to make `CopyObject` a separate call.
To be precise this issue affects also `ComposeObject` for multiple source objects. However I don't see how this can be fixed without changing other composing behavior.

aead:
This change separates the copy logic from the compose logic.
This allows users to specify SSE-C copy encryption directly.
Additionally it saves a HEAD request. 

Fixes #961

### DIFF
--- a/functional_tests.go
+++ b/functional_tests.go
@@ -5644,6 +5644,60 @@ func testEncryptedCopyObjectV2() {
 	testEncryptedCopyObjectWrapper(c)
 }
 
+func testDecryptedCopyObject() {
+	// initialize logging params
+	startTime := time.Now()
+	testName := getFuncName()
+	function := "CopyObject(destination, source)"
+	args := map[string]interface{}{}
+
+	// Instantiate new minio client object
+	c, err := minio.NewV2(
+		os.Getenv(serverEndpoint),
+		os.Getenv(accessKey),
+		os.Getenv(secretKey),
+		mustParseBool(os.Getenv(enableHTTPS)),
+	)
+	if err != nil {
+		logError(testName, function, args, startTime, "", "Minio v2 client object creation failed", err)
+		return
+	}
+
+	bucketName, objectName := randString(60, rand.NewSource(time.Now().UnixNano()), "minio-go-test-"), "object"
+	if err = c.MakeBucket(bucketName, "us-east-1"); err != nil {
+		logError(testName, function, args, startTime, "", "MakeBucket failed", err)
+		return
+	}
+
+	encryption := encrypt.DefaultPBKDF([]byte("correct horse battery staple"), []byte(bucketName+objectName))
+	_, err = c.PutObject(bucketName, objectName, bytes.NewReader(bytes.Repeat([]byte("a"), 1024*1024)), 1024*1024, minio.PutObjectOptions{
+		ServerSideEncryption: encryption,
+	})
+	if err != nil {
+		logError(testName, function, args, startTime, "", "PutObject call failed", err)
+		return
+	}
+
+	src := minio.NewSourceInfo(bucketName, objectName, encrypt.SSECopy(encryption))
+	args["source"] = src
+	dst, err := minio.NewDestinationInfo(bucketName, "decrypted-"+objectName, nil, nil)
+	if err != nil {
+		logError(testName, function, args, startTime, "", "NewDestinationInfo failed", err)
+		return
+	}
+	args["destination"] = dst
+
+	if err = c.CopyObject(dst, src); err != nil {
+		logError(testName, function, args, startTime, "", "CopyObject failed", err)
+		return
+	}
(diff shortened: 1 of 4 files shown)


pkg/encrypt/server-side.go
```diff
@@ -101,6 +101,21 @@ func NewSSEC(key []byte) (ServerSide, error) {
 	return sse, nil
 }
 
+// SSE transforms a SSE-C copy encryption into a SSE-C encryption.
+// It is the inverse of SSECopy(...).
+//
+// If the provided sse is no SSE-C copy encryption SSE returns
+// sse unmodified.
+func SSE(sse ServerSide) ServerSide {
```
harshavardhana: SSE() should be called as SSEC() here i think, since it might be confusing when we implement NewSSE()

aead: `NewSSE()` is already [there](https://github.com/minio/minio-go/blob/master/pkg/encrypt/server-side.go#L97) - as well as the `NewSSEC()` [function](https://github.com/minio/minio-go/blob/master/pkg/encrypt/server-side.go#L113)

Semantically the current solution is sound:
`encrypt.SSE(encrypt.NewSSE())` is equivalent to `encrypt.NewSSE()`
`encrypt.SSE(encrypt.DefaultPBKDF(...))` is equivalent to `encrypt.DefaultPBKDF(...)`
`encrypt.SSE(encrypt.SSECopy(encrypt.NewSSE()))` is equivalent to `encrypt.NewSSE()`
`encrypt.SSE(nil)` is equivalent to `nil`
a.s.o.

`SSE()` and `SSECopy()` are "generic" functions. Specifying `copy` only for SSE-C may be correct with respect to functionality (only SSE-C distinguish copy / non-copy headers) but probably leads  to more complexity for the client because he/she has to understand all that.
Right now `encrypt.SSE(encrypt.SSE(encrypt.SSECopy(encrypt.NewSSE())))` boils down to `encrypt.SSE()` semantically. So I don't see how clients can get confused - except if they want to do advanced things **and** don't read the function doc.

aead: So you're saying a user could confuse the function call `encrypt.SSE(..)` with a type conversion:
`encrypt.SSE(...)` like `[]byte(...)`? (A struct initialization would require `{` brackets)
 
Semantically it's exactly that - but as a function. Another (maybe more explicit) version would
be `encrypt.ToSSE(...)` - however the API would look a bit odd with `SSECopy()` and `ToSSE()`.
Or we break compatibility and add to functions to the interface to convert back and forth.
(There should be no other implementations because there is only SSE-S3, SSE-C and SSE-KMS)

harshavardhana: okay fair enough I think its fine...

functional_tests.go
```diff
@@ -5644,6 +5644,60 @@ func testEncryptedCopyObjectV2() {
 	testEncryptedCopyObjectWrapper(c)
 }
 
+func testDecryptedCopyObject() {
```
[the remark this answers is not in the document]
aead: Oh, was just copy-past mistake. Changed to `New` - IMO separate tests for different signature types are not needed...

PR #967, opened 2018-04-18
Support specifying redirect location for a object

klausenbusk:
This commit make it possiable to specify the
x-amz-website-redirect-location header for a object, which make it
possible for a object to redirect to another object in the same
bucket or to a external URL. [1]

[1] https://docs.aws.amazon.com/AmazonS3/latest/dev/how-to-page-redirect.html#page-redirect-using-rest-api

### DIFF
--- a/api-put-object.go
+++ b/api-put-object.go
@@ -33,16 +33,17 @@ import (
 
 // PutObjectOptions represents options specified by user for PutObject call
 type PutObjectOptions struct {
-	UserMetadata         map[string]string
-	Progress             io.Reader
-	ContentType          string
-	ContentEncoding      string
-	ContentDisposition   string
-	ContentLanguage      string
-	CacheControl         string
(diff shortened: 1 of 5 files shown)


api-put-object.go
```diff
@@ -84,6 +85,9 @@ func (opts PutObjectOptions) Header() (header http.Header) {
 	if opts.StorageClass != "" {
 		header[amzStorageClass] = []string{opts.StorageClass}
 	}
+	if opts.WebsiteRedirectLocation != "" {
+		header["x-amz-website-redirect-location"] = []string{opts.WebsiteRedirectLocation}
```
harshavardhana: you should define this as a constant string just like `amzStorageClass`

klausenbusk: Done

harshavardhana:
ping @nitisht

PR #974, opened 2018-04-25
fix: remove wrong policy comparison in functional tests

balamurugana:
Fixes #5852


harshavardhana:
This is fixed already in another PR, we can close this and make a release.

deekoder:
I will let @balamurugana  close instead of me doing it. Reopening hence.

harshavardhana:
@balamurugana is going to re-do this by removing some checks @deekoder which are not necessary for bucket policy verification.

PR #982, opened 2018-05-07
Fix build for httplex package merged into httpguts

dschulz:
lex/httplex is no more, it was merged into http/httpguts

Please see https://go-review.googlesource.com/c/net/+/111875


dschulz:
Already reported. Closing.

PR #986, opened 2018-05-28
fix version naming convention to have 'v' prefix

### LINKED ISSUE #985: release tag format, compatibility with the Go versioning proposal
Hey,

when testing if restic compiles with `vgo`, the prototype implementation for adding [versioning to the Go toolchain](https://github.com/golang/go/issues/24301), I noticed that some time ago you switched from release tags in the form `v1.2.3` (as required by `vgo`) to `1.2.3`, so the `v` prefix was dropped.

This leads to `vgo` selecting the `v1.0.0` release tag, which is ancient:
```
vgo: import "github.com/restic/restic/cmd/restic" ->
        import "github.com/restic/restic/internal/backend/location" ->
        import "github.com/restic/restic/internal/backend/s3" ->
        import "github.com/minio/minio-go/pkg/credentials" [/home/fd0/gopath/src/v/github.com/minio/minio-go@v1.0.0/pkg/credentials]: open /home/fd0/gopath/src/v/github.com/minio/minio-go@v1.0.0/pkg/credentials: no such file or directory
```

Would you mind adding a second tag `v6.0.1` which points to the same commit the tag `6.0.1` does, and tagging future releases with the `v` prefix? That'd be awesome :)

Thanks!

harshavardhana:
This change is needed to follow `vgo` guidelines
put forth by Go community.

https://github.com/golang/vgo/blob/master/vendor/cmd/go/internal/semver/semver.go#L6

> // semantic version strings must begin with a leading "v"

Supported format style

https://github.com/golang/vgo/blob/master/vendor/cmd/go/internal/semver/semver.go#L11

> //	vMAJOR[.MINOR[.PATCH[-PRERELEASE][+BUILD]]]

Fixes #985

### DIFF
--- a/api.go
+++ b/api.go
@@ -99,7 +99,7 @@ type Options struct {
 // Global constants.
 const (
 	libraryName    = "minio-go"
-	libraryVersion = "6.0.2"
+	libraryVersion = "v6.0.2"
 )
 
 // User Agent should always following the below style.


fd0:
Sorry for jumping in here, this change seems to be not related to what I suggested in #985: It was just about the tags in the git repository for minio-go, so from my point of view (and Go's) the string in the library does not matter at all. The version is still 6.0.1, just the version tag (in git) should be `v6.0.1`... Not sure what to do with this, at least you know now ;)

harshavardhana:
Now the tag is v6.0.2 we won't be changing the old tags is that needed as well ?

fd0:
ah, I didn't see that earlier, that's sufficient. Thanks!

PR #990, opened 2018-06-12
Added support for the start-after parameter when performing a List Objects V2 query

wd256:
When making a List Objects V2 query against Amazon S3, an object key cannot be used as a continuation-token.  Amazon S3 continuation tokens are opaque and attempting to use an object key as a continuation-token will result in an error response.

If you want to start listing objects after a particular object key, the start-after parameter needs to be used.  This pull request adds support for setting the start-after parameter when making a List Objects V2 request.

Relates to minio/minio#6032

### DIFF
--- a/api-list.go
+++ b/api-list.go
@@ -118,7 +118,7 @@ func (c Client) ListObjectsV2(bucketName, objectPrefix string, recursive bool, d
 		var continuationToken string
 		for {
 			// Get list of objects a maximum of 1000 per request.
-			result, err := c.listObjectsV2Query(bucketName, objectPrefix, continuationToken, fetchOwner, delimiter, 1000)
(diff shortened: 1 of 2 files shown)


api-list.go
```diff
@@ -171,11 +171,12 @@ func (c Client) ListObjectsV2(bucketName, objectPrefix string, recursive bool, d
 // You can use the request parameters as selection criteria to return a subset of the objects in a bucket.
 // request parameters :-
 // ---------
-// ?continuation-token - Specifies the key to start with when listing objects in a bucket.
+// ?continuation-token - Used to continue iterating over a set of objects
```
harshavardhana: can you add a comment here what  `?start-after` does?

wd256: ?start-after gets a mention on line 178

PR #995, opened 2018-06-19
Fix functional tests for Gateway behaviour

nitisht:
- Ignore Storage Class response in Listobjects as gateways may
not return expected storage class

- Use correct Etag to set Stat conditions

See https://github.com/minio/minio/issues/6051

### DIFF
--- a/functional_tests.go
+++ b/functional_tests.go
@@ -2623,8 +2623,14 @@ func testCopyObject() {
 		return
 	}
 
+	oi, err := c.StatObject(bucketName, objectName, minio.StatObjectOptions{})
+	if err != nil {
+		logError(testName, function, args, startTime, "", "StatObject failed", err)
+		return
+	}
+
 	stOpts := minio.StatObjectOptions{}
-	stOpts.SetMatchETag(objInfo.ETag)
+	stOpts.SetMatchETag(oi.ETag)
 	objInfo, err = c.StatObject(bucketName, objectName, stOpts)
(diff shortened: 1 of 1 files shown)


functional_tests.go
```diff
@@ -2623,6 +2623,22 @@ func testCopyObject() {
...
+	objInfo, err = r.Stat()
+	if err != nil {
+		logError(testName, function, args, startTime, "", "Stat failed", err)
+		return
+	}
+
+	r.Close()
+
 	stOpts := minio.StatObjectOptions{}
 	stOpts.SetMatchETag(objInfo.ETag)
```
harshavardhana: If this is required why is it working fine with server mode?

nitisht: Looks like [CopyObject to update metadata](https://github.com/minio/minio-go/blob/master/functional_tests.go#L2620) updates the Etag for Azure , while for Minio Etag doesn't change

harshavardhana: I see

functional_tests.go
```diff
@@ -2623,6 +2623,22 @@ func testCopyObject() {
 		return
 	}
 
+	// Source object
+	r, err = c.GetObject(bucketName, objectName, minio.GetObjectOptions{})
```
kannappanr: Can you do a `StatObject` instead?

nitisht: yes, updated the PR

PR #998, opened 2018-06-21
Use GCS compatible Content-Language value in functional tests

wd256:
The `testPutObjectWithContentLanguage` functional test currently fails when run against Minio in GCS gateway mode.

[This is due to GCS only allowing a specific set of values to be used for an objects Content-Language](https://cloud.google.com/storage/docs/metadata#content-language).

This pull request updates the `testPutObjectWithContentLanguage` test to use a Content-Language which is compatible with S3, GCS and Azure.


nitisht:
Thanks for the PR @wd256 LGTM

PR #1002, opened 2018-06-29
pkg/policy: detect single statement policies

harshavardhana:
Some times AWS S3 can have single statement policies
where all actions and resources are combined into a
a single statement. In general we handle multiple
statements properly, handle single statement with
bucket and object actions in the same statement
properly.


vadmeste:
@harshavardhana shouldn't honor policy Deny rules in the future ? Deny overrides Allow rule

PR #1003, opened 2018-07-01
Reduce memory consumption for streamed object: https://github.com/min…

wilyarti:
…io/minio-go/issues/989

Non-compatible fix resulting in project being forked.

Hey I don't really expect you to accept this merge, but I just wanted to show what I did to reduce memory usage.

I pass the size of the uncompressed file as a negative number, this is detected and parts are calculated based on the absolute value of this number. The file is compressed and then uploaded as a stream with no length.

This allows for adequate part sizes and results in dramatically reduced memory for streamed objects. On average the memory usage went down for 3 concurrent uploads from 1.9GiB to 200MiB.

There is only my usage case so I don't think it is worth modifying your library to suit my needs. I am willing to maintain my fork of your project to keep this feature for myself.

Thanks for the great software :)


harshavardhana:
This change is not needed as this is achieved by without doing the negative number usage. If you can provide size, Why can't you provide this number as a positive value ?

wilyarti:
The size provided as a negative number is an estimate as it is the uncompressed size. If I provide the uncompressed size the upload hangs waiting for more data. Or truncates the file if it is larger.

PR #1004, opened 2018-07-06
Add support for bucket lifecycle policies

mzdeb:
Hi, 

I've implemented basic support for bucket lifecycle policies. I know minio server currently does not support [this features](https://docs.aws.amazon.com/AmazonS3/latest/dev/manage-lifecycle-using-rest.html) however it might be useful when https://github.com/minio/minio/issues/5139 will be resolved.


mzdeb:
Thanks for review! I've added examples.

examples/s3/setbucketpolicy.go
```diff
@@ -48,4 +48,11 @@ func main() {
 	if err != nil {
 		log.Fatalln(err)
 	}
+
```
harshavardhana: Create a new file we don't need to piggyback on bucket policies example.  Create an example `examples/s3/setbucketlifecycle.go`

mzdeb: I've already done it. Sorry for confusion, please see latest commit 9c269de

vadmeste:
Great PR @mzdeb, I am just wondering why you are using policy terminology with bucket lifecycle ? I cannot find policy word here https://docs.aws.amazon.com/AmazonS3/latest/dev/object-lifecycle-mgmt.html

mzdeb:
@vadmeste True, I've no idea why I used "policy" term here. I removed from lifecycle methods "policy" word. :)

I'm not sure if file rename api-get-policy.go → api-get-bucket.go is acceptable?

harshavardhana:
> I'm not sure if file rename api-get-policy.go → api-get-bucket.go is acceptable?

you can call it `api-get-lifecycle`

mzdeb:
@harshavardhana done :) I've put those methods in ```api-get-lifecycle.go```

mzdeb:
@vadmeste is all ok now with this PR? :)

moonbamboo2:
I had an error: "A header you provided implies functionality that is not implemented" with local minio, not Amazon S3.
[testLifecycle.txt](https://github.com/minio/minio-go/files/2249443/testLifecycle.txt)

vadmeste:
@moonbamboo2 Minio server doesn't support bucket lifecycle yet.. but it will be soon https://github.com/minio/minio/issues/5139

PR #1008, opened 2018-07-14
Add check for env var

ektar:
Added check for documented env var MINIO_SHARED_CREDENTIALS_FILE


ektar:
@harshavardhana - I think formatting issues should be fixed, can you let me know if anything else is needed?

PR #1010, opened 2018-07-18
New way to update ObjectACL or Header or Meta data

pierre-emmanuelJ:
If you Put an object with nothing (no acl, meta...) and you want to update with ObjectACL it's not possible. And you don't want to use bucket policies to make it alternatively

this is an example to update a canned acl:
```Golang
src := minio.NewSourceInfo("bucketName", "object", nil)

meta := map[string]string{
	"x-amz-acl": "public-read-write",
}

// Destination object
dst, err := minio.NewDestinationInfo("bucketName", "object", nil, meta)
if err != nil {
	return err
}

// Copy object call
err = minioClient.CopyObject(dst, src)
if err != nil {
	return err
}
```

this source code copy object on itself and can update ACL, userMetaData or Header

### DIFF
--- a/api-compose-object_test.go
+++ b/api-compose-object_test.go
@@ -127,6 +127,7 @@ func TestGetUserMetaHeadersMap(t *testing.T) {
 		"x-amz-acl":           "public-read-write",
 		"content-type":        "application/binary",
 		"X-Amz-Storage-Class": "rrs",
+		"x-amz-grant-write":   "test@exo.ch",
 	}
 
 	destInfo := &DestinationInfo{"bucket", "object", nil, userMetadata}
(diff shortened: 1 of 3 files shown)


pierre-emmanuelJ:
Yes I will add a test and why not an example

pierre-emmanuelJ:
Need an example ?

api-compose-object_test.go
```diff
@@ -118,3 +119,42 @@ func TestCalculateEvenSplits(t *testing.T) {
...
+	i := 0
+
+	if _, ok := r["x-amz-metadata-directive"]; !ok {
+		t.Errorf("Test %d - metadata directive was expected but is missing", i)
+		i++
+	}
+
+	for k := range r {
+
+		println(k)
```
greut: :duck:

PR #1011, opened 2018-07-20
Add new method GetObjectACL()

pierre-emmanuelJ:
Example:

```Golang
objInfo, err := minioClient.GetObjectACL("bucketName", "objectName")
if err != nil {
	return err
}

//object ACLs is in the Metadata header with stat details
fmt.Printf("%#v\n", "objInfo.Metadata")
```

### DIFF
--- a/api-get-object-acl.go
+++ b/api-get-object-acl.go
@@ -1,3 +1,20 @@
+/*
+ * Minio Go Library for Amazon S3 Compatible Cloud Storage
+ * Copyright 2017 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


api-get-object-acl.go
```diff
@@ -0,0 +1,138 @@
+/*
+ * Minio Go Library for Amazon S3 Compatible Cloud Storage
+ * Copyright 2017 Minio, Inc.
```
kannappanr: The year has to be 2018

pierre-emmanuelJ:
@nitisht Do you want an example [here](https://github.com/minio/minio-go/tree/master/examples/s3) ?

nitisht:
> @nitisht Do you want an example here ?

Yes, please. That'd be great

PR #1013, opened 2018-07-27
[WIP]Adds Client API for SelectObjectContentOperation

Colhodm:
Current Issues/TODO:
Add Context Cleaning to the .Close() method
Change Bytes.Buffer so that it is more resizable, concerns?

Adds a Streaming API to the client for the SelectObjectContentOperation.
Tested with S3 and local minio server binary copy that supports Select operations, which can be found here: https://github.com/minio/minio/pull/6127

### DIFF
--- a/api-select.go
+++ b/api-select.go
@@ -1,0 +1,665 @@
+/*
+ * Minio Go Library for Amazon S3 Compatible Cloud Storage
+ * (C) 2018 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package minio
+
+import (
+	"bytes"
+	"context"
+	"encoding/binary"
+	"encoding/xml"
+	"errors"
+	"fmt"
+	"hash"
+	"hash/crc32"
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
+)
+
+// CSVFileHeaderInfo - is the parameter for whether to utilize headers.
+type CSVFileHeaderInfo string
+
+// Constants for file header info.
+const (
+	CSVFileHeaderInfoNone   CSVFileHeaderInfo = "NONE"
+	CSVFileHeaderInfoIgnore                   = "IGNORE"
+	CSVFileHeaderInfoUse                      = "USE"
+)
+
+// SelectCompressionType - is the parameter for what type of compression is
+// present
+type SelectCompressionType string
+
+// Constants for compression types under select API.
+const (
+	SelectCompressionNONE SelectCompressionType = "NONE"
+	SelectCompressionGZIP                       = "GZIP"
+	SelectCompressionBZIP                       = "BZIP2"
+)
+
+// CSVQuoteFields - is the parameter for how CSV fields are quoted.
+type CSVQuoteFields string
+
+// Constants for csv quote styles.
+const (
+	CSVQuoteFieldsAlways   CSVQuoteFields = "Always"
+	CSVQuoteFieldsAsNeeded                = "AsNeeded"
+)
+
+// QueryExpressionType - is of what syntax the expression is, this should only
+// be SQL
+type QueryExpressionType string
+
+// Constants for expression type.
+const (
+	QueryExpressionTypeSQL QueryExpressionType = "SQL"
+)
+
+// JSONType determines json input serialization type.
+type JSONType string
+
+// Constants for JSONTypes.
+const (
+	JSONDocumentType JSONType = "Document"
+	JSONStreamType            = "Stream"
+	JSONLinesType             = "Lines"
+)
+
+// ObjectSelectRequest - represents the input select body
+type ObjectSelectRequest struct {
+	XMLName            xml.Name `xml:"SelectObjectContentRequest" json:"-"`
+	Expression         string
+	ExpressionType     QueryExpressionType
+	InputSerialization struct {
+		CompressionType SelectCompressionType
+		CSV             struct {
+			FileHeaderInfo       CSVFileHeaderInfo
+			RecordDelimiter      string
+			FieldDelimiter       string
+			QuoteCharacter       string
+			QuoteEscapeCharacter string
+			Comments             string
+		}
+	}
+	OutputSerialization struct {
+		CSV struct {
+			QuoteFields          CSVQuoteFields
+			RecordDelimiter      string
+			FieldDelimiter       string
+			QuoteCharacter       string
+			QuoteEscapeCharacter string
+		}
+	}
+}
+
+// SelectObjectType - is the parameter which defines what type of object the
+// operation is being performed on.
+type SelectObjectType string
+
+// Constants for input data types.
+const (
+	SelectObjectTypeCSV  SelectObjectType = "CSV"
+	SelectObjectTypeJSON                  = "JSON"
+)
+
+// SelectObjectInput is a struct for specifying input options.
+type SelectObjectInput struct {
+	RecordDelimiter string
+	FieldDelimiter  string
+	Comments        string
+	FileHeaderInfo  CSVFileHeaderInfo
+}
+
+// SelectObjectOutput is a struct for specifying output options.
+type SelectObjectOutput struct {
+	RecordDelimiter string
+	FieldDelimiter  string
+}
+
+// SelectObjectOptions - Encapsulates all of the user options.
+type SelectObjectOptions struct {
+	Type   SelectObjectType
+	Input  SelectObjectInput
+	Output SelectObjectOutput
+}
+
+// preludeInfo is used for keeping track of necessary information from the
+// prelude.
+type preludeInfo struct {
+	totalLen  uint32
+	headerLen uint32
+}
+
+// headerInfo is used for keeping track of the headers from the server.
+type headerInfo struct {
+	messageType  string
+	errorType    string
+	errorMessage string
+}
+
+// SelectResults is used for the streaming responses from the server.
+type SelectResults struct {
+	Response chan interface{}
+	writer   *bytes.Buffer
+	lock     chan bool
+	stat     *StatEvent
+	progress *ProgressEvent
+	err      error
+}
+
+// RecordEvent is a struct that wraps the byte array of a single record.
+type RecordEvent struct {
+	Payload []byte
+}
+
+// ProgressEvent is a struct that wraps the byte array of a progress xml message.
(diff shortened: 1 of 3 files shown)


docs/API.md
```diff
@@ -1068,7 +1069,83 @@ for rErr := range minioClient.RemoveObjects(ctx, "my-bucketname", objectsCh) {
...
+	input := minio.SelectObjectInput{
+		RecordDelimiter: "\n",
+		FieldDelimiter:  ",",
+		FileHeaderInfo:  minio.CSVFileHeaderInfoUse,
+	}
+	output := minio.SelectObjectOutput{
+		RecordDelimiter: "\n",
+		FieldDelimiter:  ",",
+	}
+	opts := (minio.SelectObjectOptions{
```
harshavardhana: there is no reason for `(` we can simply avoid it.

docs/API.md
```diff
@@ -1068,7 +1069,83 @@ for rErr := range minioClient.RemoveObjects(ctx, "my-bucketname", objectsCh) {
...
+		Type:   minio.SelectObjectTypeCSV,
+		Input:  input,
+		Output: output,
+	})
+	temp, err := minioClient.SelectObjectContent(ctx, "sqlselectapi", "player.csv", "Select * from S3OBJECT WHERE last_name = 'James'", opts)
```
harshavardhana: temp is not a good variable name, choose a more appropriate one.

docs/API.md
```diff
@@ -1068,7 +1069,82 @@ for rErr := range minioClient.RemoveObjects(ctx, "my-bucketname", objectsCh) {
...
+	output := minio.SelectObjectOutput{
+		RecordDelimiter: "\n",
+		FieldDelimiter:  ",",
+	}
+	opts := minio.SelectObjectOptions{
+		Type:   minio.SelectObjectTypeCSV,
+		Input:  input,
+		Output: output,
+	}
+	myReader , err := minioClient.SelectObjectContent(ctx, "sqlselectapi", "player.csv", "Select * from S3OBJECT WHERE last_name = 'James'", opts)
```
donatello: This example doesnt look correct - SelectObjectContent returns a event stream according to the code, not a reader. Also the io.Pipe does not seem to be needed here.

api-select.go
```diff
@@ -0,0 +1,590 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if resp.StatusCode != http.StatusOK {
+		return nil, httpRespToErrorResponse(resp, bucketName, "")
+	}
+
+	//Process needs to return a reader to the client, so as we read from it we
+	//need to write the payload to a new reader
+	// Wrap this in a go function
```
donatello: Looks the comment refers to an older version of the code.

docs/API.md
```diff
@@ -1068,7 +1069,69 @@ for rErr := range minioClient.RemoveObjects(ctx, "my-bucketname", objectsCh) {
...
+		Type:   minio.SelectObjectTypeCSV,
+		Input:  input,
+		Output: output,
+	}
+	eventStream , err := minioClient.SelectObjectContent(ctx, "sqlselectapi", "player.csv", "Select * from S3OBJECT WHERE last_name = 'James'", opts)
+	if err != nil {
+		return
+	}
+	defer eventStream.Close()
+	go func() {
```
donatello: Go routine could be removed?

api-select.go
```diff
@@ -0,0 +1,589 @@
...
+	})
+	if err != nil {
+		return nil, err
+	}
+	if resp.StatusCode != http.StatusOK {
+		return nil, httpRespToErrorResponse(resp, bucketName, "")
+	}
+
+	// Process needs to return a eventStream to the client
+	// Wrap this in a go function so as to allow the streaming
```
donatello: Go routine is the right terminology here - not go function.

api-select.go
```diff
@@ -0,0 +1,589 @@
...
+	case "Stats":
+		myStat := &StatEvent{
+			Payload: payload,
+		}
+		streamer.Response <- myStat
+	case "Cont":
+		myCont := &ContEvent{}
+		streamer.Response <- myCont
+	case "Error":
+		streamer.err = errors.New("Error Type of " + headers.errorType + " " + headers.errorMessage)
```
donatello: Looks we should be returning true here to end the streaming as an error happened?

api-select.go
```diff
@@ -0,0 +1,589 @@
...
+						var completePayload []byte
+						completePayload, partialRecordVal, lenOfTrim = processPartial(myPayload)
+						processEvent(completePayload, headers, streamer)
+						continue
+					}
+					// if there is no longer a partial porition we can return our payload
+					// to the client and break
+					if len(myPayload) > 0 {
+						processEvent(myPayload, headers, streamer)
+					}
```
donatello: Most calls to `processEvent` are ignoring the returned value. This does not look correct when considering error handling. Also on error, it is required to close the Events.Response channel, so that the for loop that is consuming the events quits - actually this does not seem to be happening in the success case, when all results have been returned and the program should exit.

See https://play.golang.org/p/wIARKfwXJPv to understand the problem

api-select.go
```diff
@@ -0,0 +1,589 @@
...
+				payload, partialRecordVal, lenOfTrim = processPartial(payload)
+				processEvent(payload, headers, streamer)
+				// return the complete part of the message so far to the client
+				// for loop in case there are multiple sequential partial rows
+				for {
+					// Extracts the partial payload.
+					myPayload, err := extractPayload(io.LimitReader(crcReader, int64(lenOfTrim)), lenOfTrim)
+					if err != nil {
+						if streamer.err == nil {
+							streamer.err = err
```
donatello: Looks like we should return here closing the events channel - the error could be due to anything (e.g. network error, protocol error).

api-select.go
```diff
@@ -0,0 +1,627 @@
...
+	"fmt"
+	"hash"
+	"hash/crc32"
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
+)
+
+// CSVFileHeaderInfo -
```
harshavardhana: We need to explain each struct since these are exposed for public consumption.

api-select.go
```diff
@@ -0,0 +1,627 @@
...
+	}
+}
+
+// isPartial is a function which returns whether or not a message's payload is
+// partial
+func isPartial(payload []byte) bool {
+	if len(payload) < 3 {
+		return false
+	}
+	// TODO // make sure this works with Amazon again
```
harshavardhana: is this still a TODO?

api-select.go
```diff
@@ -0,0 +1,627 @@
...
+		// select block for catching a user request for closing.
+		select {
+		// Logic is basically that the .Close() method will close the streamer.lock
+		// channel, Therefore if Ok is false(the channel is closed) then we can
+		// close our response channel and return. The .Close() method blocks on the
+		// streamer.Response channel being closed, so that it is guranteed that this
+		// method has exited since it is a defered close.
+		case _, ok := <-streamer.lock:
+			if !ok {
+				defer close(streamer.Response)
```
harshavardhana: no reason to defer before return you can just close and return.

Colhodm: The reason for deferring, is that by deferring the closing of the channel until after the return, it is guaranteed that the processing thread will return before the the .Close() method returns. Otherwise without the defer, every once out of every 10 tests, I would get some weird behavior that given the timing of things, the .Close() returns before processing thread does. That might be okay, but I wasn't sure so to guarantee that the processing thread returns before the .Close() does, I used defer.

api-select.go
```diff
@@ -0,0 +1,627 @@
...
+					close(streamer.Response)
+					return
+				}
+			}
+			// Calculate the length of the payload so that the appropriate amount of
+			// bytes can be read
+			payloadLen := calcPayloadLen(prelude)
+			// Checks that there is a payload and if there is, then extracts it.
+			if payloadLen > 0 {
+				payload, err = extractPayload(io.LimitReader(crcReader, int64(payloadLen)), payloadLen)
```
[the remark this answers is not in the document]
Colhodm: Fixed, dont know why Github is not closing this ..

api-select.go
```diff
@@ -0,0 +1,637 @@
...
+						// if there is no longer a partial porition we can return our payload
+						// to the client and break
+						if len(myPayload) > 0 {
+							processEvent(myPayload, headers, streamer)
+						}
+						payload = nil
+						break
+					}
+				}
+				if err != nil {
```
donatello: This block does not seem necessary.

api-select.go
```diff
@@ -0,0 +1,637 @@
...
+					if streamer.err == nil {
+						streamer.err = err
+					}
+					close(streamer.Response)
+					return
+				}
+				// Special case for dealing with messages that are partial.
+				if isPartial(payload) {
+					payload, partialRecordVal, lenOfTrim = processPartial(payload)
+					processEvent(payload, headers, streamer)
```
donatello: This code looks a bit repetitive - it looks like it should be possible to refactor this so that there is only a single call to 
extractPayload, processPartial, processEvent within a loop. That will be easier to read and maintain.

api-select.go
```diff
@@ -0,0 +1,621 @@
...
+							streamer.err = err
+						}
+						close(streamer.Response)
+						return
+					}
+					// knits together the payload with the previous partial one.
+					payload = append(partialRecordVal, myPayload...)
+					// update state to keep track of whether partial records were
+					// processed
+					state = !isPartial(payload)
```
donatello: This state variable is not needed. As written, the for loop is exited when the isPartial condition evaluates to false. So processEvent must be called every time after the for loop.

api-select.go
```diff
@@ -0,0 +1,621 @@
...
+					if streamer.err == nil {
+						streamer.err = err
+					}
+					close(streamer.Response)
+					return
+				}
+				// Special case for dealing with messages that are partial.
+				var state bool
+				for isPartial(payload) {
+					// state indicates whether it is a partial record or not, so if it
```
donatello: Looks like an old comment.

api-select.go
Colhodm: The state variable is actually needed, because in the case that there is NOT a partial record sent, e.g it just one complete record, you want to make sure that the processEvent immediately following the isPartial, is not called. Also lets say there are 2 parts to a partial record, after the 2nd one, the record is now complete, so then we update the state and quit the for loop and send the processEvent. The reason this processEvent has to be here is because when partial records are sent only the first one follows the format of a normal message e.g prelude header payload crc, the rest follow just a format of .. payload .., so thats why it is written like this

donatello:
@Colhodm We would like improve the API interface by making the return type of the API more intelligent - see details here https://gist.github.com/donatello/9d53cc06ac9a3e840f8ec07e70fd1b37

api-select.go
```diff
@@ -0,0 +1,623 @@
...
+		} else if headerTypeName == ":error-code" {
+			myHeader = &headerInfo{
+				messageType: "Error",
+				errorType:   headerValueName,
+			}
+		} else if headerTypeName == ":error-message" {
+			myHeader.errorMessage = headerValueName
+		}
+	}
+	return myHeader, nil
```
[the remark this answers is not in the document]
Colhodm: myHeader will never be replaced in the following iteration of the loop. The 2 conditions which directly create a headerInfo reference are mutually exclusive as can be seen here:

harshavardhana:
Closing this PR for new work..

PR #1014, opened 2018-08-03
Disable ETag check when copying object < 5 Gb

### LINKED ISSUE #1015: mc returns 'At least one of the pre-conditions you specified did not hold' during a concurrent read/write on the same object
### How to reproduce ?
Running the following commands in a loop in two separate terminals:

```
while true; do 
    mc --insecure cp --encrypt-key "myminio=01234567890123456789012345678901" myminio/testbucket/file myminio/testbucket/file.cp; 
done
```

```
while true; do 
    mc --insecure cp --encrypt-key="myminio=01234567890123456789012345678901" file myminio/testbucket/file; 
done
```

### The exact error:
```
mc: <ERROR> Failed to copy `https://192.168.1.121:9000/testbucket/Dockerfile`. At least one of the pre-conditions you specified did not hold
```

vadmeste:
Currently, copying object is always protected by an ETag check.
This is pretty useful when we copy an object by parts so minio-go
can detect if the object has been changed during the copy operation.
However, this check is not needed if the object size is < 5 Gb.
This change will avoid raising an unnecessary error when an object
has been overwrited by a concurrent S3 client since the resulted
copied object will still be not corrupted.

Fixes #1015

### DIFF
--- a/api-compose-object.go
+++ b/api-compose-object.go
@@ -377,15 +377,6 @@ func (c Client) ComposeObjectWithProgress(dst DestinationInfo, srcs []SourceInfo
 				fmt.Sprintf("Client side encryption is used in source object %s/%s", src.bucket, src.object))
 		}
 
-		// Since we did a HEAD to get size, we use the ETag
-		// value to make sure the object has not changed by
-		// the time we perform the copy. This is done, only if
-		// the user has not set their own ETag match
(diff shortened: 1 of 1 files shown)


WoodyMacGuinness:
How does we disable the ETag Check?

PR #1018, opened 2018-08-17
Allow sse-s3, sse-kms headers in Put, Get and Copy operations

poornas:
Encryption headers for SSE-S3, SSE-KMS are not being passed by the sdk for Get/Put/Copy calls. This PR allows encryption headers to be marshalled for SSE-S3,SSE-KMS in addition to SSE-C

### DIFF
--- a/api-get-options.go
+++ b/api-get-options.go
@@ -44,7 +44,7 @@ func (o GetObjectOptions) Header() http.Header {
 	for k, v := range o.headers {
 		headers.Set(k, v)
 	}
-	if o.ServerSideEncryption != nil && o.ServerSideEncryption.Type() != encrypt.S3 {
+	if o.ServerSideEncryption != nil && o.ServerSideEncryption.Type() == encrypt.SSEC {
 		o.ServerSideEncryption.Marshal(headers)
 	}
 	return headers
--- a/api-put-object-multipart.go
+++ b/api-put-object-multipart.go
@@ -259,7 +259,7 @@ func (c Client) uploadPart(ctx context.Context, bucketName, objectName, uploadID
 
 	// Set encryption headers, if any.
 	customHeader := make(http.Header)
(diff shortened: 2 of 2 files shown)


api-compose-object.go
```diff
@@ -458,7 +458,12 @@ func (c Client) ComposeObjectWithProgress(dst DestinationInfo, srcs []SourceInfo
 	for i, src := range srcs {
 		h := src.Headers
 		if src.encryption != nil {
-			encrypt.SSECopy(src.encryption).Marshal(h)
+			if src.encryption.Type() == encrypt.SSEC {
```
aead: @poornas This is not necessary AFAIK. See doc and impl of: [`SSECopy`](https://github.com/minio/minio-go/blob/master/pkg/encrypt/server-side.go#L143).

kannappanr: Agree the code looks same to me

api-get-options.go
```diff
@@ -44,7 +44,7 @@ func (o GetObjectOptions) Header() http.Header {
 	for k, v := range o.headers {
 		headers.Set(k, v)
 	}
-	if o.ServerSideEncryption != nil && o.ServerSideEncryption.Type() != encrypt.S3 {
```
kannappanr: Maybe this should be `if o.ServerSideEncryption != nil && o.ServerSideEncryption.Type() == encrypt.SSEC {`

aead: I'm okay with both solutions. `o.ServerSideEncryption != nil` should cause the server to fail and the additional check `o.ServerSideEncryption.Type() == encrypt.SSEC` just drops the parameter. Whatever a typical user would prefer.

api-compose-object.go
poornas: you are correct @aead

api-get-options.go
poornas: that's a good suggestion, done.

PR #1021, opened 2018-08-30
Migrate to Go Modules

marwan-at-work:
This PR migrates the package to Go Modules and updates the path according to Semantic Import Versioning

Generated by https://github.com/marwan-at-work/mod

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,10 @@
+module github.com/minio/minio-go/v6
+
+require (
+	github.com/go-ini/ini v1.38.2
+	github.com/mitchellh/go-homedir v1.0.0
+	golang.org/x/crypto v0.0.0-20180830192347-182538f80094
+	golang.org/x/net v0.0.0-20180826012351-8a410e7b638d
+	golang.org/x/sys v0.0.0-20180830151530-49385e6e1522 // indirect
+	golang.org/x/text v0.3.0 // indirect
+)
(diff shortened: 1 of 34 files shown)


harshavardhana:
@marwan-at-work why is this a hard requirement? we shouldn't be changing the path of the directories where the packages are stored in this manner. It is not correct and prone to import issues.

Let's not take this PR in until the Go modules has been moved away from experimental and we don't want any dependency on the `mod` tool. We prefer if Go provided a mechanism already natively through their own tooling.

go.mod
```diff
@@ -0,0 +1,10 @@
+module github.com/minio/minio-go/v6
```
harshavardhana: this should be simply github.com/minio/minio-go not 'v6'

harshavardhana: In-fact this PR should be committed with a tag

harshavardhana:
> @harshavardhana totally understandable to wait until Go modules is out of experimental :) feel free to close the PR.
> 
> Just FYI, Go Modules is backwards compatible with the latest Go 1.9 and 1.10. In other words, it can still be imported without the /v6 and Go would resolve it correctly. And of course, you'd need to tag this with v6.x.x because according to Semantic Import Versioning, the git tag and the import path must match exactly on the major part.

For now, let's close this, we will revisit this when we move to go1.11 modules.

harshavardhana:
We will move when go modules become stable and fully supported.

PR #1024, opened 2018-09-13
Add new API SelectObjectContent

harshavardhana:
Original work by Arjun https://github.com/minio/minio-go/pull/1013

### DIFF
--- a/api-select.go
+++ b/api-select.go
@@ -1,0 +1,520 @@
+/*
+ * Minio Go Library for Amazon S3 Compatible Cloud Storage
+ * (C) 2018 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package minio
+
+import (
+	"bytes"
+	"context"
+	"encoding/binary"
+	"encoding/xml"
+	"errors"
+	"fmt"
+	"hash"
+	"hash/crc32"
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
+
+	"github.com/minio/minio-go/pkg/encrypt"
+	"github.com/minio/minio-go/pkg/s3utils"
+)
+
+// CSVFileHeaderInfo - is the parameter for whether to utilize headers.
+type CSVFileHeaderInfo string
+
+// Constants for file header info.
+const (
+	CSVFileHeaderInfoNone   CSVFileHeaderInfo = "NONE"
+	CSVFileHeaderInfoIgnore                   = "IGNORE"
+	CSVFileHeaderInfoUse                      = "USE"
+)
+
+// SelectCompressionType - is the parameter for what type of compression is
+// present
+type SelectCompressionType string
+
+// Constants for compression types under select API.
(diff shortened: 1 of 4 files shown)


api-select.go
```diff
@@ -0,0 +1,491 @@
...
+					return
+				}
+			}
+
+			// Calculate the length of the payload so that the appropriate amount of
+			// bytes can be read
+			payloadLen := calcPayloadLen(prelude)
+			// Checks that there is a payload and if there is, then extracts it.
+			if payloadLen > 0 {
+				payload, err = extractPayload(io.LimitReader(crcReader, int64(payloadLen)), payloadLen)
```
donatello: This will take up some memory in the client application if the payload is large (potentially 4GB), and increase latency while the payload gets buffered inside the library - it can be avoided by writing to the client directly (perhaps with a pipe).

harshavardhana: yes we can write directly @donatello

harshavardhana:
PR is updated with more memory efficient changes @donatello @kannappanr PTAL

examples/s3/selectobject.go
```diff
@@ -0,0 +1,82 @@
+// +build ignore
+
+/*
+ * Minio Go Library for Amazon S3 Compatible Cloud Storage
+ * Copyright 2015-2017 Minio, Inc.
```
kannappanr: Can you change year to 2018? Looks good otherwise

harshavardhana:
> Found a bug when running the example selectobject.go with encrypted csv file (using the server code base at PR [minio/minio#6449](https://github.com/minio/minio/pull/6449)):
> 
> ```
> go run examples/s3/selectobject.go 
> 371008.000000
> 2018/09/17 13:20:20 io: read/write on closed pipe
> exit status 1
> ```
> It seems the library is closing the reader early?

will test

api-select.go
```diff
@@ -0,0 +1,499 @@
...
+// SelectObjectOptions - represents the input select body
+type SelectObjectOptions struct {
+	XMLName              xml.Name           `xml:"SelectObjectContentRequest" json:"-"`
+	ServerSideEncryption encrypt.ServerSide `xml:"-"`
+	Expression           string
+	ExpressionType       QueryExpressionType
+	InputSerialization   struct {
+		CompressionType SelectCompressionType
+		Parquet         *struct{} `xml:"Parquet,omitempty"`
+		CSV             *struct {
```
donatello: We need to name all these structs for ease of use as a library. Otherwise, it will require referring to the library source code, just to figure out how to pass InputSerialization, OutputSerialization, etc.

harshavardhana: Done

examples/s3/selectobject.go
```diff
@@ -0,0 +1,73 @@
...
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	opts := minio.SelectObjectOptions{
+		Expression:     "select count(*) from s3object",
+		ExpressionType: minio.QueryExpressionTypeSQL,
+		InputSerialization: SelectObjectInputSerialization{
+			CompressionType: minio.SelectCompressionNONE,
+			CSVInputOptions: &CSVInputOptions{
```
donatello: This example does not compile.

examples/s3/selectobject.go
```diff
@@ -0,0 +1,73 @@
...
+	// determined based on the Endpoint value.
+	s3Client, err := minio.New("s3.amazonaws.com", "YOUR-ACCESS-KEY-HERE", "YOUR-SECRET-KEY-HERE", true)
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	opts := minio.SelectObjectOptions{
+		Expression:     "select count(*) from s3object",
+		ExpressionType: minio.QueryExpressionTypeSQL,
+		InputSerialization: SelectObjectInputSerialization{
```
kannappanr: `InputSerialization: SelectObjectInputSerialization{` should be `InputSerialization: minio.SelectObjectInputSerialization{`

PR #1029, opened 2018-09-23
Set correct etag in X-Amz-Copy-Source-If-Match for each part in ComposeObject API

### LINKED ISSUE #1028: 'At least one of the pre-conditions you specified did not hold' Error
The function '**ComposeObject**' returns the following error: 
"_At least one of the pre-conditions you specified did not hold_".

'StackOverflow.com' website Thread:
https://stackoverflow.com/questions/52371672/minio-at-least-one-of-the-pre-conditions-you-specified-did-not-hold-error

If something happens to the 'StackOverflow.com' website, then the minimal example representing the bug (source code with sample data) can be obtained here:
https://www.file-upload.com/hzthje923oh4

Thank you.

kannappanr:
Currently, all the parts are being set the same etag for the condition `X-Amz-Copy-Source-If-Match`, because of which `PUT object Part`
API fails.

Fixes #1028

### DIFF
--- a/api-compose-object.go
+++ b/api-compose-object.go
@@ -362,10 +362,10 @@ func (c Client) ComposeObjectWithProgress(dst DestinationInfo, srcs []SourceInfo
 	srcSizes := make([]int64, len(srcs))
 	var totalSize, size, totalParts int64
 	var srcUserMeta map[string]string
-	var etag string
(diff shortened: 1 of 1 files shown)


PascalVA:
Thanks for the fix, you just beat me too it by a couple of hours! Your PR fixes my use case.

PR #1031, opened 2018-10-03
Pass Etag back in core CompleteMultipartUpload call

poornas:
core.go needs to be enhanced to pass back ETag in CompleteMultipartUpload call.  

This is required by server to pass back correct ETag in the response for a CompleteMultipartOperation specifically for the case of an SSE-C encrypted object at the gateway backend (i.e pass through encryption). 

HEAD calls require SSE-C headers whereas the CompleteMultipartUpload operation does not. This change passes back ETag sent in the CompleteMultipartUpload result , so that the server can set correct ETag in the response. 

related to : https://github.com/minio/minio/issues/6323


harshavardhana:
ping @donatello

PR #1041, opened 2018-10-30
improve Quick Start Example - File Uploader

davegallant:
The example of FileUploader.go logs "Successfully created bucket" after each rerun.

It should probably only log this when it actually creates the bucket.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,8 +85,9 @@ func main() {
 		} else {
 			log.Fatalln(err)
 		}
+	} else {
+		log.Printf("Successfully created %s\n", bucketName)
 	}
-	log.Printf("Successfully created %s\n", bucketName)
 
 	// Upload the zip file
 	objectName := "golden-oldies.zip"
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -97,9 +98,9 @@ func main() {
 	n, err := minioClient.FPutObject(bucketName, objectName, filePath, minio.PutObjectOptions{ContentType:contentType})
 	if err != nil {
 		log.Fatalln(err)
+	} else {
+		log.Printf("Successfully uploaded %s of size %d\n", objectName, n)
```
kannappanr: This else is not needed.

ebozduman: Both ways have the same affect, but this is about the style. 
In all of our code, we do not use `else` if not needed.

harshavardhana: @davegallant the reason we don't worry about `else` clause here is because log.Fatalln(err) will exit the program and only upon success we get to log.Printf() . So this doesn't need to be changed. Changing the `bucket` message is correct.

davegallant: thanks. fixed it. (new to golang)

kannappanr: Thanks for the PR.

PR #1042, opened 2018-11-13
Added SetUserData to run Unit Test of Post Policy

sinhaashish:
To execute the test cases for [PR](https://github.com/minio/minio/pull/6628) added a function `SetUserData` . This will resolve the [issue ](https://github.com/minio/minio/issues/6559)

### DIFF
--- a/post-policy.go
+++ b/post-policy.go
@@ -206,6 +206,28 @@ func (p *PostPolicy) SetUserMetadata(key string, value string) error {
 	return nil
 }
 
+// SetUserData - Set user data as a key/value couple.
+// Can be retrieved through a HEAD request or an event.
+func (p *PostPolicy) SetUserData(key string, value string) error {
+	if key == "" {
+		return ErrInvalidArgument("Key is empty")
+	}
+	if value == "" {
+		return ErrInvalidArgument("Value is empty")
+	}
+	headerName := fmt.Sprintf("x-amz-%s", key)
+	policyCond := policyCondition{
+		matchType: "eq",
+		condition: fmt.Sprintf("$%s", headerName),
+		value:     value,
+	}
+	if err := p.addNewPolicy(policyCond); err != nil {
+		return err
+	}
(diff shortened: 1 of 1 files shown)


post-policy.go
```diff
@@ -206,6 +206,28 @@ func (p *PostPolicy) SetUserMetadata(key string, value string) error {
 	return nil
 }
 
+// SetUserData - Set user data as a key/value couple.
+// Can be retrieved through a HEAD request or an event.
+func (p *PostPolicy) SetUserData(key string, value string) error {
+	if strings.TrimSpace(key) == "" || key == "" {
```
kannappanr: If `key==""` is `true` then `strings.TrimSpace(key) == ""` will also be `true`

post-policy.go
```diff
@@ -206,6 +206,28 @@ func (p *PostPolicy) SetUserMetadata(key string, value string) error {
...
+		return ErrInvalidArgument("Key is empty")
+	}
+	if strings.TrimSpace(value) == "" || value == "" {
+		return ErrInvalidArgument("Value is empty")
+	}
+	headerName := fmt.Sprintf("x-amz-%s", key)
```
Praveenrajmani: Can we have canonical header prefix here ?

sinhaashish:
>Looks like we already have SetUserMetadata https://github.com/minio/minio-go/blob/774475480ffe1bffa9dae62ea98f6ea75e8cf567/post-policy.go#L189. We can use this instead.
`SetUserMetadata` contained headerName as `x-amz-meta` but the ask was for `x-amz` . so added this function.

post-policy.go
sinhaashish: Canonical header is used in minio/minio .. The test case has it small.

post-policy.go
sinhaashish: removed `strings.TrimSpace(key) == ""`

PR #1048, opened 2018-11-30
Set Content-Encoding for sigv4-streaming uploads

tipabu:
[According to AWS's docs](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html), "you must include the following headers" when performing a streaming upload:

- `x-amz-content-sha256`
- `Content-Encoding`
- `x-amz-decoded-content-length`
- `Content-Length`

Previously, we omitted `Content-Encoding`; now it will be included like the others. If a `Content-Encoding` is already provided, `aws-chunked` will be prepended.


harshavardhana:
Closing this as not required..

PR #1050, opened 2018-12-08
Correct minimum expiry time in presignURL func doc

mseshachalam:


api-presigned.go
```diff
@@ -28,7 +28,7 @@ import (
 )
 
 // presignURL - Returns a presigned URL for an input 'method'.
-// Expires maximum is 7days - ie. 604800 and minimum is 1.
+// Expires maximum is 7days - ie. 604800 and minimum is 1second.
```
harshavardhana:
```
// Expires maximum is 7days - ie. 604800 seconds and a minimum of 1 second.
```

kannappanr:
The title can also be changed to `Add unit to expiry time comment in PresignURL func`

PR #1056, opened 2019-01-01
Document AWS S3 IAM permissions for test suite

andreaso:


ebozduman:
@andreaso,

Which test suite you are referring to?

Our existing functional tests are run for every PR and they all pass.
Could you please explain the purpose a little bit more so we can understand the need better?

harshavardhana:
what @andreaso means is that if you wish to run the functional tests against AWS S3 you would need a minimal set of permissions in your IAM policy for the given user. Without this you won't be able to run functional tests properly as they will fail for access/secret keys which have reduced permissions. 

Currently, we use play.minio.io which is a public server with admin credentials, lets say you created a user `ersan` which will not have full permissions by default so you would need an Allow IAM policy such as the one documented here in this PR to be applied for the user so that you can use user credentials to run functional tests as well. It is also a way to test the policies etc and see all operations perform as expected as it would for an Admin under right IAM settings.

This is a neat addition which we should accept, @andreaso can you explain why you would need to add IAM policies a bit more.

andreaso:
@harshavardhana: What kind of explanation are you looking for? In addition to the context already being laid out by referring to `SERVER_ENDPOINT=s3.amazonaws.com`?

andreaso:
Hmm, explanation or not, I now realize that the IAM policies themselves are incomplete.

The listed IAM policy only covers running `go test -race ./...`, since that is the test which _CONTRIBUTING.md_ is telling you to run, and which do appear to do some functional testing.

Yet, taking another look at the _Makefile_, which is what _.travis.yml_ runs, I now see that there is also `go run functional_tests.go`, which requires additional permissions not covered by my PR.

By the way, should _CONTRIBUTING.md_ also mention the `go run functional_tests.go` test, and the play.minio.io admin credentials? Alternatively, should it simply tell you to run `Make checks`?

harshavardhana:
> By the way, should _CONTRIBUTING.md_ also mention the `go run functional_tests.go` test, and the play.minio.io admin credentials? Alternatively, should it simply tell you to run `Make checks`?

Yes please go ahead @andreaso - thanks

harshavardhana:
> @harshavardhana: What kind of explanation are you looking for? In addition to the context already being laid out by referring to `SERVER_ENDPOINT=s3.amazonaws.com`?

@andreaso well what you have right now is fine..

PR #1061, opened 2019-01-09
Add a "Reviewed by Hound" badge

salbertson:
screenshot 2019-01-08 18 31 45


harshavardhana:
We would like to reduce the number of badges, what we have right now is sufficient.

PR #1064, opened 2019-01-17
Set prefix, delimiter params even when empty

harshavardhana:
We have never set values which are empty on the request
because they are perhaps not useful in the List query,
but this assumption is wrong when there are restricted
policies for a given user, because empty is actually
a valid value in IAM or Bucket policy conditions.

For example, the following condition would never work with our
ListObjects API and AWS cli would work fine.
```json
            "Condition": {
                "StringEquals": {
                    "s3:prefix": [
                        "",
                        "data/",
                        "data"
                    ],
                    "s3:delimiter": [
                        "/",
                        ""
                    ]
                }
            }
```

The reason is empty or not `prefix` and `delimiter` should be
added to the query param in List operation, such that server
can use the value to validate the policies for the incoming
request.

Fixes https://github.com/minio/mc/issues/2647

### DIFF
--- a/api-list.go
+++ b/api-list.go
@@ -192,17 +192,15 @@ func (c Client) listObjectsV2Query(bucketName, objectPrefix, continuationToken s
 	// Always set list-type in ListObjects V2
 	urlValues.Set("list-type", "2")
 
-	// Set object prefix.
-	if objectPrefix != "" {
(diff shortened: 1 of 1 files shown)


harshavardhana:
PTAL @vadmeste

PR #1069, opened 2019-01-30
Strip 80/443 in getBucketLocationRequest

L11R:
There was the PR #709 with it, but works only partly. 
Currently it works only with workaround by specifying `us-east-1` region (`minio.NewWithRegion(...)`)
With this patch you can just do `minio.New(...)` and all will work flawlessly.

### DIFF
--- a/bucket-cache.go
+++ b/bucket-cache.go
@@ -18,6 +18,7 @@
 package minio
 
 import (
+	"net"
 	"net/http"
 	"net/url"
 	"path"
@@ -162,6 +163,14 @@ func (c Client) getBucketLocationRequest(bucketName string) (*http.Request, erro
 
 	// Set get bucket location always as path style.
 	targetURL := c.endpointURL
+	
+	// as it works in makeTargetURL method from api.go file
+	if h, p, err := net.SplitHostPort(targetURL.Host); err == nil {
+		if targetURL.Scheme == "http" && p == "80" || targetURL.Scheme == "https" && p == "443" {
+			targetURL.Host = h
+		}
+	}
+	
 	targetURL.Path = path.Join(bucketName, "") + "/"
 	targetURL.RawQuery = urlValues.Encode()
 
--- a/bucket-cache.go
+++ b/bucket-cache.go
@@ -163,14 +163,14 @@ func (c Client) getBucketLocationRequest(bucketName string) (*http.Request, erro
 
 	// Set get bucket location always as path style.
 	targetURL := c.endpointURL
-	
+
 	// as it works in makeTargetURL method from api.go file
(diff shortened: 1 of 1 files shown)


L11R:
Also related to #695

sinhaashish:
@L11R  I ran the below code and both `New` and `NewWithRegion` worked for me  with `master` branch not the `PR` one
```
       // s3Client, err := minio.NewWithRegion("localhost:80", "minio", "minio123", false, "us-east-1")
	s3Client, err := minio.New("localhost:80", "minio", "minio123", false)
	if err != nil {
		log.Fatalln(err)
	}

	// Set request parameters
	reqParams := make(url.Values)
	reqParams.Set("response-content-disposition", "attachment; filename=\"your-filename.pdf\"")

	// Gernerate presigned get object url.
	presignedURL, err := s3Client.PresignedGetObject("sinha", "Agreement.pdf", time.Duration(1000)*time.Second, reqParams)
	if err != nil {
		log.Fatalln(err)
	}
```
Can you elaborate , what issue you faced that led you to this PR

L11R:
I have minio at 443 port with HTTPS enabled. It works while I try to fetch list of buckets for example, but not working while I try to upload, presign, etc. I figure out that when you build url in the getBucketLocationRequest, you add port. Then I am getting wrong signature error: https://gist.github.com/L11R/4873c840d408519d0470974e3af03e8b
With NewWithRegion all works as should because it cause skipping of this method.

sinhaashish:
@L11R   Thanks for raising the PR but still i could not replicate your issue.  With port 443 and HTTPS i was able to upload the object on minio with `minio.New(...)`. 
Gist here give the details.](https://gist.github.com/sinhaashish/fd3818d85c03672a69535fe90296dd16).

What is `Server: nginx/1.15.8` in your gist file. Are you using any load balancer?  
Some more info from you would help to replicate then review the PR.

L11R:
@sinhaashish Yes, we are using Nginx as reverse proxy. All as described in your official guides.
Well, my I don't know why you cannot reproduce it. It's pretty obviously that minio-go library created different URLs in different cases. In our case with and without port. That's why we have different signatures and error. I don't know, how to fix it from server side.

L11R:
@sinhaashish oh, seems like I got it. I saw you example. You pass just an IP here: `minio.New("192.168.0.102", ...`
I didn't try it. Is it okay to just pass domain or IP here (we use domain)? I will it try tomorrow at work.

Anyway I still think that code is currently not consistent. Why we somewhere strip it and somewhere not?

L11R:
@sinhaashish okay, I tried to pass domain only. And yeah, it works. So problem resolved in my case. But I still worry for inconsistent code behavior.

PR #1070, opened 2019-02-04
Fix hook reader to seek source and hook appropriately

harshavardhana:


harshavardhana:
To test this with `mc` you need following patch

```diff
 git diff cmd/progress-bar.go
diff --git a/cmd/progress-bar.go b/cmd/progress-bar.go
index 672639f0..ae87b059 100644
--- a/cmd/progress-bar.go
+++ b/cmd/progress-bar.go
@@ -17,6 +17,7 @@
 package cmd
 
 import (
+       "errors"
        "io"
        "runtime"
        "strings"
@@ -101,6 +102,20 @@ func (p *progressBar) Set64(length int64) *progressBar {
        return p
 }
 
+func (p *progressBar) Seek(offset int64, whence int) (n int64, err error) {
+       switch whence {
+       case io.SeekStart:
+               p.ProgressBar = p.ProgressBar.Set64(offset)
+               return offset, nil
+       case io.SeekCurrent:
+               return p.ProgressBar.Add64(offset), nil
+       case io.SeekEnd:
+               p.ProgressBar.Set64(p.ProgressBar.Total - offset)
+               return p.ProgressBar.Total - offset, nil
+       }
+       return 0, errors.New("invalid argument")
+}
+
 func (p *progressBar) SetTotal(total int64) *progressBar {
        p.ProgressBar.Total = total
        return p
```

harshavardhana:
ping @vadmeste

vadmeste:
@harshavardhana I believe the testing code should be like this:
```
       case io.SeekCurrent:
+             currentOffset := p.ProgressBar.Get()
               return p.ProgressBar.Add64(currentOffset+offset), nil
```

harshavardhana:
> ```
>                return p.ProgressBar.Add64(currentOffset+offset), nil
> ```

The Add64() already adds to the currentOffset, if we do this we would be adding more.

PR #1077, opened 2019-02-22
Support HTTP 2.0 in DefaultTransport for TLS requests

harshavardhana:
This fixes the unit tests failure in the master branch

### DIFF
--- a/api-put-object-common.go
+++ b/api-put-object-common.go
@@ -34,26 +34,25 @@ func isObject(reader io.Reader) (ok bool) {
 
 // Verify if reader is a generic ReaderAt
 func isReadAt(reader io.Reader) (ok bool) {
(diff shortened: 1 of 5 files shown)


harshavardhana:
NOTE: I do not have any information on why this PR fixes the issue while master doesn't against a TLS enabled Minio server. 

So it may be http2.Transport does something, also not all tests fail it is only some of them.

I tested by uncommenting each one of them and testing individually as well. Also tested combinations of certain tests etc. None would reveal what was the problem. 

The problem simply occurs in http2 and the connection hangs after some time we see STREAM_ERROR. 

With this PR everything seems to be working properly back again.

harshavardhana:
ping @nitisht @krisis PTAL

Issue #1090, opened 2019-03-22
Please expand TraceOn() to print only errors


GiedriusS:
TraceOn() ATM prints all of the requests and responses. For debugging purposes and to reduce spam in the log messages, it would be useful to only print erroneous requests and responses to them. Please consider expanding that method. Thanks a lot! For more context, please refer to https://github.com/improbable-eng/thanos/pull/937 and the relevant issues.

szalai1:
started working on this.

harshavardhana:
This fixed by #1091

PR #1091, opened 2019-03-24
[api] enable error tracing

szalai1:
change:
 `client`'s `TraceOn` method changed. Now it has a `errorsOnly` argument.

Related issue: https://github.com/minio/minio-go/issues/1090

### DIFF
--- a/api.go
+++ b/api.go
@@ -73,8 +73,9 @@ type Client struct {
 	bucketLocCache *bucketLocationCache
 
 	// Advanced functionality.
-	isTraceEnabled bool
-	traceOutput    io.Writer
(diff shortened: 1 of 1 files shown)


szalai1:
Thanks @harshavardhana for your feedback. Changing the interface was not a good idea. I agree. I called the method `TraceErrorsOnlyOn(io.Writer)`, because it will trace errors _only_.  Also, I added `TraceErrorsOnlyOff` because enabling regular tracing after error tracing  would require:
1. `TraceOff()` 
2. `TraceOn(io)`

kannappanr:
ping @vadmeste

PR #1093, opened 2019-03-25
Add configurable PartSize for PutObject

harshavardhana:


harshavardhana:
ping @kannappanr @vadmeste PTAL

api-put-object-common.go
```diff
@@ -77,10 +77,15 @@ func optimalPartInfo(objectSize int64) (totalPartsCount int, partSize int64, las
...
+	if int64(configuredPartSize) > objectSize {
+		err = ErrEntityTooLarge(int64(configuredPartSize), objectSize, "", "")
+		return
+	}
+
 	// Use floats for part size for all calculations to avoid
 	// overflows during float64 to int64 conversions.
 	partSizeFlt := math.Ceil(float64(objectSize / maxPartsCount))
-	partSizeFlt = math.Ceil(partSizeFlt/minPartSize) * minPartSize
+	partSizeFlt = math.Ceil(partSizeFlt/float64(configuredPartSize)) * float64(configuredPartSize)
```
kannappanr: Will a user configured partsize be overridden if the object cannot be fit in 10000 parts. In that case should we fail instead of changing the partsize?

harshavardhana: We should fail though, but I am thinking more like it's a hint rather than an exact part size. It's like upto this partSize but we will still use the best possible for given number of parts.

kannappanr: In `minio-py`, we fail if that is the case.

harshavardhana: Let me think more..

harshavardhana:
ping @vadmeste @kannappanr

harshavardhana:
Ping @vadmeste @kannappanr

api-put-object-common.go
```diff
@@ -67,20 +67,41 @@ func isReadAt(reader io.Reader) (ok bool) {
...
+	if int64(configuredPartSize) > objectSize {
+		err = ErrEntityTooLarge(int64(configuredPartSize), objectSize, "", "")
+		return
+	}
+
+	if objectSize > (int64(configuredPartSize) * maxPartsCount) {
+		err = ErrInvalidArgument("Part size * max_parts(10000) is lesser than input objectSize.")
+	}
+
+	if configuredPartSize < minPartSize {
```
vadmeste: @harshavardhana minPartSize is 64mb here.. so this will return an error if the user provides 5Mb as part size

harshavardhana: Oops it should be 5MiB

harshavardhana: Fixed

harshavardhana:
PTAL @vadmeste @kannappanr

api-put-object.go
```diff
@@ -147,8 +148,12 @@ func (c Client) putObjectCommon(ctx context.Context, bucketName, objectName stri
 		return c.putObjectNoChecksum(ctx, bucketName, objectName, reader, size, opts)
 	}
 
+	if opts.PartSize == 0 {
+		opts.PartSize = minPartSize
```
kannappanr: Should `minPartSize` be renamed as `defaultPartSize`? otherwise LGTM. Ttested the PR with various part sizes.

harshavardhana: Not in this PR.

api-put-object-common.go
```diff
@@ -67,20 +67,41 @@ func isReadAt(reader io.Reader) (ok bool) {
...
 		err = ErrEntityTooLarge(objectSize, maxMultipartPutObjectSize, "", "")
 		return
 	}
+
+	if int64(configuredPartSize) > objectSize {
+		err = ErrEntityTooLarge(int64(configuredPartSize), objectSize, "", "")
+		return
+	}
+
+	if objectSize > (int64(configuredPartSize) * maxPartsCount) {
```
bunin: this always returns an error for initial `objectSize = -1` and `configuredPartSize < ~524MB`, how can I upload objects of unknown size in smaller chunks?

harshavardhana: You cannot @bunin there is no such thing as unknown file size in AWS S3 protocol. We provide that as a special feature but the limitation is we stick to the maximum size possible to upload upto 5TiB.

PR #1097, opened 2019-04-09
 Add check for period and dash in bucket name

sinhaashish:
The bucket name cannot contain underscores, end with a dash, have consecutive periods, or use dashes adjacent to periods.
[Ref ](https://docs.aws.amazon.com/awscloudtrail/latest/userguide/cloudtrail-s3-bucket-naming-requirements.html)

[Fixes ](https://github.com/minio/minio/issues/7499)


kannappanr:
`DeleteMultipleObjectsHandler` in `minio` already does this check, so we can investigate if that code can be incorporated into `makeBucket`. Until then blocking this PR.

sinhaashish:
> `DeleteMultipleObjectsHandler` in `minio` already does this check, so we can investigate if that code can be incorporated into `makeBucket`. Until then blocking this PR.

The `DeleteMultipleObjectsHandler` check calls `IsValidBucketName()` . This can not be applied while creating a bucket in minio  as there is a possibility that certains users might have buckets which are non-DNS compliant in us-east-1 and we might severely restrict them by not allowing access to these buckets. Ref - http://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html


And thus we have two functions and both are defined in `minio-go` which checks for valid bucket name while creation.
1. `CheckValidBucketName` : where the bucket name can start with `.`,  `-` and  `A-Z`
2. `CheckValidBucketNameStrict` : which is restricts the creation of bucket with . and -.

sinhaashish:
@harshavardhana  @kaankabalak  PTAL

Issue #1099, opened 2019-04-19
Missing return statement in optimalPartInfo?


MrCreosote:
https://github.com/minio/minio-go/blob/80c2fe34f1a5d3ade9fc79758f8faafaa149105b/api-put-object-common.go#L87-L89

Should there be a return here? I called PutObject with a size of -1 and a part size of 5MiB and was surprised to see that a 12GB file only had 23 parts, implying a ~500MiB part size.

I assume the code fell through the error (which would be triggered by a 5TiB size [which is assigned for -1 size values] and 5MiB parts) and calculated the part size based on the maximum object size and 10k parts.

poornas:
Thanks @MrCreosote , that's a great catch. Sending a PR to fix this.

PR #1100, opened 2019-04-19
fix: add missing return in optimalPartInfo

poornas:
Add missing return in optimalPartInfo as reported in fixes: #1099

In addition, found that the optimalPartInfo calculation has regressed from using maximal partsize when client did not configure a part size. Essentially, though the default SDK minimum part size is 64Mib, sdk should use a higher part size if possible that fits other multipart constraints for optimal chunking.

### DIFF
--- a/api-put-object-common.go
+++ b/api-put-object-common.go
@@ -79,29 +79,36 @@ func optimalPartInfo(objectSize int64, configuredPartSize uint64) (totalPartsCou
 		return
 	}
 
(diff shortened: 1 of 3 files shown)


kannappanr:
Unit test `TestPartSize` needs to be fixed.

kannappanr:
@poornas Please check the Appveyor build failure

poornas:
> @poornas Please check the Appveyor build failure

fixed @kannappanr

PR #1108, opened 2019-05-22
Move all top level go files

harshavardhana:
This is to ensure go mod works properly


harshavardhana:
Closing this PR re-doing this differently.

PR #1110, opened 2019-05-23
Allow domains with . at the end

### LINKED ISSUE #1109: s3utils.IsValidDomain rejects fully qualified domain names
What happened:

s3utils.IsValidDomain returns false for FQDNs. This is a domain name that ends in a dot -- signifying that it is not relative to a DNS suffix search list.

What I expected:

s3utils.IsValidDomain should accept FQDNs

https://github.com/minio/minio-go/blob/master/pkg/s3utils/utils.go#L50

harshavardhana:
Fixes #1109


kannappanr:
Please fix Travis failures

PR #1117, opened 2019-06-03
Handle regions returned in error responses

harshavardhana:


harshavardhana:
@kannappanr @krishnasrinivas PTAL this fixes a whole bunch of issues related to our gateway implementations with limited access.

PR #1122, opened 2019-06-14
Remove delimiter if empty to support recursion on NetApp StorageGrid …

fnox:
…S3 solution.

If delimiter is set to empty StorageGrid cannot handle recursion. AWS S3 and Minio handles this
on serverside and does not care if it is empty.

### DIFF
--- a/api-list.go
+++ b/api-list.go
@@ -195,8 +195,10 @@ func (c Client) listObjectsV2Query(bucketName, objectPrefix, continuationToken s
 	// Set object prefix, prefix value to be set to empty is okay.
 	urlValues.Set("prefix", objectPrefix)
 
-	// Set delimiter, delimiter value to be set to empty is okay.
-	urlValues.Set("delimiter", delimiter)
+	// Set delimiter, only set delimiter if delimiter is not empty. This to support NetApp StorageGrid S3 implementation.
+	if delimiter != "" {
+		urlValues.Set("delimiter", delimiter)
+	}
 
 	// Set continuation token
 	if continuationToken != "" {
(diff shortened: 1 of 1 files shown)


nitisht:
Thanks for the PR @fnox

harshavardhana:
> If delimiter is set to empty StorageGrid cannot handle recursion. AWS S3 and Minio handles this
> on serverside and does not care if it is empty.

We should fix StorageGrid @fnox - we don't wish to make these changes it also introduces a regression for a fix did here

```
commit a42b0e14697ffdcb4ef223384c1cac12738f574f
Author: Harshavardhana <harsha@minio.io>
Date:   Sun Jan 20 15:35:29 2019 +0530
```

fnox:
@harshavardhana thanks for answering. I see from your mentioned commit that previous lib actually did not send prefix/delimiter if not set. I do not fully understand the following comment though from that commit:

"For example, the following condition would never work with our
ListObjects API and AWS cli would work fine."

Do you refer to minio-go lib against a minio-server or against aws s3 (or something else)?

I came to this conclusion of pr since i compared what aws cli sends for parameters compared to minio-go lib and also mc client. This was the only difference I could indentify. 

Trying to get as much info as possible when also contacting NetApp since I'm not that deep into the specifications of s3.

harshavardhana:
> Do you refer to minio-go lib against a minio-server or against aws s3 (or something else)?

Both we test with S3 first and then MinIO in policy settings it is sometimes a requirement to support delimiter params, So this PR cannot be acceptable for us as it will break existing functionality.

Issue #1125, opened 2019-06-20
is GetObject thread safe?


jackielii:
From #598 it says it's thread safe. But another comment also says user is expected to use a lock?

I just used this simple program to test it:

```
package main

import (
	"fmt"
	"log"
	"strings"
	"sync"

	"github.com/minio/minio-go/v6"
)

var (
	testContent    = "hello test minio"
	testBucketName = "test-minio"
	testObjectName = "a/b/c.txt"
	testReader     = strings.NewReader(testContent)

	endpoint        = "play.min.io:9000"
	accessKeyID     = "Q3AM3UQ867SPQQA43P2F"
	secretAccessKey = "zuf+tfteSlswRu7BJ86wekitnifILbZam1KYY3TG"
	useSSL          = true
)

func uploadFile() {
	// Initialize minio client object.
	minioClient, err := minio.New(endpoint, accessKeyID, secretAccessKey, useSSL)
	if err != nil {
		log.Fatalln(err)
	}

	// Make a new bucket called mymusic.
	location := "us-east-1"

	err = minioClient.MakeBucket(testBucketName, location)
	if err != nil {
		// Check to see if we already own this bucket (which happens if you run this twice)
		exists, err := minioClient.BucketExists(testBucketName)
		if err == nil && exists {
			log.Printf("We already own %s\n", testBucketName)
		} else {
			log.Fatalln(err)
		}
	} else {
		log.Printf("Successfully created %s\n", testBucketName)
	}

	// Upload the zip file

	// Upload the zip file with FPutObject
	n, err := minioClient.PutObject(testBucketName, testObjectName, testReader, testReader.Size(), minio.PutObjectOptions{})
	if err != nil {
		log.Fatalln(err)
	}

	log.Printf("Successfully uploaded %s of size %d\n", testObjectName, n)
}

func main() {

	fmt.Println("vim-go")

	uploadFile()

	mc, err := minio.New(endpoint, accessKeyID, secretAccessKey, useSSL)
	if err != nil {
		log.Fatalln(err)
	}

	wg := &sync.WaitGroup{}

	read := func() {
		defer wg.Done()
		f, err := mc.GetObject(testBucketName, testObjectName, minio.GetObjectOptions{})
		if err != nil {
			panic(err)
		}
		defer f.Close()

		l := len(testContent) / 2
		b := make([]byte, l)
		_, err = f.ReadAt(b, 0)
		if err != nil {
			panic(err)
		}
	}

[...]

PR #1126, opened 2019-06-20
Data race: Use a copy instead of a pointer to url.URL

kannappanr:
Fix for the data race reported in https://github.com/minio/minio-go/issues/1125

### DIFF
--- a/bucket-cache.go
+++ b/bucket-cache.go
@@ -170,7 +170,7 @@ func (c Client) getBucketLocationRequest(bucketName string) (*http.Request, erro
 	urlValues.Set("location", "")
 
 	// Set get bucket location always as path style.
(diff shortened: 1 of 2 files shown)


bucket-cache.go
```diff
@@ -170,7 +170,7 @@ func (c Client) getBucketLocationRequest(bucketName string) (*http.Request, erro
 	urlValues.Set("location", "")
 
 	// Set get bucket location always as path style.
-	targetURL := c.endpointURL
+	targetURL := *c.endpointURL
```
harshavardhana: also needs to be changed in bucket-cache_test.go

kannappanr: done

PR #1140, opened 2019-07-10
fix error message for wrong content length on putobject

### LINKED ISSUE #1135: Provide explicit error when uploaded file size doesn't match actual size
I had an upload workflow that was functional until I started resizing the image. Unfortunately `minio-go` didn't give very helpful error messages, so it took a few hours before I realized that I was providing PutObject with the wrong file size. Just requesting a little more verbosity in errors in this area to help future devs out when they run into this mistake. Thanks.

kanagarajkm:
When putObject is called with bigger ContentLength than the actual
object size, Unexpected EOF error is thrown which is not clear.
This happens only with streaming signature.

Now the SteamingReader Read will return a specific error message
saying mismatch in content length, instead of Unexpected EOF.
For example "http: ContentLength=75444 with Body length 70444"

Fixes #1135

### DIFF
--- a/pkg/s3signer/request-signature-streaming.go
+++ b/pkg/s3signer/request-signature-streaming.go
@@ -285,7 +285,7 @@ func (s *StreamingReader) Read(buf []byte) (int, error) {
 					// bytes read from baseReader different than
 					// content length provided.
 					if s.bytesRead != s.contentLen {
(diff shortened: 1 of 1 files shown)


lukepighetti:
Thank you for the quick attention.

PR #1141, opened 2019-07-10
add `EndpointURL` method to S3 client

### LINKED ISSUE #1134: Expose server properties
`mineo-js` exposes the protocol, host, and port, which makes it easy to build your own public URLs. This doesn't work on golang because the server properties are not exposed to `Minio.Client`

https://github.com/minio/minio-js/issues/588#issuecomment-318968856

```
var publicUrl = minioClient.protocol + '//' + minioClient.host + ':' + minioClient.port + '/' + minioBucket + '/' + obj.name
```

aead:
This commit adds an EndpointURL method to
the S3 client such that calling code can
get the S3 endpoint in all contexts.

Fixes #1134

### DIFF
--- a/api.go
+++ b/api.go
@@ -187,6 +187,12 @@ func NewWithOptions(endpoint string, opts *Options) (*Client, error) {
 	return privateNew(endpoint, opts.Creds, opts.Secure, opts.Region, opts.BucketLookup)
 }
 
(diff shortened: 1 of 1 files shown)


lukepighetti:
Thanks again for the fast response.

PR #1151, opened 2019-08-20
Add user metadata headers into ObjectInfo

### LINKED ISSUE #1145: return usermeta in StatObject
It seems that PutObject support UserMeta option, but no any method to get the usermeta using minio-go.

balamurugana:
Now StatObject calls returns ObjectInfo contains user metadata which
is stripped "x-amz-meta-" prefix with first value.

Fixes #1145

### DIFF
--- a/api-stat.go
+++ b/api-stat.go
@@ -176,6 +176,17 @@ func (c Client) statObject(ctx context.Context, bucketName, objectName string, o
 	if t, err := time.Parse(http.TimeFormat, expiryStr); err == nil {
 		expTime = t.UTC()
 	}
+
+	metadata := extractObjMetadata(resp.Header)
+	userMetadata := map[string]string{}
+	const xamzmeta = "x-amz-meta-"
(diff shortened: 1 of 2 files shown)


api-stat.go
```diff
@@ -187,6 +198,7 @@ func (c Client) statObject(ctx context.Context, bucketName, objectName string, o
 		// Extract only the relevant header keys describing the object.
 		// following function filters out a list of standard set of keys
 		// which are not part of object metadata.
-		Metadata: extractObjMetadata(resp.Header),
+		Metadata:     metadata,
+		UserMetadata: userMetadata,
```
poornas: Instead of a separate UserMetadata map, we can just strip the x-amz-meta- prefix in extractObjMetadata.

poornas:
@balamurugana, functional tests for custom metadata will need to be changed.

kannappanr:
ping @balamurugana

kannappanr:
ping @balamurugana

PR #1161, opened 2019-09-18
set encoding type to url in listobjects API

kannappanr:
This will fix issues seen like the one in https://github.com/minio/mc/issues/2899
Also fix formatting in functional tests


harshavardhana:
btw we should be doing this across all our SDKs @kannappanr @vadmeste @poornas

PR #1162, opened 2019-09-18
Remove validator.go inside docs

kannappanr:
Existing code cannot be used as is.
Needs to be rewritten. OK to remove it now.


poornas:
@kannappanr , can you check travis failure?

harshavardhana:
change `Makefile: checks:` @kannappanr

Makefile
```diff
@@ -13,8 +13,5 @@ test:
 examples:
 	@mkdir -p /tmp/examples && for i in $(echo examples/s3/*); do go build -o /tmp/examples/$(basename ${i:0:-3}) ${i}; done
 
-docs:
```
harshavardhana: We should change in `checks: vet test examples docs functional-test` here as well

kannappanr: done.

PR #1165, opened 2019-09-23
Fix listObjects hang when listing special characters

harshavardhana:
This PR 1e017194bd0a9f77db34525b4592e73f29ced561 introduced
a regression in handling special character file listing, this
causes `mc ls` like commands to hang.

Fixes https://github.com/minio/mc/issues/2903

### DIFF
--- a/api-list.go
+++ b/api-list.go
@@ -128,6 +128,13 @@ func (c Client) ListObjectsV2(bucketName, objectPrefix string, recursive bool, d
 
 			// If contents are available loop through and send over channel.
 			for _, object := range result.Contents {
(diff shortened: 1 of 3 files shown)


harshavardhana:
### To reproduce

```
~ for i in $(seq 1 1000); do mc cp 1++1 myminio-local1/testbucket/prefix/1++1-${i}; done
```

```
~ mc ls myminio-local1/testbucket/prefix/
```

This will hang forever listing first 1000 entries repeatedly.

harshavardhana:
@vadmeste looks like that is only for MinIO since it's an opaque value in AWS S3

PR #1166, opened 2019-09-25
Response header returned from getObject 

### LINKED ISSUE #1106: Get the Content-Range header in (Core) GetObject
While getting a range request's response, how to get the Content-Range? The ObjectInfo struct does not seem to contain it!

sinhaashish:
Closes #1106 

getObject returns the response headers which is exposed in Core.
``` 
func (c Core) GetObject(bucketName, objectName string, opts GetObjectOptions) (io.ReadCloser, ObjectInfo, http.Header, error) {
	return c.GetObjectWithContext(context.Background(), bucketName, objectName, opts)
}
 ```
The response headers are not exposed for client.
i.e. No change in the below code.
```
// GetObject - returns an seekable, readable object.
func (c Client) GetObject(bucketName, objectName string, opts GetObjectOptions) (*Object, error) {
	return c.getObjectWithContext(context.Background(), bucketName, objectName, opts)
}
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,8 @@ matrix:
 
 before_install:
   - sudo apt-get install devscripts
-  - go get github.com/minio/minio
+  - mkdir /tmp/minio
+  - (cd /tmp/minio; GO111MODULE=on go get github.com/minio/minio)
   - sudo cp testcerts/public.crt /usr/local/share/ca-certificates/
   - sudo update-ca-certificates
   - MINIO_ACCESS_KEY=minio MINIO_SECRET_KEY=minio123 ${GOPATH}/bin/minio server --compat --quiet --certs-dir testcerts data 2>&1 > minio.log &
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by sinhaashish** [comment]
The travis is failing here stating.
```
# github.com/minio/minio/cmd/gateway/s3
../../../../pkg/mod/github.com/minio/minio@v0.0.0-20190926060824-ff726969aa89/cmd/gateway/s3/gateway-s3.go:445:17: assignment mismatch: 3 variables but l.Client.GetObject returns 4 values
The command "go get github.com/minio/minio" failed and exited with 2 during .
```
The `minio-go` travis file does a `- go get github.com/minio/minio`.  Do i need to have a `minio` PR to fix this ? If yes then which PR `minio` or `minio-go` needs to be merged first .
@harshavardhana  please suggest.

harshavardhana:
what needs to be set is `GO111MODULE=on` in the `.travis.yml` such that MinIO server builds with the correct modules. That will work in this situation

sinhaashish:
@poornas @harshavardhana PTAL

PR #1169, opened 2019-09-30
list: Decode key/prefix and marker/continuation in Core layer

vadmeste:
Some modules depending on minio-go, such as Minio Gateway S3, depends
on minio-go Core to perform listing. However a regression is found
because decoding keys and common prefixes, along with marker for list v1
and next continuation for list v2 is not happening in Core module
but in higher level.

This PR moves decoding to Core level.


vadmeste:
> You need to do this for listMultipartUploads as well

Done

Issue #1177, opened 2019-10-22
ListObjects api ignore plus character "+"


soolaugust:
Hello,

When I use ListObjects api, I found it will ignore "+" character  of object key:

![image](https://user-images.githubusercontent.com/10558124/67254476-c8649580-f4af-11e9-8c9f-e795d3291dbd.png)

In minio, it shows as following:

![image](https://user-images.githubusercontent.com/10558124/67254497-ed590880-f4af-11e9-86c3-5aeda386e52d.png)

As you can see, the plus character "+" is replaced with space.

soolaugust:
After deep research. I think things happend in following codes:

https://github.com/minio/minio-go/blob/e11840609c348b5d1e46185b7f6e98ce6e6bc313/api-list.go#L259

```golang
![image](https://user-images.githubusercontent.com/10558124/67255984-627c0c00-f4b7-11e9-9073-038797226fc5.png)
```

since golang url package will handle '+' to 'escape':

```golang
		case '+':
			if mode == encodeQueryComponent {
				t[j] = ' '
			} else {
				t[j] = '+'
			}
			j++
			i++
```

However if we use `PathUnescape`, this wouldn't happen. 

So I wanna know your opinion, thanks :)

nitisht:
feel free to send a PR @soolaugust we can discuss / review

soolaugust:
@nitisht , I opened a PR for this, please help to review, thanks :)

vadmeste:
Hey @soolaugust.

I am not able to understand exactly the issue, can you describe more how to reproduce it ?

soolaugust:
@vadmeste Hi, yes, let me explain the issue.

Currently we are pushing some files name format like `xxxiW_Sag+Mass1xxx`, when I want to get using minio-go, the `ListObjects` api will return without ''+  character, like `xxxiW_Sag Mass1xxx`. and if I want use this to get the object, it will return a 404 error.

vadmeste:
The thing is that I am not able to reproduce it:

```
package main

import (
	"fmt"

	"github.com/minio/minio-go/v6"
)

func main() {
	s3Client, err := minio.New("play.min.io:9000", "Q3AM3UQ867SPQQA43P2F", "zuf+tfteSlswRu7BJ86wekitnifILbZam1KYY3TG", true)
	if err != nil {
		fmt.Println(err)
		return
	}

	// Create a done channel to control 'ListObjects' go routine.
	doneCh := make(chan struct{})

[...]

PR #1178, opened 2019-10-22
Avoid changing '+' to space when request object

soolaugust:
Fix https://github.com/minio/minio-go/issues/1177

### DIFF
--- a/api-list.go
+++ b/api-list.go
@@ -434,21 +434,21 @@ func (c Client) listObjectsQuery(bucketName, objectPrefix, objectMarker, delimit
 	}
 
 	for i, obj := range listBucketResult.Contents {
-		listBucketResult.Contents[i].Key, err = url.QueryUnescape(obj.Key)
+		listBucketResult.Contents[i].Key, err = url.PathUnescape(obj.Key)
 		if err != nil {
 			return listBucketResult, err
 		}
 	}
 
 	for i, obj := range listBucketResult.CommonPrefixes {
-		listBucketResult.CommonPrefixes[i].Prefix, err = url.QueryUnescape(obj.Prefix)
+		listBucketResult.CommonPrefixes[i].Prefix, err = url.PathUnescape(obj.Prefix)
 		if err != nil {
 			return listBucketResult, err
 		}
 	}
 
 	if listBucketResult.NextMarker != "" {
-		listBucketResult.NextMarker, err = url.QueryUnescape(listBucketResult.NextMarker)
+		listBucketResult.NextMarker, err = url.PathUnescape(listBucketResult.NextMarker)
 		if err != nil {
 			return listBucketResult, err
 		}


soolaugust:
I notice travis-ci report an error:

image

I couldn't see any issue for this, as PathUnescape  is identical to QueryUnescape except that it does not unescape '+' to ' ' (space).

So could you help to check this, thanks :)

soolaugust:
this is not the reason, will close it. more informations, please refer to https://github.com/minio/minio-go/issues/1177

aofei:
Hi @soolaugust,

Sorry to comment in a closed PR, but why is this closed? I can't think of the reason, it seems that this PR is correct.

My scenario is to connect all of my AWS S3 API compatible storage services only using [minio/minio-go](https://github.com/minio/minio-go). I encountered a series of problems these days and eventually found out that [`minio.Core. listObjectsQuery`](https://github.com/minio/minio-go/blob/3721161a9e0f2ae10a2d8f0da88938ab2eeb99c1/api-list.go#L419) converted all `+` in my filenames into spaces, which is wrong. It's all because [`minio.Core. listObjectsQuery`](https://github.com/minio/minio-go/blob/3721161a9e0f2ae10a2d8f0da88938ab2eeb99c1/api-list.go#L419) used [`url.QueryUnescape`](https://godoc.org/net/url#QueryUnescape) instead of [`url.PathUnescape`](https://godoc.org/net/url#PathUnescape).

To reproduce this, just create a file named `example.com/foo/bar/@v/v2.3.4+incompatible.info` in any AWS S3 API compatible storage service (such as GCS) and use [`minio.Client.ListObject`](https://godoc.org/github.com/minio/minio-go#Client.ListObjects) to get it.

/cc @vadmeste here from https://github.com/minio/minio-go/issues/1177#issuecomment-544855996.

soolaugust:
Hi @aofei , in your scenario, this issue will produce. because the new minio server version have correct this issue. if you want to use minio-go to connect S3 API compatible storage services, you need to correct from server side.

but I think your advice is correct, minio-go should consider such scenario. what's your opinion @vadmeste and @harshavardhana

vadmeste:
@aofei 

If you test with AWS S3, listing objects when url encoding is set to 'url' does not return key names with '+'. It returns with '%2B'

Although I am not able to find any reference about how URL encoding works exactly, I followed the behavior of AWS S3.

But I still think using QueryUnescape or PathUnescape will behave exactly the same in this case, were you testing with GCS ?

aofei:
@vadmeste 

I just checked GCS, and it also uses "%2B". Weird...

... I can't get my cloud service provider to change because I think that would cause them to break others, so they will definitely refuse anyway.

However, I still feel that using [`url.PathUnescape`](https://godoc.org/net/url#PathUnescape) is the right approach. Because we're working on paths, aren't we? Also, replacing it with [`url.PathUnescape`](https://godoc.org/net/url#PathUnescape) solved my problem.

And, of course, I understand your concern, maybe we can do some special treatments for AWS S3's "%2B" after using [`url.PathUnescape`](https://godoc.org/net/url#PathUnescape). Like this:

```go
...
listBucketResult.Contents[i].Key, err = url.PathUnescape(obj.Key)
...
listBucketResult.Contents[i].Key = strings.Replace(listBucketResult.Contents[i].Key, "%%2B", " ", -1)
...
```

This way we can adapt to both situations. And this isn't a stupid solution, because the only difference between [`url.QueryUnescape`](https://godoc.org/net/url#QueryUnescape) and [`url.PathUnescape`](https://godoc.org/net/url#PathUnescape) is that [`url.PathUnescape`](https://godoc.org/net/url#PathUnescape) does not unescape '+' to ' ' (space).

What do you think?

harshavardhana:
@aofei if your provider doesn't work like AWS S3 then we can't do anything here. Please ask them to fix this on server.

aofei:
> @aofei if your provider doesn't work like AWS S3 then we can't do anything here. Please ask them to fix this on server.

This is really a great reply, it perfectly ignores my concerns and my proposal, thank you very much @harshavardhana.

And @vadmeste, just like you said, there doesn't seem to be a reference about how URL encoding works exactly, which shows that it may not be "standardized". So do you think we should consider the situation that using "+" instead of "%2B"?

And FYI, the [`aws/aws-sdk-go`](github.com/aws/aws-sdk-go)'s [`s3.ListObjects`](https://godoc.org/github.com/aws/aws-sdk-go/service/s3#ListObjectsInput) worked as expected. But I love [`minio/minio-go`](https://github.com/minio/minio-go) more (the [`aws/aws-sdk-go`](github.com/aws/aws-sdk-go) is too ugly). So please think about it. 🥺

harshavardhana:
@aofei you need to use `url` encoding  with aws-sdk-go - are you sure that is how you used?

PR #1180, opened 2019-10-25
Set max-keys query param only as needed

harshavardhana:
This is to ensure that S3 server implementations
can send a larger list response by default
i.e more than 1000 keys if needed.

### DIFF
--- a/api-list.go
+++ b/api-list.go
@@ -1,6 +1,6 @@
 /*
  * MinIO Go Library for Amazon S3 Compatible Cloud Storage
- * Copyright 2015-2017 MinIO, Inc.
+ * Copyright 2015-2019 MinIO, Inc.
  *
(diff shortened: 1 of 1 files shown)


api-list.go
```diff
@@ -208,12 +208,10 @@ func (c Client) listObjectsV2Query(bucketName, objectPrefix, continuationToken s
 		urlValues.Set("fetch-owner", "true")
 	}
 
-	// maxkeys should default to 1000 or less.
-	if maxkeys == 0 || maxkeys > 1000 {
```
kannappanr: Should `maxKeys` be set to `1000` if it is `<=0`

harshavardhana: No, the intention is to avoid setting such that server implementations can send their default limit. Clients don't need to limit itself. 

It is a change that we shall do on the server as well where we can send 10000 entries instead of 1000

PR #1181, opened 2019-10-28
Set default part size

### LINKED ISSUE #1176: Memleak-ish behavior of PutObjectWithContext
Hi,
I'm uploading a lot of data from a machine that has just couple of Gs of RAM.
When I call `PutObjectWithContext` GO program gets killed by OOM.

I made some experiments end it turns out that if I configure the GC to run more often it seems to work, but it isn't a real solution.

It seems that minio has somewhere a buffer that gets allocated on every read and waits for the GC to be freed. Maybe you could just use one buffer and reuse it?

klauspost:
Set the part size in the options we are passing along.

`opts` are a copy, so we are not affecting anything upstream.

Fixes #1176

### DIFF
--- a/api-put-object.go
+++ b/api-put-object.go
@@ -148,10 +148,10 @@ func (c Client) putObjectCommon(ctx context.Context, bucketName, objectName stri
 		return c.putObjectNoChecksum(ctx, bucketName, objectName, reader, size, opts)
 	}
 
-	partSize := opts.PartSize
 	if opts.PartSize == 0 {
-		partSize = minPartSize
+		opts.PartSize = minPartSize
 	}
+	partSize := opts.PartSize
 
 	if c.overrideSignerType.IsV2() {
 		if size >= 0 && size < int64(partSize) {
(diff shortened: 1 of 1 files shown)


harshavardhana:
This is not correct @klauspost when the size is -1 we should support uploading maximum object size possible by default. You are avoiding that functionality in this case the upload size will be limited to 1.5TiB instead of 5TiB which is not acceptable.

The issue #1176 is actually is bit of a stretch and not necessarily needs to be fixed IMO. 1G memory limit is not a big requirement we should simply close the issue either specify the content-length or pick a partSize 

It's not our problem to address situations when memory is 1G

klauspost:
Ah, ok, so it is because of the maximum number of parts.

Makes sense and there is a reasonable way around it.

I suppose the only reason for using a buffer is so it is possible to [retry the upload of a part](https://github.com/minio/minio-go/blob/e11840609c348b5d1e46185b7f6e98ce6e6bc313/api.go#L565).

harshavardhana:
> I suppose the only reason for using a buffer is so it is possible to [retry the upload of a part](https://github.com/minio/minio-go/blob/e11840609c348b5d1e46185b7f6e98ce6e6bc313/api.go#L565).

Correct @klauspost since we need to re-read the body anyways

Issue #1156, opened 2019-09-09
Support Web Identity providers for IAM credentials for AWS EKS


barryib:
Add support for Web Identity providers for IAM credentials for AWS EKS (IAM Roles for Service Accounts) https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html

This will be very useful in thanos https://github.com/thanos-io/thanos/issues/1494 running into an EKS cluster.

harshavardhana:
We already support that @barryib

harshavardhana:
https://godoc.org/github.com/minio/minio-go/pkg/credentials#NewIAM

kannappanr:
@barryib Closing this issue now. Please feel free to reach out to us if you have other such questions at https://slack.min.io

barryib:
I don't think it implemented. This is a new way (which come out few days ago) AWS provide IAM role to kubernetes pods and service account in EKS cluster.

This method add new env vars like

```
AWS_ROLE_ARN=arn:aws:iam::AWS_ACCOUNT_ID:role/IAM_ROLE_NAME
AWS_WEB_IDENTITY_TOKEN_FILE=/var/run/secrets/eks.amazonaws.com/serviceaccount/token
```

This was added recently in all AWS SDK. Here is the PR for the [aws-sdk-go](https://github.com/aws/aws-sdk-go/pull/2667)

More info here:
- https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts-technical-overview.html
- https://github.com/aws/containers-roadmap/issues/23
- https://github.com/aws/containers-roadmap/issues/23#issuecomment-529840651
- https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts-minimum-sdk.html

harshavardhana:
@barryib this is very specific to AWS - feel free to send a PR, we may not have cycles to address this as this moment.

kannappanr:
@barryib Closing this issue now. Please feel free to send a PR when you get a chance

PR #1183, opened 2019-11-11
Credentials: Support assuming role via WebIdentityTokenFile

saracen:
This supports the new `AWS_WEB_IDENTITY_TOKEN_FILE` and `AWS_ROLE_ARN` environment variables that allow exchanging OIDC tokens given to pods in EKS for access tokens.

I haven't introduced any new API for this, as I figured the use-case for this will be solely EKS. So this privately extends the existing `STSWebIdentity` credential provider for AWS use and is only activated when the correct environment variables are provided. Calling NewIAM() is all that is required.

I haven't actually tested this in an AWS environment yet, but figured it was worth opening for discussion early.

@barryib You requested this in #1156 - are you able to test in an AWS environment?

### DIFF
--- a/pkg/credentials/sts_web_identity.go
+++ b/pkg/credentials/sts_web_identity.go
@@ -121,7 +121,9 @@ func getWebIdentityCredentials(clnt *http.Client, endpoint, roleARN, roleSession
 		v.Set("RoleSessionName", roleSessionName)
 	}
 	v.Set("WebIdentityToken", idToken.Token)
-	v.Set("DurationSeconds", fmt.Sprintf("%d", idToken.Expiry))
+	if idToken.Expiry > 0 {
+		v.Set("DurationSeconds", fmt.Sprintf("%d", idToken.Expiry))
+	}
 	v.Set("Version", "2011-06-15")
 
 	u, err := url.Parse(endpoint)
--- a/pkg/credentials/sts_web_identity.go
+++ b/pkg/credentials/sts_web_identity.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"strconv"
 	"time"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/credentials/sts_web_identity.go
```diff
@@ -121,7 +121,11 @@ func getWebIdentityCredentials(clnt *http.Client, endpoint, roleARN, roleSession
 		v.Set("RoleSessionName", roleSessionName)
 	}
 	v.Set("WebIdentityToken", idToken.Token)
-	v.Set("DurationSeconds", fmt.Sprintf("%d", idToken.Expiry))
+	if idToken.Expiry > 0 {
+		v.Set("DurationSeconds", fmt.Sprintf("%d", idToken.Expiry))
+	} else {
+		v.Set("DurationSeconds", "3600")
```
harshavardhana: default a default global constant for this @saracen

saracen: I've updated. The alternative would be to omit DurationSeconds entirely as it is an optional parameter for AWS' implementation. But I don't know whether that's true for Minio's.

harshavardhana: That is true for MinIO as well @saracen

saracen: Okay, I'll update shortly to exclude sending the value if it is zero.

harshavardhana:
PTAL @sinhaashish @kanagarajkm

jujugrrr:
Hi @harshavardhana  @sinhaashish  @kanagarajkm, do you have any news about this PR? It's a great feature as EKS + WebIdentity is the way to go for secure IAM credentials on AWS.

jujugrrr:
Awesome, thank you

jujugrrr:
Do you know when could we see it in v6.0.45 @kanagarajkm ?

harshavardhana:
@jujugrrr soon no ETA

jujugrrr:
thank you

PR #1188, opened 2019-11-18
Implemented support for Scaleway object storage endpoints

wintamute:
I'm using the Scaleway object storage (https://www.scaleway.com/en/docs/object-storage-feature/) which is a S3 compatible storage provider. It uses regions, but no common endpoint, so I added support for this provider which automatically sets correct region based on endpoint URL.

My main reason is to use restic and minio-client (mc), since restic doesn't allow to set regions manually. I built and tested both tools against my modified sdk and both work fine.
Without these changes, the requests fail due to wrong region in the SigV4 header (defaults to us-east-1).

I also included tests and ran the included test suite.

### DIFF
--- a/api-put-bucket.go
+++ b/api-put-bucket.go
@@ -53,6 +53,12 @@ func (c Client) makeBucket(bucketName string, location string, objectLockEnabled
 			location = c.region
 		}
 	}
+	// For Scaleway endpoints, always us the custom region
+	if s3utils.IsScalewayEndpoint(*c.endpointURL) {
+		if c.region != "" {
+			location = c.region
+		}
+	}
 	// PUT bucket request metadata.
 	reqMetadata := requestMetadata{
 		bucketName:     bucketName,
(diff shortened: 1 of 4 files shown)


harshavardhana:
This is going to be a burden for us to carry @wintamute, we are not interested in expanding the scope here for scaleway.  what is the issue that you are trying to fix here?

wintamute:
The main reason is that restic doesn't support setting a region which is required for the V4 sig.
Unfortunately, there is no env variable for setting the region either.
For me the easiest way to support it was to add to Minio-go directly which also makes it possible for any other tool based on Minio-go to just work.
If you don't want to support more provides that's ok, I basically just wanted to share my changes.
So what's the best approach to support more s3 providers in a generic way?

harshavardhana:
Best approach is to let restic set a region.

wintamute:
Ok, setting the region just got merged into restic, so everything is working now without requiring changes to Minio-go.
A big thank you to the Minio team and contributors for the great tools.

PR #1189, opened 2019-11-21
Support ListObjectsWithMetadata API

harshavardhana:


harshavardhana:
Of course, needs server PR https://github.com/minio/minio/pull/8539

api-list.go
```diff
@@ -319,7 +360,7 @@ func (c Client) ListObjects(bucketName, objectPrefix string, recursive bool, don
 		var marker string
 		for {
 			// Get list of objects a maximum of 1000 per request.
```
kannappanr: What is the maximum returned list size per request? Is it 10000?

api-list.go
```diff
@@ -600,11 +641,10 @@ func (c Client) listMultipartUploadsQuery(bucketName, keyMarker, uploadIDMarker,
 	urlValues.Set("encoding-type", "url")
 
 	// maxUploads should be 1000 or less.
-	if maxUploads == 0 || maxUploads > 1000 {
-		maxUploads = 1000
+	if maxUploads > 0 {
```
kannappanr: Should there be a max number?

api-list.go
harshavardhana: maximum depends on server implementation @kannappanr

api-list.go
harshavardhana: maximum depends on server implementation @kannappanr

PR #1198, opened 2019-12-09
Retry for all network errors regardless

harshavardhana:


api.go
```diff
@@ -627,12 +627,7 @@ func (c Client) executeMethod(ctx context.Context, method string, metadata reque
 		// Initiate the request.
 		res, err = c.do(req)
 		if err != nil {
-			// For supported http requests errors verify.
-			if isHTTPReqErrorRetryable(err) {
-				continue // Retry.
-			}
-			// For other errors, return here no need to retry.
-			return nil, err
+			continue
```
krishnasrinivas: this will retry for "unable to resolve host" as well? in any case we retry for a certain number of times and return error?

harshavardhana: yes it is 10 times @krishnasrinivas

harshavardhana: We only retry at max 10 times

PR #1201, opened 2019-12-10
add maxRetry in NewWithOptions

illi-j:
This PR add `maxRetry` in the `Client` and `Options`  for indicates how many times we can retry the request


harshavardhana:
This is not needed, you can just set minio.MaxRetry in your init() function, let's not add new options with prior discussion on https://slack.minio.io

PR #1215, opened 2020-01-16
fix: allow configurable part size with -1 content-length

### LINKED ISSUE #1214: So this workaround with custom `PartSize` should work?
> This can be definitely reduced by using the PartSize as mentioned by @klauspost so if you did override the auto calculation by specifying a part size - you have to assume that your total upload size automatically reduces based on the part size.

@harshavardhana so this workaround with custom `PartSize` should work? I'm getting error with it
```go
t := time.Now().Add(retention)
opts := minio.PutObjectOptions{
    RetainUntilDate: &t,
    ContentType:     "text/csv",
    PartSize:        64 * 1024 * 1024,
}
_, err := client.PutObjectWithContext(ctx, "bucket", "key", reader, -1, opts)
if err {
    log.Fatal(err)
}
```
```
Part size * max_parts(10000) is lesser than input objectSize.
```

_Originally posted by @tetafro in https://github.com/minio/minio-go/issues/1176#issuecomment-562472216_

harshavardhana:
Fixes #1214

### DIFF
--- a/api-put-object-common.go
+++ b/api-put-object-common.go
@@ -107,7 +107,7 @@ func optimalPartInfo(objectSize int64, configuredPartSize uint64) (totalPartsCou
 
 		partSizeFlt = float64(configuredPartSize)
 		if unknownSize {
-			// If input has unkown size and part size is configured
+			// If input has unknown size and part size is configured
 			// keep it to maximum allowed as per 10000 parts.
 			objectSize = int64(configuredPartSize) * maxPartsCount
 		}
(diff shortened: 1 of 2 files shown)


api-put-object-common.go
```diff
@@ -100,7 +104,13 @@ func optimalPartInfo(objectSize int64, configuredPartSize uint64) (totalPartsCou
 			err = ErrInvalidArgument("Input part size is bigger than allowed maximum of 5GiB.")
 			return
 		}
+
 		partSizeFlt = float64(configuredPartSize)
+		if unknownSize {
+			// If input has unkown size and part size is configured
```
krisis: typo:
```suggestion
			// If input has unknown size and part size is configured
```

harshavardhana:
@poornas PTAL

PR #1218, opened 2020-01-20
fix: let net/http handle rootCAs properly by default

### LINKED ISSUE #1191: Not able to connect to server with SSL on Windows
When I try to connect to the test server with SSL as in the example in the README I get the following error (without SSL it works)
`Get https://play.min.io/: x509: certificate signed by unknown authority`

I found that it was the `crypto/x509.Verify` function that threw it. 
On line [749](https://github.com/golang/go/blob/acf3ff2e8a0ee777a35b42879c90a1d5a130988f/src/crypto/x509/verify.go#L749) of the `crypto/x509.Verify` function there is a check to use Windows own verification if `rootCAs` is nil.

The problem is that in [DefaultTransport](https://github.com/minio/minio-go/blob/7bb4ac4dfa8be91227211bca4c0386e591d828b5/transport.go#L62) the `rootCAs` is set to an empty pool and causing it not to use the Windows verify function.

I tested this by adding this code and I can successfully connect to the server and other servers:
```go
if runtime.GOOS == "windows" {
	transport, _ := minio.DefaultTransport(true)
	transport.(*http.Transport).TLSClientConfig.RootCAs = nil
	api.SetCustomTransport(transport)
}
```

harshavardhana:
Fixes #1191

### DIFF
--- a/transport.go
+++ b/transport.go
@@ -21,7 +21,6 @@ package minio
 
 import (
 	"crypto/tls"
-	"crypto/x509"
 	"net"
 	"net/http"
 	"time"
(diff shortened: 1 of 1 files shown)


vadmeste:
I guess that code was copied from mc but it is true it is not needed in minio-go, only mc needs it because mc loads certificates from `.mc\certs\CAs` which makes it hard to support both Windows local CA store & CAs in `.mc\certs\CAs`

PR #1221, opened 2020-01-24
fix: retry AccessDenied only if Region is present

harshavardhana:


kannappanr:
I am unable to reproduce this against play

harshavardhana:
The issue needs to reproduced with restricted policies @kannappanr where MakeBucket is disallowed but GetBucketLocation is allowed.

PR #1222, opened 2020-01-28
fix testListObjects functional test

poornas:
- Remove test for invalid xml character \x17 in object name

The above test fails with azure gateway storage client on Mint during response parsing because of the invalid xml character in object name.


harshavardhana:
Unfortunately, a new release is needed to get this issue fixed with mint

PR #1224, opened 2020-01-31
Make endpoint field of IAM public

### LINKED ISSUE #1223: IAM credentials broken
On 6.0.45 I am unable to get IAM-based credentials working. The problem seems to be in [getIAMRoleURL](https://github.com/minio/minio-go/blob/a8fa81f2c36397fc9dcc622f4a121776b1f15312/pkg/credentials/iam_aws.go#L170), where if the endpoint is empty, the URL consists of only the path and not the host (e.g. `/latest/meta-data/iam/security-credentials/`). I am unable to locate any usage of `defaultIAMRoleEndpoint`, and it would make sense to use it if the endpoint is an empty string.

radix-aw:
This is needed to resolve #1223.


radix-aw:
Per @harshavardhana's [feedback](https://github.com/minio/minio/pull/8921#issuecomment-580976226), I've changed solution implementation.

PR #1227, opened 2020-02-11
Bucket and object name length error message fixes

scop:


kannappanr:
the existing error messages are the ones returned by AWS S3. I think it is better to keep them as is.

harshavardhana:
Can we check ? @kannappanr it looks that might not be the case

kannappanr:
Yes, that is not the case. I initially tried it against AWS s3 using `mc` and thought this was returned by AWS s3

PR #1237, opened 2020-03-04
list: Use PathUnescape when decoding object names

vadmeste:
When url-encoding is set to 'url', minio-go should avoid decoding plain '+'
to a space in object names.

Although, '+' is generally encoded to %2B in the listing response (see AWS
response), we still can tolerate plain '+' as well.


vadmeste:
Closing, as the PR is wrong

PR #1240, opened 2020-03-05
Add (Get/Set)ObjectLockConfig API functions

poornas:
Also reverts change from #1233 as this is unused
and breaks API.

### DIFF
--- a/api-object-lock.go
+++ b/api-object-lock.go
@@ -183,8 +183,8 @@ func (c Client) SetBucketObjectLockConfig(bucketName string, mode *RetentionMode
 	return nil
 }
 
-// GetBucketObjectLockConfig gets object lock configuration of given bucket.
-func (c Client) GetBucketObjectLockConfig(bucketName string) (objectLock string, mode *RetentionMode, validity *uint, unit *ValidityUnit, err error) {
+// GetObjectLockConfig gets object lock configuration of given bucket.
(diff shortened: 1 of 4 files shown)


vadmeste:
https://docs.aws.amazon.com/AmazonS3/latest/API/archive-Type_API_ObjectLockConfiguration.html

This shows that ObjectLockEnabled is required but other fields are not required. I would say let's avoid merging this PR until we understand what means to send an xml with only ObjectLockEnabled in it

api-object-lock.go
```diff
@@ -224,9 +224,18 @@ func (c Client) GetBucketObjectLockConfig(bucketName string) (objectLock string,
...
-
-		return config.ObjectLockEnabled, mode, validity, unit, nil
+		return config.ObjectLockEnabled, validity, unit, nil
 	}
+	return config.ObjectLockEnabled, nil, nil, nil
+}
+
+// GetBucketObjectLockConfig gets object lock configuration of given bucket.
+func (c Client) GetBucketObjectLockConfig(bucketName string) (mode *RetentionMode, validity *uint, unit *ValidityUnit, err error) {
+	_, mode, validity, unit, err := c.GetObjectLockConfig(bucketName)
```
kannappanr:
```suggestion
	_, mode, validity, unit, err = c.GetObjectLockConfig(bucketName)
```

PR #1242, opened 2020-03-09
PutObject change. Adding usertags to call. Reqd for minio PR #8901.

BigUstad:


BigUstad:
Another PR will have the header checks.

PR #1247, opened 2020-03-16
fix: Context cancellation not handled

fortyanov:


retry.go
```diff
@@ -41,7 +42,7 @@ const DefaultRetryCap = time.Second * 30
 
 // newRetryTimer creates a timer with exponentially increasing
 // delays until the maximum retry attempts are reached.
-func (c Client) newRetryTimer(maxRetry int, unit time.Duration, cap time.Duration, jitter float64, doneCh chan struct{}) <-chan int {
+func (c Client) newRetryTimer(ctx context.Context, maxRetry int, unit time.Duration, cap time.Duration, jitter float64, doneCh chan struct{}) <-chan int {
```
vadmeste: It is also possible to remove doneCh here and use context instead

fortyanov:
@vadmeste hello. I made changes what did you say about.

retry.go
```diff
@@ -69,14 +70,13 @@ func (c Client) newRetryTimer(maxRetry int, unit time.Duration, cap time.Duratio
 	go func() {
 		defer close(attemptCh)
 		for i := 0; i < maxRetry; i++ {
+			attemptCh <- i + 1
+
```
harshavardhana: attemptCh should be kept as part of select otherwise your doneCh is never triggered..

fortyanov: I remove  doneCh because it not needer any more. Context is uses for breaking retries now.

harshavardhana: No the correct code should be to keep the attemptCh as part of the select @fortyanov - it won't work as expected for ctx.Done() is called.

harshavardhana: It should be 

```
select {
   case attemptCh <- i+1:
   case <-time.After(exponentialBackoffWait(i)):
   case ctx.Done():
            return
}
```

fortyanov: I made this changes.

retry.go
```diff
@@ -70,13 +71,11 @@ func (c Client) newRetryTimer(maxRetry int, unit time.Duration, cap time.Duratio
 		defer close(attemptCh)
 		for i := 0; i < maxRetry; i++ {
 			select {
-			// Attempts start from 1.
 			case attemptCh <- i + 1:
-			case <-doneCh:
-				// Stop the routine.
+			case <-time.After(exponentialBackoffWait(i)):
```
vadmeste: We can send wrong attempt number here after some timeout.

The better way is:
```
select {
case attemptCh <- i+1:
case <-ctx.Done():
   return 
}

case {
case <-time.After(exponentialBackoffWait(i)):
case <-ctx.Done():
   return
}
```

harshavardhana: This is the ideal fix

fortyanov:
@vadmeste @harshavardhana @nitisht made new PR because recreate fork for fresh minio version.

PR #1248, opened 2020-03-16
remove object tagging including status code 204/No Content

BigUstad:
DeleteObjectTagging Response HTTP/1.1 204.

### DIFF
--- a/api-object-tagging.go
+++ b/api-object-tagging.go
@@ -155,7 +155,8 @@ func (c Client) RemoveObjectTaggingWithContext(ctx context.Context, bucketName, 
 	}
 
 	if resp != nil {
-		if resp.StatusCode != http.StatusOK {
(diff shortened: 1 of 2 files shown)


api-object-tagging.go
```diff
@@ -155,7 +155,8 @@ func (c Client) RemoveObjectTaggingWithContext(ctx context.Context, bucketName,
 	}
 
 	if resp != nil {
-		if resp.StatusCode != http.StatusOK {
+		// S3 returns "204 No content" after Object tag deletion.
+		if resp.StatusCode != http.StatusOK && resp.StatusCode != http.StatusNoContent {
```
harshavardhana: We should perhaps only look for 204 and catch offending non compliant servers? @BigUstad

BigUstad: I will make this check.

PR #1249, opened 2020-03-17
extract userTags from Get/Head request

harshavardhana:


harshavardhana:
If it doesn't then we don't have to do anything @nitisht which is what the PR does. The reason it's needed is to be able to extract the tags for mc to be able to copy to the destination.

harshavardhana:
PTAL @kannappanr

api-datatypes.go
```diff
@@ -82,6 +82,9 @@ type ObjectInfo struct {
 	// x-amz-meta-* headers stripped "x-amz-meta-" prefix containing the first value.
 	UserMetadata StringMap `json:"userMetadata"`
 
+	// x-amz-tagging values in their k/v values.
+	UserTags map[string]string `json:"userTags"`
```
kannappanr: This is not being returned as part of get/head object call

kannappanr:
> If it doesn't then we don't have to do anything @nitisht which is what the PR does. The reason it's needed is to be able to extract the tags for mc to be able to copy to the destination.

By default the tags will be copied to the destination. This looks more like an option that we will set here, so we can use the `REPLACE` directive in `copyobject`

api-datatypes.go
harshavardhana: This is an extension proposal, to be returned as part of get/head call to avoid making another GetObjectTag call @kannappanr to facilitate faster mirror operations.

PR #1250, opened 2020-03-17
fix: Context cancellation not handled

fortyanov:
@vadmeste @harshavardhana @nitisht made new PR because recreate fork for fresh minio version.

### DIFF
--- a/api-get-object.go
+++ b/api-get-object.go
@@ -627,21 +627,18 @@ func (c Client) getObject(ctx context.Context, bucketName, objectName string, op
 	if err != nil {
 		return nil, ObjectInfo{}, nil, err
 	}
-
-	var header http.Header
 	if resp != nil {
 		if resp.StatusCode != http.StatusOK && resp.StatusCode != http.StatusPartialContent {
 			return nil, ObjectInfo{}, nil, httpRespToErrorResponse(resp, bucketName, objectName)
 		}
-		header = resp.Header
 	}
 
-	objectStat, err := ToObjectInfo(bucketName, objectName, header)
+	objectStat, err := ToObjectInfo(bucketName, objectName, resp.Header)
 	if err != nil {
 		closeResponse(resp)
-		return nil, objectStat, header, nil
+		return nil, objectStat, resp.Header, nil
 	}
 
 	// do not close body here, caller will close
-	return resp.Body, objectStat, header, nil
+	return resp.Body, objectStat, resp.Header, nil
 }
--- a/api-get-object.go
+++ b/api-get-object.go
@@ -627,18 +627,21 @@ func (c Client) getObject(ctx context.Context, bucketName, objectName string, op
 	if err != nil {
 		return nil, ObjectInfo{}, nil, err
 	}
+
+	var header http.Header
 	if resp != nil {
 		if resp.StatusCode != http.StatusOK && resp.StatusCode != http.StatusPartialContent {
 			return nil, ObjectInfo{}, nil, httpRespToErrorResponse(resp, bucketName, objectName)
 		}
+		header = resp.Header
 	}
 
-	objectStat, err := ToObjectInfo(bucketName, objectName, resp.Header)
+	objectStat, err := ToObjectInfo(bucketName, objectName, header)
 	if err != nil {
 		closeResponse(resp)
-		return nil, objectStat, resp.Header, nil
+		return nil, objectStat, header, nil
 	}
 
 	// do not close body here, caller will close
-	return resp.Body, objectStat, resp.Header, nil
+	return resp.Body, objectStat, header, nil
 }
--- a/api-put-object-streaming.go
+++ b/api-put-object-streaming.go
@@ -437,15 +437,18 @@ func (c Client) putObjectDo(ctx context.Context, bucketName, objectName string, 
 	if err != nil {
 		return ObjectInfo{}, err
 	}
+
+	var eTag string
 	if resp != nil {
 		if resp.StatusCode != http.StatusOK {
 			return ObjectInfo{}, httpRespToErrorResponse(resp, bucketName, objectName)
 		}
+		eTag = trimEtag(resp.Header.Get("ETag"))
 	}
 
 	var objInfo ObjectInfo
 	// Trim off the odd double quotes from ETag in the beginning and end.
-	objInfo.ETag = trimEtag(resp.Header.Get("ETag"))
+	objInfo.ETag = eTag
 	// A success here means data was written to server successfully.
 	objInfo.Size = size
 
--- a/api-put-object-streaming.go
+++ b/api-put-object-streaming.go
@@ -437,18 +437,15 @@ func (c Client) putObjectDo(ctx context.Context, bucketName, objectName string, 
 	if err != nil {
 		return ObjectInfo{}, err
 	}
-
-	var eTag string
 	if resp != nil {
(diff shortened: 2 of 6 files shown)


api.go
```diff
@@ -591,16 +591,10 @@ func (c Client) executeMethod(ctx context.Context, method string, metadata reque
...
-	doneCh := make(chan struct{}, 1)
-
-	// Indicate to our routine to exit cleanly upon return.
-	defer close(doneCh)
-
 	// Blank indentifier is kept here on purpose since 'range' without
 	// blank identifiers is only supported since go1.4
 	// https://golang.org/doc/go1.4#forrange.
-	for range c.newRetryTimer(reqRetry, DefaultRetryUnit, DefaultRetryCap, MaxJitter, doneCh) {
+	for range c.newRetryTimer(ctx, reqRetry, DefaultRetryUnit, DefaultRetryCap, MaxJitter) {
```
vadmeste: You need something similar to what was doneCh doing, which is stopping the internal goroutine for newRetryTimer.

For that you need to fork the context and call its cancel function when executeMethod() is quitting (`defer cancel()`)
So what you need is to create a new context from the passed context and run `defer cancel()`

retry.go
```diff
@@ -43,6 +43,9 @@ const DefaultRetryCap = time.Second * 30
 // newRetryTimer creates a timer with exponentially increasing
 // delays until the maximum retry attempts are reached.
 func (c Client) newRetryTimer(ctx context.Context, maxRetry int, unit time.Duration, cap time.Duration, jitter float64) <-chan int {
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
```
fortyanov: @vadmeste hope i correctly understand what you means.

retry.go
```diff
@@ -41,7 +42,10 @@ const DefaultRetryCap = time.Second * 30
 
 // newRetryTimer creates a timer with exponentially increasing
 // delays until the maximum retry attempts are reached.
-func (c Client) newRetryTimer(maxRetry int, unit time.Duration, cap time.Duration, jitter float64, doneCh chan struct{}) <-chan int {
+func (c Client) newRetryTimer(ctx context.Context, maxRetry int, unit time.Duration, cap time.Duration, jitter float64) <-chan int {
+	ctx, cancel := context.WithCancel(ctx)
```
vadmeste: This needs to be done in executeMethod() instead.

Because we need executeMethod() to tell the retry timer to stop when executeMethod is going to stop

retry.go
```diff
@@ -43,9 +43,6 @@ const DefaultRetryCap = time.Second * 30
 // newRetryTimer creates a timer with exponentially increasing
 // delays until the maximum retry attempts are reached.
 func (c Client) newRetryTimer(ctx context.Context, maxRetry int, unit time.Duration, cap time.Duration, jitter float64) <-chan int {
-	ctx, cancel := context.WithCancel(ctx)
-	defer cancel()
```
fortyanov: @vadmeste done

api.go
```diff
@@ -591,6 +591,12 @@ func (c Client) executeMethod(ctx context.Context, method string, metadata reque
 		}
 	}
 
+	// Create cancel context to control 'newRetryTimer' go routine.
+	ctx, cancel := context.WithCancel(ctx)
+
+	// Indicate to our routine to exit cleanly upon return.
+	defer cancel()
+
```
fortyanov: @vadmeste done

api.go
```diff
@@ -591,16 +591,16 @@ func (c Client) executeMethod(ctx context.Context, method string, metadata reque
 		}
 	}
 
-	// Create a done channel to control 'newRetryTimer' go routine.
-	doneCh := make(chan struct{}, 1)
+	// Create cancel context to control 'newRetryTimer' go routine.
+	ctx, cancel := context.WithCancel(ctx)
```
vadmeste: One last thing. I think it is cleaner if you call the new variable retryCtx so it won't interfere with other contexts

api.go
```diff
@@ -592,15 +592,15 @@ func (c Client) executeMethod(ctx context.Context, method string, metadata reque
 	}
 
 	// Create cancel context to control 'newRetryTimer' go routine.
-	ctx, cancel := context.WithCancel(ctx)
+	retryCtx, cancel := context.WithCancel(ctx)
```
fortyanov: @vadmeste ready

harshavardhana: Can you fix the build issues ?

kannappanr:
@fortyanov Can you please fix the build issues?

fortyanov:
Hello, i fix travis build issue, but have no idea how to fix appveyor issues

retry.go
```diff
@@ -72,8 +72,6 @@ func (c Client) newRetryTimer(ctx context.Context, maxRetry int, unit time.Durat
 		for i := 0; i < maxRetry; i++ {
 			select {
 			case attemptCh <- i + 1:
-			case <-ctx.Done():
-				return
```
fortyanov: Needs to remove it because FPutObjectWithContext should fail on short timeout, and this case hold it and avoid to receive error.

api-get-object.go
```diff
@@ -627,18 +627,21 @@ func (c Client) getObject(ctx context.Context, bucketName, objectName string, op
...
 		}
+		header = resp.Header
 	}
 
-	objectStat, err := ToObjectInfo(bucketName, objectName, resp.Header)
+	objectStat, err := ToObjectInfo(bucketName, objectName, header)
 	if err != nil {
 		closeResponse(resp)
-		return nil, objectStat, resp.Header, nil
+		return nil, objectStat, header, nil
```
fortyanov: I found that there is situation when resp == nil,  calling `resp.Header` raise panic

api-put-object-streaming.go
```diff
@@ -437,15 +437,18 @@ func (c Client) putObjectDo(ctx context.Context, bucketName, objectName string,
 	if err != nil {
 		return ObjectInfo{}, err
 	}
+
+	var eTag string
 	if resp != nil {
 		if resp.StatusCode != http.StatusOK {
 			return ObjectInfo{}, httpRespToErrorResponse(resp, bucketName, objectName)
 		}
+		eTag = trimEtag(resp.Header.Get("ETag"))
```
fortyanov: I found that there is situation when resp == nil, calling resp.Header raise panic

api-stat.go
```diff
@@ -100,11 +100,14 @@ func (c Client) statObject(ctx context.Context, bucketName, objectName string, o
...
+	var header http.Header
 	if resp != nil {
 		if resp.StatusCode != http.StatusOK && resp.StatusCode != http.StatusPartialContent {
 			return ObjectInfo{}, httpRespToErrorResponse(resp, bucketName, objectName)
 		}
+		header = resp.Header
 	}
 
-	return ToObjectInfo(bucketName, objectName, resp.Header)
+	return ToObjectInfo(bucketName, objectName, header)
```
fortyanov: I found that there is situation when resp == nil, calling resp.Header raise panic

api-get-object.go
harshavardhana: Wait it is guaranteed that resp is non nil if err is nil, I think there is some other bug internally which needs to be fixed.

This is introduced by this change @fortyanov

fortyanov: Got it, revert tests back.

fortyanov: Got it, revert tests back.

PR #1254, opened 2020-03-23
Add Object Retention related flags for CopyObject API

nitisht:


api-put-object-copy.go
```diff
@@ -22,11 +22,14 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+	"time"
 
 	"github.com/minio/minio-go/v6/pkg/encrypt"
 	"github.com/minio/minio-go/v6/pkg/s3utils"
 )
 
+var timeSentinel = time.Unix(0, 0).UTC()
```
klauspost: So this converts unit time 0 at (server) local time to UTC... which will not be the same for all servers.

api-put-object-copy.go
```diff
@@ -49,6 +52,11 @@ func (c Client) CopyObjectWithProgress(dst DestinationInfo, src SourceInfo, prog
 		header.Set(amzLegalHoldHeader, dst.opts.LegalHold.String())
 	}
 
+	if dst.opts.Mode != RetentionMode("") && dst.opts.RetainUntilDate != timeSentinel {
```
klauspost: `time.Time` values should not be compared with `==`. Use `a.Equal(b)` or `a.IsZero()` which will be the value if it wasn't set.

nitisht:
Thanks @klauspost fixing

nitisht:
> Thanks @klauspost fixing

Fixed. PTAL

PR #1258, opened 2020-04-02
Add support for x-amz-bypass-governance-retention in RemoveObjects

nitisht:


api-remove.go
```diff
@@ -179,71 +179,107 @@ func (c Client) RemoveObjectsWithContext(ctx context.Context, bucketName string,
...
-			if resp != nil {
-				if resp.StatusCode != http.StatusOK {
-					e := httpRespToErrorResponse(resp, bucketName, "")
-					errorCh <- RemoveObjectError{ObjectName: "", Err: e}
-				}
+		// Build headers.
+		headers := make(http.Header)
+		if opts.GovernanceBypass {
+			// Set the bypass goverenance retention header
+			headers.Set("x-amz-bypass-governance-retention", "True")
```
harshavardhana, line 250: "True" is "true"

harshavardhana: also lets `x-amz-bypass-governance-retention` make this a constant

api-remove.go
```diff
@@ -179,71 +179,107 @@ func (c Client) RemoveObjectsWithContext(ctx context.Context, bucketName string,
...
+		}
+
+		// Generate remove multi objects XML request
+		removeBytes := generateRemoveMultiObjectsRequest(batch)
+		// Execute GET on bucket to list objects.
+		resp, err := c.executeMethod(ctx, "POST", requestMetadata{
```
harshavardhana, line 256: use http.MethodPost instead of "POST"

nitisht:
Done, PTAL

PR #1266, opened 2020-04-08
Add get bucket versioning

clook:
Implement GetBucketVersioning

### DIFF
--- a/api-get-bucket-versioning.go
+++ b/api-get-bucket-versioning.go
@@ -1,0 +1,78 @@
+/*
+ * MinIO Go Library for Amazon S3 Compatible Cloud Storage
+ * Copyright 2020 MinIO, Inc.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


harshavardhana:
what is the purpose of just adding GetBucketVersioning? @clook

clook:
Hi @harshavardhana, the goal is to be able to get the versioning status of bucket (https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetBucketVersioning.html).

Currently, I'm using https://github.com/aminueza/terraform-provider-minio which relies on minio-go but am unable to provision a bucket with versioning configuration. I could easily make a PR for https://github.com/aminueza/terraform-provider-minio if `GetBucketVersioning` is supported.

harshavardhana:
> Currently, I'm using https://github.com/aminueza/terraform-provider-minio which relies on minio-go but am unable to provision a bucket with versioning configuration. I could easily make a PR for https://github.com/aminueza/terraform-provider-minio if `GetBucketVersioning` is supported.

@clook thanks for the explanation - it makes sense, but what is the pre-requisite of using bucket versioning with Terraform?

clook:
> @clook thanks for the explanation - it makes sense, but what is the pre-requisite of using bucket versioning with Terraform?
The pre-requesite for the Terrraform provider is to be able to compare the Terraform current known state with the real resource state and then guess if it has to apply `DisableVersioning` or `EnableVersioning` or not.
Without knowing the real resource state, we would be forced to update the resource even if no change is necessary.

clook:
Also tell me if you want some unit tests to be performed. Right now, I mostly templated over `GetBucketEncryption` and tested my branch with a PR on the actual Terraform provider.

PR #1268, opened 2020-04-14
sql: Add SQL API to set CSV/JSON options

vadmeste:
In S3 SQL, there is a difference between a 
*) non-set parameter
*) a parameter set to empty
*) a parameter set to a non empty value

For that, users need to explicity call a method to set an option, if not done, the default value in server's side will be used.


harshavardhana:
@vadmeste we cannot really make these type of API changes we should never have our users use *string it is ugly and prone to bugs.

Using " everywhere also escapes GC in unexpected places, we should avoid going down this road. We should fix this appropriately by writing a proper MarshalXML and have default values, non default values.

vadmeste:
> @vadmeste we cannot really make these type of API changes we should never have our users use *string it is ugly and prone to bugs.

so actually users will definitely need to use functions like SetQuoteCharacter() and thse functions won't take a pointer for it.

Custom xml marshal is usually hard to read and to understand but I did this before. However I still feel like it is on overkill here.

harshavardhana:
> > @vadmeste we cannot really make these type of API changes we should never have our users use *string it is ugly and prone to bugs.
> 
> so actually users will definitely need to use functions like SetQuoteCharacter() and these functions won't take a pointer.
> 
> Custom xml marshal is usually hard to read and to understand but I did this before. However I still feel like it is on overkill here.

If you ask me personally this is a horrendous problem, we should avoid getting down the road of using pointers for strings and integers it's not Go idiomatic and will lead API implementation in wrong direction. 

We should do the Marshal approach which clearly defines why we are doing what.

api-select.go
```diff
@@ -88,45 +88,140 @@ type ParquetInputOptions struct{}
...
+	QuoteFields          *CSVQuoteFields `xml:"QuoteFields"`
+	RecordDelimiter      *string         `xml:"RecordDelimiter"`
+	FieldDelimiter       *string         `xml:"FieldDelimiter"`
+	QuoteCharacter       *string         `xml:"QuoteCharacter"`
+	QuoteEscapeCharacter *string         `xml:"QuoteEscapeCharacter"`
+}
+
+// SetQuoteFields sets quote field parameter in CSV output options
+func (c *CSVOutputOptions) SetQuoteFields(v CSVQuoteFields) {
+	c.QuoteFields = &v
```
harshavardhana, line 140: The best approach is to have something like a placeholder element that allows us to signify the empty value. Implement a MarshalXML which skips this. 

We should do this because Go, unfortunately, lacks proper support for default values in XML marshaling, but it also doesn't mean that we should resort to wrong things like `*string` style.

vadmeste: > The best approach is to have something like a placeholder element that allows us to signify the empty value. Implement a MarshalXML which skips this.

Yes will do. We can also avoid breaking the API that way

vadmeste: Well, there are already some pointers in the old code, so breaking the API is inevitable.

I am doing the change only for JSON/CSV input & output options to avoid breaking the API for now.

harshavardhana: Which API are we breaking? @vadmeste

vadmeste: > Which API are we breaking? @vadmeste

The current PR does not break any API.

But to have better consistent API which needs to send another PR to:
- remove all pointers (there are some pointers style in the old code)
- make the code better (because the current one is not optimized because it tries to not break the API)

PR #1273, opened 2020-04-19
add bucket tagging apis

balamurugana:
depends on https://github.com/minio/minio-go/pull/1275


api-bucket-tagging.go
```diff
@@ -0,0 +1,147 @@
...
+	"context"
+	"encoding/xml"
+	"errors"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+
+	"github.com/minio/minio-go/v6/pkg/s3utils"
+	"github.com/minio/minio/pkg/tags"
```
kannappanr, line 30: we should not use the struct from `minio`, we should leave the tags struct defined in `minio-go` as is

balamurugana: OK. Moving `tags` packaging from `minio` to `minio-go`

kannappanr:
Can you also please add documentation and examples as part of adding a new API?

PR #1275, opened 2020-04-23
add tags package

balamurugana:
common tags package for server, minio-go and mc.

### DIFF
--- a/pkg/tags/tags.go
+++ b/pkg/tags/tags.go
@@ -1,0 +1,315 @@
+/*
+ * MinIO Cloud Storage, (C) 2020 MinIO, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package tags
+
+import (
+	"encoding/xml"
+	"io"
+	"net/url"
+	"strings"
+	"unicode/utf8"
+)
+
+// Error contains tag specific error.
+type Error interface {
+	error
+	Code() string
+}
+
+type errTag struct {
+	code    string
+	message string
+}
+
+// Code contains error code.
+func (err errTag) Code() string {
+	return err.code
+}
+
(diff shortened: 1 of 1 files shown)


pkg/tags/tags.go
```diff
@@ -0,0 +1,315 @@
...
+		}
+	}
+
+	return tagging, nil
+}
+
+// UnmarshalXML decodes XML data of tags in reader specified in
+// https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutBucketTagging.html#API_PutBucketTagging_RequestSyntax.
+// If isObject is set, it validates for object tags.
+func UnmarshalXML(reader io.Reader, isObject bool) (*Tags, error) {
```
harshavardhana, line 279: Would be more readable to call these ParseObject and ParseBucket without taking the boolean for a public API, you may use this internally.

pkg/tags/tags.go
```diff
@@ -0,0 +1,315 @@
...
+
+	if err := xml.NewDecoder(reader).Decode(tagging); err != nil {
+		return nil, err
+	}
+
+	return tagging, nil
+}
+
+// Parse decodes HTTP query formatted string into tags. A query formatted string is like "key1=value1&key2=value2".
+func Parse(s string, isObject bool) (*Tags, error) {
```
harshavardhana, line 295: Bucket tagging is not supported with query formatted string, looks like isObject is redundant here.

pkg/tags/tags.go
balamurugana: I am doing with slight change in names as `ParseBucketXML()` and `ParseObjectXML()`

pkg/tags/tags.go
balamurugana: Going to change the name as `ParseObjectTags()` without `isObject bool`

pkg/tags/tags.go
```diff
@@ -0,0 +1,324 @@
...
+// https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html#tag-restrictions
+const (
+	maxKeyLength      = 128
+	maxValueLength    = 256
+	maxObjectTagCount = 10
+	maxTagCount       = 50
+)
+
+func checkKey(key string) error {
+	if len(key) == 0 || utf8.RuneCountInString(key) > maxKeyLength || strings.Contains(key, "&") {
```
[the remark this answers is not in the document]
balamurugana: They are allowed. No change I made in this PR with previous checks. If we want stricter checks, let me know.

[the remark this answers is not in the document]
kannappanr: I think we can take this PR, but we probably should add stricter checks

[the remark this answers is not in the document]
harshavardhana: :+1:

PR #1276, opened 2020-04-24
fix: Use jsoniter instead of encoding/json

nitisht:


harshavardhana:
> @nitisht I think always declaring json as global variable is better. It is more consistent and will prevent an accidental use of both json libraries at the same time

@vadmeste the purpose to avoid that was `goimports` - it is not easy to do that and keep it sane. So it is used at the right place on purpose.

PR #1279, opened 2020-04-25
add Parse() in tags pkg

balamurugana:


kannappanr:
can we start adding better descriptions? I think this function is added so you can make use of it for `Buckettagging` Please let me know if this is wrong

PR #1281, opened 2020-04-30
fix to accept 204 as success

balamurugana:


harshavardhana:
Was the previous change not tested ?

PR #1283, opened 2020-05-04
Add custom MD5/SHA256 hasher option

klauspost:
Since there is no cleanup for the API clients using the SIMD MD5 needs to be up to the API users to manage.

Provide a function that allows to insert a custom hasher for MD5 and SHA256.

We keep using the same for both, but by default we allow to reuse the hashers.


fwessels:
Code looks good, but need to sort of the md5-simd buffer copy issue first.

harshavardhana:
> EDIT: Maybe wait until we've found out how to make md5-simd performant.

Do we need it anymore? @klauspost

klauspost:
@harshavardhana I have kept the default to be a wrapped stdlib, but added a wrapper that will allow reuse of the hashers.

I think this will be a fine improvement and it gives the flexibility to override them on clients.

harshavardhana:
PTAL @fwessels

PR #1289, opened 2020-05-15
tag: Do not validate during marshal/unmarshal

vadmeste:
This PR is needed to fix https://github.com/minio/minio/issues/9609

MinIO server depends on Tag to parse lifecycle documents.

However, some deployments already have empty values of Tag, and it was not a problem since tagging was not supported. However now, an empty key is invalid, and that's causing issues during migration.

### DIFF
--- a/pkg/tags/tags.go
+++ b/pkg/tags/tags.go
@@ -103,32 +103,6 @@ func (tag Tag) Validate() error {
 	return checkValue(tag.Value)
 }
 
-// MarshalXML encodes to XML data.
-func (tag Tag) MarshalXML(e *xml.Encoder, start xml.StartElement) error {
-	if err := tag.Validate(); err != nil {
-		return err
-	}
-
-	type subTag Tag // to avoid recursively calling MarshalXML()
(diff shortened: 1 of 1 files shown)


harshavardhana:
wait ins't Tags allowed to be empty in Lifecycle? @vadmeste

pkg/tags/tags.go
```diff
@@ -103,32 +103,6 @@ func (tag Tag) Validate() error {
 	return checkValue(tag.Value)
 }
 
-// MarshalXML encodes to XML data.
-func (tag Tag) MarshalXML(e *xml.Encoder, start xml.StartElement) error {
```
balamurugana, line 107: `Tag` type was added primarily for `minio/pkg/bucket/lifecycle`. I would recommend to move `Tag` type to `minio/pkg/bucket/lifecycle` and `tags` to deal with `TagSet`

harshavardhana:
@vadmeste is this PR needed anymore?

vadmeste:
No, I defined Tag in server's side, so this PR is not useful anymore. Closing.

PR #1290, opened 2020-05-21
Putobject with PutObjectOptions

BigUstad:
Replacing ObjectOptions, used in gateway and other places


kannappanr:
@BigUstad Please fix the build errors. You need to fix `core_test.go` and functional tests also

PR #1299, opened 2020-06-04
Global context fix

ebozduman:
This is the minio-go side of the fix for bringing in global context for all apis. The mc side PR is minio/mc#3240.
Since mc PR#3240 fix has dependency on this PR, this one needs to be merged before PR#3240 can be merged.


ebozduman:
Looking into build failures...

PR #1300, opened 2020-06-04
Implement Versioning support

vadmeste:
#### *1.*   GetObject

Download object can receive a version id to download a particular object version:

`GetObject(bucketName, objectName, GetObjectOptions{VersionID: versionID})`

#### *2.*  StatObject

 Stating object can receive a particular version id:

`StatObject(bucketName, objectName, StatObjectOptions{GetObjectOptions{VersionID: versionId}})` 

(In a breaking change, we need to define StatObjectOptions as type alais of GetObjectOptions)

#### *3.*  RemoveObject

 Removing object can remove a particular object

`RemoveObject(bucketName, objectName, RemoveObjectOptions{VersionID: versionId})`

#### *4.*  ListObjectVersions

 Listing objects versions has a new API function introduced:

```
type ObjectVersionInfo struct {
       ETag string 
       Key          string   
       LastModified time.Time
       Size         int64  
       Owner Owner
       StorageClass string
       IsLatest       bool
       IsDeleteMarker bool
       VersionID      string
       Err error
}

ListObjectVersions(bucketName, prefix string, recursive bool, doneCh <-chan struct{}) <-chan ObjectVersionInfo
```

#### *5.* CopyObject 
Setting the version ID of the source object during copy

`NewSourceInfo(bucket, object).SetVersionID(versionID)` to specify a version ID of the source object.

#### *6.* ComposeObject 

UploadPart Copy with versioning id is the same as regular copy: `NewSourceInfo(bucket, object).SetVersionID(versionID)`

#### *7.* Object Tagging
```
type PutObjectTaggingOptions struct {
    VersionID string
}
PutObjectTaggingWithOptions(ctx context.Context, bucketName, objectName string, objectTags map[string]string, opts PutObjectTaggingOptions) error
```

```
type GetObjectTaggingOptions struct {
    VersionID string
}

GetObjectTaggingWithOptions(ctx context.Context, bucketName, objectName string, opts GetObjectTaggingOptions) (map[string]string, error)
```

```
type RemoveObjectTaggingOptions struct {
    VersionID string
}

RemoveObjectTaggingWithOptions(ctx context.Context, bucketName, objectName string, opts RemoveObjectTaggingOptions) error
```


api-compose-object.go
```diff
@@ -173,6 +173,7 @@ func (d *DestinationInfo) getUserMetaHeadersMap(withCopyDirectiveHeader bool) ma
 // server-side copying APIs.
 type SourceInfo struct {
```
harshavardhana, line 174: We should deprecate this SourceInfo instead it should become SourceOptions and DestinationInfo should become DestinationOptions

harshavardhana:
Looks like there is a bug in the PR @vadmeste locally running the functional_test results in 

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x6ae01d]

goroutine 1 [running]:
github.com/json-iterator/go.(*nonEmptyInterfaceCodec).Encode(0xcb7f20, 0xc0001b21d0, 0xc0000d8690)
	/home/harsha/mygo/src/github.com/json-iterator/go/feature_reflect_native.go:337 +0x2d
github.com/json-iterator/go.(*structFieldEncoder).Encode(0xc00020b460, 0xc0001b21c0, 0xc0000d8690)
	/home/harsha/mygo/src/github.com/json-iterator/go/feature_reflect_object.go:1033 +0x5c
github.com/json-iterator/go.(*structEncoder).Encode(0xc00020be60, 0xc0001b21c0, 0xc0000d8690)
	/home/harsha/mygo/src/github.com/json-iterator/go/feature_reflect_object.go:1070 +0xe4
github.com/json-iterator/go.(*structEncoder).EncodeInterface(0xc00020be60, 0x893920, 0xc0001b21c0, 0xc0000d8690)
	/home/harsha/mygo/src/github.com/json-iterator/go/feature_reflect_object.go:1090 +0xe2
github.com/json-iterator/go.(*Stream).WriteVal(0xc0000d8690, 0x893920, 0xc0001b21c0)
	/home/harsha/mygo/src/github.com/json-iterator/go/feature_reflect.go:234 +0xdd
github.com/json-iterator/go.(*emptyInterfaceCodec).EncodeInterface(0xcb7f20, 0x893920, 0xc0001b21c0, 0xc0000d8690)
	/home/harsha/mygo/src/github.com/json-iterator/go/feature_reflect_native.go:309 +0x3f
github.com/json-iterator/go.(*sortKeysMapEncoder).Encode(0xc0000d6880, 0xc00129e7b0, 0xc0000d8690)
	/home/harsha/mygo/src/github.com/json-iterator/go/feature_reflect_map.go:187 +0x313
github.com/json-iterator/go.WriteToStream(0x8293a0, 0xc00129e7b0, 0xc0000d8690, 0x9f48a0, 0xc0000d6880)
	/home/harsha/mygo/src/github.com/json-iterator/go/feature_reflect.go:48 +0xfa
github.com/json-iterator/go.(*sortKeysMapEncoder).EncodeInterface(0xc0000d6880, 0x8293a0, 0xc00129e7b0, 0xc0000d8690)
	/home/harsha/mygo/src/github.com/json-iterator/go/feature_reflect_map.go:227 +0x55
github.com/json-iterator/go.(*Stream).WriteVal(0xc0000d8690, 0x8293a0, 0xc00129e7b0)
	/home/harsha/mygo/src/github.com/json-iterator/go/feature_reflect.go:234 +0xdd
github.com/json-iterator/go.(*emptyInterfaceCodec).EncodeInterface(0xcb7f20, 0x8293a0, 0xc00129e7b0, 0xc0000d8690)
	/home/harsha/mygo/src/github.com/json-iterator/go/feature_reflect_native.go:309 +0x3f
```

harshavardhana:
@vadmeste you need to fix this test

```
diff --git a/functional_tests.go b/functional_tests.go
index b2b823f..fa152c2 100644
--- a/functional_tests.go
+++ b/functional_tests.go
@@ -865,7 +865,8 @@ func testStatObjectWithVersioning() {
                        logError(testName, function, args, startTime, "", "error during HEAD object, unexpected ETag", err)
                        return
                }
-               if statInfo.LastModified != results[i].LastModified {
+               // We don't care about nano-second precision
+               if results[i].LastModified.Unix() != statInfo.LastModified.Unix() {
                        logError(testName, function, args, startTime, "", "error during HEAD object, unexpected Last-Modified", err)
                        return
                }
@@ -999,7 +1000,8 @@ func testGetObjectWithVersioning() {
                        logError(testName, function, args, startTime, "", "error during HEAD object, unexpected ETag", err)
                        return
                }
-               if statInfo.LastModified != results[i].LastModified {
+               // We don't care about nano-second precision
+               if results[i].LastModified.Unix() != statInfo.LastModified.Unix() {
                        logError(testName, function, args, startTime, "", "error during HEAD object, unexpected Last-Modified", err)
                        return
                }
```

Also the RemoveObjectWithOptions() needs to support cleanup of the bucket, since object-lock automatically enables versioning.

api-compose-object.go
harshavardhana: @vadmeste can you make this change? ^^

vadmeste: > @vadmeste can you make this change? ^^

It is already done in a separate PR,  https://github.com/minio/minio-go/pull/1305.. it just needs API change approval

klauspost:
@vadmeste @harshavardhana 

We need to give access to the generated Version ID returned from upload operations. Having to do a `ListObjectVersions` both opens up to races and is of course also an un-needed roundtrip.

harshavardhana:
> @vadmeste @harshavardhana 
> 
> We need to give access to the generated Version ID returned from upload operations. Having to do a `ListObjectVersions` both opens up to races and is of course also an un-needed roundtrip.
> 
>  

That access is primarily for replication it is generally not advisable to be used by clients ever. 

If they do use it, they cannot expect List to be consistent.

vadmeste:
> That access is primarily for replication it is generally not advisable to be used by clients ever.
> 
> If they do use it, they cannot expect List to be consistent.

Are you saying that we should not return the version ID after a PutObject() call ? there is a workaround suggested by Klaus which works already

klauspost:
I can't imagine I am the only one who will need access to them. This is an API SDK, we should not assume how it will be used.

Anyone doing anything auditing related would want to record the actual version id of the uploaded object and a separate call is prone to races.

harshavardhana:
> I can't imagine I am the only one who will need access to them. This is an API SDK, we should not assume how it will be used.
> 

It will not happen @klauspost there shouldn't be any public API where this feature of versionId provided by client should be exposed. 

> Anyone doing anything auditing related would want to record the actual version id of the uploaded object and a separate call is prone to races.

All other functionality is already present and AWS compatible so no issues there.

klauspost:
> All other functionality is already present and AWS compatible so no issues there.

It *is* an issue and has nothing to do with compatibility. I want to know the version of the file I just uploaded. There is *no* way to get that without risking someone else having added another version - and it is also a waste of resources to have to do a separate call even if that wasn't a problem.

> It will not happen @klauspost there shouldn't be any public API where this feature of versionId provided by client should be exposed.

Well. It is returned by the [aws sdk](https://docs.aws.amazon.com/sdk-for-go/v2/api/service/s3/#PutObjectOutput) so that isn't a problem.

Your reasoning makes no sense whatsoever. This is not just some magic, it is about providing basic access to the information just uploaded.

harshavardhana:
> > All other functionality is already present and AWS compatible so no issues there.
> 
> It _is_ an issue and has nothing to do with compatibility. I want to know the version of the file I just uploaded. There is _no_ way to get that without risking someone else having added another version - and it is also a waste of resources to have to do a separate call even if that wasn't a problem.
> 

This will be provided by the PutObject refactor which I think is not yet done by @vadmeste because he was trying to keep compatibility which is not needed as we need to move to v7 API.

> > It will not happen @klauspost there shouldn't be any public API where this feature of versionId provided by client should be exposed.
> 
> Well. It is returned by the [aws sdk](https://docs.aws.amazon.com/sdk-for-go/v2/api/service/s3/#PutObjectOutput) so that isn't a problem.
> 
> Your reasoning makes no sense whatsoever. This is not just some magic, it is about providing basic access to the information just uploaded.
> 
> Do you want me to fork this just so I can get access to that information?

That information will be returned by PutObject() that is the change @vadmeste has to bring in PutObject should return a conclusive result struct which has all the necessary values filled in not just length but also ETag, VersionId, LifecycleExpiration value, etc. 

It is a limitation of existing API that warp or any application to keep track of it and use it.

klauspost:
@harshavardhana 

> he was trying to keep compatibility which is not needed as we need to move to v7 API.

The change proposed does not break compatibility with v6, but obviously a returned struct is much cleaner.

api-list.go
```diff
@@ -458,6 +458,213 @@ func (c Client) ListObjectsWithContext(ctx context.Context, bucketName,
...
+			// Listing ends result is not truncated, return right here.
+			if !result.IsTruncated {
+				return
+			}
+		}
+	}(resultCh)
+	return resultCh
+}
+
+// listObjectVersions - (List Object Versions) - List some or all (up to 1000) of the exiting objects
```
poornas, line 570:
```suggestion
// listObjectVersions - (List Object Versions) - List some or all (up to 1000) of the existing objects
```

api-remove.go
```diff
@@ -249,7 +271,7 @@ func (c Client) removeObjects(ctx context.Context, bucketName string, objectsCh
 		}
 		if err != nil {
 			for _, b := range batch {
-				errorCh <- RemoveObjectError{ObjectName: b, Err: err}
+				errorCh <- RemoveObjectError{ObjectName: b.Key, Err: err}
```
poornas, line 274: shouldn't the version attempted to be deleted be sent here in the errorCh too?

docs/API.md
```diff
@@ -1375,6 +1438,64 @@ for rErr := range minioClient.RemoveObjectsWithOptions("my-bucketname", objectsC
...
+
+Parameters
+
+|Param   |Type   |Description   |
+|:---|:---| :---|
+|`bucketName` | _string_ | Name of the bucket  |
+|`objectsCh` | _chan minio.ObjectVersion_ | Channel of objects and versions to be removed  |
+|`opts` | _minio.RemoveObjectsOptions_ | Objects deletion general options |
+
+__minio.RemoveObjecVersion__
```
poornas, line 1453:
```suggestion
__minio.RemoveObjectVersion__
```

api-remove.go
vadmeste: > shouldn't the version attempted to be deleted be sent here in the errorCh too?

Indeed, good catch!

vadmeste:
This PR is rebased against V7, please review again..

vadmeste:
> This PR is rebased against V7, please review again..

Just found out functional tests are failing with MinIO server master, investigating..

vadmeste:
> Just found out functional tests are failing with MinIO server master, investigating..

Fixed already

vadmeste:
Rebased. Can I get some reviewers here ?

harshavardhana:
> Rebased. Can I get some reviewers here ?

Yes sure @vadmeste the PR looks ready we should take it in.

api-compose-object.go
```diff
@@ -88,45 +88,7 @@ func filterCustomMeta(userMeta map[string]string) (map[string]string, error) {
...
-		bucket: bucket,
-		object: object,
-		opts:   opts,
-	}, nil
-}
-
-// NewDestinationInfoWithOptions - creates a compose-object/copy-source
-// destination info object.
-func NewDestinationInfoWithOptions(bucket, object string, destOpts DestInfoOptions) (d DestinationInfo, err error) {
+func NewDestinationInfo(bucket, object string, destOpts DestInfoOptions) (d DestinationInfo, err error) {
```
harshavardhana, line 91: NewDestinationInfo and NewSourceInfo are too generic, I think they should be renamed as part of CopyObject() APIs

It should be `NewCopyDestinationInfo` and `NewCopySourceInfo`

api-put-object-copy.go
```diff
@@ -29,12 +29,12 @@ import (
...
 
 // CopyObject - copy a source object into a new object
-func (c Client) CopyObject(ctx context.Context, dst DestinationInfo, src SourceInfo) error {
+func (c Client) CopyObject(ctx context.Context, dst DestinationInfo, src SourceInfo) (UploadInfo, error) {
 	return c.CopyObjectWithProgress(ctx, dst, src, nil)
 }
 
 // CopyObjectWithProgress is like CopyObject with additional progress bar.
-func (c Client) CopyObjectWithProgress(ctx context.Context, dst DestinationInfo, src SourceInfo, progress io.Reader) error {
+func (c Client) CopyObjectWithProgress(ctx context.Context, dst DestinationInfo, src SourceInfo, progress io.Reader) (UploadInfo, error) {
```
harshavardhana, line 37: We should remove this as well there should be only one API `CopyObject` which optionally takes progress as input as part of the Options()

api-compose-object.go
harshavardhana: In-fact I think Info itself is a bit too much we should generally move to DestinationCopyOptions struct and have a function to Validate() and Transform it if needed internally instead of called 

NewDestinationInfo() and then passing that to CopyObject()

api-put-object-copy.go
```diff
@@ -29,12 +29,12 @@ import (
 )
 
 // CopyObject - copy a source object into a new object
-func (c Client) CopyObject(ctx context.Context, dst DestinationInfo, src SourceInfo) error {
+func (c Client) CopyObject(ctx context.Context, dst DestinationInfo, src SourceInfo) (UploadInfo, error) {
```
harshavardhana, line 32: This should be just ObjectInfo not UploadInfo

api-datatypes.go
```diff
@@ -104,10 +114,37 @@ type ObjectInfo struct {
...
+	IsLatest       bool
+	IsDeleteMarker bool
+	VersionID      string `xml:"VersionId"`
+
 	// Error
 	Err error `json:"-"`
 }
 
+// ObjectVersionInfo container for object version result.
+type ObjectVersionInfo struct {
```
harshavardhana, line 127: There is no reason to have ObjectVersionInfo() - ObjectInfo carries all that is needed for ObjectVersionInfo

api-list.go
```diff
@@ -423,6 +423,213 @@ func (c Client) ListObjects(ctx context.Context, bucketName, objectPrefix string
...
+//   api := client.New(....)
+//   // Create a done channel.
+//   doneCh := make(chan struct{})
+//   defer close(doneCh)
+//   // Recurively list all objects in 'mytestbucket'
+//   recursive := true
+//   for message := range api.ListObjectVersions(ctx, "mytestbucket", "prefix", recursive, doneCh) {
+//       fmt.Println(message)
+//   }
+func (c Client) ListObjectVersions(ctx context.Context, bucketName, prefix string, recursive bool, doneCh <-chan struct{}) <-chan ObjectVersionInfo {
```
harshavardhana, line 438: We don't need another ListObjectVersions() API instead we only need one API 

```
ListObjects(ctx context.Context, bucketName, opts ListObjectsOptions)

type ListObjectsOptions {
    Type Version1/Version2
    WithVersions bool
    WithMetadata bool
    Prefix string
    Recursive bool
    Marker string
    MaxKeys int // represents max-keys to be fetched per request
```

We don't need doneCh anymore we use context everywhere that is the canceller not doneCh

api-put-object-streaming.go
```diff
@@ -56,13 +56,13 @@ func (c Client) putObjectMultipartStream(ctx context.Context, bucketName, object
...
 			if size > maxSinglePutObjectSize {
-				return 0, errEntityTooLarge(size, maxSinglePutObjectSize, bucketName, objectName)
+				return UploadInfo{}, errEntityTooLarge(size, maxSinglePutObjectSize, bucketName, objectName)
 			}
 			// Fall back to uploading as single PutObject operation.
 			return c.putObject(ctx, bucketName, objectName, reader, size, opts)
 		}
 	}
-	return n, err
+	return info, err
```
poornas, line 65: UploadInfo struct does not have size - whereas ObjectInfo set both size and ETag previously in PutObject. Think we should continue to have size in the returned struct as applications may be using this today.

harshavardhana: There is more stuff missing,  please join #minio-go-v7 for more information

api-put-object-multipart.go
```diff
@@ -361,9 +357,14 @@ func (c Client) completeMultipartUpload(ctx context.Context, bucketName, objectN
...
 			// xml parsing failure due to presence an ill-formed xml fragment
-			return completeMultipartUploadResult, err
+			return UploadInfo{}, err
 		}
-		return completeMultipartUploadResult, completeMultipartUploadErr
+		return UploadInfo{}, completeMultipartUploadErr
 	}
-	return completeMultipartUploadResult, nil
+
+	return UploadInfo{
```
klauspost, line 365: For streaming multipart it would be nice if we got the size back as well.

api-datatypes.go
```diff
@@ -86,10 +100,7 @@ type ObjectInfo struct {
 	UserTags map[string]string `json:"userTags"`
```
kannappanr, line 100: How will you get `userTags`?

api-put-object-streaming.go
vadmeste: > UploadInfo struct does not have size - whereas ObjectInfo set both size and ETag previously in PutObject. Think we should continue to have size in the returned struct as applications may be using this today.

Done

api-list.go
vadmeste: > We don't need another ListObjectVersions() API instead we only need one API
> 
> ```
> ListObjects(ctx context.Context, bucketName, opts ListObjectsOptions)
> 
> type ListObjectsOptions {
>     Type Version1/Version2
>     WithVersions bool
>     WithMetadata bool
>     Prefix string
>     Recursive bool
>     Marker string
>     MaxKeys int // represents max-keys to be fetched per request
> ```
> 
> We don't need doneCh anymore we use context everywhere that is the canceller not doneCh

Done

api-datatypes.go
vadmeste: > There is no reason to have ObjectVersionInfo() - ObjectInfo carries all that is needed for ObjectVersionInfo

yeah, deleted

api-put-object-copy.go
vadmeste: > This should be just ObjectInfo not UploadInfo

Plan changed after discussion, it is UploadInfo now

api-put-object-copy.go
vadmeste: > We should remove this as well there should be only one API `CopyObject` which optionally takes progress as input as part of the Options()

Done in a separate PR

api-compose-object.go
vadmeste: > In-fact I think Info itself is a bit too much we should generally move to DestinationCopyOptions struct and have a function to Validate() and Transform it if needed internally instead of called
> 
> NewDestinationInfo() and then passing that to CopyObject()

This will be done in a separate PR

harshavardhana:
c752eca3ae9682be0ee23842d99367cd37e88b97 merged locally.

PR #1301, opened 2020-06-09
remove usages of debug.FreeOSMemory()

vcabbage:
This can be detrimental to an application due to:
* Forces a GC on every execution and blocks until the GC is complete.
* Returns OS memory immediately instead of allowing the runtime to reuse it.

The above two issues aren't limited only to the Minio code but the entire application.

This was introduced in #781. It's unclear whether the changes were tested without `debug.FreeOSMemory()`. Removing `debug.FreeOSMemory()` may be fine even in the memory constrained use case. If not, the user can tune `GOGC` to run GC more aggressively. Additionally, a `sync.Pool` could be reintroduced as the `sync.Pool` implementation has improved in the intervening years.

### DIFF
--- a/api-put-object-multipart.go
+++ b/api-put-object-multipart.go
@@ -28,7 +28,6 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
-	"runtime/debug"
 	"sort"
 	"strconv"
 	"strings"
@@ -98,7 +97,6 @@ func (c Client) putObjectMultipartNoStream(ctx context.Context, bucketName, obje
 
 	// Create a buffer.
 	buf := make([]byte, partSize)
-	defer debug.FreeOSMemory()
 
 	for partNumber <= totalPartsCount {
 		// Choose hash algorithms to be calculated by hashCopyN,
--- a/api-put-object-streaming.go
+++ b/api-put-object-streaming.go
@@ -24,7 +24,6 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"runtime/debug"
 	"sort"
 	"strings"
 
@@ -262,7 +261,6 @@ func (c Client) putObjectMultipartStreamOptionalChecksum(ctx context.Context, bu
 
 	// Create a buffer.
 	buf := make([]byte, partSize)
-	defer debug.FreeOSMemory()
 
 	// Avoid declaring variables in the for loop
 	var md5Base64 string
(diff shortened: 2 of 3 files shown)


harshavardhana:
> This can be detrimental to an application due to:
> 
> * Forces a GC on every execution and blocks until the GC is complete.
> * Returns OS memory immediately instead of allowing the runtime to reuse it.
> 
> The above two issues aren't limited only to the Minio code but the entire application.
> 
> This was introduced in #781. It's unclear whether the changes were tested without `debug.FreeOSMemory()`. Removing `debug.FreeOSMemory()` may be fine even in the memory constrained use case. If not, the user can tune `GOGC` to run GC more aggressively. Additionally, a `sync.Pool` could be reintroduced as the `sync.Pool` implementation has improved in the intervening years.

The best way to do this avoid passing content-length with `-1` @vcabbage pass the right content length it was added for this specific requirement. 

Changing this you should provide different alternative write sync.Pool implementation etc.

vcabbage:
@harshavardhana As far as I can tell this code will be executed regardless of whether `-1` is passed. `putObjectMultipart` calls `putObjectMultipartNoStream`.

No library intended for general usage should call `debug` functions.

There's no test defining the behavior that you're expecting, so adding a `sync.Pool` may or may not "fix" anything. That said, if you still want me to add a pool, I can.

harshavardhana:
> No library intended for general usage should call `debug` functions.

That was freely used because primarily for `-1` use-cases even with sync.Pool implementation is not that worth it
because of how many varied objects can come in, so it may not really be optimal in general.

vcabbage:
I suspect that the majority of users on a recent version of Go will have no memory issues with `debug.FreeOSMemory()` removed. The current reusing of a buffer seems like a good, efficient solution to me. I'd contend that anyone needing to run in a memory constrained environment should use `GOGC` (or `debug.SetGCPercent` in their application code).

In Go 1.13 the `sync.Pool` implementation was improved such that it doesn't clear the entire pool every GC. As long as buffers of similar size are stored in a pool it'd hypothetically reduce memory allocations.

I'm a little unclear on where we stand with this PR. Are you asking for the PR to be adjusted or do you not want the change?

Thanks for taking the time to review :slightly_smiling_face:

harshavardhana:
> I suspect that the majority of users on a recent version of Go will have no memory issues with `debug.FreeOSMemory()` removed. The current reusing of a buffer seems like a good, efficient solution to me. I'd contend that anyone needing to run in a memory constrained environment should use `GOGC` (or `debug.SetGCPercent` in their application code).
> 
> In Go 1.13 the `sync.Pool` implementation was improved such that it doesn't clear the entire pool every GC. As long as buffers of similar size are stored in a pool it'd hypothetically reduce memory allocations.
> 
> I'm a little unclear on where we stand with this PR. Are you asking for the PR to be adjusted or do you not want the change?
> 
> Thanks for taking the time to review

Just removing this is fine for now @vcabbage

klauspost:
👍 Nice catch!

PR #1302, opened 2020-06-10
Support changes for mc global context fix

ebozduman:
This PR is created to support mc global context changes, minio/mc#3240

### DIFF
--- a/api-remove.go
+++ b/api-remove.go
@@ -28,17 +28,22 @@ import (
 	"github.com/minio/minio-go/v6/pkg/s3utils"
 )
 
-// RemoveBucket deletes the bucket name.
+// RemoveBucket is the wrapper for RemoveBucketWithContext
+func (c Client) RemoveBucket(bucketName string) error {
+	return c.RemoveBucketWithContext(context.Background(), bucketName)
+}
+
+// RemoveBucketWithContext deletes the bucket name.
 //
 //  All objects (including all object versions and delete markers).
 //  in the bucket must be deleted before successfully attempting this request.
-func (c Client) RemoveBucket(bucketName string) error {
+func (c Client) RemoveBucketWithContext(ctx context.Context, bucketName string) error {
 	// Input validation.
 	if err := s3utils.CheckValidBucketName(bucketName); err != nil {
 		return err
 	}
 	// Execute DELETE on bucket.
-	resp, err := c.executeMethod(context.Background(), "DELETE", requestMetadata{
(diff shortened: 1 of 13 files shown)


api-remove.go
```diff
@@ -183,32 +188,6 @@ func (c Client) RemoveObjectsWithContext(ctx context.Context, bucketName string,
 	return errorCh
 }
 
-// RemoveObjectsWithOptionsContext - Identical to RemoveObjects call, but accepts context to
-// facilitate request cancellation and options to bypass governance retention
-func (c Client) RemoveObjectsWithOptionsContext(ctx context.Context, bucketName string, objectsCh <-chan string, opts RemoveObjectsOptions) <-chan RemoveObjectError {
```
kannappanr, line 188: Let's keep this name itself, otherwise this is a breaking change

core.go
```diff
@@ -46,13 +46,13 @@ func NewCore(endpoint string, accessKeyID, secretAccessKey string, secure bool)
 // ListObjects - List all the objects at a prefix, optionally with marker and delimiter
 // you can further filter the results.
 func (c Core) ListObjects(bucket, prefix, marker, delimiter string, maxKeys int) (result ListBucketResult, err error) {
-	return c.listObjectsQuery(bucket, prefix, marker, delimiter, maxKeys)
+	return c.listObjectsQuery(context.Background(), bucket, prefix, marker, delimiter, maxKeys)
```
poornas, line 49: suggest adding a WIthContext variant for this method and ListObjectsV2

core.go
```diff
@@ -101,7 +101,7 @@ func (c Core) NewMultipartUpload(bucket, object string, opts PutObjectOptions) (
 
 // ListMultipartUploads - List incomplete uploads.
 func (c Core) ListMultipartUploads(bucket, prefix, keyMarker, uploadIDMarker, delimiter string, maxUploads int) (result ListMultipartUploadsResult, err error) {
-	return c.listMultipartUploadsQuery(bucket, keyMarker, uploadIDMarker, prefix, delimiter, maxUploads)
+	return c.listMultipartUploadsQuery(context.Background(), bucket, keyMarker, uploadIDMarker, prefix, delimiter, maxUploads)
```
poornas, line 104: suggest adding a WithContext variant, for this and other calls in core.go that do not pass context currently.

api-remove.go
ebozduman: ok

PR #1306, opened 2020-06-19
Add replication status and version to Put/Copy opts

poornas:
### 
This is companion PR to https://github.com/minio/minio/pull/9882

### DIFF
--- a/api-put-object.go
+++ b/api-put-object.go
@@ -53,6 +53,8 @@ type PutObjectOptions struct {
 	LegalHold               LegalHoldStatus
 	SendContentMd5          bool
 	DisableMultipart        bool
+	ReplicationVersionID    string
+	ReplicationStatus       string
 }
 
 // getNumThreads - gets the number of threads to be used in the multipart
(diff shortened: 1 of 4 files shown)


kannappanr:
should replication-status be set by the client/source ? Based on the input, can the server not set this?

harshavardhana:
> should replication-status be set by the client/source ? Based on the input, can the server not set this?

We are not implementing proprietary APIs to do that @kannappanr  we are simply leveraging existing APIs and minimally extend them.

api-put-object.go
```diff
@@ -53,6 +53,8 @@ type PutObjectOptions struct {
 	LegalHold               LegalHoldStatus
 	SendContentMd5          bool
 	DisableMultipart        bool
+	ReplicationVersionID    string
```
harshavardhana, line 56: @poornas this should be documented as to be carefully used by applications which know what they are trying to do.

WE should also articulate that this is an extension to versioning implementation by AWS.

PR #1311, opened 2020-06-25
Put/Compose/Copy returns UploadInfo for uploaded/copied objects

vadmeste:
PutObject()/ComposeObject/CopyObject are returning (UploadInfo, error) which
 contains the ETag/VersionID of the uploaded or the newly copied object.

**WARNING:** 
PutObject() does not return the uploaded size anymore. I do not believe it is useful.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,12 +3,13 @@ module github.com/minio/minio-go/v7
 go 1.12
 
 require (
+	github.com/dustin/go-humanize v1.0.0
 	github.com/json-iterator/go v1.1.9
 	github.com/minio/md5-simd v1.1.0
-	github.com/minio/minio-go/v6 v6.0.57 // indirect
 	github.com/minio/sha256-simd v0.1.1
 	github.com/mitchellh/go-homedir v1.1.0
-	github.com/sirupsen/logrus v1.6.0 // indirect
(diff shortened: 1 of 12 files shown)


go.mod
```diff
@@ -3,12 +3,13 @@ module github.com/minio/minio-go/v7
...
 
 require (
+	github.com/dustin/go-humanize v1.0.0
 	github.com/json-iterator/go v1.1.9
 	github.com/minio/md5-simd v1.1.0
-	github.com/minio/minio-go/v6 v6.0.57 // indirect
 	github.com/minio/sha256-simd v0.1.1
 	github.com/mitchellh/go-homedir v1.1.0
-	github.com/sirupsen/logrus v1.6.0 // indirect
+	github.com/sirupsen/logrus v1.6.0
```
klauspost, line 11: Is this intentional? Are they imported elsewhere?

Asking since I don't see any changed imports in your code.

vadmeste: > s this intentional? Are they imported elsewhere?
> 
> Asking since I don't see any changed imports in your code.

Good catch! it seems that this comes from some untracked git files (my test code)

vadmeste:
This change is cherry-picked in this versioning PR: https://github.com/minio/minio-go/pull/1300

Closing this.

PR #1314, opened 2020-06-29
Add APIs to set/get replication config

poornas:
These APIs are companion to [minio PR](https://github.com/minio/minio/pull/9882)


api-get-bucket-replication.go
```diff
@@ -0,0 +1,77 @@
...
+		bucketName:  bucketName,
+		queryValues: urlValues,
+	})
+
+	defer closeResponse(resp)
+	if err != nil {
+		return "", err
+	}
+
+	if resp != nil {
```
vadmeste, line 64: This check is not needed, resp will always be not nil if err == nil

api-get-bucket-replication.go
```diff
@@ -0,0 +1,77 @@
...
+import (
+	"context"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+
+	"github.com/minio/minio-go/v7/pkg/s3utils"
+)
+
+// GetBucketReplication fetches bucket replication configuration
```
vadmeste, line 29: Let's add that it can return an empty string with nil error if the bucket does not have any configuration.

api-put-bucket.go
```diff
@@ -477,3 +477,84 @@ func (c Client) EnableVersioning(ctx context.Context, bucketName string) error {
...
+		contentMD5Base64: sumMD5Base64(b),
+	}
+
+	// Execute PUT to upload a new bucket replication config.
+	resp, err := c.executeMethod(ctx, "PUT", reqMetadata)
+	defer closeResponse(resp)
+	if err != nil {
+		return err
+	}
+	if resp != nil {
```
vadmeste, line 530: Checking resp is not needed here

api-put-bucket.go
```diff
@@ -477,3 +479,95 @@ func (c Client) EnableVersioning(ctx context.Context, bucketName string) error {
...
 	return c.setVersioning(ctx, bucketName, versionDisableConfig, versionDisableConfigLen, versionDisableConfigMD5Sum, versionDisableConfigSHA256)
 }
+
+// RemoveBucketReplication removes a replication config on an existing bucket.
+func (c Client) RemoveBucketReplication(ctx context.Context, bucketName string) error {
+	return c.SetBucketReplication(ctx, bucketName, replication.Config{}, "")
+}
+
+// SetBucketReplication sets a replication config on an existing bucket.
+func (c Client) SetBucketReplication(ctx context.Context, bucketName string, cfg replication.Config, replicationArn string) error {
```
harshavardhana, line 489: We should start with `ReplicationOptions` as input @poornas for the API and this `opts.Validate()` would validate all the options..

api-put-bucket.go
```diff
@@ -477,3 +479,95 @@ func (c Client) EnableVersioning(ctx context.Context, bucketName string) error {
...
+	if err := s3utils.CheckValidBucketName(bucketName); err != nil {
+		return err
+	}
+
+	// Get resources properly escaped and lined up before
+	// using them in http request.
+	urlValues := make(url.Values)
+	urlValues.Set("replication", "")
+	headers := make(http.Header)
+	headers.Add("x-minio-replication-arn", replicationArn)
```
harshavardhana, line 518: we should make this a const elsewhere in code to indicate why is this needed? since this is specific for MinIO?

api-put-bucket.go
```diff
@@ -477,3 +479,95 @@ func (c Client) EnableVersioning(ctx context.Context, bucketName string) error {
...
+	reqMetadata := requestMetadata{
+		bucketName:       bucketName,
+		queryValues:      urlValues,
+		customHeader:     headers,
+		contentBody:      replicationReader,
+		contentLength:    int64(len(b)),
+		contentMD5Base64: sumMD5Base64(b),
+	}
+	// Execute PUT to upload a new bucket replication config.
+	resp, err := c.executeMethod(ctx, "PUT", reqMetadata)
```
harshavardhana, line 538: use `http.MethodPut` all the http constants..

docs/API.md
```diff
@@ -2129,6 +2129,129 @@ if err != nil {
...
+using `mc admin bucket replication set` to associate the bucket `bucketName` with the replication endpoint. Next, this call requires a replication ARN which can be fetched from MinIO server using `mc admin bucket remote`
+
+__Parameters__
+
+|Param   |Type   |Description   |
+|:---|:---| :---|
+|`ctx`  | _context.Context_  | Custom context for timeout/cancellation of the call|
+|`bucketName` | _string_  |Name of the bucket|
+|`cfg` | _replication.Config_  |Replication configuration to be set |
+|`replicationARN` | _string_  |Replication ARN from MinIO server identifying the replication endpoint |
```
harshavardhana, line 2145: I think personally we should extend the

```
<ReplicationConfiguration>
   <Role>string</Role>
   <ReplicationArn></ReplicationArn> --> optional for AWS
   <Rule>
      <DeleteMarkerReplication>
         <Status>Disabled</Status>
      </DeleteMarkerReplication>
      <Destination>
         <Bucket>string</Bucket>
         <StorageClass>string</StorageClass>
      </Destination>
      <Filter>
         <And>
            <Prefix>string</Prefix>
            <Tag>
               <Key>string</Key>
               <Value>string</Value>
            </Tag>
            ...
         </And>
         <Prefix>string</Prefix>
         <Tag>
            <Key>string</Key>
            <Value>string</Value>
         </Tag>
      </Filter>
      <ID>string</ID>
      <Prefix>string</Prefix>
      <Priority>integer</Priority>
      <Status>string</Status>
   </Rule>
</ReplicationConfiguration>
```

docs/API.md
```diff
@@ -2129,6 +2129,129 @@ if err != nil {
...
+if err != nil {
+    fmt.Println(err)
+    return
+}
+```
+
+<a name="GetBucketReplication"></a>
+
+### GetBucketReplication(ctx context.Context, bucketName string) (replication.Config, error)
+Get replication config on a bucket.
```
harshavardhana, line 2202:
```suggestion
Get current replication config on a bucket.
```

go.mod
```diff
@@ -3,14 +3,13 @@ module github.com/minio/minio-go/v7
 go 1.12
 
 require (
-	github.com/json-iterator/go v1.1.9
+	github.com/json-iterator/go v1.1.10
 	github.com/minio/md5-simd v1.1.0
-	github.com/minio/minio-go/v6 v6.0.57 // indirect
+	github.com/minio/minio v0.0.0-20200702003552-fabd0469585a // indirect
```
harshavardhana, line 8: we shouldn't be depending on `minio/minio` not sure why this is coming in? @poornas

docs/API.md
poornas: this is a good idea, makes it cleaner. Done

api-put-bucket.go
poornas: Not really required as of now - however I added an opts for future proofing.

api-put-bucket.go
poornas: No longer needed as this ARN will be part of the configuration

api-put-bucket.go
```diff
@@ -474,3 +475,93 @@ func (c Client) EnableVersioning(ctx context.Context, bucketName string) error {
 func (c Client) DisableVersioning(ctx context.Context, bucketName string) error {
 	return c.setVersioning(ctx, bucketName, versionDisableConfig, versionDisableConfigLen, versionDisableConfigMD5Sum, versionDisableConfigSHA256)
 }
+
+// ReplicationOptions represents options specified by user for Replication call. This is not needed
+// as of now and present for future proofing API call
+type ReplicationOptions struct{}
+
+// RemoveBucketReplication removes a replication config on an existing bucket.
+func (c Client) RemoveBucketReplication(ctx context.Context, bucketName string) error {
```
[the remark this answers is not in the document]
poornas: That's a great suggestion - done.

api-put-object.go
```diff
@@ -119,7 +122,9 @@ func (opts PutObjectOptions) Header() (header http.Header) {
 	if opts.ReplicationStatus != "" {
 		header.Set(amzBucketReplicationStatus, opts.ReplicationStatus)
 	}
-
+	if opts.ReplicationMTime != timeSentinel {
```
harshavardhana, line 125: can it not be `ReplicationMTime.IsZero()` ? @poornas more simpler

docs/API.md
```diff
@@ -538,6 +538,7 @@ __minio.PutObjectOptions__
...
 | `opts.ReplicationStatus`                | _string_               | Specify replication status of object. This option is intended for internal use by MinIO server to extend the replication API implementation by AWS. This option should not be set unless the application is aware of intended use.                                                                                                             |
+| `opts.ReplicationMTime`                | _string_               | Specify replication modTime of source to be used for object. This option is intended for internal use by MinIO server to extend the replication API implementation by AWS. This option should not be set unless the application is aware of intended use.                                                                                                             |
```
harshavardhana, line 541: This should be time.Time @poornas not string

docs/API.md
```diff
@@ -538,6 +538,7 @@ __minio.PutObjectOptions__
 | `opts.PartSize`                | _uint64_               | Specify a custom part size used for uploading the object                                                                                                                           |
 | `opts.ReplicationVersionID`                | _string_               | Specify VersionID of object to replicate.This option is intended for internal use by MinIO server to extend the replication API implementation by AWS. This option should not be set unless the application is aware of intended use.                                                                                              |
 | `opts.ReplicationStatus`                | _string_               | Specify replication status of object. This option is intended for internal use by MinIO server to extend the replication API implementation by AWS. This option should not be set unless the application is aware of intended use.                                                                                                             |
```
harshavardhana, line 540: can we have a type for ReplicationStatus?

docs/API.md
```diff
@@ -538,6 +538,7 @@ __minio.PutObjectOptions__
 | `opts.PartSize`                | _uint64_               | Specify a custom part size used for uploading the object                                                                                                                           |
 | `opts.ReplicationVersionID`                | _string_               | Specify VersionID of object to replicate.This option is intended for internal use by MinIO server to extend the replication API implementation by AWS. This option should not be set unless the application is aware of intended use.                                                                                              |
```
harshavardhana, line 539: We should validate that on the client that this is `uuid` - since this is MinIO specific extension.

api-put-object.go
poornas: definitely.

docs/API.md
poornas: done

docs/API.md
poornas: added a validation here.

docs/API.md
```diff
@@ -537,7 +537,8 @@ __minio.PutObjectOptions__
...
+| `opts.ReplicationStatus`                | _minio.ReplicationStatus_ | Specify replication status of object. This option is intended for internal use by MinIO server to extend the replication API implementation by AWS. This option should not be set unless the application is aware of intended use.                                                                                                             |
+| `opts.ReplicationMTime`                | _time.Time_               | Specify replication modTime of source to be used for object. This option is intended for internal use by MinIO server to extend the replication API implementation by AWS. This option should not be set unless the application is aware of intended use.                                                                                                             |
```
harshavardhana, line 541:
```suggestion
| `opts.ReplicationMTime`                | _time.Time_               | Preserve source modTime on the replicated object. This option is intended for internal use only by MinIO server to comply with AWS bucket replication implementation. This option should not be set unless the application is aware of intended use.                                                                                                             |
```

docs/API.md
```diff
@@ -537,7 +537,8 @@ __minio.PutObjectOptions__
...
-| `opts.ReplicationStatus`                | _string_               | Specify replication status of object. This option is intended for internal use by MinIO server to extend the replication API implementation by AWS. This option should not be set unless the application is aware of intended use.                                                                                                             |
+| `opts.ReplicationStatus`                | _minio.ReplicationStatus_ | Specify replication status of object. This option is intended for internal use by MinIO server to extend the replication API implementation by AWS. This option should not be set unless the application is aware of intended use.                                                                                                             |
```
harshavardhana, line 540:
```suggestion
| `opts.ReplicationStatus`                | _minio.ReplicationStatus_ | Set replication status of object. This option is intended for internal use by MinIO server to extend the replication API implementation by AWS. This option should not be set unless the application is aware of intended use.                                                                                                             |
```

PR #1315, opened 2020-06-30
Custom metadata copy issue

ebozduman:
Copying an object with user defined metadata between s3 servers fails with:

```
mc: <ERROR> Failed to copy `https://play.minio.io:9000/test/test`. Cannot add both Sss and x-amz-meta-Sss keys as custom metadata
That is because we try to copy the same metadata information twice, as user defined metadata raw name, Sss, and as already converted and added into metadata with the standard name, x-amz-meta-Sss.
```

So, the fix is to ignore if the user defined metadata appears more than once in `UserMetadata` list.

### DIFF
--- a/api-compose-object.go
+++ b/api-compose-object.go
@@ -79,7 +79,7 @@ func filterCustomMeta(userMeta map[string]string) (map[string]string, error) {
 			k = k[len("x-amz-meta-"):]
 		}
 		if _, ok := m[k]; ok {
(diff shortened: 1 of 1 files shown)


harshavardhana:
You cannot send PRs to release branch @ebozduman PRs should be merged to master first and then be backported/cherry-picked  to release.

harshavardhana:
you should push again to CI to run

kannappanr:
@ebozduman Please resolve the conflict

Issue #1321, opened 2020-07-01
STS Web Identity Credentials not cached


gregoryfranklin:
When using `credentials.IAM{}` with AWS_WEB_IDENTITY_TOKEN_FILE, the credentials are not cached resulting in an AssumeRoleWithWebIdentity STS call on every request.

#1183 added the ability to use the AWS_WEB_IDENTITY_TOKEN_FILE environment variable with `credentials.IAM{}`
https://github.com/minio/minio-go/blob/v6.0.56/pkg/credentials/iam_aws.go#L83
It does this by creating the STSWebIdentity credentials provider in `func (m *IAM) Retrieve()` and calling `Retrieve()` on that
https://github.com/minio/minio-go/blob/v6.0.56/pkg/credentials/iam_aws.go#L107

However, when you call IsExpired() in the IAM credentials provider it always returns true because the expiry has been set on the STSWebIdentity credentials provider not on the IAM credentials provider.

For reference, in the case of ECS and EC2 roles, the expiry is set correctly
https://github.com/minio/minio-go/blob/v6.0.56/pkg/credentials/iam_aws.go#L140

In my case, the impact of this is that Thanos runs extremely when using IAM Roles for Service Accounts because it is making an STS call on every request to S3.
https://github.com/thanos-io/thanos/issues/2097

harshavardhana:
@gregoryfranklin feel free to send a fix

PR #1322, opened 2020-07-01
Cache STS Web Identity Credentials

gregoryfranklin:
When using the STSWebIdentity credentials provider from the IAM
credentials provider, we need to set the expiry time of the credentials.

We therefore expose the expiry time in the STSWebIdentity credentials
provider so that it can be used by the IAM credentials provider.

Calls to IsExpired() on the IAM credentials provider will then work as
they would have if it had been called on the underlying STSWebIdentity
provider.  Therefore caching the credentials as intended.

Fixes minio/minio-go#1321

### DIFF
--- a/pkg/credentials/iam_aws.go
+++ b/pkg/credentials/iam_aws.go
@@ -106,7 +106,7 @@ func (m *IAM) Retrieve() (Value, error) {
 
 		stsWebIdentityCreds, err := creds.Retrieve()
 		if err == nil {
-			m.SetExpiration(creds.Expiration(), 0)
+			m.SetExpiration(creds.Expiration(), DefaultExpiryWindow)
 		}
 		return stsWebIdentityCreds, err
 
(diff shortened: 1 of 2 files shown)


pkg/credentials/iam_aws.go
```diff
@@ -104,7 +104,11 @@ func (m *IAM) Retrieve() (Value, error) {
 			},
 		}
 
-		return creds.Retrieve()
+		stsWebIdentityCreds, err := creds.Retrieve()
+		if err == nil {
+			m.SetExpiration(creds.Expiration(), 0)
```
harshavardhana, line 109:
```suggestion
			m.SetExpiration(creds.Expiration(), DefaultExpiryWindow)
```

gregoryfranklin: DefaultExpiryWindow has already been applied.

`creds.Expiration()` returns `m.expiration` which is set by `m.SetExpiration(a.Result.Credentials.Expiration, DefaultExpiryWindow)` in `STSWebIdentity`

I figured we didn't want to add DefaultExpiryWindow twice.

harshavardhana: If this is a new value, we should apply here as well.

PR #1323, opened 2020-07-02
Fix errInvalidArgument change

h8liu:


h8liu:
It was breaking the build.

PR #1327, opened 2020-07-02
refactor: Single MakeBucket API with options

vadmeste:
Use the following struct as options when creating a bucket:

MakeBucketOptions {
   Region string
   ObjectLocking bool
}

MakeBucket(ctx context.Context, bucketName string, opts MakeBucketOptions) error

### DIFF
--- a/api-put-bucket.go
+++ b/api-put-bucket.go
@@ -126,6 +126,14 @@ func (c Client) doMakeBucket(ctx context.Context, bucketName string, location st
 	return nil
 }
 
+// MakeBucketOptions holds all options to tweak bucket creation
+type MakeBucketOptions struct {
+	// Bucket location
+	Region string
+	// Enable object locking
+	ObjectLocking bool
+}
+
 // MakeBucket creates a new bucket with bucketName with a context to control cancellations and timeouts.
 //
 // Location is an optional argument, by default all buckets are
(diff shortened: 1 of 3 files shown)


api-put-bucket.go
```diff
@@ -126,26 +126,23 @@ func (c Client) doMakeBucket(ctx context.Context, bucketName string, location st
...
 //
 // Location is an optional argument, by default all buckets are
 // created in US Standard Region.
 //
 // For Amazon S3 for more supported regions - http://docs.aws.amazon.com/general/latest/gr/rande.html
 // For Google Cloud Storage for more supported regions - https://cloud.google.com/storage/docs/bucket-locations
-func (c Client) MakeBucketWithObjectLock(ctx context.Context, bucketName string, location string) (err error) {
-	return c.makeBucket(ctx, bucketName, location, true)
+func (c Client) MakeBucket(ctx context.Context, bucketName string, opts MakeBucketOptions) (err error) {
+	return c.makeBucket(ctx, bucketName, opts.Region, opts.ObjectLocking)
```
harshavardhana, line 145:
```suggestion
	return c.makeBucket(ctx, bucketName, opts.Region, opts.ObjectLocking)
```

change this to use `opts` as well no need to preserve old wrapper

PR #1328, opened 2020-07-02
api.PresignedPostPolicy() > Don't make request to minio if it's not needed.

andytruong:


api-presigned.go
```diff
@@ -114,10 +114,14 @@ func (c Client) PresignedPostPolicy(ctx context.Context, p *PostPolicy) (u *url.
...
 	bucketName := p.formData["bucket"]
-	// Fetch the bucket location.
-	location, err := c.getBucketLocation(ctx, bucketName)
-	if err != nil {
-		return nil, nil, err
+
+	// Fetch the bucket location, we only need this value for AWS.
+	location := ""
+	if s3utils.IsAmazonEndpoint(*c.endpointURL) {
+		location, err = c.getBucketLocation(ctx, bucketName)
```
harshavardhana, line 121: This is a wrong code, location is true for all S3 vendors. It is only default settings MinIO doesn't have a region.

But you can configure a region and this call will ignore it.

andytruong: location is only used by 

```
u, err = c.makeTargetURL(bucketName, "", location, …
```

Inside `makeTargetURL()`, location is only used for AWS bucket:

```
if s3utils.IsAmazonEndpoint(*c.endpointURL) {
```

harshavardhana: That is a misconception location is used for signature as well. This will break existing users.

andytruong: I don't think so. It's clearly location value is not used for non-AWS vendors. Let me try adding test case.

andytruong: @harshavardhana I added test case for this API. I hope it's not breaking anything.

harshavardhana: The problem you are not understanding is location is needed to generate the signature your PR will break if MinIO server is configured with a region MINIO_REGION setting.

harshavardhana: Its all the way down - this will fail 

> 	credential := signer.GetCredential(accessKeyID, location, t, signer.ServiceTypeS3)

PR #1338, opened 2020-07-15
reimplement all Bucket config APIs

harshavardhana:


harshavardhana:
I removed ReplicationReqOpts on purpose as they are not idiomatic for all our existing APIs for buckets, any new options should be sent as part of replicationConfig instead of opts for the request.

harshavardhana:
> I removed ReplicationReqOpts on purpose as they are not idiomatic for all our existing APIs for buckets, any new options should be sent as part of replicationConfig instead of opts for the request.

Either we add `Options` in all APIs or we don't add it right now and bring in when its needed.

api-bucket-policy.go
```diff
@@ -0,0 +1,158 @@
...
+		contentLength: int64(len(b)),
+	}
+
+	// Execute PUT to upload a new bucket policy.
+	resp, err := c.executeMethod(ctx, "PUT", reqMetadata)
+	defer closeResponse(resp)
+	if err != nil {
+		return err
+	}
+	if resp != nil {
```
vadmeste, line 77: if resp != nil doesn't seem to be needed

api-bucket-policy.go
```diff
@@ -0,0 +1,158 @@
...
+	// using them in http request.
+	urlValues := make(url.Values)
+	urlValues.Set("policy", "")
+
+	// Content-length is mandatory for put policy request
+	policyReader := strings.NewReader(policy)
+	b, err := ioutil.ReadAll(policyReader)
+	if err != nil {
+		return err
+	}
```
vadmeste, line 62: I don't think this block is needed, besides we should read from policyReader and then use it again.

api-bucket-policy.go
```diff
@@ -0,0 +1,158 @@
...
+
+	reqMetadata := requestMetadata{
+		bucketName:    bucketName,
+		queryValues:   urlValues,
+		contentBody:   policyReader,
```
vadmeste, line 67: strings.NewReader(policy)

api-bucket-policy.go
```diff
@@ -0,0 +1,158 @@
...
+		contentLength: int64(len(b)),
```
vadmeste, line 68: len(policy)

api-bucket-replication.go
```diff
@@ -0,0 +1,168 @@
...
+	replication, err := xml.Marshal(cfg)
+	if err != nil {
+		return err
+	}
+	// Content-length is mandatory for put replication request
+	replicationReader := bytes.NewReader(replication)
+	b, err := ioutil.ReadAll(replicationReader)
+	if err != nil {
+		return err
+	}
```
vadmeste, line 72: This block doesn't seem to be needed

api-bucket-replication.go
```diff
@@ -0,0 +1,168 @@
...
+	reqMetadata := requestMetadata{
+		bucketName:       bucketName,
+		queryValues:      urlValues,
+		contentBody:      replicationReader,
```
vadmeste, line 76: bytes.NewReader(replication)

api-bucket-replication.go
```diff
@@ -0,0 +1,168 @@
...
+		contentLength:    int64(len(b)),
```
vadmeste, line 77: len(replication)

pkg/notification/info.go
```diff
@@ -0,0 +1,61 @@
+package notification
```
vadmeste, line 1: Licence header is missing

api-bucket-policy.go
```diff
@@ -0,0 +1,158 @@
...
+	"net/url"
+	"strings"
+
+	"github.com/minio/minio-go/v7/pkg/s3utils"
+)
+
+// SetBucketPolicy sets the access permissions on an existing bucket.
+func (c Client) SetBucketPolicy(ctx context.Context, bucketName, policy string) error {
+	// Input validation.
+	if err := s3utils.CheckValidBucketName(bucketName); err != nil {
```
kannappanr, line 32: This check is not required as it is checked in both `removeBucketPolicy` and `putBucketPolicy`

api-bucket-replication.go
```diff
@@ -0,0 +1,168 @@
...
+
+// RemoveBucketReplication removes a replication config on an existing bucket.
+func (c Client) RemoveBucketReplication(ctx context.Context, bucketName string) error {
+	return c.removeBucketReplication(ctx, bucketName)
+}
+
+// SetBucketReplication sets a replication config on an existing bucket.
+func (c Client) SetBucketReplication(ctx context.Context, bucketName string, cfg replication.Config) error {
+	// Input validation.
+	if err := s3utils.CheckValidBucketName(bucketName); err != nil {
```
kannappanr, line 40: This check is also not required

pkg/sse/sse.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+}
+
+// NewConfigurationSSEKMS initializes a new SSE-KMS configuration
+func NewConfigurationSSEKMS(kmsMasterKey string) *Configuration {
+	return &Configuration{
+		Rules: []Rule{
+			{
+				Apply: ApplySSEByDefault{
+					KmsMasterKeyID: "my-masterkey",
```
kannappanr, line 60:
```suggestion
					KmsMasterKeyID: kmsMasterKey,
```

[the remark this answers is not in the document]
harshavardhana: No it is kmsMasterKeyId @vadmeste

api-bucket-policy.go
harshavardhana: old code @vadmeste not changing it.

api-bucket-policy.go
harshavardhana: old code @vadmeste not changing it right now

pkg/notification/info.go
kannappanr, line 1: copyright info missing

api-lifecycle.go
```diff
@@ -0,0 +1,156 @@
+/*
+ * MinIO Go Library for Amazon S3 Compatible Cloud Storage
+ * Copyright 2015-2017 MinIO, Inc.
```
kannappanr, line 3:
```suggestion
 * Copyright 2020 MinIO, Inc.
```

api-lifecycle.go
```diff
@@ -0,0 +1,156 @@
...
+	"net/url"
+
+	"github.com/minio/minio-go/v7/pkg/lifecycle"
+	"github.com/minio/minio-go/v7/pkg/s3utils"
+)
+
+// SetBucketLifecycle set the lifecycle on an existing bucket.
+func (c Client) SetBucketLifecycle(ctx context.Context, bucketName string, config *lifecycle.Configuration) error {
+	// Input validation.
+	if err := s3utils.CheckValidBucketName(bucketName); err != nil {
```
kannappanr, line 35: This check is not required. Otherwise the checks in removeBucketLifecyle and putBucketLifecylce should be removed

harshavardhana:
Addressed all comments @kannappanr @vadmeste @poornas

PR #1344, opened 2020-07-18
fix: removing lifecycle properly

harshavardhana:


harshavardhana:
> LGTM but not sure why CI fails

Because of a bug in MinIO @vadmeste

PR #1351, opened 2020-07-22
lifecycle: Fix marshaling expiration date/days

vadmeste:
Setting a new bucket lifecycle was not working properly, the reason
is that a zero time.Time is still parsed in expiration date. Fixing it.

### DIFF
--- a/pkg/lifecycle/lifecycle.go
+++ b/pkg/lifecycle/lifecycle.go
@@ -180,11 +180,11 @@ type ExpirationDays int
 
 // MarshalXML encodes number of days to expire if it is non-zero and
 // encodes empty string otherwise
-func (eDays *ExpirationDays) MarshalXML(e *xml.Encoder, startElement xml.StartElement) error {
-	if *eDays == ExpirationDays(0) {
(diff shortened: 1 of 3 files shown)


pkg/lifecycle/lifecycle.go
```diff
@@ -195,8 +195,8 @@ type ExpirationDate struct {
 
 // MarshalXML encodes expiration date if it is non-zero and encodes
 // empty string otherwise
-func (eDate *ExpirationDate) MarshalXML(e *xml.Encoder, startElement xml.StartElement) error {
-	if *eDate == (ExpirationDate{time.Time{}}) {
+func (eDate ExpirationDate) MarshalXML(e *xml.Encoder, startElement xml.StartElement) error {
+	if eDate.Time.IsZero() {
```
harshavardhana, line 199: This code is borrowed from server @vadmeste does this mean it needs to be fixed on server as well?

vadmeste: > This code is borrowed from server @vadmeste does this mean it needs to be fixed on server as well?

Yes

PR #1356, opened 2020-08-01
Add methods to set/remove replication rules

poornas:
Needed for https://github.com/minio/mc/pull/3335


go.mod
```diff
@@ -3,16 +3,14 @@ module github.com/minio/minio-go/v7
 go 1.12
 
 require (
-	github.com/dustin/go-humanize v1.0.0 // indirect
 	github.com/google/uuid v1.1.1
 	github.com/json-iterator/go v1.1.10
 	github.com/minio/md5-simd v1.1.0
+	github.com/minio/minio v0.0.0-20200731035804-2174a228351a
```
harshavardhana, line 9: minio-go cannot depend on server at any point in time @poornas ?

harshavardhana: perhaps missing ` go mod tidy?`

go.mod
```diff
@@ -3,16 +3,23 @@ module github.com/minio/minio-go/v7
 go 1.12
 
 require (
-	github.com/dustin/go-humanize v1.0.0 // indirect
 	github.com/google/uuid v1.1.1
 	github.com/json-iterator/go v1.1.10
+	github.com/klauspost/cpuid v1.3.1 // indirect
```
harshavardhana, line 8: none of these are used by `minio-go` I am not able to see where is this coming from?

poornas: it is coming from the md5-simd dependency

PR #1360, opened 2020-08-05
fix: change AddRule method in replication

poornas:
to inherit priority and replication Arn
if rule ID is specified in the options.

Needed for mc [PR](https://github.com/minio/mc/pull/3335)

### DIFF
--- a/pkg/replication/replication.go
+++ b/pkg/replication/replication.go
@@ -75,8 +75,7 @@ func (opts Options) Tags() []Tag {
 type Config struct {
 	XMLName xml.Name `xml:"ReplicationConfiguration" json:"-"`
 	Rules   []Rule   `xml:"Rule" json:"Rules"`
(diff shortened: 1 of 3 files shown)


pkg/replication/replication.go
```diff
@@ -108,16 +109,24 @@ func (c *Config) AddRule(opts Options) error {
 	if opts.Disable {
 		status = Disabled
 	}
-	tokens := strings.Split(opts.Arn, ":")
+	arnStr := opts.Arn
+	if opts.Arn == "" {
+		arnStr = c.ReplicationARN
+	}
+	tokens := strings.Split(arnStr, ":")
```
harshavardhana, line 116: can you verify the number of tokens expected as well? @poornas

poornas: that's the next check

PR #1371, opened 2020-08-17
[doc] fix doc fun-index error

KevinSmile:
fix wrong index of functions in doc.

In the function-index table of the doc:
1. missing index for `PresignedHeadObject` function
2. duplicate index for `PutObject` function
3. wrong columns for many functions on the 2nd-row
(I deleted the `Encrypted Object operations` column as it's incomplete and confusing.)

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -58,29 +58,29 @@ func main() {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


KevinSmile:
@harshavardhana

PR #1373, opened 2020-08-19
Add missing context value to example

klauspost:
`BucketExists` was missing a context value.

We place it in a variable to make the code cleaner.


vadmeste (approved):
LGTM

PR #1378, opened 2020-08-28
Send single DELETE requests if object name has xml invalid chars

vadmeste:
Currently, it is not possible to remove some objects using the multi
delete API if those objects contain some weird characters.

In an XML document, some characters are not allowed to be there and
there is no way to escape them in XML 1.0, currently used by S3.
Therefore, some objects that have those invalid characters won't be
deleted in a multi-delete request.

To solve this, we need to check for object names and send individual
DELETE requests. The object name will be encoded in the request URL and
can be removed as expected

### DIFF
--- a/api-remove.go
+++ b/api-remove.go
@@ -196,7 +196,7 @@ func (c Client) RemoveObjects(ctx context.Context, bucketName string, objectsCh 
 
 // Return true if the character is within the allowed characters in an XML 1.0 document
 // The list of allowed characters can be found here: https://www.w3.org/TR/xml/#charsets
-func isXMLValidChar(r rune) (inrange bool) {
+func validXMLChar(r rune) (ok bool) {
 	return r == 0x09 ||
 		r == 0x0A ||
 		r == 0x0D ||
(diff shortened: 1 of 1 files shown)


api-remove.go
```diff
@@ -189,6 +194,26 @@ func (c Client) RemoveObjects(ctx context.Context, bucketName string, objectsCh
 	return errorCh
 }
 
+// Return true if the character is within the allowed characters in an XML 1.0 document
+// The list of allowed characters can be found here: https://www.w3.org/TR/xml/#charsets
+func isXMLValidChar(r rune) (inrange bool) {
```
klauspost, line 199:
```suggestion
func validXMLChar(r rune) (ok bool) {
```

api-remove.go
```diff
@@ -189,6 +194,26 @@ func (c Client) RemoveObjects(ctx context.Context, bucketName string, objectsCh
...
+	return r == 0x09 ||
+		r == 0x0A ||
+		r == 0x0D ||
+		r >= 0x20 && r <= 0xD7FF ||
+		r >= 0xE000 && r <= 0xFFFD ||
+		r >= 0x10000 && r <= 0x10FFFF
+}
+
+func hasXMLInvalidChar(str string) bool {
```
klauspost, line 208:
```suggestion
func hasInvalidXMLChar(str string) bool {
```

klauspost (approved):
Nice! Just a few cosmetics.

PR #1381, opened 2020-09-01
lifecycle: Fix parsing Tag in documents

vadmeste:
Tag was omitted while parsing a lifecycle document in JSON format
because Tag field was unintentionally omitted in field description.

This causes an issue in mc ilm command.


klauspost (approved):
LGTM

PR #1390, opened 2020-09-25
Rename opts vars for common use by transition and replication

poornas:


api-put-object.go
```diff
@@ -72,10 +72,10 @@ type PutObjectOptions struct {
 	LegalHold               LegalHoldStatus
 	SendContentMd5          bool
 	DisableMultipart        bool
-	ReplicationVersionID    string
-	ReplicationETag         string
+	SourceVersionID         string
+	SourceETag              string
 	ReplicationStatus       ReplicationStatus
-	ReplicationMTime        time.Time
```
harshavardhana, line 78: Can we simply move this into a separate struct and reference it here. Such as AdvancedOptions - so that you are free to change this struct instead of modifying `PutObjectOptions` on a continuous basis.

It is better to just hide it, add a comment to indicate that not to be used by client applications etc.

harshavardhana (changes requested):

api-put-object.go
poornas: that's a great idea - done!

api-put-object.go
```diff
@@ -72,10 +81,7 @@ type PutObjectOptions struct {
 	LegalHold               LegalHoldStatus
 	SendContentMd5          bool
 	DisableMultipart        bool
-	ReplicationVersionID    string
-	ReplicationETag         string
-	ReplicationStatus       ReplicationStatus
-	ReplicationMTime        time.Time
+	Internal                AdvancedPutOptions
```
harshavardhana, line 84:
```suggestion
        // To be not used by external applications
	Internal                AdvancedPutOptions
```

api-get-options.go
```diff
@@ -25,13 +25,18 @@ import (
...
 // during GET requests.
 type GetObjectOptions struct {
-	headers                 map[string]string
-	ServerSideEncryption    encrypt.ServerSide
-	VersionID               string
-	ReplicationDeleteMarker bool
+	headers              map[string]string
+	ServerSideEncryption encrypt.ServerSide
+	VersionID            string
+	Internal             AdvancedGetOptions
```
harshavardhana, line 39:
```suggestion
        // To be not used by external applications
	Internal             AdvancedGetOptions
```

harshavardhana (changes requested):

harshavardhana (approved):

PR #1404, opened 2020-10-29
replication: add source-proxy header if opts set

poornas:
- this is basically needed in a two way replication scenario and needs to be set to stop lookup in other site unless this header is sent by other site.

### DIFF
--- a/api-get-options.go
+++ b/api-get-options.go
@@ -28,6 +28,7 @@ import (
 //AdvancedGetOptions for internal use by MinIO server - not intended for client use.
 type AdvancedGetOptions struct {
 	ReplicationDeleteMarker bool
+	ReplicationProxyRequest bool
 }
 
 // GetObjectOptions are used to specify additional headers or options
@@ -52,6 +53,11 @@ func (o GetObjectOptions) Header() http.Header {
 	}
 	if o.ServerSideEncryption != nil && o.ServerSideEncryption.Type() == encrypt.SSEC {
 		o.ServerSideEncryption.Marshal(headers)
+	}
+	// this header is set for active-active replication scenario where GET/HEAD
+	// to site A is proxy'd to site B if object/version missing on site A.
+	if o.Internal.ReplicationProxyRequest {
+		headers.Set(minIOBucketReplicationProxyRequest, "true")
 	}
 	return headers
 }
--- a/constants.go
+++ b/constants.go
@@ -87,4 +87,5 @@ const (
 
 	minIOBucketSourceETag              = "X-Minio-Source-Etag"
 	minIOBucketReplicationDeleteMarker = "X-Minio-Source-DeleteMarker"
+	minIOBucketReplicationProxyRequest = "X-Minio-Source-Proxy-Request"
 )


pkg/replication/replication.go
```diff
@@ -296,6 +301,14 @@ func (c *Config) EditRule(opts Options) error {
 		}
 	}
 
+	if opts.ReplicateSync != "" {
+		switch opts.ReplicateSync {
```
vadmeste, line 305: Maybe let's add a default clause which returns an error

pkg/replication/replication.go
```diff
@@ -145,6 +147,9 @@ func (c *Config) AddRule(opts Options) error {
 	if c.Role == "" {
 		c.Role = arnStr
 	}
+	if opts.ReplicateSync == "sync" {
```
vadmeste, line 150: Let's return an error if replicate sync is not recognized, also let me make a public constant for "sync" and "async" so mc or other can uses it.

api-get-options.go
```diff
@@ -53,6 +54,11 @@ func (o GetObjectOptions) Header() http.Header {
 	if o.ServerSideEncryption != nil && o.ServerSideEncryption.Type() == encrypt.SSEC {
 		o.ServerSideEncryption.Marshal(headers)
 	}
+	// this header is set for active-active replication scenario where GET/HEAD
+	// to site A is proxy'd to site B if object/version missing on site A.
+	if o.Internal.ReplicationProxyRequest {
```
vadmeste, line 59: Will this be needed by MinIO server replication code ?

vadmeste (commented):
Some comments

pkg/replication/replication.go
poornas: done

pkg/replication/replication.go
poornas: done

api-get-options.go
poornas: yes

pkg/replication/replication.go
```diff
@@ -145,6 +152,9 @@ func (c *Config) AddRule(opts Options) error {
 	if c.Role == "" {
 		c.Role = arnStr
 	}
+	if opts.ReplicateSync == ReplicateSync {
+		c.ReplicateSync = true
```
harshavardhana, line 156:
```suggestion
		c.ReplicateSync = opts.ReplicateSync == ReplicateSync
```

pkg/replication/replication.go
```diff
@@ -296,6 +306,16 @@ func (c *Config) EditRule(opts Options) error {
...
 	}
 
+	if opts.ReplicateSync != "" {
+		switch opts.ReplicateSync {
+		case ReplicateSync:
+			c.ReplicateSync = true
+		case ReplicateAsync: //default mode
+			c.ReplicateSync = false
+		default:
+			return fmt.Errorf("ReplicateSync state should be either [sync|async]")
```
harshavardhana, line 316: Isn't default mode Async? @poornas

pkg/replication/replication.go
```diff
@@ -296,6 +306,16 @@ func (c *Config) EditRule(opts Options) error {
 		}
 	}
 
+	if opts.ReplicateSync != "" {
+		switch opts.ReplicateSync {
+		case ReplicateSync:
+			c.ReplicateSync = true
```
harshavardhana, line 312:
```suggestion
			c.ReplicateSync = opts.ReplicateSync == ReplicateSync
```

can be written in one line? and validation for `ReplicateSync` value to be outside of this check?

pkg/replication/replication.go
```diff
@@ -59,6 +64,7 @@ type Options struct {
 	IsSCSet                bool
 	ReplicateDeletes       string // replicate versioned deletes
 	ReplicateDeleteMarkers string // replicate soft deletes
+	ReplicateSync          string // whether synchronous replication is specified.
```
harshavardhana, line 67: why not make this `boolean` to start with?  - you don't need to bring in `mc` logic here. `mc` can still support words like `sync` and `async` but API wise this doesn't need to be `string`

harshavardhana (changes requested):

pkg/replication/replication.go
poornas: this is needed for `mc replicate edit` option to make out whether user is opting for synchronous/asynchronous replication. It would not be possible to make out with a boolean state if user is changing it or left it unset

pkg/replication/replication.go
poornas: yes

pkg/replication/replication.go
```diff
@@ -81,9 +87,10 @@ func (opts Options) Tags() []Tag {
 // Config - replication configuration specified in
 // https://docs.aws.amazon.com/AmazonS3/latest/dev/replication-add-config.html
 type Config struct {
-	XMLName xml.Name `xml:"ReplicationConfiguration" json:"-"`
-	Rules   []Rule   `xml:"Rule" json:"Rules"`
-	Role    string   `xml:"Role" json:"Role"`
+	XMLName       xml.Name `xml:"ReplicationConfiguration" json:"-"`
+	Rules         []Rule   `xml:"Rule" json:"Rules"`
+	Role          string   `xml:"Role" json:"Role"`
+	ReplicateSync bool     `xml:"ReplicateSync" json:"ReplicateSync"`
```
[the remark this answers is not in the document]
poornas: Both aws sdk go and aws cli seem to have validations to reject any fields other than Rules/Role - so this change will break the sdk compatibility

pkg/replication/replication.go
poornas: your suggestion would work if this option is rule specific - it is intended at the bucket level and user can specify --sync flag on any rule and it applies to the configuration rather than at a rule prefix level.

pkg/replication/replication.go
poornas: as mentioned above,the ReplicateSync setting is at bucket level rather than rule level, hence this check.

harshavardhana:
is this PR waiting on someone @poornas ?

poornas:
> is this PR waiting on someone @poornas ?

waiting on reviews :-)

pkg/replication/replication.go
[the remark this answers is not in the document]
vadmeste: > Both aws sdk go and aws cli seem to have validations to reject any fields other than Rules/Role - so this change will break the sdk compatibility

maybe let's hide this field then (omitempty), I guess it will break only when someone enables this flag

kannappanr (approved):

harshavardhana (changes requested):
Should we not have headers 

X-Minio-Replication (as prefix) instead of just X-Minio looks like we have added some headers already that might conflict and are too generic.

poornas:
> Should we not have headers
> 
> X-Minio-Replication (as prefix) instead of just X-Minio looks like we have added some headers already that might conflict and are too generic.

`X-Minio-Source-` is the prefix being used here, which is already being used for replication

harshavardhana:
> `X-Minio-Source-` is the prefix being used here, which is already being used for replication

Correct that is still a bit generic @poornas not sure if the time has passed to change these.

harshavardhana (approved):

PR #1409, opened 2020-11-12
Add opts to core CopyObject api

poornas:
needed for https://github.com/minio/minio/pull/10880 to pass version id in core CopyObject call.

### DIFF
--- a/api-compose-object.go
+++ b/api-compose-object.go
@@ -231,7 +231,7 @@ func (c Client) copyObjectDo(ctx context.Context, srcBucket, srcObject, destBuck
 	}
 	if dstOpts.Internal.SourceVersionID != "" {
 		if _, err := uuid.Parse(dstOpts.Internal.SourceVersionID); err != nil {
-			return ObjectInfo{}, err
+			return ObjectInfo{}, errInvalidArgument(err.Error())
 		}
 		urlValues := make(url.Values)
 		urlValues.Set("versionId", dstOpts.Internal.SourceVersionID)
--- a/api-compose-object.go
+++ b/api-compose-object.go
@@ -28,6 +28,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/google/uuid"
 	"github.com/minio/minio-go/v7/pkg/encrypt"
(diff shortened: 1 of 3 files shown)


api-compose-object.go
```diff
@@ -210,16 +211,38 @@ func (c Client) copyObjectDo(ctx context.Context, srcBucket, srcObject, destBuck
...
+	}
+
+	reqMetadata := requestMetadata{
+		bucketName:   destBucket,
+		objectName:   destObject,
+		customHeader: headers,
+	}
+	if dstOpts.Internal.SourceVersionID != "" {
+		if _, err := uuid.Parse(dstOpts.Internal.SourceVersionID); err != nil {
+			return ObjectInfo{}, err
```
harshavardhana, line 234:
```suggestion
			return ObjectInfo{}, errInvalidArgument(err.Error())
```

harshavardhana (changes requested):

harshavardhana (approved):

api-compose-object.go
```diff
@@ -210,16 +211,38 @@ func (c Client) copyObjectDo(ctx context.Context, srcBucket, srcObject, destBuck
...
+		bucketName:   destBucket,
+		objectName:   destObject,
+		customHeader: headers,
+	}
+	if dstOpts.Internal.SourceVersionID != "" {
+		if _, err := uuid.Parse(dstOpts.Internal.SourceVersionID); err != nil {
+			return ObjectInfo{}, errInvalidArgument(err.Error())
+		}
+		urlValues := make(url.Values)
+		urlValues.Set("versionId", dstOpts.Internal.SourceVersionID)
```
vadmeste, line 237: LGTM,

Otherwise, is this style already released ? Why not using 'minIOBucketSourceVersionID' for consistency with others, also this is more prone to fail with AWS, probably in other use cases in the future.

poornas: all other API's use a query string for version id - so trying to stay consistent with that. With AWS, this version id would never be set

vadmeste (approved):
LGTM

Issue #1413, opened 2020-11-23
Context timeout not handled in FGetObject


alexcocia:
When setting a context timeout of x duration, the request made to FGetObject always lasts 30 secs. Context is not returned until that time is passed

The following code returns the error after 30 secs:
```
        scalityContext, cancel := context.WithTimeout(context.Background(), 5 * time.Second)
	defer cancel()

	err := scality.client.FGetObject(scalityContext, bucket, itemName, filePath, minio.GetObjectOptions{})
	if err != nil {
		return errorhandler.CreateInternalError(fmt.Sprintf("Error in Scality getting item: [%s] bucket: [%s]", itemName, bucket), err)
	}
```

harshavardhana:
fixed by #1414

PR #1414, opened 2020-11-23
Fix - context timeout not launched on get bucket location #1413

alexcocia:
Using context on request, in order to handle context timeouts

### DIFF
--- a/bucket-cache.go
+++ b/bucket-cache.go
@@ -97,12 +97,12 @@ func (c Client) getBucketLocation(ctx context.Context, bucketName string) (strin
 	}
 
 	// Initialize a new request.
-	req, err := c.getBucketLocationRequest(bucketName)
+	req, err := c.getBucketLocationRequest(ctx, bucketName)
 	if err != nil {
 		return "", err
 	}
 
-	req = req.WithContext(ctx)
+	//req = req.WithContext(ctx)
 	// Initiate the request.
 	resp, err := c.do(req)
 	defer closeResponse(resp)
@@ -170,7 +170,7 @@ func processBucketLocationResponse(resp *http.Response, bucketName string) (buck
 }
 
 // getBucketLocationRequest - Wrapper creates a new getBucketLocation request.
-func (c Client) getBucketLocationRequest(bucketName string) (*http.Request, error) {
+func (c Client) getBucketLocationRequest(ctx context.Context, bucketName string) (*http.Request, error) {
 	// Set location query.
 	urlValues := make(url.Values)
 	urlValues.Set("location", "")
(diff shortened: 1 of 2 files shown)


bucket-cache.go
```diff
@@ -102,6 +102,7 @@ func (c Client) getBucketLocation(ctx context.Context, bucketName string) (strin
 		return "", err
 	}
 
+	req = req.WithContext(ctx)
 	// Initiate the request.
 	resp, err := c.do(req)
 	defer closeResponse(resp)
```
harshavardhana, line 108: it is better to use http.NewRequestWithContext() instead of req.WithContext @alexcocia

harshavardhana (changes requested):

bucket-cache.go
harshavardhana: Here is the change I would expect

```diff
git diff
diff --git a/bucket-cache.go b/bucket-cache.go
index 7d485a6..156150f 100644
--- a/bucket-cache.go
+++ b/bucket-cache.go
@@ -97,7 +97,7 @@ func (c Client) getBucketLocation(ctx context.Context, bucketName string) (strin
        }
 
        // Initialize a new request.
-       req, err := c.getBucketLocationRequest(bucketName)
+       req, err := c.getBucketLocationRequest(ctx, bucketName)
        if err != nil {
                return "", err
        }
@@ -169,7 +169,7 @@ func processBucketLocationResponse(resp *http.Response, bucketName string) (buck
 }
 
 // getBucketLocationRequest - Wrapper creates a new getBucketLocation request.
-func (c Client) getBucketLocationRequest(bucketName string) (*http.Request, error) {
+func (c Client) getBucketLocationRequest(ctx context.Context, bucketName string) (*http.Request, error) {
        // Set location query.
        urlValues := make(url.Values)
        urlValues.Set("location", "")
@@ -198,7 +198,7 @@ func (c Client) getBucketLocationRequest(bucketName string) (*http.Request, erro
        }
 
        // Get a new HTTP request for the method.
-       req, err := http.NewRequest(http.MethodGet, urlStr, nil)
+       req, err := http.NewRequestWithContext(ctx, http.MethodGet, urlStr, nil)
        if err != nil {
                return nil, err
        }
```

alexcocia: New commit provided applying your code. Thanks a lot, that was a quick response!

bucket-cache.go
```diff
@@ -97,11 +97,12 @@ func (c Client) getBucketLocation(ctx context.Context, bucketName string) (strin
 	}
 
 	// Initialize a new request.
-	req, err := c.getBucketLocationRequest(bucketName)
+	req, err := c.getBucketLocationRequest(ctx, bucketName)
 	if err != nil {
 		return "", err
 	}
 
+	//req = req.WithContext(ctx)
```
harshavardhana, line 105: Not needed since you made the change already can be removed

```suggestion
```

harshavardhana (changes requested):

alexcocia:
Totally right. Forgot to remove it after tests on my env.

harshavardhana (approved):

klauspost (approved):
LGTM

PR #1427, opened 2020-12-14
Add SourceSelectionCriteria to replication config

poornas:
This PR is needed for viewing this additional config in mc

### DIFF
--- a/pkg/replication/replication.go
+++ b/pkg/replication/replication.go
@@ -195,6 +195,14 @@ func (c *Config) AddRule(opts Options) error {
 		},
 		DeleteMarkerReplication: DeleteMarkerReplication{Status: dmStatus},
 		DeleteReplication:       DeleteReplication{Status: vDeleteStatus},
+		// MinIO enables replica metadata syncing by default in the case of bi-directional replication to allow
+		// automatic failover as the expectation in this case is that replica and source should be identical.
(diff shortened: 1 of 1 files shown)


pkg/replication/replication.go
```diff
@@ -195,6 +195,10 @@ func (c *Config) AddRule(opts Options) error {
 		},
 		DeleteMarkerReplication: DeleteMarkerReplication{Status: dmStatus},
 		DeleteReplication:       DeleteReplication{Status: vDeleteStatus},
+		SourceSelectionCriteria: SourceSelectionCriteria{
+			ReplicaModifications: ReplicaModifications{
+				Status: Enabled},
```
harshavardhana, line 200:
```suggestion
				Status: Enabled,
				},
```

harshavardhana: BTW should we do this by default?

harshavardhana (changes requested):

pkg/replication/replication.go
poornas: If we do not enable this by default, we will need additional flags in mc to support toggling this status. It seems to make sense for bi-directional replication to keep source and replica in sync

harshavardhana: Hmm good - add a comment on why we enforce it @poornas - and also why it makes sense by referencing the link from AWS - such that our choice is to make it default

poornas: done

harshavardhana (approved):

PR #1429, opened 2020-12-16
Fix Bug on Events Configuration Equal function

cesnietor:
If Config.Events was nil, it would cause an null pointer exception

- [X] Tests included

### DIFF
--- a/pkg/notification/notification.go
+++ b/pkg/notification/notification.go
@@ -180,20 +180,28 @@ func EqualFilterRuleList(a, b []FilterRule) bool {
 
 // Equal returns whether this `Config` is equal to another defined by the passed parameters
 func (t *Config) Equal(events []EventType, prefix, suffix string) bool {
-	//Compare events
+	if t == nil {
+		return false
+	}
+
+	// Compare events
 	passEvents := EqualEventTypeList(t.Events, events)
 
-	//Compare filters
-	var newFilter []FilterRule
+	// Compare filters
+	var newFilterRules []FilterRule
 	if prefix != "" {
-		newFilter = append(newFilter, FilterRule{Name: "prefix", Value: prefix})
+		newFilterRules = append(newFilterRules, FilterRule{Name: "prefix", Value: prefix})
 	}
 	if suffix != "" {
-		newFilter = append(newFilter, FilterRule{Name: "suffix", Value: suffix})
+		newFilterRules = append(newFilterRules, FilterRule{Name: "suffix", Value: suffix})
 	}
 
-	passFilters := EqualFilterRuleList(t.Filter.S3Key.FilterRules, newFilter)
-	// if it matches events and filters, mark the index for deletion
+	var currentFilterRules []FilterRule
+	if t.Filter != nil {
+		currentFilterRules = t.Filter.S3Key.FilterRules
+	}
+
+	passFilters := EqualFilterRuleList(currentFilterRules, newFilterRules)
 	return passEvents && passFilters
 }
 
--- a/pkg/notification/notification_test.go
+++ b/pkg/notification/notification_test.go
@@ -389,6 +389,139 @@ func TestConfiguration_RemoveQueueByArnEventsPrefixSuffix(t *testing.T) {
 			},
 			wantErr: true,
 		},
+		{
+			name: "Queue Configuration Removed with nil Filter",
+			fields: fields{
+				XMLName:       xml.Name{},
+				LambdaConfigs: nil,
+				TopicConfigs:  nil,
+				QueueConfigs: []QueueConfig{
+					{
+						Config: Config{
+							ID: "",
+							Arn: Arn{
+								Partition: "minio",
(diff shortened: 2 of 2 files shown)


pkg/notification/notification.go
```diff
@@ -192,9 +192,18 @@ func (t *Config) Equal(events []EventType, prefix, suffix string) bool {
...
+		if t.Filter != nil {
+			passFilters := EqualFilterRuleList(t.Filter.S3Key.FilterRules, newFilter)
+			return passEvents && passFilters
+		}
+		// if Config is nil, they are not equal
+		return false
+	}
+
+	// if matches events and t.filter == nil
+	return (t.Filter == nil) && passEvents
```
harshavardhana, line 206: You should invert this logic 

```diff
diff --git a/pkg/notification/notification.go b/pkg/notification/notification.go
index 05c79bb..e37fbe3 100644
--- a/pkg/notification/notification.go
+++ b/pkg/notification/notification.go
@@ -180,10 +180,15 @@ func EqualFilterRuleList(a, b []FilterRule) bool {
 
 // Equal returns whether this `Config` is equal to another defined by the passed parameters
 func (t *Config) Equal(events []EventType, prefix, suffix string) bool {
-       //Compare events
+       // Config is nil return not equal.
+       if t == nil {
+               return false
+       }
+
+       // Compare events
        passEvents := EqualEventTypeList(t.Events, events)
 
-       //Compare filters
+       // Compare filters
        var newFilter []FilterRule
        if prefix != "" {
                newFilter = append(newFilter, FilterRule{Name: "prefix", Value: prefix})
@@ -192,6 +197,11 @@ func (t *Config) Equal(events []EventType, prefix, suffix string) bool {
                newFilter = append(newFilter, FilterRule{Name: "suffix", Value: suffix})
        }
 
+       // If filter is provided, config should have a filter if not its false.
+       if len(newFilter) > 0 && t.Filter == nil {
+               return false
+       }
+
        passFilters := EqualFilterRuleList(t.Filter.S3Key.FilterRules, newFilter)
        // if it matches events and filters, mark the index for deletion
        return passEvents && passFilters
```

A much simpler change

harshavardhana (changes requested):

pkg/notification/notification.go
cesnietor: the first one won't work since Config can be nil and event can be empty as well as prefix and suffix which then should return`true`.
This one:
```
if prefix == "" && suffix == "" {
               return passEvents
       }
```
won't work cause the Config can have Filter and hence that would give a false positive.
Tried both but Tests don't pass with your suggestions. @harshavardhana

harshavardhana: @cesnietor t can never be nil if that is nil - all bets are off and everything fails i.e t.Events will also fail so Config should be always allocated with Arn - if that is not the case then Equal should always be false.

harshavardhana (changes requested):
Here is the simplified diff

```diff
diff --git a/pkg/notification/notification.go b/pkg/notification/notification.go
index d150636..b17e6c5 100644
--- a/pkg/notification/notification.go
+++ b/pkg/notification/notification.go
@@ -180,30 +180,29 @@ func EqualFilterRuleList(a, b []FilterRule) bool {
 
 // Equal returns whether this `Config` is equal to another defined by the passed parameters
 func (t *Config) Equal(events []EventType, prefix, suffix string) bool {
-       //Compare events
+       if t == nil {
+               return false
+       }
+
+       // Compare events
        passEvents := EqualEventTypeList(t.Events, events)
 
-       //Compare filters
-       var newFilter []FilterRule
+       // Compare filters
+       var newFilterRules []FilterRule
        if prefix != "" {
-               newFilter = append(newFilter, FilterRule{Name: "prefix", Value: prefix})
+               newFilterRules = append(newFilterRules, FilterRule{Name: "prefix", Value: prefix})
        }
        if suffix != "" {
-               newFilter = append(newFilter, FilterRule{Name: "suffix", Value: suffix})
+               newFilterRules = append(newFilterRules, FilterRule{Name: "suffix", Value: suffix})
        }
 
-       if len(newFilter) > 0 {
-               // compare newFilter with Config.Filter
-               if t.Filter != nil {
-                       passFilters := EqualFilterRuleList(t.Filter.S3Key.FilterRules, newFilter)
-                       return passEvents && passFilters
-               }
-               // if Config is nil, they are not equal
-               return false
+       var currentFilterRules []FilterRule
+       if t.Filter != nil {
+               currentFilterRules = t.Filter.S3Key.FilterRules
        }
 
-       // if matches events and t.filter == nil
-       return (t.Filter == nil) && passEvents
+       passFilters := EqualFilterRuleList(currentFilterRules, newFilterRules)
+       return passEvents && passFilters
 }
```

cesnietor:
@harshavardhana  I've addressed it and added more tests for the function itself.

harshavardhana (approved):

Alevsk (approved):
LGTM

PR #1430, opened 2020-12-17
encrypt: fix SSE-KMS HTTP context header

aead:
This commit fixes a bug in the SSE-KMS.
The current SSE-KMS implementation sets
the wrong header for the SSE-KMS context.

Now, the correct header is set:
```
X-Amz-Server-Side-Encryption-Context
```


vadmeste (approved):
LGTM

PR #1434, opened 2020-12-21
sse: re-implement SSE-C, SSE-S3 and SSE-KMS

aead:
This commit re-implements the server-side
encryption implementation.

Therefore, the bucket encryption configuration
is moved from `pkg/sse` to `encrypt`.

Further, the `encrypt.ServerSide` interface has
been extended to be more flexible for S3 GET / PUT / COPY
operations.

The reason for this refactoring is that the SSE-KMS implementation
has been broken. It was not possible to specify e.g. the SSE-KMS
context during GET / HEAD operations with the previous abstraction.


api-compose-object.go
```diff
@@ -114,7 +114,7 @@ func (opts CopyDestOptions) Marshal(header http.Header) {
 	}
 
 	if opts.Encryption != nil {
-		opts.Encryption.Marshal(header)
+		opts.Encryption.MarshalCOPY2(header)
```
poornas, line 117: it will be more readable to rename MarshalCOPY -> MarshalCOPYSrc and MarshalCOPY2 -> MarshalCOPYDest

poornas (changes requested):

balamurugana:
@aead As the PR is a refactor, how about using implementation like https://github.com/minio/minio-py/blob/master/minio/sse.py ?

api-compose-object.go
aead: My idea was `MarshalCOPY2` reads like `MarshalCOPY to` referring to `copy from` / `copy to`.
However, I agree that it may be too confusing when reading code using the interface...

Will try to find a better name (that also is a idiomatic Go name)

aead:
> As the PR is a refactor, how about using implementation like https://github.com/minio/minio-py/blob/master/minio/sse.py ?

The PR more or less does that. However, the python implementation seems broken as well. For example: It seems to send the SSE-KMS key ID and algorithm for GET requests. This is not correct. It must only send the context (if a context is present). Otherwise, it must not send the context as well.

pkg/encrypt/server-side.go
```diff
@@ -60,139 +52,229 @@ type PBKDF func(password, salt []byte) ServerSide
...
+	md5Sum := md5.Sum(s)
+
+	if op == CopySource {
+		h.Add(amzServerSideEncryptionCopyCustomerAlgorithm, "AES256")
+		h.Add(amzServerSideEncryptionCopyCustomerKey, base64.StdEncoding.EncodeToString(s))
+		h.Add(amzServerSideEncryptionCopyCustomerKeyMD5, base64.StdEncoding.EncodeToString(md5Sum[:]))
+	} else {
+		h.Add(amzServerSideEncryptionCustomerAlgorithm, "AES256")
+		h.Add(amzServerSideEncryptionCustomerKey, base64.StdEncoding.EncodeToString(s))
+		h.Add(amzServerSideEncryptionCustomerKeyMD5, base64.StdEncoding.EncodeToString(md5Sum[:]))
```
poornas, line 278: just reuse s.MarshalPUT for the else block?

pkg/encrypt/server-side.go
```diff
@@ -60,139 +52,229 @@ type PBKDF func(password, salt []byte) ServerSide
...
-	keyMD5 := md5.Sum(s[:])
-	h.Set(sseCustomerAlgorithm, "AES256")
-	h.Set(sseCustomerKey, base64.StdEncoding.EncodeToString(s[:]))
-	h.Set(sseCustomerKeyMD5, base64.StdEncoding.EncodeToString(keyMD5[:]))
+// MarshalGET adds the SSE-KMS server-side encryption HTTP
+// header containing the encryption context, if any.
+func (s ssekms) MarshalGET(h http.Header) {
+	if s.context != nil {
+		h.Add(amzServerSideEncryptionContext, base64.StdEncoding.EncodeToString(s.context))
+	}
```
poornas, line 222: was looking at headers needed for kms in aws GetObject - the encryption context doesn't need to be specified on GET per [this doc](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObject.html). Is this needed by KES?

poornas (changes requested):

pkg/encrypt/server-side.go
aead: See: 
https://docs.aws.amazon.com/kms/latest/developerguide/concepts.html#encrypt_context

> When you include an encryption context in an encryption request, it is cryptographically bound to the ciphertext such that the same encryption context is required to decrypt (or decrypt and re-encrypt) the data. If the encryption context provided in the decryption request is not an exact, case-sensitive match, the decrypt request fails. Only the order of the key-value pairs in the encryption context can vary.

aead: S3 uses the AWS-KMS. AWS-KMS requires the context to decrypt a ciphertext (data encryption key) since the context is crypto. bound to the ciphertext.
Therefore, S3 GET / HEAD / PUT / COPY operations may require the context.

poornas: ok

poornas (approved):
LGTM

PR #1436, opened 2020-12-23
Allow setting SourceVersionID for multipart uploads

thepax:
Currently SourceVersionID works for streaming PutObject only.
This patch allows using SourceVersionID for multipart uploads.

### DIFF
--- a/api-put-object-multipart.go
+++ b/api-put-object-multipart.go
@@ -32,6 +32,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/google/uuid"
 	"github.com/minio/minio-go/v7/pkg/encrypt"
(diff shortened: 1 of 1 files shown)


harshavardhana:
This is an internal API -  under what situation are you using this ?

thepax:
It would be used for a parallel replication. Single stream replication would be really slow on long distance.
I'm working on proof-of-concept side container.

harshavardhana:
> It would be used for a parallel replication. Single stream replication would be really slow on long distance.
> I'm working on proof-of-concept side container.

It is not expected that you attempt these things. Please open issues on minio with current challenges in server side replication.

These fields are internal and shouldn't be used by client apps. If used may be broken in future we do not allow any API compatibility for them.

PR #1439, opened 2021-01-07
lifecycle: Fix infinite NoncurrentTransition xml marshalling

vadmeste:
A missing type wrapper to avoid an infinite recursive call of MarshalXML
is missing for NoncurrentTransition struct type. This commit will add
it.

Also add NoncurrentTransition.IsDaysNull() API similarly to NonCurrentExpiration.


klauspost (approved):
;)

harshavardhana (approved):

PR #1441, opened 2021-01-13
lifecycle: Add IsDisabled to ExpireDeleteMarker

vadmeste:
This is needed to fix a bug in mc while setting expiration delete marker


klauspost (approved):
👍🏼

nitisht (approved):

Issue #1442, opened 2021-01-13
Thanos - BaseFetcher: iter bucket: You are not authorized to perform this operation


narasimman-elu:
Deploying Thanos(either as a side-car or store) China region is not successful.

Unfortunately minio does not give any error message. 

Error Message from Thanos:
level=warn ts=2021-01-13T11:02:50.446721205Z caller=intrumentation.go:54 msg="changing probe status" status=not-ready reason="bucket store initial sync: sync block: BaseFetcher: iter bucket: You are not authorized to perform this operation"
level=info ts=2021-01-13T11:02:50.446764001Z caller=http.go:65 service=http/server component=store msg="internal server is shutting down" err="bucket store initial sync: sync block: BaseFetcher: iter bucket: You are not authorized to perform this operation"

narasimman-elu:
After spending a week the issue has been found in minio package.

minio uses credentials pkg to get the aws credentials access in 4 different ways to authorize with tokens.
[Credentials](https://github.com/thanos-io/thanos/blob/855c50886659b7d6441d5fba5034004c4188af9f/docs/storage.md#credentials)
The preferred way which most organisation would like to use IAM approch.
This approach uses token from the serviceaccount and updates to AWS_WEB_IDENTITY_FILE env variable.

minio parses these env variables to decode the token using sts.
The sts endpoint that is used in minio is hardcoded.

```
if len(os.Getenv("AWS_REGION")) > 0 {
				endpoint = "https://sts." + os.Getenv("AWS_REGION") + ".amazonaws.com"
```
and the default is
`
defaultSTSRoleEndpoint      = "https://sts.amazonaws.com"`

This allows only the regions that are expect china regions.
incase if the Beiging region is configured then it does not uses the special sts china endpoint [China (Beijing) Region Endpoints](https://docs.amazonaws.cn/en_us/general/latest/gr/cnnorth_region.html)

harshavardhana:
fixed by #1444

PR #1444, opened 2021-01-13
[1442] sts endpoint fix for china regions

narasimman-elu:
The pull request for issue #1442 
This fixes the sts endpoind issue for china provided the AWS_REGION environment shall be set by pod configs.
The fix is based on v7.0.7

### DIFF
--- a/pkg/credentials/iam_aws.go
+++ b/pkg/credentials/iam_aws.go
@@ -83,7 +83,7 @@ func (m *IAM) Retrieve() (Value, error) {
 	case len(os.Getenv("AWS_WEB_IDENTITY_TOKEN_FILE")) > 0:
 		if len(endpoint) == 0 {
 			if len(os.Getenv("AWS_REGION")) > 0 {
-				if strings.HasPrefix(os.Getenv("AWS_REGION"), "cn") {
+				if strings.HasPrefix(os.Getenv("AWS_REGION"), "cn-") {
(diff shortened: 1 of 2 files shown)


pkg/credentials/iam_aws.go
```diff
@@ -82,7 +83,11 @@ func (m *IAM) Retrieve() (Value, error) {
 	case len(os.Getenv("AWS_WEB_IDENTITY_TOKEN_FILE")) > 0:
 		if len(endpoint) == 0 {
 			if len(os.Getenv("AWS_REGION")) > 0 {
-				endpoint = "https://sts." + os.Getenv("AWS_REGION") + ".amazonaws.com"
+				if strings.HasPrefix(os.Getenv("AWS_REGION"), "cn") {
```
vadmeste, line 86: `cn-` instead of `cn`

vadmeste: By the way, normally you can pass endpoint using NewIAM(endpoint string) API, that would be enough

narasimman-elu: Thanos is one of the client neither set endpoint while creeating IAM and nor using NewIAM function.
May be Thanos should consider to fix this issue also from Thanos side.

narasimman-elu: > `cn-` instead of `cn`

Have considered this review point and fixed.

poornas (approved):

harshavardhana (approved):

PR #1448, opened 2021-02-04
tagSet.String() shows the encoded form of all tags

vadmeste:
It does not make sense to show a list of tags while key/value
are not encoded.

e.g.  key==val, you don't know if '=' belongs to the key or to the value


klauspost (approved):
LGTM

harshavardhana (approved):

PR #1449, opened 2021-02-04
Add support for extra signed headers in Presign() API

### LINKED ISSUE #1415: Support for custom signed headers as part of Presign URLs
I'm trying to generate Presigned URLs that require the user to provide matching headers upon usage. The current `PresignXXX` methods don't allow me to specify extra headers that should be part of the signature. 

`ReqParams` aren't sufficient because the value of the params are included in the returned presign URL. Whereas with headers, only the header key is returned.

donatello:
Fixes #1415


donatello:
Needs a version bump as an exported function signature has changed.

api-presigned.go
```diff
@@ -30,7 +30,7 @@ import (
 
 // presignURL - Returns a presigned URL for an input 'method'.
 // Expires maximum is 7days - ie. 604800 and minimum is 1.
-func (c Client) presignURL(ctx context.Context, method string, bucketName string, objectName string, expires time.Duration, reqParams url.Values) (u *url.URL, err error) {
+func (c Client) presignURL(ctx context.Context, method string, bucketName string, objectName string, expires time.Duration, reqParams url.Values, extraHeaders http.Header) (u *url.URL, err error) {
```
ebozduman, line 33: Am I missing something here? Providing support at the internal function level for `presignURL()` is not good enough, because the end user has only access to the following 3 apis as documented in [MinIO Go Client API Reference](https://docs.minio.io/docs/golang-client-api-reference.html) :
- PresignedGetObject
- PresignedPutObject
- PresignedHeadObject

When end-user calls one of the above apis, s/he would expect to pass the extra headers to the api.

As the continuation of the above comment, we also need to modify the corresponiding scripts; `examples/s3/presignedgetobject.go`, `examples/s3/presignedheadobject.go`, and `examples/s3/presignedputobject.go`.

ebozduman (changes requested):

api-presigned.go
donatello: @ebozduman

> Providing support at the internal function level for presignURL() is not good enough, because the end user has only access to the following 3 apis as documented in MinIO Go Client API Reference :

We also export `Presign()` (see [here](https://pkg.go.dev/github.com/minio/minio-go/v7#Client.Presign)) though we don't officially document it on our docs.min.io site. Since this change is expected to be useful only to some users, I've added this feature there. This helps prevent breaking backwards compatibility - however I am open to changing this if we think this can be more universally useful. CC: @harshavardhana

ebozduman: OK. 
I'll run some tests and get back to you.

ebozduman: I tested with my script and tried GET and PUT methods. They both passed successfully.
I also used Swagger tool to add the header on the browser and it was successful.
LGTM

ebozduman (approved):

api-presigned.go
```diff
@@ -90,14 +91,14 @@ func (c Client) PresignedPutObject(ctx context.Context, bucketName string, objec
...
 
-// Presign - returns a presigned URL for any http method of your choice
-// along with custom request params. URL can have a maximum expiry of
-// upto 7days or a minimum of 1sec.
-func (c Client) Presign(ctx context.Context, method string, bucketName string, objectName string, expires time.Duration, reqParams url.Values) (u *url.URL, err error) {
-	return c.presignURL(ctx, method, bucketName, objectName, expires, reqParams)
+// Presign - returns a presigned URL for any http method of your choice along
+// with custom request params and extra signed headers. URL can have a maximum
+// expiry of upto 7days or a minimum of 1sec.
+func (c Client) Presign(ctx context.Context, method string, bucketName string, objectName string, expires time.Duration, reqParams url.Values, extraHeaders http.Header) (u *url.URL, err error) {
```
klauspost, line 100: This would require a major version bump.

You could add `PresignHeader`  which doesn't change API with a note to update this on next major version (where we maybe should switch to sending params as a struct).

`extraHeaders ...http.Header` is another, but uglier option since it also allows to send multiple headers which we should account for then.

donatello:
PTAL @klauspost @ebozduman @harshavardhana

Apologies for the huge delay on this one. I've added a new `PresignHeader()` func to avoid a major version bump and noted in the comment that we should be adding it to Presign().

harshavardhana (approved):

klauspost (commented):
Fails after merge. Otherwise LGTM.

PR #1450, opened 2021-02-10
Set versionID in core CopyObject if specified

poornas:
To copy a specific version, X-Amz-Copy-Source needs to have a
query string with source version id.


harshavardhana (approved):

harshavardhana:
Build failure is expected with latest release @poornas can you `go get` this in your PR?

PR #1455, opened 2021-02-19
fix: support MINIO_ROOT* envs

harshavardhana:


pkg/credentials/env_minio.go
```diff
@@ -22,10 +22,12 @@ import "os"
 // A EnvMinio retrieves credentials from the environment variables of the
 // running process. EnvMinioironment credentials never expire.
 //
-// EnvMinioironment variables used:
+// Environment variables used:
 //
 // * Access Key ID:     MINIO_ACCESS_KEY.
```
vadmeste, line 27: Shouldn't we remove the deprecated env vars ?

vadmeste (commented):
One comment

pkg/credentials/env_minio.go
harshavardhana: No not everyone is expected to migrate, and we shall never remove these ENVs from MinIO either @vadmeste

vadmeste (approved):
LGTM

kannappanr (approved):

PR #1463, opened 2021-03-11
Add an interface for Object{} to improve the testability

nikolaistraessle:
Hi guys,

I've created an interface for the Object{} to improve the testability.

No additional tests are required for this feature.

Any feedback is welcome.

Many thanks

### DIFF
--- a/api-get-object.go
+++ b/api-get-object.go
@@ -29,8 +29,16 @@ import (
 	"github.com/minio/minio-go/v7/pkg/s3utils"
 )
 
+type ObjectInterface interface {
+	Read(b []byte) (n int, err error)
+	Stat() (ObjectInfo, error)
+	ReadAt(b []byte, offset int64) (n int, err error)
+	Seek(offset int64, whence int) (n int64, err error)
+	Close() (err error)
+}
+
 // GetObject wrapper function that accepts a request context
-func (c Client) GetObject(ctx context.Context, bucketName, objectName string, opts GetObjectOptions) (io.Reader, error) {
+func (c Client) GetObject(ctx context.Context, bucketName, objectName string, opts GetObjectOptions) (ObjectInterface, error) {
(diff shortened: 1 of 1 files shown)


api-get-object.go
```diff
@@ -29,8 +29,16 @@ import (
...
+type ObjectInterface interface {
+	Read(b []byte) (n int, err error)
+	Stat() (ObjectInfo, error)
+	ReadAt(b []byte, offset int64) (n int, err error)
+	Seek(offset int64, whence int) (n int64, err error)
+	Close() (err error)
+}
+
 // GetObject wrapper function that accepts a request context
-func (c Client) GetObject(ctx context.Context, bucketName, objectName string, opts GetObjectOptions) (*Object, error) {
```
klauspost, line 33: "Accept Interfaces, but return implementations", I seem to remember from somewhere. So we don't want to return interfaces.

While we could have the interface definition in this package, there technically isn't any problem in this being completely in your own code.

klauspost (changes requested):
Not sure if there is any reason to have this in the package.

Please provide your use case where the interface cannot simply be in your own code?

api-get-object.go
nikolaistraessle: If the return type is not an interface, it is possible to mock this method, but it is not possible to specify an object{} as the return value of the mock.

A better solution can be to change the return type to io.reader instead an own interface, then it is fully possible to mock it.

klauspost: `ObjectInterface` is the return type you are proposing. That is an interface.

However, you can convert the returned (*Object) to your interface in your calling code.

Or if you are looking to mock the Client it can easily be done with a shallow wrapper.

nikolaistraessle: @klauspost your right. I will close this pull request.

PR #1466, opened 2021-03-18
Add 200 to possible response codes for putBucketPolicy

joewhite86:
Hi guys,

we are working with Ceph and are getting a 200 instead of a 204 for `putBucketPolicy` if successful.

Even though Amazon defines a 200 as response status code in their documentation, they even have a 204 in their response example: https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutBucketPolicy.html

I could not find a test for PutBucketPolicy, so let me know if I should add one.


joewhite86:
We are using Ceph 14.2.11 if that matters.

harshavardhana (approved):

PR #1472, opened 2021-04-02
Add new method to get a STSWebIdentity credential assuming a role

sylr:
Hi,

The purpose of this PR is to be able to use a role in one AWS account which has no abilities except assuming a second role in a different account which has the correct policy to manage S3 buckets.

I've yet to test it but If someone sees this and thinks this is already possible then I would appreciate if you could show me how it can be done.

Otherwise I'd welcome your inputs about this :)

Regards.

### DIFF
--- a/pkg/credentials/sts_web_identity.go
+++ b/pkg/credentials/sts_web_identity.go
@@ -103,6 +103,27 @@ func NewSTSWebIdentity(stsEndpoint string, getWebIDTokenExpiry func() (*WebIdent
 	}), nil
 }
 
+// NewSTSWebIdentityWithAssumeRole is similar to NewSTSWebIdentity but allows to
+// specify a role to assume.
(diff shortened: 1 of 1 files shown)


harshavardhana (commented):
We already have

```go
// NewSTSAssumeRole returns a pointer to a new
// Credentials object wrapping the STSAssumeRole.
func NewSTSAssumeRole(stsEndpoint string, opts STSAssumeRoleOptions) (*Credentials, error) {
        if stsEndpoint == "" {
                return nil, errors.New("STS endpoint cannot be empty")
        }
        if opts.AccessKey == "" || opts.SecretKey == "" {
                return nil, errors.New("AssumeRole credentials access/secretkey is mandatory")
        }
        return New(&STSAssumeRole{
                Client: &http.Client{
                        Transport: http.DefaultTransport,
                },
                STSEndpoint: stsEndpoint,
                Options:     opts,
        }), nil
}
```

what is the point of this?

sylr:
I want to use STS Web Identity token, not Access or Secret keys.

PR #1488, opened 2021-05-11
lifecycle: And xml tag should always show both Prefix & Tag

vadmeste:
<And> is lifecycle is used when you want to specify a prefix and tag or
multiple tags with empty prefix (Prefix tag inside And should always be
there even if it is empty). Therefore, fields inside <And> should
always be visible for the S3 server, hence removing omitempty from xml
tag.

Signed-off-by: Anis Elleuch <anis@min.io>


kannappanr (approved):
LGTM

harshavardhana (approved):

PR #1501, opened 2021-06-09
fix: prevent out of memory by explicitly allocating buffer

copierrj:
This PR prevents out of memory errors by explicitly allocating a reasonably sized buffer.

The function io.CopyN allocates a buffer (indirectly in a helper function, see: https://golang.org/src/io/io.go#L413) with the same size as the provided value for param 'n'. 

Even for very large values of 'n' (read: downloading very large objects) this is usually not a problem as the required memory isn't allocated until it is actually written to. Because Reader.Read implementation are highly unlikely to use the entire buffer provided (or even a substantial part it for very large buffers), everything works just fine.

Under certain circumstances this large buffer however becomes a serious problem, leading to inevitable out of memory issues. Unfortunately this is quite difficult to reproduce as the number of bytes written by a single call to Reader.Read is largely outside the sphere of influence of a programmer.

What consistently triggered the problem in our setup was a combination of large objects (few dozens of GiBs), high bandwidth between client en server (10Gbps), plenty of cpu power available on the client side and a sudden increase in load on the server lowering the upload speed of individual requests.

### DIFF
--- a/api-get-object-file.go
+++ b/api-get-object-file.go
@@ -107,7 +107,8 @@ func (c Client) FGetObject(ctx context.Context, bucketName, objectName, filePath
 	}
 
 	// Write to the part file.
-	if _, err = io.CopyN(filePart, objectReader, objectStat.Size); err != nil {
+	buffer := make([]byte, 1024 * 1024)
+	if _, err = io.CopyBuffer(filePart, io.LimitReader(objectReader, objectStat.Size), buffer); err != nil {
 		return err
 	}
 


harshavardhana:
The way you interpreting is not correct - here the allocation happens only for the 32k buffer, not the entire l.N

```go
	if buf == nil {
		size := 32 * 1024
		if l, ok := src.(*LimitedReader); ok && int64(size) > l.N {
			if l.N < 1 {
				size = 1
			} else {
				size = int(l.N)
			}
		}
		buf = make([]byte, size)
	}
```

The fix you did won't fix anything other than use 1MiB buffer instead of 32KiB buffer. 

I don't think you have root caused your memory issues properly here. Please share how you arrived at this we can help you root cause it properly.

copierrj:
@harshavardhana you're right. I misread the relevant code. I'll revisit the tests we did and report back.

harshavardhana:
> @harshavardhana you're right. I misread the relevant code. I'll revisit the tests we did and report back.

Thanks closing the PR

PR #1502, opened 2021-06-14
Add custom headers to ListObjectsOptions.

klauspost:
Make it possible to add custom headers to ListObject calls.

### DIFF
--- a/api-list.go
+++ b/api-list.go
@@ -640,6 +640,9 @@ type ListObjectsOptions struct {
 // key-value pair will be part of the HTTP GET request
 // headers.
 func (o *ListObjectsOptions) Set(key, value string) {
+	if o.headers == nil {
+		o.headers = make(http.Header)
+	}
 	o.headers.Set(key, value)
 }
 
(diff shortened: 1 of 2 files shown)


harshavardhana (approved):

poornas (approved):

api-list.go
```diff
@@ -629,6 +632,15 @@ type ListObjectsOptions struct {
...
 	UseV1 bool
+
+	headers http.Header
+}
+
+// Set adds a key value pair to the options. The
+// key-value pair will be part of the HTTP GET request
+// headers.
+func (o *ListObjectsOptions) Set(key, value string) {
+	o.headers.Set(key, value)
```
vadmeste, line 643: This will crash if o.headers is not initialized. You nedd to initialize it here I guess.

vadmeste (changes requested):
One comment

api-list.go
harshavardhana, line 643:
```suggestion
        if o.headers == nil {
                o.headers = make(http.Header)
        }
	o.headers.Set(key, value)
```

harshavardhana (changes requested):

harshavardhana (approved):

PR #1506, opened 2021-06-21
mint: Add concurrent upload test

klauspost:
Fails before https://github.com/minio/minio/pull/12541 - fixed with PR.

### DIFF
--- a/functional_tests.go
+++ b/functional_tests.go
@@ -38,6 +38,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/dustin/go-humanize"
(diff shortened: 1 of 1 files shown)


functional_tests.go
```diff
@@ -1054,6 +1055,153 @@ func testGetObjectWithVersioning() {
...
+	}
+
+	sort.SliceStable(results, func(i, j int) bool {
+		return results[i].Size < results[j].Size
+	})
+
+	sort.SliceStable(buffers, func(i, j int) bool {
+		return len(buffers[i]) < len(buffers[j])
+	})
+
```
harshavardhana, line 1157: BTW stable sort is only needed if we are sorting with > 1 dimensions - it's enough to use `sort.Sort()` here since these are disconnected slices.

harshavardhana (changes requested):

functional_tests.go
klauspost: @harshavardhana Yeah, or if there are tiebreaks, not here, though.

harshavardhana (approved):

PR #1507, opened 2021-06-24
fix: use default IAM endpoint when fetching IMDSv2 token

iler-dev:
If a custom IAM endpoint is not specified, the IAM credential provider did not use the default endpoint to fetch the IMDSv2 token. This changes sets the default earlier so it is used for fetching the token.

### DIFF
--- a/pkg/credentials/iam_aws.go
+++ b/pkg/credentials/iam_aws.go
@@ -181,10 +181,6 @@ type ec2RoleCredRespBody struct {
 // be sent to fetch the rolling access credentials.
 // http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html
 func getIAMRoleURL(endpoint string) (*url.URL, error) {
-	if endpoint == "" {
-		endpoint = defaultIAMRoleEndpoint
-	}
-
 	u, err := url.Parse(endpoint)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


pkg/credentials/iam_aws.go
```diff
@@ -181,10 +181,6 @@ type ec2RoleCredRespBody struct {
 // be sent to fetch the rolling access credentials.
 // http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html
 func getIAMRoleURL(endpoint string) (*url.URL, error) {
-	if endpoint == "" {
-		endpoint = defaultIAMRoleEndpoint
-	}
-
```
harshavardhana, line 187: why is this removed?

harshavardhana (changes requested):

pkg/credentials/iam_aws.go
```diff
@@ -281,6 +277,10 @@ func fetchIMDSToken(client *http.Client, endpoint string) (string, error) {
 // If the credentials cannot be found, or there is an error
 // reading the response an error will be returned.
 func getCredentials(client *http.Client, endpoint string) (ec2RoleCredRespBody, error) {
+	if endpoint == "" {
+		endpoint = defaultIAMRoleEndpoint
+	}
```
harshavardhana, line 282: can you share the document where it says defaultIAMRoleEndpoint is the default for IMDS token?

harshavardhana (changes requested):

pkg/credentials/iam_aws.go
iler-dev: It seemed most appropriate to hoist the block up to `getCredentials()`. I can keep it, but it seems like dead code given it appears that only `getCredentials()` calls `getIAMRoleURL()`.

pkg/credentials/iam_aws.go
iler-dev: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html

pkg/credentials/iam_aws.go
harshavardhana: okay

pkg/credentials/iam_aws.go
harshavardhana: ah I see its same endpoint.. for newer feature as well.

harshavardhana (approved):

PR #1520, opened 2021-07-21
sts: add credential provider for AssumeRoleWithCertificate

aead:
This commit adds an STS S3 credential provider that fetches
S3 credentials from an STS API endpoint using a TLS client
certificate.

Ref: minio/minio#12748

### DIFF
--- a/pkg/credentials/sts-tls-identity.go
+++ b/pkg/credentials/sts-tls-identity.go
@@ -1,0 +1,192 @@
+// MinIO Go Library for Amazon S3 Compatible Cloud Storage
+// Copyright 2021 MinIO, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package credentials
+
+import (
+	"crypto/tls"
+	"encoding/xml"
+	"errors"
+	"io"
+	"net"
+	"net/http"
+	"net/url"
+	"strconv"
+	"time"
+)
+
+// CertificateIdentityOption is an optional AssumeRoleWithCertificate
+// parameter - e.g. a custom HTTP transport configuration or S3 credental
+// livetime.
+type CertificateIdentityOption func(*STSCertificateIdentity)
+
+// CertificateIdentityWithTransport returns a CertificateIdentityOption that
+// customizes the STSCertificateIdentity with the given http.RoundTripper.
+func CertificateIdentityWithTransport(t http.RoundTripper) CertificateIdentityOption {
+	return CertificateIdentityOption(func(i *STSCertificateIdentity) { i.Client.Transport = t })
+}
+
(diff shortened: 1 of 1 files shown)


pkg/credentials/sts-tls-identity.go
```diff
@@ -0,0 +1,150 @@
...
+	//   }
+	Client http.Client
+}
+
+var _ Provider = (*STSWebIdentity)(nil) // compiler check
+
+// NewSTSCertificateIdentity returns a STSCertificateIdentity that authenticates
+// to the given STS endpoint with the given TLS certificate and retrieves and
+// rotates S3 credentials.
+func NewSTSCertificateIdentity(endpoint string, certificate tls.Certificate) (*Credentials, error) {
```
harshavardhana, line 56: Can you add options pattern here? @aead since we need to provide `durationSeconds` as well here. 

```
func NewSTSCertificateIdentity(opts ...Options)
```

pkg/credentials/sts-tls-identity.go
```diff
@@ -0,0 +1,150 @@
...
+	return New(&STSCertificateIdentity{
+		STSEndpoint: endpoint,
+		Client: http.Client{
+			Transport: &http.Transport{
+				Proxy: http.ProxyFromEnvironment,
+				DialContext: (&net.Dialer{
+					Timeout:   30 * time.Second,
+					KeepAlive: 30 * time.Second,
+				}).DialContext,
+				ForceAttemptHTTP2:     true,
```
harshavardhana, line 72: why would we need to do HTTP2?

pkg/credentials/sts-tls-identity.go
```diff
@@ -0,0 +1,150 @@
...
+				MaxIdleConns:          100,
+				IdleConnTimeout:       90 * time.Second,
+				TLSHandshakeTimeout:   10 * time.Second,
+				ExpectContinueTimeout: 1 * time.Second,
```
harshavardhana, line 76: perhaps too low @aead 5secs is okay

pkg/credentials/sts-tls-identity.go
```diff
@@ -0,0 +1,150 @@
...
+	if endpoint == "" {
+		return nil, errors.New("STS endpoint cannot be empty")
+	}
+	if _, err := url.Parse(endpoint); err != nil {
+		return nil, err
+	}
+	return New(&STSCertificateIdentity{
+		STSEndpoint: endpoint,
+		Client: http.Client{
+			Transport: &http.Transport{
```
harshavardhana, line 66: We should make this Transport configuration here since this is a very specific mTLS based API.  We need to provide this in some way configurable. 

So 

```
WithTransport(t Transport) *Options {
}
```

Will work

harshavardhana (changes requested):

harshavardhana:
Should we also filter out certificates here for CAs before sending it to server ? @aead

pkg/credentials/sts-tls-identity.go
aead: That field does *not* require `HTTP/2`. It says that the client will attempt HTTP/2. HTTP/2 has several adv. compared to HTTP/1.x. If client and server support HTTP/2 it will be used. Otherwise, it won't.

aead:
> Should we also filter out certificates here for CAs before sending it to server ? @aead

That we shouldn't do for two reasons:
 1. The client may need to send an intermediate CA certificate. The client certificate may be issued by an intermediate CA but the server only has the root CA that issued the intermediate CA. `Client Cert <---> Intermediate CA <---> Root CA`.
    The server would need the intermediate CA certificate to verify the chain.
 2. Such filtering should happen on the server side. There can always be clients that don't do the filtering.

harshavardhana (approved):

PR #1528, opened 2021-08-05
Update Replication API to allow multi site replication.

poornas:
In addition this PR change ResetBucketReplication API to add Arn as query parameter to reset
replication for specific targets


kannappanr (approved):
LGTM

PR #1530, opened 2021-08-10
Accept ExpiredObjectDeleteMarker from lifecycle policy input as json

krisis:


harshavardhana (approved):

vadmeste:
Agree that DeleteMarker is confusing, but not sure if this breaks anything or not.

pkg/lifecycle/lifecycle.go
```diff
@@ -281,7 +281,7 @@ type Expiration struct {
 	XMLName      xml.Name           `xml:"Expiration,omitempty" json:"-"`
 	Date         ExpirationDate     `xml:"Date,omitempty" json:"Date,omitempty"`
 	Days         ExpirationDays     `xml:"Days,omitempty" json:"Days,omitempty"`
-	DeleteMarker ExpireDeleteMarker `xml:"ExpiredObjectDeleteMarker,omitempty"`
```
vadmeste, line 284: It looks like L292 needs to be updated as well

vadmeste (commented):
One comment

krisis:
> Agree that DeleteMarker is confusing, but not sure if this breaks anything or not.
@vadmeste it does as we don't parse the input.json as expected. This can be seen through the following example.
```
$ cat /tmp/lifecycle.json 
{
  "Rules": [
    {
      "ID": "expire-non-current",
      "Status": "Enabled",
      "Expiration": {
        "ExpiredObjectDeleteMarker": true
      },
      "NoncurrentVersionExpiration": {
        "NoncurrentDays": 30
      }
    }
  ]
}

$ mc ilm import minio/smoke-test < /tmp/lifecycle.json
Lifecycle configuration imported successfully to `minio/smoke-test`.
$ mc ilm export minio/smoke-test
{
 "Rules": [
  {
   "AbortIncompleteMultipartUpload": {},
   "Expiration": {
    "Date": "0001-01-01T00:00:00Z",
    "DeleteMarker": false
   },
   ...
   ]
 }
```
You can see that Expiration.DeleteMarker in the above json is `false` instead of `true`.

vadmeste:
> You can see that Expiration.DeleteMarker in the above json is false instead of true.

Yes, what I am saying is that it was DeleteMarker since the beginning so some users already use it. I can send another PR which understands both DeleteMarker & ExpiredObjectDeleteMarker

krisis:
> > You can see that Expiration.DeleteMarker in the above json is false instead of true.
> 
> Yes, what I am saying is that it was DeleteMarker since the beginning so some users already use it. I can send another PR which understands both DeleteMarker & ExpiredObjectDeleteMarker

Okay. Should we also marshal json (for `mc ilm export ...`) depending on whether the input json used DeleteMarker/ExpiredObjectDeleteMarker?

vadmeste:
> Okay. Should we also marshal json (for mc ilm export ...) depending on whether the input json used DeleteMarker/ExpiredObjectDeleteMarker?

No, exporting should show ExpiredObjectDeleteMarker but importing should understand DeleteMarker as well

krisis:
@vadmeste okay. Please send the PR you mentioned above. I shall close this one.

PR #1531, opened 2021-08-11
Fix error returned by statObject for delete markers

poornas:


vadmeste (approved):
Does this fix any replication issue in server side ? just asking

api-stat.go
```diff
@@ -99,11 +99,11 @@ func (c Client) statObject(ctx context.Context, bucketName, objectName string, o
 	if err != nil {
 		return ObjectInfo{}, err
 	}
-	deleteMarker := resp.Header.Get(amzDeleteMarker) == "true"
 
 	if resp != nil {
+		deleteMarker := resp.Header.Get(amzDeleteMarker) == "true"
 		if resp.StatusCode != http.StatusOK && resp.StatusCode != http.StatusPartialContent {
-			if resp.StatusCode == http.StatusBadRequest && opts.VersionID != "" && deleteMarker {
+			if resp.StatusCode == http.StatusMethodNotAllowed && opts.VersionID != "" && deleteMarker {
```
harshavardhana, line 106: I am still not sure why the subsequent code doesn't work correctly

harshavardhana (changes requested):

api-stat.go
poornas: The problem is httpRespToErrorResponse error is used here which returns 405 as Error code in the response instead of using Minio error code MethodNotAllowed.  The error translation to object error doesn't work on server side due to this.

harshavardhana: While this change is okay to take in, I am not sure why is it not possible to do this on the server side

```
if objInfo.IsDeleteMarker && resp.StatusCode == http.StatusMethodNotAllowed && resp.Code == http.StatusText(http.StatusMethodNotAllowed)
```

?

harshavardhana (changes requested):

harshavardhana (approved):

PR #1532, opened 2021-08-13
fix: allow force delete options for bucket and objects

harshavardhana:


klauspost (approved):
LGTM

api-remove.go
```diff
@@ -29,6 +29,52 @@ import (
...
+// RemoveBucketWithOptions deletes the bucket name, additionally
+// takes options to force purging a bucket or a prefix when the
+// endpoint is MinIO server
+//
+// All objects (including all object versions and delete markers)
+// in the bucket will be deleted if bucket options set ForceDelete
+// as true.
+func (c Client) RemoveBucketWithOptions(ctx context.Context, bucketName string, opts BucketOptions) error {
+	// Input validation.
+	if err := s3utils.CheckValidBucketName(bucketName); err != nil {
```
vadmeste, line 47: The comment says it can remove a prefix but we are validating this as a bucket name, maybe what we can do is we call this function: RemoveWithOptions() and add a Prefix string in BucketOptions, which can be called RemoveOptions I guess.

vadmeste (commented):
Some comments

poornas (approved):
LGTM

api-remove.go
harshavardhana: BucketName is the most important requirement - prefix doesn't need validation since its allowed to be empty.

harshavardhana: That comment is wrong

PR #1533, opened 2021-08-17
sts: Always slash expiry time by 80%

vadmeste:
Not tested yet.

By default, STS expiration time is reduced by 10 seconds, but this is
not always good since an S3 call can be called near expiration time but
evaluated after the expiration time - the window here is 10 seconds
which can be not enough to upload some large data to an S3 server before
this latter rejects it with bad (expired) credentials.

This commit will always slash expiration time by 80% instead.

### DIFF
--- a/pkg/credentials/credentials.go
+++ b/pkg/credentials/credentials.go
@@ -22,8 +22,13 @@ import (
 	"time"
 )
 
-// STSVersion sts version string
-const STSVersion = "2011-06-15"
+const (
+	// STSVersion sts version string
+	STSVersion = "2011-06-15"
+
+	// How much duration to slash from the given expiration duration
+	defaultExpiryWindow = 0.8
+)
 
 // A Value is the AWS credentials value for individual credential fields.
 type Value struct {
@@ -82,10 +87,15 @@ type Expiry struct {
 // the expiration time given to ensure no requests are made with expired
 // tokens.
 func (e *Expiry) SetExpiration(expiration time.Time, window time.Duration) {
-	e.expiration = expiration
-	if window > 0 {
-		e.expiration = e.expiration.Add(-window)
+	if e.CurrentTime == nil {
+		e.CurrentTime = time.Now
 	}
+	cut := window
+	if cut < 0 {
+		expireIn := expiration.Sub(e.CurrentTime())
+		cut = time.Duration(float64(expireIn) * (1 - defaultExpiryWindow))
+	}
+	e.expiration = expiration.Add(-cut)
 }
 
 // IsExpired returns if the credentials are expired.
--- a/pkg/credentials/iam_aws.go
+++ b/pkg/credentials/iam_aws.go
@@ -38,8 +38,9 @@ import (
 // prior to the credentials actually expiring. This is beneficial
 // so race conditions with expiring credentials do not cause
 // request to fail unexpectedly due to ExpiredTokenException exceptions.
-// When set to -1, which is the default value, refreshing will be triggered
-// when 80% of the elapsed time until the actual expiration time is passed.
+// DefaultExpiryWindow can be used as parameter to (*Expiry).SetExpiration.
+// When used the tokens refresh will be triggered when 80% of the elapsed
+// time until the actual expiration time is passed.
 const DefaultExpiryWindow = -1
 
 // A IAM retrieves credentials from the EC2 service, and keeps track if
--- a/pkg/credentials/iam_aws.go
+++ b/pkg/credentials/iam_aws.go
@@ -38,7 +38,9 @@ import (
 // prior to the credentials actually expiring. This is beneficial
 // so race conditions with expiring credentials do not cause
 // request to fail unexpectedly due to ExpiredTokenException exceptions.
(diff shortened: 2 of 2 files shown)


pkg/credentials/iam_aws.go
```diff
@@ -33,11 +33,12 @@ import (
...
 )
 
-// DefaultExpiryWindow - Default expiry window.
+// Deprecated: DefaultExpiryWindow - Default expiry window.
 // ExpiryWindow will allow the credentials to trigger refreshing
 // prior to the credentials actually expiring. This is beneficial
 // so race conditions with expiring credentials do not cause
 // request to fail unexpectedly due to ExpiredTokenException exceptions.
+// Deprecated in August 2021.
 const DefaultExpiryWindow = time.Second * 10 // 10 secs
```
klauspost, line 42:
```suggestion
const DefaultExpiryWindow = -1
```

pkg/credentials/credentials.go
```diff
@@ -81,11 +86,18 @@ type Expiry struct {
...
 func (e *Expiry) SetExpiration(expiration time.Time, window time.Duration) {
-	e.expiration = expiration
-	if window > 0 {
-		e.expiration = e.expiration.Add(-window)
+	if e.CurrentTime == nil {
+		e.CurrentTime = time.Now
+	}
+	expireIn := expiration.Sub(e.CurrentTime())
+	if expireIn < 0 {
+		panic("expiration date is too soon")
```
klauspost, line 97: Don't panic. Libraries shouldn't panic. Send it as requested.

klauspost (changes requested):
I think it would be more reasonable to only use the 80% when the `DefaultExpiryWindow` is used.

I would keep it, but have it be `-1`. If the `window` sent is equal, use 80%.

Don't deprecate it and allow previous behavior to work as expected.

What about also having a default minimum time?

pkg/credentials/credentials.go
```diff
@@ -81,11 +86,18 @@ type Expiry struct {
...
+//
+// window argument is deprecated in August 2021
 func (e *Expiry) SetExpiration(expiration time.Time, window time.Duration) {
-	e.expiration = expiration
-	if window > 0 {
-		e.expiration = e.expiration.Add(-window)
+	if e.CurrentTime == nil {
+		e.CurrentTime = time.Now
+	}
+	expireIn := expiration.Sub(e.CurrentTime())
```
klauspost, line 95:
```suggestion
	expireIn := time.Until(expiration)
```

vadmeste: I am actually not sure why we have CurrentTime in the first place, that's why I am using it.

pkg/credentials/credentials.go
vadmeste: > Don't panic. Libraries shouldn't panic. Send it as requested.

Yes you are right, fixed

pkg/credentials/iam_aws.go
vadmeste: Good idea, done

pkg/credentials/credentials.go
klauspost: @vadmeste Yeah, lets keep it.

pkg/credentials/iam_aws.go
```diff
@@ -33,12 +33,13 @@ import (
...
 )
 
-// DefaultExpiryWindow - Default expiry window.
+// Deprecated: DefaultExpiryWindow - Default expiry window.
 // ExpiryWindow will allow the credentials to trigger refreshing
 // prior to the credentials actually expiring. This is beneficial
 // so race conditions with expiring credentials do not cause
 // request to fail unexpectedly due to ExpiredTokenException exceptions.
-const DefaultExpiryWindow = time.Second * 10 // 10 secs
+// Deprecated in August 2021.
```
klauspost, line 41: Not deprecated, but update docs to reflect the new functionality of expiring when 80% has been passed.

pkg/credentials/credentials.go
vadmeste: > @vadmeste Yeah, lets keep it.

It looks like it is used for testing.

pkg/credentials/iam_aws.go
```diff
@@ -38,7 +38,9 @@ import (
 // prior to the credentials actually expiring. This is beneficial
 // so race conditions with expiring credentials do not cause
 // request to fail unexpectedly due to ExpiredTokenException exceptions.
-const DefaultExpiryWindow = time.Second * 10 // 10 secs
+// When set to -1, which is the default value, refreshing will be triggered
+// when 80% of the elapsed time until the actual expiration time is passed.
+const DefaultExpiryWindow = -1
```
klauspost, line 43:
```suggestion
// DefaultExpiryWindow can be used as parameter to (*Expiry).SetExpiration.
// When used the tokens refresh will be triggered when 80% 
// of the elapsed time until the actual expiration time is passed.
const DefaultExpiryWindow = -1
```

(Remove text above)

klauspost (commented):
Doc update. Rest LGTM.

harshavardhana (approved):

klauspost (approved):

PR #1534, opened 2021-08-19
Rename BucketOptions to RemoveBucketOptions

vadmeste:
RemoveBucketOptions should be a better name for removing a bucket.


klauspost (approved):
LGTM

harshavardhana:
Breaking change ship has sailed for now, we can slate this for v8 change in future.

PR #1538, opened 2021-09-02
Implement RestoreObject API

vadmeste:
Implement the following API https://docs.aws.amazon.com/AmazonS3/latest/API/API_RestoreObject.html

### DIFF
--- a/api-datatypes.go
+++ b/api-datatypes.go
@@ -86,6 +86,14 @@ type UploadInfo struct {
 	ExpirationRuleID string
 }
 
+// RestoreInfo contains information of the restore operation of an archived object
+type RestoreInfo struct {
+	// Is the restoring operation is still ongoing
+	OngoingRestore bool
+	// When the restored copy of the archived object will be removed
+	ExpiryTime time.Time
+}
+
 // ObjectInfo container for object metadata.
 type ObjectInfo struct {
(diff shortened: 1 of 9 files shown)


klauspost (approved):
LGTM, just a minor change.

vadmeste:
@harshavardhana @klauspost I added a separate commit which parses x-amz-restore header in a HEAD request. PTAL again.

api-datatypes.go
```diff
@@ -85,6 +86,12 @@ type UploadInfo struct {
 	ExpirationRuleID string
 }
 
+// RestoreInfo contains information of the restore operation of an archived object
+type RestoreInfo struct {
+	OngoingRestore bool
+	ExpiryTime     time.Time
```
klauspost, line 92: It is not obvious what "ExpiryTime" is. Add comments to explain fields.

Do we know the start time of the restore. Would be nice to have, if feasible.

vadmeste: > It is not obvious what "ExpiryTime" is. Add comments to explain fields.

Done. Restore API will actually get a copy of the archived object in the first time. The object will be automatically removed from the fast tier after x days.

> Do we know the start time of the restore. Would be nice to have, if feasible.

No, S3 doesn't expose that.

klauspost (approved):
LGTM

harshavardhana (approved):

PR #1551, opened 2021-09-18
fix type of channel in RemoveObjects signature

elahe-dastan:
I was using your docs to use "RemoveObjects" in my project but I got an error and found the actual type of the channel is "ObjectInfo".


vadmeste (approved):
Thanks @elahe-dastan

harshavardhana (approved):

PR #1555, opened 2021-09-21
simplify healthcheck implementation

harshavardhana:


api.go
```diff
@@ -642,22 +643,17 @@ func (c Client) executeMethod(ctx context.Context, method string, metadata reque
 				continue // Retry.
 			}
 
-			if atomic.LoadInt32(&c.healthCheck) != unknown && IsNetworkOrHostDown(err, false) {
-				select {
-				case c.healthCheckCh <- struct{}{}:
-				default:
-				}
+			if IsNetworkOrHostDown(err, false) {
```
poornas, line 646: shouldn't set this offline unless health check is enabled

api.go
```diff
@@ -642,22 +643,17 @@ func (c Client) executeMethod(ctx context.Context, method string, metadata reque
...
 		}
 		// Initiate the request.
 		res, err = c.do(req)
 		if err != nil {
-			if atomic.LoadInt32(&c.healthCheck) != unknown && IsNetworkOrHostDown(err, false) {
-				select {
-				case c.healthCheckCh <- struct{}{}:
-				default:
-				}
+			if IsNetworkOrHostDown(err, false) {
```
poornas, line 655:
```suggestion
			if atomic.LoadInt32(&c.healthCheck) != unknown && IsNetworkOrHostDown(err, false) {
```

poornas: same reason - client may not have enabled health check

poornas (changes requested):

api.go
```diff
@@ -397,66 +393,56 @@ func (c *Client) hashMaterials(isMd5Requested bool) (hashAlgos map[string]md5sim
...
 	probeBucketName := randString(60, rand.NewSource(time.Now().UnixNano()), "probe-health-")
 	go func(duration time.Duration) {
 		timer := time.NewTimer(duration)
 		defer timer.Stop()
 		for {
 			select {
 			case <-ctx.Done():
-				close(c.healthCheckCh)
-				atomic.StoreInt32(&c.healthCheck, unknown)
+				atomic.StoreInt32(&c.healthStatus, online)
```
poornas, line 428: ```c.healthCheck = false```
after health check context is canceled.

api.go
```diff
@@ -397,66 +393,56 @@ func (c *Client) hashMaterials(isMd5Requested bool) (hashAlgos map[string]md5sim
...
 				return
 			case <-timer.C:
 
 				timer.Reset(duration)
 				// Do health check the first time and ONLY if the connection is marked offline
-				if c.IsOffline() || c.lastOnline.IsZero() {
+				if c.IsOffline() {
```
poornas, line 434:
```suggestion
				if c.IsOffline() && c.healthCheck {
```

api.go
```diff
@@ -589,6 +575,14 @@ var successStatus = []int{
 // request upon any error up to maxRetries attempts in a binomially
 // delayed manner using a standard back off algorithm.
 func (c Client) executeMethod(ctx context.Context, method string, metadata requestMetadata) (res *http.Response, err error) {
+	if c.IsOffline() {
```
poornas, line 578: if healthcheck is not enabled, this will always return error since default is offline

harshavardhana:
Now its properly fixed as per expectation @poornas PTAL

klauspost (approved):
LGTM

api.go
```diff
@@ -588,7 +587,15 @@ var successStatus = []int{
 // executeMethod - instantiates a given method, and retries the
 // request upon any error up to maxRetries attempts in a binomially
 // delayed manner using a standard back off algorithm.
-func (c Client) executeMethod(ctx context.Context, method string, metadata requestMetadata) (res *http.Response, err error) {
+func (c *Client) executeMethod(ctx context.Context, method string, metadata requestMetadata) (res *http.Response, err error) {
+	if c.IsOffline() {
+		return nil, &url.Error{
+			Op:  method,
+			URL: c.endpointURL.String(),
+			Err: errors.New("Connection closed by foreign host " + c.endpointURL.String() + ". Retry again."),
```
vadmeste, line 595: Can we change this error message to avoid making it very standard ? Also why this is always foreign host that closed connection, it looks like c.IsOffline() can return true after networking connectivity.

vadmeste (commented):
Two comments

api.go
```diff
@@ -404,59 +402,54 @@ const (
...
 					_, err := c.getBucketLocation(context.Background(), probeBucketName)
 					if err != nil && IsNetworkOrHostDown(err, false) {
-						atomic.CompareAndSwapInt32(&c.healthCheck, online, offline)
-					} else {
-						switch ToErrorResponse(err).Code {
-						case "NoSuchBucket", "AccessDenied", "":
-							c.lastOnline = time.Now()
-							atomic.CompareAndSwapInt32(&c.healthCheck, offline, online)
-						}
+						c.markOffline()
```
poornas, line 444: since this is already offline, no need to mark offline again

vadmeste: > since this is already offline, no need to mark offline again

yeah but I think if IsNetworkOrHostDown() doesn't return true, then we should switch online in all cases

poornas (approved):

PR #1558, opened 2021-09-27
replication: Allow null version to be specified in options

poornas:
This is for supporting existing object replication

### DIFF
--- a/api-compose-object.go
+++ b/api-compose-object.go
@@ -243,8 +243,10 @@ func (c *Client) copyObjectDo(ctx context.Context, srcBucket, srcObject, destBuc
 		customHeader: headers,
 	}
 	if dstOpts.Internal.SourceVersionID != "" {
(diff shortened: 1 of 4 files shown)


api-compose-object.go
```diff
@@ -243,8 +243,10 @@ func (c *Client) copyObjectDo(ctx context.Context, srcBucket, srcObject, destBuc
 		customHeader: headers,
 	}
 	if dstOpts.Internal.SourceVersionID != "" {
-		if _, err := uuid.Parse(dstOpts.Internal.SourceVersionID); err != nil {
-			return ObjectInfo{}, errInvalidArgument(err.Error())
+		if dstOpts.Internal.SourceVersionID != "null" {
```
harshavardhana, line 246: `null` should be stored as `const` @poornas

harshavardhana (changes requested):

harshavardhana (approved):

PR #1563, opened 2021-10-06
[WIP] credentials: Retry AWS IAM failure

vadmeste:
This change the chain provider behavior to retry errRetryRetrieve
error returned by a provider.

Also it moves NewRetryTimer to a separate pacakge, pkg/utils


vadmeste:
No need for review. It needs some discussion first.

We need to go over all providers and make them return two errors, errIgnore and errRetry.

- errIgnore when ~/.aws/credentials is not found
- errRetry when a connection to IAM endpoint is timed out or DNS was unresolvable.

PR #1564, opened 2021-10-06
Retry getting creds from IAM AWS in case of error

vadmeste:
When AWS_WEB_IDENTITY_TOKEN_FILE is specified, a call will be made to
AWS IAM to generate credentials, which can fail due to networking issue.

This PR will retry the request until succeed.

A more comprenhensive fix for all other types of providers should come.
The chain provider should retry with an exponential backoff. Providers
should also return a specific error to ask the chain provider to retry
the request.


pkg/credentials/iam_aws.go
```diff
@@ -116,11 +116,19 @@ func (m *IAM) Retrieve() (Value, error) {
...
+		//
+		// Loop until getting a valid credentials
+		for {
+			stsWebIdentityCreds, err := creds.Retrieve()
+			if err == nil {
+				m.SetExpiration(creds.Expiration(), DefaultExpiryWindow)
+				return stsWebIdentityCreds, nil
+			}
+
+			time.Sleep(10 * time.Second)
```
harshavardhana, line 130: We should log the error @vadmeste perhaps by taking some custom logger function here as part of the API such that we can log the errors here.

harshavardhana (changes requested):

pkg/credentials/iam_aws.go
vadmeste: > We should log the error @vadmeste perhaps by taking some custom logger function here as part of the API such that we can log the errors here.

For logging, the only thing that I can think of is something like this:
`SetCustomLogger(fn func(error))`

But also I wanted to send a more comprehensive PR today, https://github.com/minio/minio-go/pull/1563 but I want to discuss it first. Maybe let's merge this first and then I can send another PR with more careful testing.

harshavardhana: Sure atleast we need logging here @vadmeste

pkg/credentials/iam_aws.go
```diff
@@ -116,11 +117,20 @@ func (m *IAM) Retrieve() (Value, error) {
...
+		//
+		// Loop until getting a valid credentials
+		for {
+			stsWebIdentityCreds, err := creds.Retrieve()
+			if err == nil {
+				m.SetExpiration(creds.Expiration(), DefaultExpiryWindow)
+				return stsWebIdentityCreds, nil
+			}
+
+			log.Printf("unable to fetch STS credentials: %s", err.Error())
```
harshavardhana, line 131: We can't log like this @vadmeste .. it needs to be a custom logger.

harshavardhana (changes requested):

pkg/credentials/iam_aws.go
vadmeste: > it needs to be a custom logger.

This is more complex than it looks. That is why I wanted a separate PR for this, since we are also going to release today, I did not want to risk it.

PR #1584, opened 2021-11-15
Remove duplicated code

### LINKED ISSUE #1583: StatObject validates names twice
Either `statObject` should not be doing the validation, or `StatObject` is redundant, because the first 7 lines of each function are identical.

https://github.com/minio/minio-go/blob/daea25af3ddbba88ddd7201c052d538348e4676a/api-stat.go#L60-L80

klauspost:
Clean up duplicate internal function.

Fixes #1583


harshavardhana:
This causes stack overflow @klauspost since we expose StatObject() lower level

klauspost:
@harshavardhana ok, easy fix.

harshavardhana (approved):

PR #1585, opened 2021-11-16
Report non-XML formated error details

simonswine:
This avoids hiding further error context in the case when it is coming from a layer, which is not necessarily using the S3 XML error format.

Note: The separate commit also ensures an unrelated test case is called

### DIFF
--- a/api-error-response.go
+++ b/api-error-response.go
@@ -18,8 +18,11 @@
 package minio
 
 import (
+	"bytes"
 	"encoding/xml"
 	"fmt"
+	"io"
+	"io/ioutil"
 	"net/http"
 )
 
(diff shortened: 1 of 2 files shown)


api-error-response.go
```diff
@@ -156,10 +171,14 @@ func httpRespToErrorResponse(resp *http.Response, bucketName, objectName string)
 				Key:        objectName,
 			}
 		default:
+			msg := resp.Status
+			if len(errBody) > 0 {
+				msg = string(errBody)
+			}
```
klauspost, line 177: Lets keep the message to the first 1K:
```suggestion
			if len(errBody) > 0 {
				msg = string(errBody)
				if len(msg) > 1024 {
					msg = msg[:1024] + "..."
				}
			}
```

api-error-response.go
```diff
@@ -98,6 +101,18 @@ const (
 	reportIssue = "Please report this issue at https://github.com/minio/minio-go/issues."
 )
 
+// xmlDecodeAndBody reads the whole body up to a certain limit and tries to XML
+// decode it. In any case the read body will be returned.
+func xmlDecodeAndBody(bodyReader io.Reader, v interface{}) ([]byte, error) {
+	// read the whole body (up to a certain limit)
```
klauspost, line 107:
```suggestion
	// read the whole body (up to 1MB)
```

api-error-response.go
```diff
@@ -98,6 +101,18 @@ const (
...
+	const maxBodyLength = int64(1024 * 1024)
```
klauspost, line 108:
```suggestion
	const maxBodyLength = 1 << 20
```

api-error-response.go
```diff
@@ -98,6 +101,18 @@ const (
 	reportIssue = "Please report this issue at https://github.com/minio/minio-go/issues."
 )
 
+// xmlDecodeAndBody reads the whole body up to a certain limit and tries to XML
+// decode it. In any case the read body will be returned.
```
klauspost, line 105:
```suggestion
// xmlDecodeAndBody reads the whole body up to 1MB and
// tries to XML decode it into v.
// The body that was read and any error from reading or decoding is returned. 
```

klauspost (changes requested):
Good idea. Some minor changes.

klauspost:
@simonswine Seems like some formatting is required. Otherwise LGTM.

simonswine:
> @simonswine Seems like some formatting is required. Otherwise LGTM.

should be good now

klauspost (approved):

harshavardhana (approved):

PR #1601, opened 2021-12-17
remove redirector for S3 API calls

harshavardhana:
S3 APIs behave in non-standard style
not following HTTP spec for redirections

Refer here

https://github.com/aws/aws-sdk-go/blob/main/service/s3/s3manager/bucket_region.go#L117

```
// Disable HTTP redirects to prevent an invalid 301 from eating the response
// because Go's HTTP client will fail, and drop the response if an 301 is
// received without a location header. S3 will return a 301 without the
// location header for HeadObject API calls.
req.DisableFollowRedirects = true
```

Fixes the issue https://github.com/minio/minio/issues/13918

### DIFF
--- a/api.go
+++ b/api.go
@@ -182,67 +182,6 @@ func (r *lockedRandSource) Seed(seed int64) {
 	r.lk.Unlock()
 }
 
-// Redirect requests by re signing the request.
-func (c *Client) redirectHeaders(req *http.Request, via []*http.Request) error {
-	if len(via) >= 5 {
-		return errors.New("stopped after 5 redirects")
-	}
-	if len(via) == 0 {
-		return nil
-	}
-	lastRequest := via[len(via)-1]
-	var reAuth bool
-	for attr, val := range lastRequest.Header {
-		// if hosts do not match do not copy Authorization header
-		if attr == "Authorization" && req.Host != lastRequest.Host {
-			reAuth = true
-			continue
-		}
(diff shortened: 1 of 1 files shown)


klauspost:
@harshavardhana But the linked code is only for `GetBucketRegion`. It seems like this only applies for this call, not others.

It would seem that this change would be for *all* calls, which I could imagine would give a lot of regressions.

harshavardhana:
> @harshavardhana But the linked code is only for `GetBucketRegion`. It seems like this only applies for this call, not others.
> 
> It would seem that this change would be for *all* calls, which I could imagine would give a lot of regressions.

There is no redirection supported in S3 API. We added it for support some niche usecase.

I am just removing that nothing should regress anywhere AFAIK.

harshavardhana:
> @harshavardhana But the linked code is only for `GetBucketRegion`. It seems like this only applies for this call, not others.
> 
> It would seem that this change would be for *all* calls, which I could imagine would give a lot of regressions.

GetBucketRegion is the first call you need to do before you do any bucket operations.

The redirection logic doesn't work with S3 protocol due to payload signature etc.

We had a hack on our end which worked to some extent for some calls that do not have a payload or body.

harshavardhana:
PTAL @donatello @vadmeste

donatello (approved):

poornas (approved):

PR #1602, opened 2021-12-22
parse error responses for STS requests

harshavardhana:
provide more descriptive error messages
to the caller instead of just HTTP status


poornas (approved):
LGTM

donatello (approved):
Much needed - we can probably be even more specific based on this framework

PR #1615, opened 2022-01-27
Allow to pass headers in Presigned API

lanzrein:
- Allow to pass extraHeaders to PresignedPutObject and PresignedGetObject
- Allow to pass reqParams to PresignedPutObject
- Add an example to illustrate the use of passing extraHeaders with PresignedPutObject. 

This feature will allow to create presigned URLs with extra headers. With this we can pass the headers necessary for KMS encryption in an AWS S3 setting.

### DIFF
--- a/api-presigned.go
+++ b/api-presigned.go
@@ -66,11 +66,11 @@ func (c *Client) presignURL(ctx context.Context, method string, bucketName strin
 // data without credentials. URL can have a maximum expiry of
 // upto 7days or a minimum of 1sec. Additionally you can override
 // a set of response headers using the query parameters.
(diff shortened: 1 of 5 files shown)


klauspost:
👎🏼  This is a breaking change. 

Use `(*Client).PresignHeader` to generate presigned urls with extra headers.

harshavardhana:
> 👎🏼  This is a breaking change. 
> 
> Use `(*Client).PresignHeader` to generate presigned urls with extra headers.

We are not making breaking changes right now @lanzrein feel free to use the lower API for now.

Or write a different API name for each such change.

lanzrein:
thanks for the feedback I will close this MR.

PR #1619, opened 2022-02-08
Add API function to get resync replication status

poornas:


harshavardhana (approved):

api-bucket-replication.go
```diff
@@ -256,3 +256,41 @@ func (c *Client) resetBucketReplicationOnTarget(ctx context.Context, bucketName
...
+		return rinfo, httpRespToErrorResponse(resp, bucketName, "")
+	}
+	respBytes, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return rinfo, err
+	}
+
+	if err := json.Unmarshal(respBytes, &rinfo); err != nil {
+		return rinfo, err
+	}
```
harshavardhana, line 294: Actually a cleaner way is to use `json.NewDecoer(resp.Body).Decode(&rinfo)` @poornas

harshavardhana (changes requested):

api-bucket-replication.go
poornas: addressed this

harshavardhana (approved):

PR #1624, opened 2022-02-22
fix: ipv6 support without 'port' add missing brackets

harshavardhana:


klauspost (approved):

api.go
```diff
@@ -869,16 +869,24 @@ func (c *Client) makeTargetURL(bucketName, objectName, bucketLocation string, is
...
 
+	var isIPV6 bool
 	// Strip port 80 and 443 so we won't send these ports in Host header.
 	// The reason is that browsers and curl automatically remove :80 and :443
 	// with the generated presigned urls, then a signature mismatch error.
 	if h, p, err := net.SplitHostPort(host); err == nil {
 		if scheme == "http" && p == "80" || scheme == "https" && p == "443" {
 			host = h
+			if ip := net.ParseIP(h); ip != nil {
+				isIPV6 = ip.To16() != nil
```
vadmeste, line 880: maybe let's do:
```
if ipTo16() != nil {
  host = "[" + host + "]"
}
```
it seems easier

vadmeste (commented):
One comment, LGTM otherwise

api.go
```diff
@@ -875,10 +875,14 @@ func (c *Client) makeTargetURL(bucketName, objectName, bucketLocation string, is
 	if h, p, err := net.SplitHostPort(host); err == nil {
 		if scheme == "http" && p == "80" || scheme == "https" && p == "443" {
 			host = h
+			if ip := net.ParseIP(h); ip != nil {
```
vadmeste, line 878: I think what is missing here is to check if this is an ipv6 address, so we need just another if

harshavardhana: that is already added ParseIP parses IPv4 and IPv6 and then ip.To16() says that its IPv6 @vadmeste

vadmeste: > that is already added ParseIP parses IPv4 and IPv6 and then ip.To16() says that its IPv6 @vadmeste

what I am saying is that there is no ip.To16() in your code right now

harshavardhana: done fixed.

poornas (approved):

PR #1627, opened 2022-03-08
retry 499 error if seen by client

harshavardhana:


vadmeste:
I really don't think this will be ever seen by a client. 

499 is a nginx error code and it happens when the client closes the connection before it receives the response. https://www.nginx.com/resources/wiki/extending/api/http/

Same thing for MinIO
```
ErrClientDisconnected: {   
                Code:           "ClientDisconnected", 
                Description:    "Client disconnected before response was ready",
                HTTPStatusCode: 499, // No official code, use nginx value. 
        }, 
```

harshavardhana:
> I really don't think this will be ever seen by a client.
> 
> 499 is a nginx error code and it happens when the client closes the connection before it receives the response. https://www.nginx.com/resources/wiki/extending/api/http/
> 
> Same thing for MinIO
> 
> ```
> ErrClientDisconnected: {   
>                 Code:           "ClientDisconnected", 
>                 Description:    "Client disconnected before response was ready",
>                 HTTPStatusCode: 499, // No official code, use nginx value. 
>         }, 
> ```

The idea here is if seen @vadmeste

vadmeste (approved):
LGTM

PR #1631, opened 2022-04-10
Update api-list.go

shanggx:
If value of delimiter in urlValues is empty, http request to ctyun (电信天翼云OOS对象存储) will be failed with error 400 (Bad Request) and error message "The length of delimiter must be 1."


harshavardhana:
This is incompatible with AWS S3, fix your storage vendor to handle this properly. `delimiter` can be set and it can be empty string according to AWS S3 specs.

PR #1635, opened 2022-04-11
More fixes in ipv6 parsing in get bucket location

vadmeste:
Also fix the endpoint validation. The code was assuming that any address
has ':' in it means it has a port but this is not true.


harshavardhana (changes requested):
Please add unit tests for the new changes

harshavardhana (approved):

kannappanr (approved):

PR #1638, opened 2022-04-13
keep stat() and list() lastModified upto same resolution

### LINKED ISSUE #1632: Inconsistent LastModified field
I had a brief search and could not find a similar issue.

When fetching object info via `client.StatObject` and `client.ListObjects` the last modified field is not equal. For `StatObject` the LastModified field only seems to be accurate to the second, but `ListObjects` seems to be accurate to the millisecond.

For example the below test:
```go
func TestLastModified(t *testing.T) {
	ctx := context.Background()
	const bucket = "testing"
	const path = "mc_tests/foo"

	uploadInfo, err := c.PutObject(ctx, bucket, path, strings.NewReader("foo"), 3, minio.PutObjectOptions{})
	if err != nil {
		t.Fatal(err)
	}
	fmt.Println(uploadInfo.LastModified)

	info, err := c.StatObject(ctx, bucket, path, minio.GetObjectOptions{})
	if err != nil {
		t.Fatal(err)
	}
	fmt.Println(info.LastModified)

	ch := c.ListObjects(ctx, bucket, minio.ListObjectsOptions{
		Prefix: path,
	})
	for obj := range ch {
		if obj.Err != nil {
			t.Fatal(obj.Err)
		}
		fmt.Println(obj.LastModified)
	}
	
	t.Fail()
}
```

outputs:

```sh
0001-01-01 00:00:00 +0000 UTC
2022-04-10 20:51:06 +0000 UTC
2022-04-10 20:51:06.821 +0000 UTC
```

I'm unsure as to why the upload info has the `LastModified` field but it's not populated, but that's not as important!

This leads to inconsistent behaviour when comparing the last modified times. Is this behaviour intended?

The Minio server this test is using is running `Version: 2022-04-01T03:41:39Z`

harshavardhana:
fixes #1632

### DIFF
--- a/api-list.go
+++ b/api-list.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"time"
 
 	"github.com/minio/minio-go/v7/pkg/s3utils"
 )
@@ -239,6 +240,7 @@ func (c *Client) listObjectsV2Query(ctx context.Context, bucketName, objectPrefi
 		if err != nil {
 			return listBucketResult, err
 		}
+		listBucketResult.Contents[i].LastModified = listBucketResult.Contents[i].LastModified.Truncate(time.Millisecond)
 	}
 
 	for i, obj := range listBucketResult.CommonPrefixes {
@@ -384,7 +386,7 @@ func (c *Client) listObjectVersions(ctx context.Context, bucketName string, opts
 				info := ObjectInfo{
 					ETag:           trimEtag(version.ETag),
 					Key:            version.Key,
-					LastModified:   version.LastModified,
+					LastModified:   version.LastModified.Truncate(time.Millisecond),
 					Size:           version.Size,
 					Owner:          version.Owner,
 					StorageClass:   version.StorageClass,
@@ -600,6 +602,7 @@ func (c *Client) listObjectsQuery(ctx context.Context, bucketName, objectPrefix,
 		if err != nil {
 			return listBucketResult, err
 		}
+		listBucketResult.Contents[i].LastModified = listBucketResult.Contents[i].LastModified.Truncate(time.Millisecond)
 	}
 
 	for i, obj := range listBucketResult.CommonPrefixes {


jhwz:
Quick note (I'm the OP of #1632) I found in my tests that to keep that same behaviour I needed to use:

```go
LastModified.Add(-time.Millisecond * 500).Round(time.Second)
```
instead of 
```go
LastModified.Round(time.Second)
```
to floor the time, not round it. Not sure if that will solve the tests not completing

harshavardhana:
No it shouldn't be Round instead it should be Truncate() @jhwz

jhwz:
> No it shouldn't be Round instead it should be Truncate() @jhwz

Ah right, that works too

vadmeste:
@harshavardhana can't we make Stat() to show more precise time instead of truncating to second ? 

I don't think it will break anything since is this a go library and time.Parse(time.RFC3339, "2022-04-16T13:37:01.511096Z") works just fine (time.RFC3339 without Nano)

I think it is just better to keep time precision here rather than losing it (not sure what is the effect for bucket replication here as well)

harshavardhana:
> @harshavardhana can't we make Stat() to show more precise time instead of truncating to second ?
> 
> I don't think it will break anything since is this a go library and time.Parse(time.RFC3339, "2022-04-16T13:37:01.511096Z") works just fine (time.RFC3339 without Nano)
> 
> I think it is just better to keep time precision here rather than losing it (not sure what is the effect for bucket replication here as well)

The problem is there is no time precision allowed in AWS S3. It's only up to second precision @vadmeste 

That's why millisecond values don't make sense.

harshavardhana:
Tested this with AWS S3 @vadmeste the ListObjects doesn't have the millisecond resolution

```
AWS_REGION=us-east-1 ./example 
2022/04/17 13:34:42 first page results:
Name:           hosts
ListObjectsV2 -> Last modified:  2022-04-17 20:33:17 +0000 UTC
Size:           303
Storage class:  STANDARD

HeadObject() Last Modified:  2022-04-17 20:33:17 +0000 UTC
```

```go
package main

import (
	"context"
	"fmt"
	"log"

	"github.com/aws/aws-sdk-go-v2/aws"
	"github.com/aws/aws-sdk-go-v2/config"
	"github.com/aws/aws-sdk-go-v2/service/s3"
)

func main() {
	// Load the Shared AWS Configuration (~/.aws/config)
	// cfg, err := config.LoadDefaultConfig(context.TODO(), config.WithSharedConfigProfile("minio"), config.WithEndpointResolverWithOptions(customResolver))
	cfg, err := config.LoadDefaultConfig(context.TODO(), config.WithSharedConfigProfile("s3"))
	if err != nil {
		log.Fatal(err)
	}
	
	// Create an Amazon S3 service client
	client := s3.NewFromConfig(cfg, func(opts *s3.Options) {
		opts.UsePathStyle = true
	})

	// Get the first page of results for ListObjectsV2 for a bucket
	output, err := client.ListObjectsV2(context.TODO(), &s3.ListObjectsV2Input{
		Bucket: aws.String("vadmeste"),
		Prefix: aws.String("hosts"),
	})
	if err != nil {
		log.Fatal(err)
	}

	log.Println("first page results:")
	for _, item := range output.Contents {
		fmt.Println("Name:          ", *item.Key)
		fmt.Println("ListObjectsV2 -> Last modified: ", *item.LastModified)
		fmt.Println("Size:          ", item.Size)
		fmt.Println("Storage class: ", item.StorageClass)
		fmt.Println("")
	}

	houtput, err := client.HeadObject(context.TODO(), &s3.HeadObjectInput{
		Bucket: aws.String("vadmeste"),
		Key:    aws.String("hosts"),
	})

	fmt.Println("HeadObject() Last Modified: ", *houtput.LastModified)
}
```

So this PR is valid and we should also perhaps remove millisecond resolution from the MinIO server. 

// cc @kannappanr @klauspost

vadmeste (approved):
LGTM

PR #1640, opened 2022-04-16
Add PresignOptions

dvaldivia:
This PR adds `PresignOptions` to `PresignedGetObject`, `PresignedHeadObject`, `PresignHeader` and `presignURL`

Any other function using `presignURL` will pass an empty `PresignOptions`

`PresignOptions` mimics the possible configurations the same way `GetObject` exposes via `GetObjectOptions`

The main need for this is to allow to configure the Region (or Bucket Location) before the signed request is created which avoids doing a round trip to whatever the endpoint is to figure out the bucket location.

This allows to sign URL to whichever endpoint the MinIO Client is configured to.

Signed-off-by: Daniel Valdivia <hola@danielvaldivia.com>

### DIFF
--- a/api-presigned.go
+++ b/api-presigned.go
@@ -30,7 +30,7 @@ import (
 
 // presignURL - Returns a presigned URL for an input 'method'.
 // Expires maximum is 7days - ie. 604800 and minimum is 1.
-func (c *Client) presignURL(ctx context.Context, method string, bucketName string, objectName string, expires time.Duration, reqParams url.Values, extraHeaders http.Header) (u *url.URL, err error) {
+func (c *Client) presignURL(ctx context.Context, method string, bucketName string, objectName string, expires time.Duration, opts PreSignOptions) (u *url.URL, err error) {
 	// Input validation.
 	if method == "" {
 		return nil, errInvalidArgument("method cannot be empty.")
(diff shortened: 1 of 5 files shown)


harshavardhana (changes requested):
This is not going to work as this a breaking change @dvaldivia whatever region you want the presign URL you should simply have the right MiniO client instance configured with the right region.

Passing custom regions is wrong and non idiomatic when you have a constructor region configured already.

dvaldivia:
The problem I'm trying to address is signing URLs for a domain that the client cannot resolve to, and I can only achieve that by avoiding the `get bucket location call` that is executed with every new request, if the region is passed manually the logic will simply not execute, and if the wrong region is passed, for S3 endpoints it will fail, but for MinIO endpoints things will work just fine.

The particular use case is a MinIO inside kubernetes that needs to sign a URL for it's public domain, which either cannot be resolved or resolves on a different port from what `MINIO_DOMAIN` tells the server. MinIO is happy to resolve properly signed URLs hitting it through any request domain as long as the signature matches, which is what I'm trying to leverage here @harshavardhana

harshavardhana:
> The problem I'm trying to address is signing URLs for a domain that the client cannot resolve to, and I can only achieve that by avoiding the `get bucket location call` that is executed with every new request, if the region is passed manually the logic will simply not execute, and if the wrong region is passed, for S3 endpoints it will fail, but for MinIO endpoints things will work just fine.
> 
> The particular use case is a MinIO inside kubernetes that needs to sign a URL for it's public domain, which either cannot be resolved or resolves on a different port from what `MINIO_DOMAIN` tells the server. MinIO is happy to resolve properly signed URLs hitting it through any request domain as long as the signature matches, which is what I'm trying to leverage here @harshavardhana

You need to avoid get bucket location call by making sure to have the right MiniO client instance for the domain and also the region pre configured.

Region is a constructor level parameter it shouldn't be willy nilly changed in this manner per API call in the SDK.

If you need region specific presigned URLs all you need to region specific MinIO client.  Get bucket location is never called if MinIO client is configured with the expected region.

dvaldivia:
it's not the case in the `presignURL()` call hierarchy, the `requestMetadata.bucketLocation` is never populated and causes the `get bucket location`

https://github.com/minio/minio-go/blob/master/api.go#L696

unless we add the check for region on this block, which would be easier, do you prefer that approach?

PR #1644, opened 2022-04-26
Add support for a special header `x-minio-purge-on-delete`

Praveenrajmani:
this header can be passed to purge the objects on deletion


api-remove.go
```diff
@@ -124,6 +124,7 @@ type RemoveObjectOptions struct {
 	GovernanceBypass bool
 	VersionID        string
 	Internal         AdvancedRemoveOptions
+	PurgeOnDelete    bool
```
klauspost, line 127: Document that this is a minio specific extension.

harshavardhana:
Corresponding server PR is closed.

PR #1645, opened 2022-04-27
Add TestMultipartUpload

zkep:
Add TestMultipartUpload

### DIFF
--- a/core_test.go
+++ b/core_test.go
@@ -20,6 +20,7 @@ package minio
 import (
 	"bytes"
 	"context"
+	"io"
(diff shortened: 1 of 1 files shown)


harshavardhana (changes requested):
what are trying to add here that is not covered by existing tests? @zkep

zkep:
> what are trying to add here that is not covered by existing tests? @zkep

listObjectParts，PutObjectPart   is not covered

harshavardhana (approved):

PR #1649, opened 2022-05-04
Remove `Range` header after `Seek` to origin

flobernd:
When reading data using `ReadAt()` and seeking to the origin using `Seek(0, 0)` afterwards, the old `Range` header is still included in the request. This causes any subsequent `Read()` operation to start at the offset previously passed to `ReadAt()`.

The commit in this PR fixes that behaviour.

### DIFF
--- a/api-get-object.go
+++ b/api-get-object.go
@@ -199,6 +199,9 @@ func (c *Client) GetObject(ctx context.Context, bucketName, objectName string, o
 							opts.SetRange(req.Offset, req.Offset+int64(len(req.Buffer))-1)
 						} else if req.Offset > 0 { // Range is set with respect to the offset.
 							opts.SetRange(req.Offset, 0)
+						} else {
+							// Remove range header if already set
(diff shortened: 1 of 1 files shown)


klauspost:
@flobernd Thanks for the contribution. This seems to trigger some tests failures.

```
=== RUN   TestSetHeader
Error:     get-options_test.go:55: Test 5: Expected range header 'bytes=0-0', but got ''
```
This seems functionally different. `bytes=0-0` would return 1 byte. So this should be investigated, if the test is wrong or the change.

```
=== RUN   TestGetObjectCore
Error:     core_test.go:206: Error: GetObject read shorter bytes before reaching EOF, want 946344, got 1
```

This also seems something that should be double checked.

flobernd:
@klauspost Let's try a different way :-)

klauspost (approved):
LGTM

harshavardhana (approved):

PR #1654, opened 2022-05-25
Add funcs to request credentials using AssumeRoleWithCustomToken

donatello:


pkg/credentials/sts_custom_identity.go
```diff
@@ -0,0 +1,146 @@
...
+// CustomTokenResult - Contains temporary creds and user metadata.
+type CustomTokenResult struct {
+	Credentials struct {
+		AccessKey    string    `xml:"AccessKeyId" json:"accessKey,omitempty"`
+		SecretKey    string    `xml:"SecretAccessKey" json:"secretKey,omitempty"`
+		Expiration   time.Time `xml:"Expiration" json:"expiration,omitempty"`
+		SessionToken string    `xml:"SessionToken" json:"sessionToken,omitempty"`
+	} `xml:",omitempty"`
+
+	AssumedUser string `xml:",omitempty"`
```
krisis, line 38: Shouldn't the json field name for `AssumedUser` be in pascal case too?
```suggestion
	AssumedUser string `xml:",omitempty" json:"assumedUser,omitempty"`
```

donatello: Actually I'm not sure why we need a json instance here - I am removing it. Even the other related STS types don't really need it.

krisis (approved):

harshavardhana (approved):

PR #1656, opened 2022-05-31
Add ClusterHealth API to check the cluster health

Praveenrajmani:
This will hit `/minio/health/cluster` anonymous API to check the cluster health

### DIFF
--- a/api-health.go
+++ b/api-health.go
@@ -1,0 +1,69 @@
+/*
+ * MinIO Go Library for Amazon S3 Compatible Cloud Storage
+ * Copyright 2015-2022 MinIO, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package minio
+
+import (
+	"context"
+	"net/http"
+	"strconv"
+)
+
+// ClusterHealthResult represents the cluster health result
+type ClusterHealthResult struct {
+	Healthy         bool
+	MaintenanceMode bool
+	WriteQuorum     int
+	HealingDrives   int
+}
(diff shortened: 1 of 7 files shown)


harshavardhana (changes requested):
why don't we enhance the HealthCheck API that runs on a continuous basis?

```
// HealthCheck starts a healthcheck to see if endpoint is up. Returns a context cancellation function
// and and error if health check is already started
func (c *Client) HealthCheck(hcDuration time.Duration) (context.CancelFunc, error) {
        if atomic.LoadInt32(&c.healthStatus) == online {
                return nil, fmt.Errorf("health check is running")
        }
        if hcDuration < 1*time.Second {
                return nil, fmt.Errorf("health check duration should be atleast 1 second")
        }
        ctx, cancelFn := context.WithCancel(context.Background())
        atomic.StoreInt32(&c.healthStatus, online)
        probeBucketName := randString(60, rand.NewSource(time.Now().UnixNano()), "probe-health-")
        go func(duration time.Duration) {
                timer := time.NewTimer(duration)
                defer timer.Stop()
                for {
                        select {
                        case <-ctx.Done():
                                atomic.StoreInt32(&c.healthStatus, unknown)
                                return
                        case <-timer.C:
                                timer.Reset(duration)
                                // Do health check the first time and ONLY if the connection is marked offline
                                if c.IsOffline() {
                                        gctx, gcancel := context.WithTimeout(context.Background(), 3*time.Second)
                                        _, err := c.getBucketLocation(gctx, probeBucketName)
                                        gcancel()
                                        if IsNetworkOrHostDown(err, false) {
                                                // Still network errors do not need to do anything.
                                                continue
                                        }
                                        switch ToErrorResponse(err).Code {
                                        case "NoSuchBucket", "AccessDenied", "":
                                                atomic.CompareAndSwapInt32(&c.healthStatus, offline, online)
                                        }
                                }
                        }
                }
        }(hcDuration)
        return cancelFn, nil
}
```

and also `mc ready` is it one time activity or would it basically run for some time and exit? 

Because if we need `mc ready` to do only one time we can just do it using `curl`

klauspost:
@Praveenrajmani @harshavardhana Why are we adding minio specific calls to `minio-go` and not `madmin-go`?

harshavardhana:
> @Praveenrajmani @harshavardhana Why are we adding minio specific calls to `minio-go` and not `madmin-go`?

Actually we can add it to madmin-go that might be the right place for this.

Praveenrajmani:
As discussed, we can re-use the existing `HealthCheck` implementation. Closing this PR.

PR #1657, opened 2022-05-31
Add Policy to STSAssumeRoleOptions

rgcostea:
Allow the policy to be specified when invoking the STS AssumeRole API.

Note that the STS AssumeRole implementation in minio already supports assigning a policy.  This change simply provide a way in which to invoke it.  Also note this has been tested locally and works as expected with different types of policies.


rgcostea:
@harshavardhana , anything else I need to do here?

harshavardhana (approved):

donatello (approved):

PR #1658, opened 2022-06-04
replace github.com/mitchellh/go-homedir by os

Humm42:
os.UserHomeDir provides a way to get the user's home directory for
different platforms. Using it drops a dependency.

This doesn't tidy up examples/*/go.sum, as they aren't up-to-date
anyway. That should be done separately.


pkg/credentials/file_aws_credentials.go
```diff
@@ -62,7 +61,7 @@ func (p *FileAWSCredentials) Retrieve() (Value, error) {
 	if p.Filename == "" {
 		p.Filename = os.Getenv("AWS_SHARED_CREDENTIALS_FILE")
 		if p.Filename == "" {
-			homeDir, err := homedir.Dir()
+			homeDir, err := os.UserHomeDir()
```
harshavardhana, line 64: homedir.Dir() solves more than os.UserHomeDir() - what about other situations when HOME env is not set? @Humm42

harshavardhana (changes requested):

pkg/credentials/file_aws_credentials.go
Humm42: Then, there was a choice not to set `$HOME`, which should be respected. A fine fallback would be to treat `/` as home directory. We shouldn’t try to find some home directory if the one place it actually must be specified doesn’t. The whole using different fallbacks for finding some home directory is fragile and takes away meaning of `$HOME` given to it by historic practice and by POSIX.

harshavardhana: That is the whole point of that package, it exists for a reason and why we use it.

Your PR is simply taking away something that was already given to our users.

Humm42: The dependency was added in 5d7ee332f62e (2017-05-12). `os.UserHomeDir` was introduced in https://github.com/golang/go/commit/fa1a49aa556d87d41315329811de328e19022608 (2018-10-03). That is reason to believe that the dependency was chosen not because of its behavior of having those fallbacks, but because there had not been anything akin to `os.UserHomeDir`.

The existing behavior is broken. I doubt there are many, if any, people relying on that. I see the removal of a dependency, and additionally fixing the behavior, as outweighing that. That is giving something to the users: One dependency less.

Humm42: I mixed up minio/go-homedir and mitchellh/go-homedir. The dependency on the latter was 06dcf064d9e3b1dfd21b9a01a822da7665a9c971. The former’s behavior I can’t check, as it is neither on GitHub nor on proxy.golang.org. That shouldn’t matter, though.

harshavardhana: I can't even understand what you are talking about here 

> The existing behavior is broken. 

what is broken? - I don't see any bug fix here.  

The problem I have with the PR is that there is no good justification on why we should use os.UserHomeDir is rather a poor implementation of how HomeDir's are available in the Unix world.

If you look at the implementation here from mitchell/go-homedir - https://github.com/mitchellh/go-homedir/blob/main/homedir.go#L148 - it is extensive and covers quite a few areas which os.UserHomeDir lacks

Unless there is a specific problem that you faced, I don't see why we should be using os.UserHomeDir

Humm42: In Unix world, you can get the home directory by:

- _getpwuid_, which typically reads from a local `/etc/passwd`, but might also be able to use stuff such as NSS for different mechanisms. This is an option for implementation of _login_. In some platform-dependent way, the sysadmin makes it work. This is potentially costly—might use the network, might do whatever to retrieve information from wherever.
- `$HOME`, which is set by _login_. This is what POSIX calls “home directory” and what programs use to get the home directory. It’s what can be changed to make programs use a different directory and what can be unset to make programs error out or try to use `/`.

That’s historic practice. That’s how things work. Trying to use `getpwuid` would be bad enough, but depending on some command being in `PATH` is worse: You depend on some little executable. You might not notice clearning `HOME` somewhere, someone removes the apparently useless _getent_, the shell expects `HOME` to be set and suddenly your current working directory is being treated as your home directory. That’s bad. This stuff must be reliable and easily understood.

Humm42: In my opinion, mitchellh/go-homedir is bad. It should not do what it does do. If so, it should be fixed and you should use the standard library instead, which would then do the same thing. If not, the standard library should be fixed and you should use the standard library, which would then do the same thing. If both behaviors are fine for different usecases, you should use the standard library, as that implements the behavior correct for this case.

harshavardhana (approved):

PR #1663, opened 2022-06-10
Support detecting region from FIPS URLs

### LINKED ISSUE #1662: Uploading files using FIPS endpoints is broken
The following code, inspired from https://docs.min.io/docs/golang-client-quickstart-guide.html will fail because it is using a FIPS endpoint and the GetRegionFromUrl method doesn't handle FIPS endpoints.

```go
package main

import (
	"context"
	"log"
	"os"

)

func main() {
	ctx := context.Background()

	endpoint := "s3-fips.us-east-1.amazonaws.com"
	useSSL := true

	accessKeyID := os.Getenv("AWS_ACCESS_KEY_ID")
	secretAccessKey := os.Getenv("AWS_SECRET_ACCESS_KEY")
	bucketName := os.Getenv("AWS_BUCKET_NAME")

	// Initialize minio client object.
	minioClient, err := minio.New(endpoint, &minio.Options{
		Creds:  credentials.NewStaticV4(accessKeyID, secretAccessKey, ""),
		Secure: useSSL,
	})
	if err != nil {
		log.Fatalln(err)
	}

	objectName := "test_file.zip"
	filePath := "/tmp/test_file.zip"
	contentType := "application/zip"

	// Upload the zip file with FPutObject
	info, err := minioClient.FPutObject(ctx, bucketName, objectName, filePath, minio.PutObjectOptions{ContentType: contentType})
	if err != nil {
		log.Fatalln(err)
	}

	log.Printf("Successfully uploaded %s of size %d\n", objectName, info.Size)
}
```

```shell
$ go run FileUploader.go
2022/06/10 17:02:30 The authorization header is malformed; the region 'fips.us-east-1' is wrong; expecting 'us-east-1'
exit status 1
```

balasankarc:
Attempt to fix #1662

### DIFF
--- a/pkg/s3utils/utils.go
+++ b/pkg/s3utils/utils.go
@@ -89,6 +89,12 @@ var amazonS3HostHyphen = regexp.MustCompile(`^s3-(.*?).amazonaws.com$`)
 // amazonS3HostDualStack - regular expression used to determine if an arg is s3 host dualstack.
 var amazonS3HostDualStack = regexp.MustCompile(`^s3.dualstack.(.*?).amazonaws.com$`)
 
+// amazonS3HostFIPS - regular expression used to determine if an arg is s3 FIPS host.
+var amazonS3HostFIPS = regexp.MustCompile(`^s3-fips.(.*?).amazonaws.com$`)
+
+// amazonS3HostFIPSDualStack - regular expression used to determine if an arg is s3 FIPS host dualstack.
+var amazonS3HostFIPSDualStack = regexp.MustCompile(`^s3-fips.dualstack.(.*?).amazonaws.com$`)
+
 // amazonS3HostDot - regular expression used to determine if an arg is s3 host in . style.
 var amazonS3HostDot = regexp.MustCompile(`^s3.(.*?).amazonaws.com$`)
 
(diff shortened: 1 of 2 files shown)


pkg/s3utils/utils.go
```diff
@@ -126,6 +132,18 @@ func GetRegionFromURL(endpointURL url.URL) string {
 	if len(parts) > 1 {
 		return parts[1]
 	}
+	if IsAmazonFIPSUSEastWestEndpoint(endpointURL) {
```
balasankarc, line 135: Unlike FIPS endpoiints in Govcloud, the normal FIPS endpoints are available in two locations - `us-east-1` and `us-west-1`. So, we need to extract out which one it is and return a value accordingly (as opposed to Gov cloud where we could just hardcode the `"us-gov-west-1"` location)

balasankarc:
@harshavardhana Could you please take a look? Thanks. (Apologies if pinging you directly wasn't the right thing to do. I checked the Contributor's guide and couldn't find what I should be doing to request reviews)

harshavardhana (approved):

balasankarc:
@harshavardhana Thanks a lot for the quick action on this one. Much appreciated.

BTW, is a new patch release planned anytime soon? Asking because this issue essentially prevents users from using FIPS endpoints for the GitLab Runner cache - https://gitlab.com/gitlab-org/gitlab-runner/-/issues/29112

PR #1665, opened 2022-06-16
Add API function to return unreplicated objects info

poornas:
for `mc replicate diff`

### DIFF
--- a/api-bucket-replication.go
+++ b/api-bucket-replication.go
@@ -289,3 +289,54 @@ func (c *Client) GetBucketReplicationResyncStatus(ctx context.Context, bucketNam
 	}
 	return rinfo, nil
 }
+
+// BucketReplicationDiff - gets diff for non-replicated entries.
+func (c *Client) BucketReplicationDiff(ctx context.Context, bucketName string, opts replication.DiffOpts) <-chan replication.DiffInfo {
+	diffCh := make(chan replication.DiffInfo)
+
+	// start a routine to start reading line by line.
+	go func(diffCh chan<- replication.DiffInfo) {
+		defer close(diffCh)
+		urlValues := make(url.Values)
+		urlValues.Set("replication-diff", "")
+		if opts.Verbose {
+			urlValues.Set("verbose", "true")
+		}
+		if opts.ARN != "" {
+			urlValues.Set("arn", opts.ARN)
+		}
+		if opts.Prefix != "" {
+			urlValues.Set("prefix", opts.Prefix)
+		}
+		// Execute GET on bucket to get replication config.
+		resp, err := c.executeMethod(ctx, http.MethodPost, requestMetadata{
+			bucketName:  bucketName,
+			queryValues: urlValues,
+		})
+		if err != nil {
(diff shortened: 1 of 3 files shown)


klauspost:
Is this an official API? If not, we should add a notice about this in the docs.

poornas:
> Is this an official S3 API? If not, we should add a notice about this in the docs.

This is a MinIO only extension - added a reference to this in the docs.

api-bucket-replication.go
```diff
@@ -289,3 +289,53 @@ func (c *Client) GetBucketReplicationResyncStatus(ctx context.Context, bucketNam
 	}
 	return rinfo, nil
 }
+
+// BucketReplicationDiff - gets diff for unreplicated entries.
```
harshavardhana, line 293:
```suggestion
// BucketReplicationDiff - gets diff for non-replicated entries.
```

api-bucket-replication.go
```diff
@@ -289,3 +289,53 @@ func (c *Client) GetBucketReplicationResyncStatus(ctx context.Context, bucketNam
...
+func (c *Client) BucketReplicationDiff(ctx context.Context, bucketName string, opts replication.DiffOpts) <-chan replication.DiffInfo {
+	diffCh := make(chan replication.DiffInfo)
+
+	// start a routine to start reading line by line.
+	go func(diffCh chan<- replication.DiffInfo) {
+		defer close(diffCh)
+		urlValues := make(url.Values)
+		urlValues.Set("replication-diff", "")
```
harshavardhana, line 301:
```suggestion
		urlValues.Set("replication-diff", "")
```

Since the API is for replication for a more terse API this should be just `diff` @poornas

harshavardhana: Just like arn is not called `replication-arn`

api-bucket-replication.go
```diff
@@ -289,3 +289,53 @@ func (c *Client) GetBucketReplicationResyncStatus(ctx context.Context, bucketNam
...
+			bucketName:  bucketName,
+			queryValues: urlValues,
+		})
+		if err != nil {
+			diffCh <- replication.DiffInfo{Err: err}
+			return
+		}
+
+		if resp.StatusCode != http.StatusOK {
+			closeResponse(resp)
```
harshavardhana, line 319: you can do `defer closeResponse(resp)` instead of calling it everywhere, since we shall close response body only upon return.

harshavardhana (changes requested):

api-bucket-replication.go
poornas: api route is like `minio:9001/bucket/?replication-diff=` using just "diff" would be too terse

api-bucket-replication.go
poornas: done

api-bucket-replication.go
harshavardhana: yeah but why it would be just "replication-diff" ? isn't it just `diff` anyways?

poornas:
closing this PR - moving this API to madmin-go

PR #1666, opened 2022-06-17
use context to wait for cancelation on GetObject()

harshavardhana:
it is potentially possible that the caller has canceled
the context, however the GetObject() go-routine will
be stuck forever since no one is reading anymore.


klauspost (approved):
LGTM

vadmeste (approved):
LGTM

PR #1669, opened 2022-06-27
always listen on context canceled waiting on channels

harshavardhana:


api-put-object-streaming.go
```diff
@@ -156,8 +156,14 @@ func (c *Client) putObjectMultipartStreamFromReadAt(ctx context.Context, bucketN
...
 	for w := 1; w <= opts.getNumThreads(); w++ {
 		go func(w int, partSize int64) {
-			// Each worker will draw from the part channel and upload in parallel.
-			for uploadReq := range uploadPartsCh {
+			for {
+				var uploadReq uploadPartReq
+				select {
+				case <-ctx.Done():
+					return
+				case uploadReq = <-uploadPartsCh:
```
vadmeste, line 164: I guess you need to return if uploadPartsCh is closed

vadmeste (approved):
LGTM

PR #1671, opened 2022-06-30
Add transfer progress timeout

klauspost:
Add default 5 minute timeout for transfer progress for sending and receiving bodies.

Proposed fix for https://github.com/minio/mc/issues/4133

### DIFF
--- a/transport.go
+++ b/transport.go
@@ -100,9 +100,13 @@ type transportTimeout struct {
 }
 
 func (t *transportTimeout) RoundTrip(request *http.Request) (*http.Response, error) {
-	request, ctx, cancel := newReqAliveChecker(request, t.timeout)
+	request, ctx, cancel, status := newReqAliveChecker(request, t.timeout)
 	resp, err := t.parent.RoundTrip(request)
 	if err != nil {
+		if status != nil && atomic.LoadUint32(status) == 2 {
+			// If we canceled, change the error.
+			err = context.DeadlineExceeded
+		}
 		cancel()
 		return resp, err
 	}
@@ -113,19 +117,20 @@ func (t *transportTimeout) RoundTrip(request *http.Request) (*http.Response, err
 // This will measure time between body reads.
 // When the timeout is exceeded the request will be canceled.
 // A new request with a new context is returned.
-func newReqAliveChecker(req *http.Request, timeout time.Duration) (request *http.Request, ctx context.Context, cancel context.CancelFunc) {
+func newReqAliveChecker(req *http.Request, timeout time.Duration) (request *http.Request, ctx context.Context, cancel context.CancelFunc, status *uint32) {
 	ctx, cancel = context.WithCancel(req.Context())
 	req = req.WithContext(ctx)
 
 	if req.Body == nil || timeout <= 0 {
-		return req, ctx, cancel
+		return req, ctx, cancel, nil
 	}
-	req.Body = &reqAliveChecker{timeout: timeout, rc: req.Body, ctx: ctx, cancel: cancel}
-	return req, ctx, cancel
+	ac := reqAliveChecker{timeout: timeout, rc: req.Body, ctx: ctx, cancel: cancel}
+	req.Body = &ac
+	return req, ctx, cancel, &ac.timedOut
 }
 
 type reqAliveChecker struct {
-	timedOut uint32
+	timedOut uint32 // 0 = ok, 1 = upstream cancel, 2 = timeout
 	timeout  time.Duration
 	rc       io.ReadCloser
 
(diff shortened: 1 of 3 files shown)


transport.go
```diff
@@ -82,3 +82,185 @@ var DefaultTransport = func(secure bool) (*http.Transport, error) {
...
+		return 0, context.DeadlineExceeded
+	}
+	n, err = a.rc.Read(p)
+	// Start on first request
+	if err == nil {
+		if a.reset == nil {
+			update := make(chan struct{}, 1)
+			a.reset = update
+			go func() {
+				t := time.NewTimer(a.timeout)
```
harshavardhana, line 146:
```suggestion
				t := time.NewTimer(a.timeout)
				defer t.Stop()
```

Missing Stop() upon return?

transport.go
```diff
@@ -82,3 +82,185 @@ var DefaultTransport = func(secure bool) (*http.Transport, error) {
...
+						}
+						// Reset on update...
+						t.Reset(a.timeout)
+					case <-a.ctx.Done():
+						atomic.StoreUint32(&a.timedOut, 1)
+						a.cancel()
+					case <-t.C:
+						atomic.StoreUint32(&a.timedOut, 1)
+						a.cancel()
+					}
```
harshavardhana, line 162: shouldn't a t.Reset() be done after t.C has fired?

harshavardhana (changes requested):
This would require a good amount of testing - perhaps almost all Go programs after being affected by this kind of behavior. 

Not sure if everyone should be writing this.

transport.go
klauspost: No, we have cancelled. If would like to return, but that wouldn't be safe right now, as we could be blocking on `a.reset <-`.

Hmmm.. maybe if we make `a.reset <- struct{}{}` non-blocking (which would be fine anyway).

klauspost: PTAL

klauspost:
> This would require a good amount of testing - perhaps almost all Go programs after being affected by this kind of behavior.

Yeah, seen it plenty of times. Most network failures will cause the broken connections to be detected, but if you are reading and all packets to the destination is lost you end up like this.

harshavardhana:
Looking at this PR once again why can't we do `net.Conn` based Deadline Reader/Writer @klauspost ? because we do get access to net.Conn if we use a custom Dialer

net.Dialer.Control - provides this exact functionality and may be even simpler to do?  WDYT?

klauspost:
@harshavardhana I don't like platform specific hacks, and AFAICT it is already set by the dialer, but not 100% effective.

harshavardhana:
> @harshavardhana I don't like platform specific hacks, and AFAICT it is already set by the dialer, but not 100% effective.

what I meant is quite simple 

```
package deadlineconn

// DeadlineConn - is a generic stream-oriented network connection supporting buffered reader and read/write timeout.
type DeadlineConn struct {
	net.Conn
	readTimeout            time.Duration // sets the read timeout in the connection.
	writeTimeout           time.Duration // sets the write timeout in the connection.
}

// Sets read timeout
func (c *DeadlineConn) setReadTimeout() {
	if c.readTimeout != 0 {
		c.SetReadDeadline(time.Now().UTC().Add(c.readTimeout))
	}
}

func (c *DeadlineConn) setWriteTimeout() {
	if c.writeTimeout != 0 {
		c.SetWriteDeadline(time.Now().UTC().Add(c.writeTimeout))
	}
}

// Read - reads data from the connection using wrapped buffered reader.
func (c *DeadlineConn) Read(b []byte) (n int, err error) {
	c.setReadTimeout()
	n, err = c.Conn.Read(b)
	return n, err
}

// Write - writes data to the connection.
func (c *DeadlineConn) Write(b []byte) (n int, err error) {
	c.setWriteTimeout()
	n, err = c.Conn.Write(b)
	return n, err
}

// NewDeadlineConn - creates a new connection object wrapping net.Conn with deadlines.
func NewDeadlineConn(c net.Conn, readTimeout, writeTimeout time.Duration) *DeadlineConn {
	return &DeadlineConn{
		Conn:                   c,
		readTimeout:            readTimeout,
		writeTimeout:           writeTimeout,
	}
}
```

And you get to keep this as a custom dialer like

```
// NewCustomDialContext setups a custom dialer for any external communication and proxies.
func NewCustomDialContext(dialTimeout time.Duration) DialContext {
        return func(ctx context.Context, network, addr string) (net.Conn, error) {
                dialer := &net.Dialer{
                        Timeout: dialTimeout,
                }
                conn, err := dialer.DialContext(ctx, network, addr)
                if err != nil {
                        return nil, err 
                }
                return  NewDeadlineConn(10*time.Second, 10*time.Second), nil
        }
}
```

harshavardhana:
I just this is way simpler than dealing with the http.Transport.RoundTrip which looks like pretty complicated the way it needs to be handled.

klauspost:
Not sure how reliable it is. But I think we should try the simplest solution first.

PR #1682, opened 2022-07-30
Do not ignore error when Instance Metadata service doesn't exist

wlan0:
When running local tests using minio-go, where IMDS doesn't exist, ignoring this error leads to test failures.

### DIFF
--- a/pkg/credentials/iam_aws.go
+++ b/pkg/credentials/iam_aws.go
@@ -289,7 +289,10 @@ func getCredentials(client *http.Client, endpoint string) (ec2RoleCredRespBody, 
 	}
 
 	// https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html
-	token, _ := fetchIMDSToken(client, endpoint)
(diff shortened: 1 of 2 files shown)


wlan0:
Here is the description of the error: https://github.com/kopia/kopia/pull/2213#issuecomment-1197296465

harshavardhana (approved):

kannappanr (approved):

vadmeste:
@wlan0 tests are failing, most likley tests are wrong, I don't believe your PR is causing any real regression

wlan0:
@vadmeste I've fixed the tests

wlan0:
https://github.com/minio/minio-go/pull/1682/files#diff-ff6260224bd798cd39fd6253c421cc34a9075249e82bc8b1c2fa82b720dc195fL92 

This server isn't really useful because there is a required step before this server is called, which uses IMDSv2. This server will always return `400 Bad Request` for all IMDSv2 requests. 

I'm going to update the PR to remove the `initTestServer` as a separate commit. I can revert if a different solution is desired

PR #1684, opened 2022-08-01
fix: remote target is permanently marked offline by mistake

garenchan:
### Issue

The steps to reproduce the issue are as follows:
1. Stop the remote target.
2. Upload a file to local target and we will see in the log that `remote target is offline`.
3. Wait a while to ensure that a health check is performed.
4. Start the remote target and verify that it is accessible.
5. No matter how long we wait, the local target always considers the remote target to be offline.

### Reason

When the remote target or network is down, the health check timer will not be reset. So even if the remote target or network recovers later, it will be permanently marked offline by mistake.

### DIFF
--- a/api.go
+++ b/api.go
@@ -384,13 +384,11 @@ func (c *Client) HealthCheck(hcDuration time.Duration) (context.CancelFunc, erro
 					gctx, gcancel := context.WithTimeout(context.Background(), 3*time.Second)
 					_, err := c.getBucketLocation(gctx, probeBucketName)
 					gcancel()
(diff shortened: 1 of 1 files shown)


vadmeste (approved):
LGTM! thanks @garenchan

garenchan:
@vadmeste, why don't we run unit tests to find these issues?

harshavardhana:
> @vadmeste, why don't we run unit tests to find these issues?

@garenchan feel free to write one.

harshavardhana (approved):

PR #1686, opened 2022-08-02
allow marking offline minio.Client externally

harshavardhana:


harshavardhana:
We do not need this for now.

PR #1688, opened 2022-08-26
convert LDAP/JWT Identity from query to form body

harshavardhana:
HTTP calls get logged through proxies, avoid
sensitive data getting logged by using request
body instead of query params.


klauspost (approved):
LGTM

a1466d44-d3dc-4c0b-90c7-315b088731d7:
Quick question from the uninformed: Is this automagicaly going to work on the minio service side as well, because I cant find a matching change there?

harshavardhana:
> Quick question from the uninformed: Is this automagicaly going to work on the minio service side as well, because I cant find a matching change there?

Yes it will.

PR #1699, opened 2022-09-24
fix: use seeker to avoid memory without multipart upload

harshavardhana:
refer minio/minio#15754

### DIFF
--- a/api-put-object-streaming.go
+++ b/api-put-object-streaming.go
@@ -419,6 +419,7 @@ func (c *Client) putObject(ctx context.Context, bucketName, objectName string, r
 		return UploadInfo{}, errInvalidArgument("MD5Sum cannot be calculated with size '-1'")
 	}
 
+	var readSeeker io.Seeker
 	if size > 0 {
 		if isReadAt(reader) && !isObject(reader) {
 			seeker, ok := reader.(io.Seeker)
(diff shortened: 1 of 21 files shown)


api-put-object-streaming.go
```diff
@@ -434,19 +434,29 @@ func (c *Client) putObject(ctx context.Context, bucketName, objectName string, r
 
 	var md5Base64 string
 	if opts.SendContentMd5 {
-		// Create a buffer.
-		buf := make([]byte, size)
+		// Calculate md5sum.
+		hash := c.md5Hasher()
+
+		seeker, ok := reader.(io.Seeker)
```
klauspost, line 440: This is a bit complicated to make safe (and I see the code above also makes that mistake).

But you should do a `Seek(0, io.SeekCurrent)` to get the current offset. You might be given a ReadSeeker that has been read from, meaning we should not return to absolute 0 when resetting.

klauspost (changes requested):

api-put-object-streaming.go
harshavardhana: That was not the intention, the intention was simply to honor io.NewSectionReader's offset input.  Made the code more safe.

api-put-object-streaming.go
```diff
@@ -419,6 +419,7 @@ func (c *Client) putObject(ctx context.Context, bucketName, objectName string, r
 		return UploadInfo{}, errInvalidArgument("MD5Sum cannot be calculated with size '-1'")
 	}
 
+	var readSeeker io.Seeker
```
klauspost, line 422: This gets set when `reader` is `io.ReaderAt` *and* `io.Seeker`.

It seems like it should also be possible with it just being `io.Seeker` (but still excluding stdin/out).

Maybe we can add that in the future.

klauspost (approved):

api-put-object-streaming.go
harshavardhana: No it will only get set for the `io.SectionReader`

PR #1701, opened 2022-09-28
fail properly while parsing duplicate tags

harshavardhana:
AWS S3 API expects duplicate keys to be rejected,
due to our usage of url.ParseQuery() this was skipped 
since URL query parser simply appends duplicate keys 
as key with multiple values, this is not desirable.

Add more input validation and restrictions as per
AWS S3 API.

### DIFF
--- a/pkg/tags/tags.go
+++ b/pkg/tags/tags.go
@@ -1,5 +1,6 @@
 /*
- * MinIO Cloud Storage, (C) 2020 MinIO, Inc.
+ * MinIO Go Library for Amazon S3 Compatible Cloud Storage
+ * Copyright 2020-2022 MinIO, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -20,6 +21,8 @@ import (
 	"encoding/xml"
 	"io"
 	"net/url"
+	"regexp"
+	"sort"
 	"strings"
 	"unicode/utf8"
 )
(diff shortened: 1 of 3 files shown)


pkg/tags/tags.go
```diff
@@ -311,25 +341,47 @@ func ParseObjectXML(reader io.Reader) (*Tags, error) {
 	return unmarshalXML(reader, true)
 }
 
+func (tags *tagSet) parseTags(tgs string) (err error) {
+	for tgs != "" {
+		var key string
+		key, tgs, _ = strings.Cut(tgs, "&")
```
klauspost, line 347: Yeah, a bit too soon. When 1.20 is out we can use it :)

harshavardhana: done

.github/workflows/go-windows.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        go-version: [1.17.x]
+        go-version: [1.18.x, 1.19.x]
```
klauspost, line 17:
```suggestion
        go-version: [1.17.x, 1.18.x, 1.19.x]
```
IMO we should keep 3 active versions for `minio-go`. For other libs (madmin, mc, minio) we can be more aggressive.

Implementing a `strings.Cut` isn't really a big deal.

harshavardhana: I thought go1.17 is deprecated? @klauspost

harshavardhana: Okay got back go1.17 for now.

klauspost (approved):

PR #1710, opened 2022-10-16
Allow empty contentTypeStartsWith in PostPolicy

skirsten:
The S3 spec allows this:

https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-HTTPPOSTConstructPolicy.html#sigv4-ConditionMatching
> Matching Any Content
> To configure the POST policy to allow any content within a form field, use starts-with with an empty value ("").

With this change
```go
policy.SetContentTypeStartsWith("")
```
allows the uploader to specify any content-type they want.

### DIFF
--- a/functional_tests.go
+++ b/functional_tests.go
@@ -4216,10 +4216,6 @@ func testPresignedPostPolicy() {
 		logError(testName, function, args, startTime, "", "SetContentType did not fail for invalid conditions", err)
 		return
 	}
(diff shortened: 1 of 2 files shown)


vadmeste:
Thanks for the PR @skirsten 

Can you add a comment to the function description to say empty is supported ?

harshavardhana (approved):

vadmeste (approved):
LGTM

PR #1715, opened 2022-10-21
Set additional header for deletemarker replication

poornas:
to check if target is ready to receive the object version


harshavardhana (approved):

vadmeste (approved):
LGTM

PR #1718, opened 2022-10-26
fix: allow any content using starts-with with an empty value

fortuneforme:
Thanks for the PR @skirsten 

```text
commit 4acfae712723fae025df6b841f9cc0e80c8ca464
Author: Simon Kirsten <1972314+skirsten@users.noreply.github.com>
Date:   Fri Oct 21 22:38:04 2022 +0200

    Allow empty contentTypeStartsWith in PostPolicy (#1710)
```

- But it's still not allowed to add an empty value in `addNewPolicy`, so i supported it when using `starts-with`.
- And by the way, this PR will allow empty `keyStartsWith` in `PostPolicy`.

### DIFF
--- a/functional_tests.go
+++ b/functional_tests.go
@@ -4204,10 +4204,6 @@ func testPresignedPostPolicy() {
 		logError(testName, function, args, startTime, "", "SetKey did not fail for invalid conditions", err)
 		return
 	}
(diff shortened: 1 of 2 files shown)


minio-trusted (approved):

harshavardhana (approved):

vadmeste (approved):
LGTM

PR #1734, opened 2022-11-30
feat: volcengine TOS also support virtualhost.

hustclf:
Hi, I am using [volcengine TOS](https://www.volcengine.com/docs/6349/147050) (a s3 compatible Object storage from bytedance), mc and [argo workflows](https://github.com/argoproj/argo-workflows) recently,  and find out volcengine TOS only support VirtualHost because path-style is being [deprecated](https://aws.amazon.com/cn/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/) by aws s3 in the future. 

It will be convenient if mc will detect it by default, and It will also make argo workflows S3ArtifactRepository available based on volcengine TOS. :)

### DIFF
--- a/pkg/s3utils/utils.go
+++ b/pkg/s3utils/utils.go
@@ -65,9 +65,8 @@ func IsValidIP(ip string) bool {
 	return net.ParseIP(ip) != nil
 }
 
-// IsVirtualHostSupported - verifies if bucketName can be part of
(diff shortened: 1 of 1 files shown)


harshavardhana:
https://github.com/minio/minio-go/pull/2057

PR #1741, opened 2022-12-14
PutObject() to return a wrapped error for io.Reader errors

vadmeste:
PutObject() does not differentiate between internal errors 
or errors coming from io.Reader argument, which makes t
roubleshooting harder.

Add a prefix for the returned error during reading io.Reader
 argument.

Example:
Before this commit:

```
mc: <ERROR> Failed to copy `http://192.168.1.113:9000/testbucket/testobject.2`. Put "http://localhost:9001/testbucket/testobject.2": Resource requested is unreadable, please reduce your request rate
```

After this commit:
```
mc: <ERROR> Failed to copy `http://192.168.1.113:9000/testbucket/testobject.2`. Put "http://localhost:9001/testbucket/testobject.2": read: Resource requested is unreadable, please reduce your request rate
```

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -666,3 +666,20 @@ func (h *hashReaderWrapper) Read(p []byte) (n int, err error) {
 	}
 	return n, err
 }
+
+// wrapReader embeds io.Reader but returns an error
+// with 'read error' prefix. This is needed because
+// net/http Do() request does not have a way to indicate
+// a way to indicate this is an error from reading the
+// request body.
+type wrapReader struct {
+	io.Reader
+}
+
+func (wr wrapReader) Read(p []byte) (n int, e error) {
+	n, e = wr.Reader.Read(p)
+	if e != nil && e != io.EOF { // do not wrap io.EOF for callers not using errors.Is() yet
+		e = fmt.Errorf("read: %w", e)
+	}
+	return
+}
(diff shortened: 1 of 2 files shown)


utils.go
```diff
@@ -666,3 +666,20 @@ func (h *hashReaderWrapper) Read(p []byte) (n int, err error) {
...
+// net/http Do() request does not have a way to indicate
+// a way to indicate this is an error from reading the
+// request body.
+type wrapReader struct {
+	io.Reader
+}
+
+func (wr wrapReader) Read(p []byte) (n int, e error) {
+	n, e = wr.Reader.Read(p)
+	if e != nil && e != io.EOF { // do not wrap io.EOF for callers not using errors.Is() yet
```
harshavardhana, line 681: maybe also io.ErrUnexpectedEOF @vadmeste

vadmeste: > maybe also io.ErrUnexpectedEOF @vadmeste

I don't worry too much about io.ErrUnexpectedEOF because the caller will have to stop the execution and return the error but it is not the case with with io.EOF

klauspost (approved):

harshavardhana:
tests didn't finish can you rebase this PR? @vadmeste

PR #1742, opened 2022-12-15
PutObject to allow upload with a given size less than io.Reader content

vadmeste:
This commit will allow upload of io.Reader content when size is specified 
and it is less than what io.Reader can provide.

e.g.

```
 // Upload an object.
 _, err = c.PutObject(context.Background(), "bucket", "object", 
                bytes.NewBuffer([]byte("abcd")), 3, minio.PutObjectOptions{})
```

### DIFF
--- a/api.go
+++ b/api.go
@@ -813,7 +813,11 @@ func (c *Client) newRequest(ctx context.Context, method string, metadata request
 	if metadata.contentLength == 0 {
 		req.Body = nil
 	} else {
-		req.Body = io.NopCloser(metadata.contentBody)
+		body := metadata.contentBody
+		if metadata.contentLength > 0 {
+			body = io.LimitReader(body, metadata.contentLength)
+		}
+		req.Body = io.NopCloser(body)
 	}
 
 	// Set incoming content-length.
--- a/functional_tests.go
+++ b/functional_tests.go
@@ -10565,6 +10565,79 @@ func testStorageClassMetadataCopyObject() {
 	successLogger(testName, function, args, startTime).Info()
 }
 
+// Test put object with a size shorter than the io.Reader content
+func testPutObjectLessLength() {
+	// initialize logging params
+	startTime := time.Now()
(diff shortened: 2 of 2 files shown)


klauspost:
IMO we shouldn't be hiding this error. We cannot tell if the user intends to upload 3 or 4 bytes - silently choosing one could lead to data loss.

vadmeste:
> We cannot tell if the user intends to upload 3 or 4 bytes 

The size is passed as argument in PutObject()

klauspost:
@vadmeste Yes, but we should inform that there is a mismatch. 

The caller can do the same wrapping if they don't want an error. As an SDK we should not silently ignore consistency issues.

vadmeste:
>  @vadmeste Yes, but we should inform that there is a mismatch.
> The caller can do the same wrapping if they don't want an error. As an SDK we should not silently ignore consistency issues.

Well, I think this is not intuitive, also this means it is not possible to split an io.Reader stream into multiple objects

klauspost:
> it is not possible to split an io.Reader stream into multiple objects

What do you mean? You can still use LimitedReader upstream.

harshavardhana:
@klauspost I think @vadmeste is thinking from io.CopyN() style which can arbitrarily copy lesser bytes from the reader.

klauspost:
I still don't understand. 

We should not hide user errors. If the user claims the size to be `n` bytes, the Reader should provide `n` bytes. No more, no less.

vadmeste:
- If 'n' is more the io.Reader content size, then yes we should return an error to inform the user
- If 'n' is less than io.Reader content size, then this is not necessarly a user error. A user can ask to upload only a certain number of bytes, similar to io.CopyN(). The user is explicitly setting a size in PutObject API here

However, I noticed that PutObject() has a io.Seeker inside and seeks to zero offset before the upload.. so it looks like this PR will have unexpected behavior at least for now. I guess we can close this PR.

PR #1743, opened 2022-12-22
Support V10 generated config

gtpho:
It's currently not possible to use `credentials.NewFileMinioClient` with a V10 generated config because it expects `hosts` instead of `aliases`. This commit uses `aliases` if the version is 10.


harshavardhana (approved):

vadmeste (approved):
LGTM

PR #1745, opened 2023-01-05
Add parallel buffered multipart upload

klauspost:
Add multipart upload mode that will buffer a number of parts in memory and fill/upload them in parallel.

Use `ConcurrentStreamParts` to enable and `NumThreads` to control the number of buffers. `PartSize` will control the size of each buffer.

### DIFF
--- a/api-put-object-streaming.go
+++ b/api-put-object-streaming.go
@@ -28,6 +28,7 @@ import (
 	"net/url"
 	"sort"
 	"strings"
+	"sync"
 
 	"github.com/google/uuid"
 	"github.com/minio/minio-go/v7/pkg/s3utils"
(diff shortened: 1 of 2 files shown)


api-put-object-streaming.go
```diff
@@ -446,6 +449,193 @@ func (c *Client) putObjectMultipartStreamOptionalChecksum(ctx context.Context, b
...
+
+	// Cancel all when an error occurs.
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	// Calculate the optimal parts info for a given size.
+	partSize := int64(maxPartSize)
+	if opts.PartSize > 0 {
+		partSize = int64(opts.PartSize)
+	}
```
harshavardhana, line 480: we should be using this?

```
        totalPartsCount, partSize, lastPartSize, err := OptimalPartInfo(size, opts.PartSize)
        if err != nil {
                return UploadInfo{}, err
        }
```

maxPartSize is 5Gi by default which is a bit too much?

harshavardhana (changes requested):

harshavardhana (approved):

harshavardhana (approved):

PR #1749, opened 2023-01-10
Use provided sha256 hasher

klauspost:
Use client provided hasher for SHA256 signatures.

By default this is "minio/sha256-simd" or "crypto/sha256" for FIPS.

Bonus: Skip tests if env var is missing.

### DIFF
--- a/pkg/signer/request-signature-streaming.go
+++ b/pkg/signer/request-signature-streaming.go
@@ -248,7 +248,7 @@ func (s *StreamingReader) addSignedTrailer(h http.Header) {
 
 	s.sh256.Reset()
 	s.sh256.Write(s.chunkBuf)
-	chunckChecksum := hex.EncodeToString(s.sh256.Sum(nil))
+	chunkChecksum := hex.EncodeToString(s.sh256.Sum(nil))
 	// Compute chunk signature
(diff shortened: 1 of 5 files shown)


harshavardhana (approved):

harshavardhana (changes requested):

vadmeste:
@klauspost looks like this needs one more fix

pkg/signer/request-signature-streaming.go
```diff
@@ -294,6 +304,7 @@ func StreamingSignV4(req *http.Request, accessKeyID, secretAccessKey, sessionTok
 		chunkNum:        1,
 		totalChunks:     int((dataLen+payloadChunkSize-1)/payloadChunkSize) + 1,
 		lastChunkSize:   int(dataLen % payloadChunkSize),
+		sh256:           sh256,
```
vadmeste, line 307: Maybe let's define a default sha256 if sh256 is nil ? this is a public package

klauspost: We always set a fallback if unset: https://github.com/minio/minio-go/blob/efc53fd9e0f1d982354d3168aa437165b19fd110/api.go#L253

harshavardhana (approved):

PR #1752, opened 2023-01-13
allow re-use of sha256Pool and md5Pool in streaming uploads

harshavardhana:


vadmeste (approved):
LGTM

api-put-object-streaming.go
```diff
@@ -552,52 +555,63 @@ func (c *Client) putObjectMultipartStreamParallel(ctx context.Context, bucketNam
 			return UploadInfo{}, rerr
 		}
 
-		// Calculate md5sum.
 		customHeader := make(http.Header)
-		if !opts.SendContentMd5 {
-			// Add CRC32C instead.
+		// Calculates hash sums while copying partSize bytes into cw.
```
klauspost, line 559: You are moving the expensive hash calculation from the multithreaded goroutines to the single-threaded reader routine. I don't see how that is an improvement.

Also streaming the SHA256 distributes the calculation on the upload, instead of doing everything as a single block.

klauspost (changes requested):
I also don't see where the pool is missing.

The `return` is a good change, and I see another issue. I will send a PR with the "useful" stuff.

api-put-object-streaming.go
klauspost: Also you are doing content-sha AND streaming SHA.

harshavardhana: This is similar to other upload code, content sha256 is only done as needed when requested you are not reading the code right.

Hash materials only returns the checksums when it's required. This code is correct otherwise other code will be wrong.

Each buffer is read serially anyways   I don't see how it is parallel processing. Compute here is serialized because the buffer is serialized so it means that we are not wasting CPU cycles during upload instead when uploading we just upload() and no compute.

The problem right now is big buffers we use too much CPU here causing overall slowness

klauspost: > Each buffer is read serially anyways I don't see how it is parallel processing. 

You are switching from `[reading][reading][reading]` to `[reading][slow checksumming][reading][slow checksumming]`. This will block the incoming data. SHA and MD5 are typically < 1GB/s, so the blocking will be significant.  CRC32 is at memory speed, typically 30GB/s+.

You don't reduce the amount of checksumming, you are just making it slower.

SHA256 is already added per block when uploaded with HTTP, so there is no need to do a full part SHA256 as well. There are faster checksums, which will validate the uploaded part.

PR #1764, opened 2023-01-27
Update minio.Core API

donatello:
These breaking changes are added:

- Add custom headers parameter for PutObjectPart

- Expose upload info in CompleteMultipartUpload

PR #1765, opened 2023-01-27
Add CRC32 checksums as trailer

donatello:
This change is made for the multipart parallel (non-seekable) stream
upload.


donatello:
Needs https://github.com/minio/minio-go/pull/1764

klauspost (changes requested):
Just send it as a custom header like before.

There is no need for all this complication here. Any CPU will do the CRC at memory read speed or close to it.

donatello:
@klauspost 👍🏾 Closing this one for now. Won't try to add this again, unless we can demonstrate at least some improvement.

PR #1768, opened 2023-02-07
Bump Go versions

klauspost:
Deprecate Go 1.17.x and add Go 1.20.x

This is may also fix the issues with staticcheck.

Edit: Staticcheck requires Go 1.19 - ditching Go 1.18 check.


harshavardhana (approved):

vadmeste (approved):
LGTM

PR #1775, opened 2023-02-24
custom header

molei20021:
add custom header global

### DIFF
--- a/api.go
+++ b/api.go
@@ -96,6 +96,7 @@ type Client struct {
 	healthStatus int32
 
 	trailingHeaderSupport bool
+	headers               map[string]string
 }
 
 // Options for New method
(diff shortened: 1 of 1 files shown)


harshavardhana:
@molei20021 please explain why this is needed.

molei20021:
> @molei20021 please explain why this is needed first..

Because minio do not support subpath, this can help to identify which backend of minio service to call for proxy server

vadmeste:
@molei20021 if your applications will need to do an extra thing to select a specific backend, then you can simply create a subdomain for each backend server. That should work as well.

harshavardhana:
> > @molei20021 please explain why this is needed first..
> 
> Because minio do not support subpath, this can help to identify which backend of minio service to call for proxy server

You need to base that off from access credentials map - this is not correct API pattern.

PR #1782, opened 2023-03-01
Detect S3 region from non amazon endpoint

fatpat:
when using S3 endpoints from providers other than amazon the region is not extracted from the endpoint and the first request returns `HTTP 400` with `AuthorizationHeaderMalformed` and the client gather the region from the XML and then retry the request with the correct region.

It would be nice to extract the region from the endpoint instead of having to set it in `minio.Options` as it's redundant when the region is already in the endpoint.


harshavardhana:
You should set the right region for your client constructor. We did it for AWS S3 because they have well defined domains and regions.

We do not want to generalize for all endpoints.

harshavardhana:
https://github.com/minio/minio-go/pull/1786 This should be what you would want here, so write your own detection patterns here.

PR #1783, opened 2023-03-01
Detect S3 region from OVH Cloud endpoints

fatpat:
as discussed in #1782, a generic solution is not acceptable.

here is the regex that match all [OVH CLOUD S3](https://www.ovhcloud.com/en/public-cloud/object-storage/) endpoints.

Working at ovh in the public storage team, I can assure the endpoints matches `s3.region(.sub)?.cloud.ovh.(net|us)`. Futur endpoints will follow the same pattern.


harshavardhana:
As I mentioned this behavior is only meant for Amazon S3, OVH cloud is an alternative to AWS S3. We don't plan to have first class support here all clouds today and future.

I will bring in a change where you can provide your own regex for region lookup during MinIO constructor initialization so that we don't pollute this SDK further.

PR #1784, opened 2023-03-03
fix: getBucketLocation using virtual hosted style for Ksyun KS3

ChaosBee:
[Ksyun KS3](https://en.ksyun.com/nv/product/KS3.html) —— Object Storage Service by Kingsoft Cloud.

The getBucketLocation API is accessed using virtual hosted style in KS3 by default.

refer:  #1253

### DIFF
--- a/bucket-cache.go
+++ b/bucket-cache.go
@@ -194,8 +194,8 @@ func (c *Client) getBucketLocationRequest(ctx context.Context, bucketName string
 
 	var urlStr string
 
-	// only support Aliyun OSS for virtual hosted path,  compatible  Amazon & Google Endpoint
-	if isVirtualHost && s3utils.IsAliyunOSSEndpoint(targetURL) {
+	// only support Aliyun OSS & Ksyun KS3 for virtual hosted path,  compatible  Amazon & Google Endpoint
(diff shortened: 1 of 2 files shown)


harshavardhana:
We are not going to add more endpoints here. Aliyun is where we will stop.

ChaosBee:
OK. Thank you for your response. 
What a pity. I believe that host-style access is better than path-style access. So is it possible that minio-sdk or minio will add a variable changing the style of get-bucket-location?

harshavardhana:
All you need to do is tell how the bucket lookups must happen

https://github.com/minio/minio-go/blob/master/api.go#LL107C6-L107C6

There is no reason to have new magic added into codebase. However I would perhaps go ahead and allow extending the checks to ensure users pass their own functions here instead of polluting the SDK further.

ChaosBee:
As your mention, I tried the BucketLoookup Variable before, but it does no effect on the _urlStr_ that get_bucket_location use.

https://github.com/minio/minio-go/blob/master/bucket-cache.go#L199

harshavardhana:
Yes that is correct because an S3 service must implement both path and dns style. The region is not known prior to getBucketLocation() - however can fix the behavior here. to use DNS style for GetBucketLocation() requests if it works?

harshavardhana:
https://github.com/minio/minio-go/pull/1785 this should work for you @ChaosBee

PR #1793, opened 2023-03-29
For the document

cqqqq777:
Updated Chinese documents and collated English documents


klauspost (commented):
Thanks for the update. I will see if I can locate one to do a review.

jiuker (changes requested):
This pr is a old version from #1794 . Let review it as #1794

PR #1795, opened 2023-03-30
Fix listobjects with metadata

klauspost:
`UserMetadata` did not match actual format sent, so this has been broken for a while.

Change type of `UserTags`, so it can be unmarshaled from XML.

Fix for minio server that matches this will be sent.

Tested with mc.


klauspost:
https://github.com/minio/minio/pull/16906 is the server side.

PR #1796, opened 2023-03-31
Presign return invalid url

ihatiko:
Hello.

Expectation:
/some-bucket/subdirectory-path1/subdirectory-path2/test.jpg

Reality:
 /some-bucket//subdirectory-path1/subdirectory-path2/test.jpg

`
d, err := client.PreSign(context.Background(), "some-bucket", "/subdirectory-path1/subdirectory-path2/test.jpg")
fmt.Println(d.Path) // returning /some-bucket//subdirectory-path1/subdirectory-path2/test.jpg
`

this fix trim objectName path, to correct urlPath

### DIFF
--- a/api-presigned.go
+++ b/api-presigned.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"net/http"
 	"net/url"
+	"strings"
 	"time"
 
 	"github.com/minio/minio-go/v7/pkg/s3utils"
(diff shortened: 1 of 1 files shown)


klauspost:
Yes. In S3 object keys are "just text", so even if MinIO cannot handle objects with a `/` as the first character they are technically allowed.

ihatiko:
but if you work with url, its not correct behaviour

harshavardhana:
@ihatiko it is correct, you should normalize the object names as necessary outside of this SDK.

PR #1808, opened 2023-04-15
listen bucket: Support empty events from the server

vadmeste:
Ask the server to send empty events to clear the buffer 
scanner which can be filled with whitespaces but without
 new lines when the bucket is idle and without any events


harshavardhana:
PTAL at the conflict @vadmeste

vadmeste:
There is another PR similar to this already merged, 

https://github.com/minio/minio-go/pull/1812

I may sent two PRs already, this is not needed.

PR #1817, opened 2023-05-03
Add PutObject CRC & trailing checksum tests

klauspost:
Add tests for https://github.com/minio/minio/pull/16484 (required to pass)

This tests trailing checksums and has been validated against AWS S3.

PutObject will now add CRC when client supports it, MD5 is not set, not Google and we don't add SHA256 to chunks.

### DIFF
--- a/pkg/signer/request-signature-v4.go
+++ b/pkg/signer/request-signature-v4.go
@@ -289,7 +289,7 @@ func signV4(req http.Request, accessKeyID, secretAccessKey, sessionToken, locati
 			req.Header.Add("X-Amz-Trailer", strings.ToLower(k))
 		}
 
-		req.TransferEncoding = []string{"aws-chunked"}
+		req.Header.Set("Content-Encoding", "aws-chunked")
 		req.Header.Set("x-amz-decoded-content-length", strconv.FormatInt(req.ContentLength, 10))
 	}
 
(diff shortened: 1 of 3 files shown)


klauspost:
Passing now.

pkg/signer/request-signature-v4.go
```diff
@@ -289,7 +289,7 @@ func signV4(req http.Request, accessKeyID, secretAccessKey, sessionToken, locati
 			req.Header.Add("X-Amz-Trailer", strings.ToLower(k))
 		}
 
-		req.TransferEncoding = []string{"aws-chunked"}
+		req.Header.Set("Content-Encoding", "aws-chunked")
```
harshavardhana, line 292: are you sure that this won't break anything?

harshavardhana (changes requested):

pkg/signer/request-signature-v4.go
klauspost: @harshavardhana It is the correct value. 

From https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html

> [Y]ou must include the following headers:
> [...] Content-Encoding: Set the value to `aws-chunked`.

harshavardhana: This basically broke if you remember last time? @klauspost

klauspost: @harshavardhana No. The only thing I remember was that what we tried didn't make any difference (since this is the wrong field). Please refresh my memory.

harshavardhana (approved):

PR #1826, opened 2023-05-17
implement fan-out API

harshavardhana:


vadmeste:
@harshavardhana it is not very clear  what io.Reader contains exactly, (there is no Content-Length field in PutObjectFanOutRequest), can you explain ?

klauspost:
@vadmeste I don't believe mime-post needs a `Content-Length` - and when not set it is just a streaming request.

We should ensure that we reliably detected truncated requests, but AFAIR mime sends the content length of each value, so we should be fairly safe here.

harshavardhana:
> @vadmeste I don't believe mime-post needs a `Content-Length` - and when not set it is just a streaming request.
> 
> We should ensure that we reliably detected truncated requests, but AFAIR mime sends the content length of each value, so we should be fairly safe here.

@vadmeste mime-post doesn't need content length it is up to the server to read and write from the stream as the data arrives.

api-put-object-fan-out.go
```diff
@@ -0,0 +1,145 @@
...
+	VersionID    string     `json:"versionId,omitempty"`
+	LastModified *time.Time `json:"lastModified,omitempty"`
+	Error        error      `json:"error,omitempty"`
+}
+
+// PutObjectFanOut - is a variant of PutObject instead of writing a single object from a single
+// stream multiple objects are written, defined via a list of PutObjectFanOutRequests. Each entry
+// in PutObjectFanOutRequest carries an object keyname and its relevant metadata if any. `Key` is
+// mandatory, rest of the other options in PutObjectFanOutRequest are optional.
+func (c *Client) PutObjectFanOut(ctx context.Context, bucket string, body io.Reader, fanOutReq []PutObjectFanOutRequest) ([]PutObjectFanOutResponse, error) {
```
klauspost, line 63: It should be possible to add Form elements. Maybe something like:

```suggestion

// PutObjectFanOutRequest contains parameters for PutObjectFanOut requests.
type PutObjectFanOutRequest struct {
	// Bucket is the destination bucket.
	Bucket  string
	
	// Targets contains all targets to fan out to. 
	Targets []FanOutTarget
	
	// Data should return the data to upload.
	Data io.Reader
	
	// UserData can contain additional fields that will be added to the form.
	UserData map[string]string
}

// PutObjectFanOut - is a variant of PutObject instead of writing a single object from a single
// stream multiple objects are written, defined via a list of PutObjectFanOutRequests. Each entry
// in PutObjectFanOutRequest carries an object keyname and its relevant metadata if any. `Key` is
// mandatory, rest of the other options in PutObjectFanOutRequest are optional.
func (c *Client) PutObjectFanOut(ctx context.Context, o PutObjectFanOutRequest) ([]PutObjectFanOutResponse, error) {
	policy := NewPostPolicy()
	policy.SetBucket(o.Bucket)
	policy.SetKey(strconv.FormatInt(time.Now().UnixNano(), 16))
	for k, v := range o.UserData {
		policy.SetUserData(strings.TrimPrefix(strings.ToLower(k), "amz-x-"), v)
	}
```

harshavardhana: The reason I kept it this way is to match input and output slices to be an exact match - putting all of them into a big struct makes it a bit wide IMO. 

We do not set anything on the form field itself as there is no value in adding that, these are directly added per object in the JSON value itself.

harshavardhana: It also means to match that we are sending JSON line by line instead of a single large JSON struct.

harshavardhana: Also would like the `bucket` to be an actual arg, and the body as well to be separate. To designate them as mandatory args not optional - just like all our other APIs

```
APIName(context, bucketname, objectname, io.Reader, opts)
```

In this case since object is configurable it is part of `opts...`

```
APIName(context, bucketname, io.Reader, opts...)
```

docs/API.md
```diff
@@ -504,6 +504,46 @@ if err != nil {
...
+
+__minio.PutObjectFanOutResponse__
+
+| Field          | Type       | Description                                                          |
+|:---------------|:-----------|:---------------------------------------------------------------------|
+| `Key`          | _string_   | Name of the object                                                   |
+| `ETag`         | _string_   | ETag opaque unique value of the object                               |
+| `VersionID`    | _string_   | VersionID of the uploaded object                                     |
+| `LastModified` | _time.Time | Last modified time of the latest object                              |
+| `Error`        | _error_    | Is only non `nil` only when the fan-out for a specific object failed |
```
poornas, line 545:
```suggestion
| `Error`        | _error_    | Is non `nil` only when the fan-out for a specific object failed |
```

poornas (changes requested):

poornas (approved):

harshavardhana:
okay confirmed with @abperiasamy we go with the current API for now.

donatello (approved):

PR #1827, opened 2023-05-19
Add mint PostObject checksum

klauspost:
Mint test for: https://github.com/minio/minio/pull/17244  (released)

Validated against AWS S3.

PR #1828, opened 2023-05-23
add fan-Out API support for encryption, checksums

harshavardhana:


harshavardhana:
Based on your proposal - keeping few things 

- per object, metadata, tags, and retention can be applied independently
- encryption is set for the whole fan-out
- checksum is applied for the whole fan-out

@klauspost - keeps the API functional name in-line with the common pattern across this SDK.

klauspost (approved):
Nice! I will update https://github.com/minio/minio-go/pull/1827 to use the new methods when merged.

PR #1835, opened 2023-05-30
fix: Send correct host header when an endpoint is IPv4

vadmeste:
Currently, when the specified endpoint is an IP v4 address and the port is 80/443, 
minio-go sends a Host header with a bracket, which is wrong.

An ipV4 can be converted to ip6 in golang library, therefore using ip.To4() shoud 
return nil if the passed IP is v6.


klauspost (approved):
LGTM.

harshavardhana (approved):

PR #1838, opened 2023-06-13
Fix multipart upload with encryption context

alanprot:
Similar to https://github.com/minio/minio-go/pull/1777

The context header is also not allowed when completing the multipart upload on S3 and should be removed.

https://docs.aws.amazon.com/AmazonS3/latest/API/API_CompleteMultipartUpload.html

I also created a unit test to check all those behaviors.


go.mod
```diff
@@ -11,15 +11,19 @@ require (
 	github.com/minio/sha256-simd v1.0.1
 	github.com/rs/xid v1.5.0
 	github.com/sirupsen/logrus v1.9.2
+	github.com/stretchr/testify v1.7.0
 	golang.org/x/crypto v0.9.0
```
harshavardhana, line 15: Please don't bring in testify, please use Go's testing package instead.

alanprot: ok

alanprot: Done!

harshavardhana (approved):

klauspost (approved):
LGTM

alanprot:
thanks

PR #1839, opened 2023-06-15
Fix fanout response unmarshalable error type

klauspost:
Type `error` cannot be JSON unmarshaled, so it should not be used in responses.

Change the type to a  string.

Not backwards compatible.


vadmeste (approved):
LGTM

harshavardhana (approved):

PR #1843, opened 2023-06-22
Added missing Buckets Events

bexsoft:
## What does this do?

Added missing bucket events for `replica` & `ilm` according to supported events list in https://min.io/docs/minio/linux/reference/minio-mc/mc-event-add.html#mc-event-supported-events


klauspost (approved):
LGTM

dvaldivia (approved):

Issue #1813, opened 2023-04-28
Running Stat on GetObject before reading discards range


klauspost:
Running the code:

```Go
		gopts := minio.GetObjectOptions{}
		err = gopts.SetRange(100, 1000)
		if err != nil {
				panic(err)
		}
		fmt.Println("HDR:", gopts.Header())
		c.TraceOn(os.Stdout)
		r, err := c.GetObject(context.Background(), bucketName, objectName, gopts)
		if err != nil {
				panic(err)
		}
		
		if false { // enable to get correct size
			_, err := io.ReadAll(r)
			if err != nil {
				logError(testName, function, args, startTime, "", "Read failed", err)
				return
			}
		}
		
		st, err := r.Stat()
		if err != nil {
				panic(err)
		}
		fmt.Println("Size:", st.Size)
```

This discards the range and the reported size is the full object. Enabling the read gives the expected size (901 bytes).

jiuker:
https://github.com/minio/minio-go/blob/ab51b38e91a09d5c803ae2b45ec722e6b4071e00/api-get-object.go#L152-L156
```golang
					// First request is a Stat or Seek call.
					// Only need to run a StatObject until an actual Read or ReadAt request comes through.

					// Remove range header if already set, for stat Operations to get original file size.
					delete(opts.headers, "Range")
					objectInfo, err = c.StatObject(gctx, bucketName, objectName, StatObjectOptions(opts))
					if err != nil {
						resCh <- getResponse{
							Error: err,
						}
						// Exit the go-routine.
						return
					}
					etag = objectInfo.ETag
					// Send back the first response.
					resCh <- getResponse{
						objectInfo: objectInfo,
					}
```
Looks like it's expected. @harshavardhana cc

klauspost:
@jiuker 

A) Calling Stat before the read shouldn't discard the range. That is a clear bug.

B) `Size()` should return a consistent value whether Read has been called or not. It seems this is intended to be the full object size.

PR #1846, opened 2023-06-29
Enable stat operations for specific ranges

r-scheele:
ref issue number: #1813

`Stat()` was designed to always work on the original file size, and not a portion of the object


klauspost (approved):
LGTM

api-get-object.go
```diff
@@ -127,9 +127,6 @@ func (c *Client) GetObject(ctx context.Context, bucketName, objectName string, o
 				} else {
 					// First request is a Stat or Seek call.
 					// Only need to run a StatObject until an actual Read or ReadAt request comes through.
-
-					// Remove range header if already set, for stat Operations to get original file size.
-					delete(opts.headers, "Range")
```
vadmeste, line 132: Does setting range change the stat output ? Let's get @harshavardhana  review here. But I agree, this is not the right place to delete a header here

klauspost: See linked bug. It creates a weird inconsistency where the Stat response changes based on whether you have read from the returned object or not.

vadmeste: > See linked bug. It creates a weird inconsistency where the Stat response changes based on whether you have read from the returned object or not.

Yes, I agree, in general it is just wrong to remove a header at this location. But my question is why the code was there in the first place, so just asking for @harshavardhana input

harshavardhana: This code was there to ensure that we always store the original file size not the size of the range.

We only need to honor range for the body when 'reading' but Stat() output was using the total size to ensure Seek() can return appropriate errors.

If we honor Stat() to honor the range required for the body then the problems might show up with Seek().

harshavardhana: Remember Stat() is like (*os.File).Stat() which returns the original size of the object not the range you asked. 

The range is basically here given for situations where the caller simply doesn't care about Stat() or Seek()

However users like mc don't really use such low level usage. It relies on Seek() to have correct behavior and a Seek() must not change the Stat() result this is why the range header is removed here. Since we are interested in the original size of the object.

klauspost: @harshavardhana Then calling Stat after reading cannot reuse the the data info from GetObject should be changed to always do HeadObject if there is a range.

harshavardhana: Correct we should fix that @klauspost it must be "independent" of any existing state.

klauspost: @r-scheele Let's do that then. If a range was set and we fetched some data, discard the information we already got and perform a clean HeadObject.

r-scheele: Alright @klauspost, I'll look into it.

api-get-object.go
```diff
@@ -415,7 +414,8 @@ func (o *Object) Stat() (ObjectInfo, error) {
 	if o.prevErr != nil && o.prevErr != io.EOF || o.isClosed {
 		return ObjectInfo{}, o.prevErr
 	}
-
+	o.isStarted = false
+	o.objectInfoSet = false
```
klauspost, line 418: Can't you check if range has been set and only discard it if it has?

klauspost (approved):
Thanks! That is more in line with the change I expected.

harshavardhana:
Please fix the tests @r-scheele

klauspost:
```
2023-07-06T11:10:48.3248895Z === RUN   TestCoreCopyObject
2023-07-06T11:10:48.3294758Z ##[error]    core_test.go:472: Error: number of bytes in stat does not match, want 32768, got 0
2023-07-06T11:10:48.3304410Z --- FAIL: TestCoreCopyObject (0.03s)
2023-07-06T11:10:48.3304745Z === RUN   TestCoreCopyObjectPart
2023-07-06T11:10:48.3305160Z --- PASS: TestCoreCopyObjectPart (0.41s)
2023-07-06T11:10:48.3305442Z === RUN   TestCorePutObject
2023-07-06T11:10:48.3306219Z ##[error]    core_test.go:737: Error: number of bytes in stat does not match, want 32768, got 0
2023-07-06T11:10:48.3306961Z --- FAIL: TestCorePutObject (0.03s)
```

functional_tests.go
```diff
@@ -2170,43 +2171,12 @@ func testPutObjectWithChecksums() {
...
-			return
-		}
-		r, err = c.GetObject(context.Background(), bucketName, objectName, gopts)
-		if err != nil {
-			logError(testName, function, args, startTime, "", "GetObject failed", err)
-			return
-		}
-
-		b, err = io.ReadAll(r)
+		// Discard the object
```
klauspost, line 2174: Why did you delete all this?

Checking what checksum range requests return seems valuable.

pkg/tags/tags_test.go
```diff
@@ -39,7 +39,7 @@ func TestParseTags(t *testing.T) {
 			2,
 		},
 		{
-			" store forever =false&factory=true",
+			"_store_forever_=false&factory=true",
```
klauspost, line 42: What is this change about?

r-scheele: The test was failing, so I noticed that fixed it.

functional_tests.go
r-scheele: It was deleted during testing, please check again.

api-get-object.go
```diff
@@ -129,19 +131,39 @@ func (c *Client) GetObject(ctx context.Context, bucketName, objectName string, o
 					// Only need to run a StatObject until an actual Read or ReadAt request comes through.
 
 					// Remove range header if already set, for stat Operations to get original file size.
-					delete(opts.headers, "Range")
-					objectInfo, err = c.StatObject(gctx, bucketName, objectName, StatObjectOptions(opts))
-					if err != nil {
+					_, rangeHeaderPresent := opts.headers["Range"]
```
minio-trusted, line 134:
```suggestion
					_, ok := opts.headers["Range"]
```

api-get-object.go
```diff
@@ -129,19 +131,39 @@ func (c *Client) GetObject(ctx context.Context, bucketName, objectName string, o
...
+					if rangeHeaderPresent {
```
minio-trusted, line 135:
```suggestion
					if ok {
```

api-get-object.go
```diff
@@ -129,19 +131,39 @@ func (c *Client) GetObject(ctx context.Context, bucketName, objectName string, o
...
+						}
+						etag = objectInfo.ETag
+						// Send back the first response.
 						resCh <- getResponse{
-							Error: err,
+							objectInfo: objectInfo,
+						}
+					} else {
+						// The range header is not present, continue with the existing objectInfo.
+						delete(opts.headers, "Range")
```
minio-trusted, line 152:
```suggestion
```

Not needed in `else {`

minio-trusted (changes requested):

PR #1857, opened 2023-07-17
allow users to provide their custom httptrace.ClientTrace

harshavardhana:
providing custom httptrace allows users to debug low
level details such as dial, dns and connection errors 
more granularly.


klauspost (approved):

vadmeste (approved):
LGTM

PR #1868, opened 2023-08-05
allow skipping errors while creating snowball archive

harshavardhana:


klauspost (commented):
I guess this could work if the error is on the write side.

However, I don't really like the idea of returning 200, when some of the files haven't been written. It could hide a lot of problems.

harshavardhana:
> However, I don't really like the idea of returning 200, when some of the files haven't been written. It could hide a lot of problems.

it is when you don't care about source data if it's readable or not, very rarely practical but still applicable when you are recovering as much as you can from source to destination.

@klauspost

klauspost:
@harshavardhana Yes, I see that. I do however think for the majority you want 200 to mean "your files are safe".

Another header? - I don't think modifying the body is an option.

harshavardhana:
> @harshavardhana Yes, I see that. I do however think for the majority you want 200 to mean "your files are safe".
> 
> Another header? - I don't think modifying the body is an option.

This is still optional right now default. It's only useful rarely by some tools such as ziptransfer that I wrote

klauspost (approved):

PR #1870, opened 2023-08-10
disable mandatory V2 Google signature

### LINKED ISSUE #1800: Google storage signature issue
https://github.com/minio/minio-go/blob/5b159884739dd2c6330b423378de330775f3c93b/api.go#L158

All requsts to Google Storage lead to signature error:
`The request signature we calculated does not match the signature you provided. Check your Google secret key and signing method.`

As of September 2020, Google Cloud Storage only supports request signing using version 4 of the Signature Version.

loorke:
Any suggestion why we wouldn't want to do it? I've already encountered some issues while presigning upload URLs for Google Storage. To be more precise, Google force you to include _Content-Type_ header into the signature, and you can't specify it while calling _Client.PresignHeader()_ because of this implicit limitation of which me and my colleagues were unaware, resulting into a considerable waste of time.

Fixes #1800


klauspost (approved):
LGTM

harshavardhana (approved):

PR #1872, opened 2023-08-11
update library dependencies to fix CVEs

### LINKED ISSUE #1871: CVE-2023-3978 in golang.org/x/net@v0.12.0
Hi!

There is a vulnerability found in golang.org/x/net before v0.13.0: https://pkg.go.dev/vuln/GO-2023-1988. Is there any plan to upgrade golang.org/x/net?

Though the affected package is golang.org/x/net/html which I think is not likely to be used by minio-go, it would help a lot if the potential vulnerability can be eliminated due to the vulnerability scanning strategy on our product.

harshavardhana:
fixes #1871


klauspost (approved):
I guess just testing the lower and upper version is fine.

PR #1873, opened 2023-08-14
api: retry GET and HEAD requests without request bodies

MichaHoffmann:
The current retry logic only retries if we have a seekable request body. This makes it so that for example `ListBuckets` will not be retried. This PR fixes this by allowing GET and HEAD requests to be retried if they have no request body. 

Related to: https://github.com/thanos-io/thanos/issues/6535


MichaHoffmann:
Closing ~ misread the code, it already retries those

PR #1874, opened 2023-08-16
replication: additional fields for metrics

poornas:


harshavardhana (approved):

pkg/replication/replication.go
```diff
@@ -803,33 +855,52 @@ type ReplQueueStats struct {
...
+func (q ReplQueueStats) Workers() (tot WorkerStat) {
 	for _, node := range q.Nodes {
-		workers += int64(node.ActiveWorkers)
+		tot.Avg += node.Workers.Avg
+		tot.Curr += node.Workers.Curr
+		if tot.Max < node.Workers.Max {
+			tot.Max = node.Workers.Max
+		}
+	}
+	tot.Avg /= float32(len(q.Nodes))
```
harshavardhana, line 866: check for situations like `q.Nodes` is zero @poornas to avoid potential crashes if any.

harshavardhana (changes requested):
we simply need to have some validation on the data structure received such that we can avoid panics at runtime.

Add some validation before we can trust the structs, return error early and then we can safely process the stats.

harshavardhana (approved):

jiuker:
It broken something to mc. @poornas @harshavardhana

PR #1877, opened 2023-08-22
allow IMDSv2 endpoint to fail, fallback to IMDSv1

### LINKED ISSUE #1866: Unable to authenticate using node role
Cause and effect:
During the investigation of https://github.com/grafana/tempo/issues/2743 I was able to track down the problem to specific commits in mino-go https://github.com/minio/minio-go/pull/1626 and https://github.com/minio/minio-go/pull/1682.
Those two together causes us not being able to authenticate, and subsequent Tempo error. 

If that is expected behavior (as it seems according to this Cortex commit https://github.com/cortexproject/cortex/pull/4897) the issue lies in the usage of the library.

harshavardhana:
fixes #1866


klauspost (approved):

vadmeste (approved):
LGTM

PR #1881, opened 2023-08-23
replication: fix dependency issues

poornas:
adding back deprecated fields for metrics to avoid dependency issues


harshavardhana (approved):

vadmeste (approved):
LGTM

PR #1893, opened 2023-10-23
Support replication of deleted ILM expiry rules

shtripat:


zveinn (approved):

api-bucket-lifecycle.go
```diff
@@ -24,14 +24,15 @@ import (
...
+	"strconv"
 	"time"
 
 	"github.com/minio/minio-go/v7/pkg/lifecycle"
 	"github.com/minio/minio-go/v7/pkg/s3utils"
 )
 
 // SetBucketLifecycle set the lifecycle on an existing bucket.
-func (c *Client) SetBucketLifecycle(ctx context.Context, bucketName string, config *lifecycle.Configuration) error {
+func (c *Client) SetBucketLifecycle(ctx context.Context, bucketName string, config *lifecycle.Configuration, expiryRuleRemoved bool) error {
```
harshavardhana, line 35: No breaking APIs please @shtripat :-) 

Please add a new API or move `expiryRuleRemoved` into `lifecycle.Configuration` as another variable.

harshavardhana (changes requested):

api-bucket-lifecycle.go
```diff
@@ -48,15 +49,16 @@ func (c *Client) SetBucketLifecycle(ctx context.Context, bucketName string, conf
...
 }
 
 // Saves a new bucket lifecycle.
-func (c *Client) putBucketLifecycle(ctx context.Context, bucketName string, buf []byte) error {
+func (c *Client) putBucketLifecycle(ctx context.Context, bucketName string, buf []byte, expiryRuleRemoved bool) error {
 	// Get resources properly escaped and lined up before
 	// using them in http request.
 	urlValues := make(url.Values)
 	urlValues.Set("lifecycle", "")
+	urlValues.Set("expiryRuleRemoved", strconv.FormatBool(expiryRuleRemoved))
```
harshavardhana, line 61:
```suggestion
	urlValues.Set("expiryRuleRemoved", strconv.FormatBool(expiryRuleRemoved))
```

We shouldn't add this by default no other S3 provider will recognize this and may even error out including AWS S3.

harshavardhana, line 61: also why does it say `expiryRuleRemoved` ?

harshavardhana (changes requested):

api-bucket-lifecycle.go
shtripat: I think I will move this under `lifecycle.Configuration` to keep it non breaking.

api-bucket-lifecycle.go
shtripat: If we move the flag under `lifecycle.Configuration` we dont need to set the explicitly anymore.

shtripat: Actually `SetBucketLifecycle` gets invoked for ILM add/edit as well as ILM remove scenarios. So to differentiate setting of ILM rule whether its removal or not, I set the flag. It gets passed as true only in case of ILM delete scenario. Could there be a better name, we can change this name.

pkg/lifecycle/lifecycle.go
```diff
@@ -449,8 +449,9 @@ type Rule struct {
 
 // Configuration is a collection of Rule objects.
 type Configuration struct {
-	XMLName xml.Name `xml:"LifecycleConfiguration,omitempty" json:"-"`
-	Rules   []Rule   `xml:"Rule"`
+	XMLName           xml.Name `xml:"LifecycleConfiguration,omitempty" json:"-"`
+	Rules             []Rule   `xml:"Rule"`
+	ExpiryRuleRemoved bool     `xml:"ExpiryRuleRemoved,omitempty" json:"-"`
```
harshavardhana, line 454:
```suggestion
	ExpiryRuleRemoved *bool     `xml:"ExpiryRuleRemoved,omitempty" json:"-"`
```

Keep it pointer such that `omitempty` works

harshavardhana (changes requested):

pkg/lifecycle/lifecycle.go
shtripat: oh yes. will change

harshavardhana:
So what does this `ExpiryRuleRemoved` do ? is this to indicate that this ILM expiration must be replicated? what is it about?

shtripat:
> So what does this `ExpiryRuleRemoved` do ? is this to indicate that this ILM expiration must be replicated? what is it about?

Yes, this would be used for replication of ILM expiry rules removal to other sites. In this case when we remove the rules, we set current time at `expiryUpdatedAt` and using that this removal is replicated to other sites as other sites would have older/no value of `expiryUpdatedAt`

poornas (approved):

harshavardhana:
> Yes, this would be used for replication of ILM expiry rules removal to other sites. In this case when we remove the rules, we set current time at `expiryUpdatedAt` and using that this removal is replicated to other sites as other sites would have older/no value of `expiryUpdatedAt`

So this is a ILM configuration?

shtripat:
> > Yes, this would be used for replication of ILM expiry rules removal to other sites. In this case when we remove the rules, we set current time at `expiryUpdatedAt` and using that this removal is replicated to other sites as other sites would have older/no value of `expiryUpdatedAt`
> 
> So this is a ILM configuration?

yes

harshavardhana:
So this option is like enabling ILM site replication vs not?

shtripat:
> So this option is like enabling ILM site replication vs not?

Kind of. For enabling replication of ILM rule removal across sites in case of site replication enabled. If not this flag doesnt have any impact. Also this flag is not saved as part of ILM config, rather considered for setting `expiryUpdatedAt` field of ILM config. Because in case of ILM rules removal as well we want to update the `expiryUpdatedAt` field for the ILM config so that while replication of expiry rules, we know that there was a removal with latest time and other sites as well should remove the rules.

shtripat:
This PR wont be needed anymore with changes in `mc` as part of https://github.com/minio/mc/pull/4698/commits/6c94adcab7f8895800708776c4ae6bbbed077297

PR #1897, opened 2023-11-07
Addressing a few vulnerability issues

safaci2000:
Addressing the following current issue related to golang version.

govulncheck ./...  
Scanning your code and 163 packages across 14 dependent modules for known vulnerabilities...

=== Informational ===

Found 3 vulnerabilities in packages that you import, but there are no call
stacks leading to the use of these vulnerabilities. You may not need to
take any action. See https://pkg.go.dev/golang.org/x/vuln/cmd/govulncheck
for details.

Vulnerability #1: GO-2023-2102
    HTTP/2 rapid reset can cause excessive work in net/http
  More info: https://pkg.go.dev/vuln/GO-2023-2102
  Standard library
    Found in: net/http@go1.21
    Fixed in: net/http@go1.21.3

Vulnerability #2: GO-2023-2045
    Memory exhaustion in QUIC connection handling in crypto/tls
  More info: https://pkg.go.dev/vuln/GO-2023-2045
  Standard library
    Found in: crypto/tls@go1.21
    Fixed in: crypto/tls@go1.21.1

Vulnerability #3: GO-2023-2044
    Panic when processing post-handshake message on QUIC connections in
    crypto/tls
  More info: https://pkg.go.dev/vuln/GO-2023-2044
  Standard library
    Found in: crypto/tls@go1.21
    Fixed in: crypto/tls@go1.21.1

No vulnerabilities found.

Share feedback at https://go.dev/s/govulncheck-feedback.


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/minio/minio-go/v7
 
-go 1.17
+go 1.21.3
```
harshavardhana, line 3:
```suggestion
go 1.19
```

harshavardhana (changes requested):

go.mod
safaci2000: 1.19 suffers from the same vulnerabilities unfortunately.  I saw that you had some CICD to be backward compatible with 1.19 which might be a bigger change.  I'm happy to accept the changes but they won't solve the issue.

harshavardhana:
I don't think this PR doing anything, closing.

PR #1900, opened 2023-11-19
Add Expires to PutObject opts

vadmeste:
Add Expires field while uploading an object

### DIFF
--- a/.github/workflows/vulncheck.yml
+++ b/.github/workflows/vulncheck.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go-version: [ 1.21.x ]
+        go-version: [ 1.21.4 ]
     steps:
     - name: Check out code into the Go module directory
       uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


.github/workflows/vulncheck.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go-version: [ 1.21.0 ]
+        go-version: [ 1.21.x ]
```
harshavardhana, line 17:
```suggestion
        go-version: [ 1.21.4 ]
```

harshavardhana (changes requested):
Vulncheck must be set to exact version of Go

harshavardhana (approved):

MB175:
I tried using this option but its not working

MB175:
@klauspost  I just did it looks like the header is set correctly also the "Expires" field is set. Looks like the issue lies within my minio server which is still serving the file after expiration strange enough.

So no worries your changes work as expected, sorry for the disturbance

klauspost:
@MB175 The header has no functional meaning other than being stored on the object. It can be picked up by proxies, etc.

MinIO will do nothing more than store it and return whatever value it was set to. It is not a TTL. See [Object Lifecycle Management](https://min.io/docs/minio/linux/administration/object-management/object-lifecycle-management.html#object-expiration) on how to expire objects via lifecycle. This is similar to AWS S3.

PR #1907, opened 2023-11-29
lifecycle: Incl. empty Prefix when all else empty

krisis:
## Steps to reproduce
- ` mc ilm rule add --expire-days 1 --size-gt 512KiB  --debug myminio/bucket`
- This fails even when tested on AWS S3.

The bug here is that Filter's `MarshalXML` method includes an empty `Prefix` tag when we have a non-empty size based predicate present.


vadmeste (approved):
LGTM

harshavardhana (approved):

PR #1909, opened 2023-12-04
Added bucket throttle configuration struct

shtripat:


shtripat:
Not needed as suggested by @harshavardhana . Closing!

PR #1910, opened 2023-12-05
support IAM credentials to be fully cutomizable not just via ENVs

harshavardhana:
currently customizing the behavior of credentials.IAM for

- Container authorization style
- EC2 Metadata service credentials
- EKS credentials

Was via environment variables, allow this to be changed in a more declarative manner that can be used and remembered in application.


vadmeste (approved):
LGTM

PR #1911, opened 2023-12-11
add support for sso auth

jchorl:
This PR adds credential support for retrieving AWS SSO credentials and using those against a bucket store. I also added a test.

SSO is becoming increasingly common for [many benefits](https://aws.amazon.com/iam/identity-center/). I figured I'd try a PR to support it in MinIO. I can understand if this feels too coupled with a single provider, figured I'd try anyway.

This is not a full-fledged implementation, but it works and was enough to get me off the ground running. I'm sure there are edge-cases and configs that are not covered here. I noticed that AWS sdk is not vendored in this project, so I opted to write an implementation from scratch.


jchorl:
@harshavardhana is it an explicit goal of `minio` not to vendor in aws libraries? They'd be really useful here to avoid reimplementing the mechanics of SSO. I noticed some activity on https://github.com/minio/minio-go/issues/1930 and was willing to pick this back up.

harshavardhana:
> @harshavardhana is it an explicit goal of `minio` not to vendor in aws libraries? They'd be really useful here to avoid reimplementing the mechanics of SSO. I noticed some activity on #1930 and was willing to pick this back up.

aws-sdk-go is a bloatware we cannot really depend on that

pkg/credentials/file_aws_credentials.go
```diff
@@ -149,9 +296,17 @@ func loadProfile(filename, profile string) (*ini.Section, error) {
 	if err != nil {
 		return nil, err
 	}
+
 	iniProfile, err := config.GetSection(profile)
 	if err != nil {
-		return nil, err
+		// aws allows specifying the profile as [profile myprofile]
+		if strings.Contains(err.Error(), "does not exist") {
+			iniProfile, err = config.GetSection(fmt.Sprintf("profile %s", profile))
```
klauspost, line 304: nit: Maybe this shouldn't over-write the original error if it fails...

pkg/credentials/file_test.go
```diff
@@ -147,6 +152,67 @@ func TestFileAWS(t *testing.T) {
...
+func TestFileAWSSSO(t *testing.T) {
+	tmpDir, err := os.MkdirTemp("", "minio-sso-")
+	if err != nil {
+		t.Errorf("Creating temp dir: %+v", err)
+	}
+
+	// the file path is the sso-profile, "main", sha1-ed
+	os.WriteFile(
+		path.Join(tmpDir, "b28b7af69320201d1cf206ebf28373980add1451.json"),
+		[]byte(`{"startUrl": "https://testacct.awsapps.com/start", "region": "us-test-2", "accessToken": "my-access-token", "expiresAt": "2020-01-11T00:00:00Z"}`),
```
klauspost, line 164: Is the "startURL" used for a call?

Just making sure that we don't have tests making external calls.

harshavardhana:
Too old, closing since there is no PR activity

harshavardhana:
Too old, closing since there is no PR activity

PR #1925, opened 2024-01-18
fix: support more type to umnarshaJSON

jiuker:
support more type to umnarshaJSON
fix: https://github.com/minio/mc/issues/3786
or 
```log
ReadString: expects " or n, but found [, error found in #1 byte of ...|[true]|..., bigger context ...|[true]|...
```
`[true]` will unmarsha here.

### DIFF
--- a/pkg/set/stringset.go
+++ b/pkg/set/stringset.go
@@ -153,18 +153,18 @@ func (set StringSet) MarshalJSON() ([]byte, error) {
 // If 'data' contains JSON string, the set contains the string as one element.
 // If 'data' contains Other JSON types, JSON parse error is returned.
 func (set *StringSet) UnmarshalJSON(data []byte) error {
-	sl := []string{}
+	sl := []interface{}{}
 	var err error
 	if err = json.Unmarshal(data, &sl); err == nil {
 		*set = make(StringSet)
 		for _, s := range sl {
-			set.Add(s)
+			set.Add(fmt.Sprintf("%v", s))
 		}
 	} else {
-		var s string
+		var s interface{}
 		if err = json.Unmarshal(data, &s); err == nil {
 			*set = make(StringSet)
-			set.Add(s)
+			set.Add(fmt.Sprintf("%v", s))
 		}
 	}
 


shtripat (approved):
lgtm

harshavardhana (approved):

harshavardhana:
Can you add a test-case here @jiuker ?

pkg/set/stringset.go
```diff
@@ -149,22 +149,19 @@ func (set StringSet) MarshalJSON() ([]byte, error) {
...
-// If 'data' contains Other JSON types, JSON parse error is returned.
 func (set *StringSet) UnmarshalJSON(data []byte) error {
-	sl := []string{}
+	sl := []interface{}{}
 	var err error
 	if err = json.Unmarshal(data, &sl); err == nil {
 		*set = make(StringSet)
 		for _, s := range sl {
-			set.Add(s)
+			set.Add(fmt.Sprintf("%v", s))
```
harshavardhana, line 158: this is not correct here @jiuker we cannot make `[]string{}` into a single string

harshavardhana (changes requested):

harshavardhana (commented):
False alarm this will work fine.

harshavardhana (approved):

harshavardhana (approved):

pkg/policy/bucket-policy_test.go
```diff
@@ -210,6 +210,31 @@ func TestUnmarshalBucketPolicy(t *testing.T) {
...
+		"Action": [
+			"s3:PutObject"
+		],
+		"Resource": [
+			"arn:aws:s3:::mytest/*"
+		],
+		"Condition": {
+			"Null": {
+				"s3:x-amz-server-side-encryption": [
+					true
```
harshavardhana, line 233: can you add a test for `"true"` as per the original issue?

harshavardhana: Like these https://docs.aws.amazon.com/AmazonS3/latest/userguide/amazon-s3-policy-keys.html

```

{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Deny",
            "Principal": "*",
            "Action": "s3:PutObject",
            "Resource": [
                "arn:aws:s3:::DOC-EXAMPLE-BUCKET1",
                "arn:aws:s3:::DOC-EXAMPLE-BUCKET1/*"
            ],
            "Condition": {
                "NumericLessThan": {
                    "s3:TlsVersion": 1.2
                }
            }
        }
    ]
}					
```

harshavardhana (changes requested):

harshavardhana (approved):

PR #1939, opened 2024-02-22
Added valid allowed internal headers

shtripat:
Needed for replication of SSE-C objects to other sites while site replication.

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -512,6 +512,20 @@ func isAmzHeader(headerKey string) bool {
 	return strings.HasPrefix(key, "x-amz-meta-") || strings.HasPrefix(key, "x-amz-grant-") || key == "x-amz-acl" || isSSEHeader(headerKey) || strings.HasPrefix(key, "x-amz-checksum-")
 }
 
+var supportedInternalHeaders = map[string]bool{
+	"x-minio-internal-server-side-encryption-sealed-key":     true,
+	"x-minio-internal-inline-data":                           true,
+	"x-minio-internal-server-side-encryption-seal-algorithm": true,
+	"x-minio-internal-server-side-encryption-iv":             true,
+	// Add more supported headers here.
+	// Must be lower case.
+}
+
+// isValidInternalHeader returns true if header is one of valid internal headers
+func isValidInternalHeader(headerKey string) bool {
+	return supportedInternalHeaders[strings.ToLower(headerKey)]
+}
+
 // supportedQueryValues is a list of query strings that can be passed in when using GetObject.
 var supportedQueryValues = map[string]bool{
 	"attributes":                   true,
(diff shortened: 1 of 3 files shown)


utils.go
```diff
@@ -512,6 +512,20 @@ func isAmzHeader(headerKey string) bool {
...
 }
 
+var supportedInternalHeaders = map[string]bool{
+	"x-minio-internal-server-side-encryption-sealed-key":     true,
+	"x-minio-internal-inline-data":                           true,
+	"x-minio-internal-server-side-encryption-seal-algorithm": true,
+	"x-minio-internal-server-side-encryption-iv":             true,
+	// Add more supported headers here.
+	// Must be lower case.
+}
```
harshavardhana, line 522: We shouldn't mention it in this fashion, instead it must be `x-minio-replication-server-side-encryption-sealed-key` follow this for others as well. 

Convert this on the target to relevant internal values.

harshavardhana (changes requested):

utils.go
shtripat: ok, will do the changes

utils.go
```diff
@@ -512,6 +512,20 @@ func isAmzHeader(headerKey string) bool {
 	return strings.HasPrefix(key, "x-amz-meta-") || strings.HasPrefix(key, "x-amz-grant-") || key == "x-amz-acl" || isSSEHeader(headerKey) || strings.HasPrefix(key, "x-amz-checksum-")
 }
 
+var supportedReplicationHeaders = map[string]bool{
+	"x-minio-replication-server-side-encryption-sealed-key":     true,
+	"x-minio-replication-inline-data":                           true,
```
harshavardhana, line 517:
```suggestion
	"x-minio-replication-inline-data":                           true,
```

is this needed? @shtripat

harshavardhana (changes requested):

utils.go
shtripat: may not be. will try removing and testing.

shtripat: removed now as not needed.

utils.go
```diff
@@ -512,6 +512,19 @@ func isAmzHeader(headerKey string) bool {
...
+var supportedReplicationHeaders = map[string]bool{
+	"x-minio-replication-server-side-encryption-sealed-key":     true,
+	"x-minio-replication-server-side-encryption-seal-algorithm": true,
+	"x-minio-replication-server-side-encryption-iv":             true,
+	// Add more supported headers here.
+	// Must be lower case.
+}
+
+// isValidReplicationHeader returns true if header is one of valid internal headers
+func isValidReplicationHeader(headerKey string) bool {
```
harshavardhana, line 524:
```suggestion
func isValidReplicationEncryptionHeader(headerKey string) bool {
```

harshavardhana (approved):

PR #1945, opened 2024-03-01
Allow disabling dual-stack endpoints for Amazon S3

### LINKED ISSUE #1766: make endpoints configurable
Hello friends,

currently the minio-go client uses dualstack urls as default.

https://github.com/minio/minio-go/blob/f5f1604f764e0543943961efefac0e6dc56ee4b2/s3-endpoints.go#L36

Is that on purpose? Is it possible to disable this?
I didn't found a possiblility to get rid of the "dualstack" host part.

narqo:
Closes #1766

This PR adds a new configuration method, that enables users to switch off the use of dual-stack endpoints, when they use the client against Amazon S3.

I wanted to keep the name of the config aligned with how AWS SDKs and CLI call it, even though the default value is reversed. That is, AWS SDKs call it like `UseDualStack bool` or `withDualstackEnabled(bool)`.

### DIFF
--- a/s3-endpoints.go
+++ b/s3-endpoints.go
@@ -1,6 +1,6 @@
 /*
  * MinIO Go Library for Amazon S3 Compatible Cloud Storage
- * Copyright 2015-2017 MinIO, Inc.
+ * Copyright 2015-2024 MinIO, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


s3-endpoints.go
```diff
@@ -48,17 +50,19 @@ var awsS3EndpointMap = map[string]string{
...
-func getS3Endpoint(bucketLocation string) (s3Endpoint string) {
+func getS3Endpoint(bucketLocation string, useDualstack bool) (s3Endpoint string) {
 	s3Endpoint, ok := awsS3EndpointMap[bucketLocation]
 	if !ok {
 		// Default to 's3.dualstack.us-east-1.amazonaws.com' endpoint.
 		s3Endpoint = "s3.dualstack.us-east-1.amazonaws.com"
 	}
+	if !useDualstack {
+		// Trim 'dualstack' from the endpoint if disabled.
+		s3Endpoint = "s3." + s3Endpoint[len("s3.dualstack."):]
```
harshavardhana, line 65:
```suggestion
		s3Endpoint = "s3." + s3Endpoint[len("s3.dualstack."):]
```

Please have a separate map with other set of valid endpoints. Let's avoid string tricks.

harshavardhana (changes requested):

s3-endpoints.go
narqo: I'm concerned that having two separate maps will lead to human-errors in the future. So I updated the changes, keeping a single map, which holds both kinds of endpoints. PHAL

harshavardhana (approved):

narqo:
@harshavardhana thank you for the review. Is there anything left here, or how can I help to move this forward and land the fixes to main.

klauspost (approved):

PR #1946, opened 2024-03-05
Added `x-minio-replication-actual-object-size` to allowed headers

shtripat:
Need to send actual object size while replicating a multipart object as parts would be replicated as is with their encrypted content and content length. Later at target while creating xl.meta for the completed object, we need the actual object size be set as `X-Minio-Internal-actual-size`

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -517,6 +517,7 @@ var supportedReplicationEncryptionHeaders = map[string]bool{
 	"x-minio-replication-server-side-encryption-seal-algorithm": true,
 	"x-minio-replication-server-side-encryption-iv":             true,
 	"x-minio-replication-encrypted-multipart":                   true,
+	"x-minio-replication-object-size":                           true,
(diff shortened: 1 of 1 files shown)


utils.go
```diff
@@ -517,6 +517,7 @@ var supportedReplicationEncryptionHeaders = map[string]bool{
 	"x-minio-replication-server-side-encryption-seal-algorithm": true,
 	"x-minio-replication-server-side-encryption-iv":             true,
 	"x-minio-replication-encrypted-multipart":                   true,
+	"x-minio-replication-object-size":                           true,
```
harshavardhana, line 520:
```suggestion
	"x-minio-replication-actual-object-size":                           true,
```

harshavardhana (changes requested):

klauspost (approved):

harshavardhana (approved):

PR #1949, opened 2024-03-15
Add support Tencent Cloud COS vistual hosted style.

jorben:
Tencent Cloud COS does not support Path-style, so we need to use virtual-hosted-style.

refer to
[Tencent Cloud COS Regions and Access Endpoints](https://www.tencentcloud.com/document/product/436/6224)
[Tencent Cloud COS path-style 域名相关问题](https://cloud.tencent.com/document/faq/436/102489)


harshavardhana:
We are no longer accepting these endpoints in this fashion, we have to make room for other vendors if needed but via a different configuration. 

We cannot keep adding new functions in this manner. Thanks for the PR but we can't accept it in current form.

jswxstw:
> We cannot keep adding new functions in this manner. Thanks for the PR but we can't accept it in current form.

I have also encountered this issue where `BucketLookupAuto` policy cannot accurately find the appropriate domain style.
@harshavardhana Emmm, what form do you perfer to fix this issue?

harshavardhana:
https://github.com/minio/minio-go/pull/2057 check here how to fix this.

PR #1958, opened 2024-04-25
remove sha256-simd support we do not need it anymore

harshavardhana:
we have moved to go1.21 as the least Go version we would support 
for minio-go/v7 - and move to crypto/sha256


fwessels (approved):
LGTM

donatello (approved):

sergeevabc:
Does crypto/sha256 work on processors without SSE4 and AVX?

harshavardhana:
> Does crypto/sha256 work with processors without SSE4 and AVX?

This processor is from 2007, sure you can use it don't expect performance.

sergeevabc:
> This processor is from 2007, sure you can use it don't expect performance.

Performance is good, thank you. Keep it that way, because there are a [lot of places](https://en.wikipedia.org/wiki/Favela) outside your rich neighborhoods where people don't upgrade regularly, they have other priorities and opportunities.

PR #1959, opened 2024-04-26
Add support for DelMarkerExpiration element

krisis:
pkg/lifecycle:
- Add marshal/unmarshal support for DelMarkerExpiration action

pkg/notification:
- Add bucket event type for deletion of all versions of an object with DEL marker for current version "s3:LifecycleDelMarkerExpiration:Delete"

For https://github.com/minio/minio/pull/19481

### DIFF
--- a/pkg/notification/notification.go
+++ b/pkg/notification/notification.go
@@ -50,6 +50,7 @@ const (
 	ObjectRemovedAll                                   EventType = "s3:ObjectRemoved:*"
 	ObjectRemovedDelete                                EventType = "s3:ObjectRemoved:Delete"
 	ObjectRemovedDeleteMarkerCreated                   EventType = "s3:ObjectRemoved:DeleteMarkerCreated"
+	ILMDelMarkerExpirationDelete                       EventType = "s3:LifecycleDelMarkerExpiration:Delete"
 	ObjectReducedRedundancyLostObject                  EventType = "s3:ReducedRedundancyLostObject"
 	ObjectTransitionAll                                EventType = "s3:ObjectTransition:*"
 	ObjectTransitionFailed                             EventType = "s3:ObjectTransition:Failed"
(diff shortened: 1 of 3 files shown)


pkg/notification/notification.go
```diff
@@ -50,6 +50,7 @@ const (
 	ObjectRemovedAll                                   EventType = "s3:ObjectRemoved:*"
 	ObjectRemovedDelete                                EventType = "s3:ObjectRemoved:Delete"
 	ObjectRemovedDeleteMarkerCreated                   EventType = "s3:ObjectRemoved:DeleteMarkerCreated"
+	ILMDelMarkerExpirationDelete                       EventType = "s3:LifecycleDelMarkerExpiration:Delete"
```
vadmeste, line 53: It looks to me s3:ObjectRemoved:AllVersions is better

vadmeste (commented):
One comment

pkg/notification/notification.go
krisis: I wanted to move away from overloading `s3:ObjectRemoved:*` for ILM expiry related object removal to the notification event names used in [S3](https://docs.aws.amazon.com/AmazonS3/latest/userguide/notification-how-to-event-types-and-destinations.html). 
```
s3:LifecycleExpiration:*
s3:LifecycleExpiration:Delete
s3:LifecycleExpiration:DeleteMarkerCreated
```
I do see that this event name doesn't capture that all the versions of this object was expired. I could change this to `s3:LifecycleDelMarkerExpiration:DeleteAllVersions`. Would that work for you?

vadmeste: I think let's be consistent with the other regular ILM expiration events, I think s3:ObjectRemoved will be triggered instead in that case.

vadmeste (approved):
LGTM

harshavardhana (approved):

PR #1960, opened 2024-05-03
`x-amz-server-side-encryption-aws-kms-key-id` allowed header

shtripat:


shtripat:
How did I miss it, the entry already there in front of me ;)

PR #1961, opened 2024-05-15
Support custom request headers

olljanat:
First step to implement https://github.com/minio/minio/issues/19698 by allowing client to set `User-Agent` and `X-Forwarded-For` headers for all requests.

### DIFF
--- a/api.go
+++ b/api.go
@@ -100,6 +100,9 @@ type Client struct {
 	healthStatus int32
 
 	trailingHeaderSupport bool
+
+	// Custom headers to be used for all requests. Leave nil to use standard.
+	CustomRequestHeaders http.Header
 }
 
 // Options for New method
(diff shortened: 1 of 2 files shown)


olljanat:
Those are standard HTTP headers and MinIO rely on them in it's code. Blocking ability to use them here does not provide any extra security because any customized client can use what ever values they want on those.

That why there is this warning
> The aws:Referer, aws:SourceIp, and aws.UserAgent keys may be easily spoofed and therefore pose a potential security risk

in https://min.io/docs/minio/linux/administration/identity-access-management/policy-based-access-control.html#id6

However, if would be way to get those to MinIO backend when SFTP is used then it is possible to block those connection from certain IPs and example completely from SFTP clients to some buckets with policies like I listed in https://github.com/minio/minio/issues/19698

PR #1967, opened 2024-05-28
feat: support tags for postPolicy Upload

jiuker:
feat: support tags for postPolicy Upload
related: https://github.com/minio/minio/pull/19816

### DIFF
--- a/post-policy.go
+++ b/post-policy.go
@@ -19,12 +19,14 @@ package minio
 
 import (
 	"encoding/base64"
+	"errors"
 	"fmt"
 	"net/http"
 	"strings"
 	"time"
 
 	"github.com/minio/minio-go/v7/pkg/encrypt"
+	"github.com/minio/minio-go/v7/pkg/tags"
 )
 
 // expirationDateFormat date format for expiration key in json policy.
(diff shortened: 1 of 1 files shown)


shtripat (approved):
lgtm

post-policy.go
```diff
@@ -152,6 +152,23 @@ func (p *PostPolicy) SetCondition(matchType, condition, value string) error {
 	return errInvalidArgument("Invalid condition in policy")
 }
 
+// SetTagging - Sets tagging for the object for this policy based upload.
+func (p *PostPolicy) SetTagging(tagging string) error {
+	if strings.TrimSpace(tagging) == "" || tagging == "" {
```
vadmeste, line 157: `|| tagging == ""` seems to be redundant

vadmeste (commented):
I am not too much into adding a new API since we can have multiple operators here, other than `eq`

The user can use SetCondition() I believe (after fixing it)

post-policy.go
ramondeklein: I see that it's done more often in this file. I'm not a big fan of checking these trimmed values. It's not very likely that clients will issue a space or a newline, so the check is almost never necessary. And if they do specify it, then the check will fail on the back-end. I would prefer to remove all `string.TrimSpace(..)` in this file.

ramondeklein (approved):

jiuker:
> I am not too much into adding a new API since we can have multiple operators here, other than `eq`
> 
> The user can use SetCondition() I believe (after fixing it)

`SetCondition` will not support `tagging`. And according to the minio pr @ramondeklein mention the document introduced `tagging` is a single special field. So I want to add that into a function @vadmeste

post-policy.go
jiuker: I have no idea. @harshavardhana @klauspost  any advice?

harshavardhana (approved):

PR #1970, opened 2024-06-10
Keep all x-minio- headers as-is

klauspost:
Having a whitelist for minio headers is pointless. Remove so we don't have to keep these in sync.


harshavardhana (commented):
Shouldn't we make sure to remove x-minio-internal?

klauspost:
@harshavardhana They already get sent, just with a `x-amz-meta-` prefix. 

So the current implementation doesn't give any double safety. They should be filtered out elsewhere (and are AFAICT).

harshavardhana (approved):

Issue #1968, opened 2024-06-03
Sec vul CVE-2022-28948 from pkg/credentials > ini.v1 > stretchr/testify > yaml.v3


amaciejk:
We are using latest minio-go v7.0.70 and are seeing:
https://github.com/advisories/GHSA-hp87-p4gw-j4gq
https://nvd.nist.gov/vuln/detail/CVE-2022-28948

![image](https://github.com/minio/minio-go/assets/42215217/1d062ddf-4171-4015-a0c2-0ee03f22e164)

Dep tree is:
```
github.com/minio/minio-go/v7/pkg/credentials
gopkg.in/ini.v1
gopkg.in/ini.v1.test
github.com/stretchr/testify/assert
gopkg.in/yaml.v3
```

an updated go.sum shows:
```
github.com/minio/minio-go/v7 v7.0.70 h1:1u9NtMgfK1U42kUxcsl5v0yj6TEOPR497OAQxpJnn2g=
github.com/minio/minio-go/v7 v7.0.70/go.mod h1:4yBA8v80xGA30cfM3fz0DKYMXunWl/AV/6tWEs9ryzo=
github.com/stretchr/testify v1.7.0 h1:nwc3DEeHmmLAfoZucVR881uASk0Mfjw8xYJ99tb5CcY=
github.com/stretchr/testify v1.7.0/go.mod h1:6Fq8oRcR53rry900zMqJjRRixrwX3KX962/h/Wwjteg=
gopkg.in/ini.v1 v1.67.0 h1:Dgnx+6+nfE+IfzjUEISNeydPJh9AXNNsWbGP9KzCsOA=
gopkg.in/ini.v1 v1.67.0/go.mod h1:pNLf8WUiyNEtQjuu5G5vTm06TEv9tsIgeAvK8hOrP4k=
gopkg.in/yaml.v3 v3.0.0-20200313102051-9f266ea9e77c h1:dUUwHk2QECo/6vqA44rthZ8ie2QXMNeKRTHCNY2nXvo=
gopkg.in/yaml.v3 v3.0.0-20200313102051-9f266ea9e77c/go.mod h1:K4uyk7z7BCEPqu6E+C64Yfv1cQ7kz7rIZviUmN+EgEM=
```

2020 seems quite old compared to 2022 or latest even.

Looks like stretchr/testify have fixed in v1.8.4 or later: 
https://github.com/stretchr/testify/issues/1532
so perhaps it is ini.v1 that needs fixing?

[...]

PR #1977, opened 2024-06-25
upgrade all deps and replace gopkg.in for ini with go module

marktheunissen:
- Should fix https://github.com/minio/minio-go/issues/1968
- the `ini` package was still using `gopkg.in` directly in file `file_aws_credentials.go` so switched that to go modules
- updated the testify package


pkg/credentials/file_aws_credentials.go
```diff
@@ -26,7 +26,7 @@ import (
 	"strings"
 	"time"
 
-	ini "gopkg.in/ini.v1"
+	"github.com/go-ini/ini"
```
marktheunissen, line 29: @harshavardhana FYI, I think this might have used the `gopkg` syntax years ago, perhaps before `go mod` supported Github properly? I looked into it a bit but couldn't see why.

`github.com/go-ini/ini v1.67.0` in go.mod should be an equal replacement

harshavardhana (approved):

PR #1985, opened 2024-07-10
fix: GCS endpoint validation to allow ports

ammarhusain-sol:
fix: Fixed the endpoint validation issue for GCS if port number is added to the URL

### DIFF
--- a/pkg/s3utils/utils.go
+++ b/pkg/s3utils/utils.go
@@ -226,7 +226,7 @@ func IsGoogleEndpoint(endpointURL url.URL) bool {
 	if endpointURL == sentinelURL {
 		return false
 	}
-	return endpointURL.Host == "storage.googleapis.com"
+	return endpointURL.Hostname() == "storage.googleapis.com"
 }
 
 // Expects ascii encoded strings - from output of urlEncodePath
--- a/pkg/s3utils/utils_test.go
+++ b/pkg/s3utils/utils_test.go
@@ -284,6 +284,8 @@ func TestIsGoogleEndpoint(t *testing.T) {
 		// valid inputs.
 		{"http://storage.googleapis.com", true},
 		{"https://storage.googleapis.com", true},
+		{"http://storage.googleapis.com:80", true},
+		{"https://storage.googleapis.com:443", true},
 	}
 
 	for i, testCase := range testCases {
--- a/utils_test.go
+++ b/utils_test.go
@@ -114,6 +114,7 @@ func TestGetEndpointURL(t *testing.T) {
 		{"192.168.1.1:9000", false, "http://192.168.1.1:9000", nil, true},
 		{"192.168.1.1:9000", true, "https://192.168.1.1:9000", nil, true},
 		{"s3.amazonaws.com:443", true, "https://s3.amazonaws.com:443", nil, true},
+		{"storage.googleapis.com:443", true, "https://storage.googleapis.com:443", nil, true},
 		{"[::1]", false, "http://[::1]", nil, true},
 		{"[::1]", true, "https://[::1]", nil, true},
 		{"[::1]:80", false, "http://[::1]:80", nil, true},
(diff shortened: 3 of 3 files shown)


harshavardhana (changes requested):
why do you add a port number? for perfectly simple URL like 

```
storage.googleapis.com
```
?

ammarhusain-sol:
In the case of a URL with added port number, the validation check fails even though the endpoint is `storage.googleapis.com`

harshavardhana:
> In the case of a URL with added port number, the validation check fails even though the endpoint is `storage.googleapis.com`

yes but why add the port, `http` is port 80 and `https` is port 443, why specify it again?

ammarhusain-sol:
You are saying by default http and https requests are handled through 80 and 443 so why do we have to mention it explicitly that is what you are asking right?
The reason I have added this here is that restricting the users not to add port number in the request endpoint I thought of as a limitation so I added also I faced this problem only for the GCS endpoint alone not for the AWS S3 endpoint, that is another reason why I added it.

pkg/s3utils/utils.go
```diff
@@ -31,6 +31,7 @@ import (
 
 // Sentinel URL is the default url value which is invalid.
 var sentinelURL = url.URL{}
+var host string
```
harshavardhana, line 34: this is wrong - please remove

pkg/s3utils/utils.go
```diff
@@ -223,10 +224,11 @@ func IsAmazonPrivateLinkEndpoint(endpointURL url.URL) bool {
 
 // IsGoogleEndpoint - Match if it is exactly Google cloud storage endpoint.
 func IsGoogleEndpoint(endpointURL url.URL) bool {
+	host := HasPortNumber(endpointURL)
 	if endpointURL == sentinelURL {
 		return false
 	}
-	return endpointURL.Host == "storage.googleapis.com"
+	return host == "storage.googleapis.com"
```
harshavardhana, line 231:
```suggestion
	return endpointURL.Hostname() == "storage.googleapis.com"
```

pkg/s3utils/utils.go
```diff
@@ -409,3 +411,13 @@ func CheckValidObjectName(objectName string) error {
...
+
+//checks for port number
+func HasPortNumber(endpointURL url.URL) string {
+	if colonIndex := strings.LastIndex(endpointURL.Host, ":"); colonIndex != -1 {
+		host = endpointURL.Host[:colonIndex]
+	} else {
+		host = endpointURL.Host
+	}
+	return host
+}
```
harshavardhana, line 423: this is not needed

harshavardhana (changes requested):

harshavardhana:
> You are saying by default http and https requests are handled through 80 and 443 so why do we have to mention it explicitly that is what you are asking right?
> The reason I have added this here is that restricting the users not to add port number in the request endpoint I thought of as a limitation so I added also I faced this problem only for the GCS endpoint alone not for the AWS S3 endpoint, that is another reason why I added it.

I added review comments, we can take it in with the requested changes.

ammarhusain-sol:
@harshavardhana I made the changes.

harshavardhana (approved):

harshavardhana:
> @harshavardhana I made the changes.

Please run gofumpt and goimports

harshavardhana (approved):

go.mod
```diff
@@ -9,15 +9,18 @@ require (
...
 	github.com/klauspost/cpuid/v2 v2.2.6 // indirect
 	github.com/stretchr/testify v1.3.0 // indirect
-	golang.org/x/sys v0.18.0 // indirect
-	golang.org/x/text v0.14.0 // indirect
+	golang.org/x/mod v0.19.0 // indirect
+	golang.org/x/sync v0.7.0 // indirect
+	golang.org/x/sys v0.22.0 // indirect
+	golang.org/x/text v0.16.0 // indirect
+	golang.org/x/tools v0.23.0 // indirect
 )
```
harshavardhana, line 26: do not update `go mod` if you don't intend to, please revert these changes.

harshavardhana (changes requested):

go.mod
ammarhusain-sol: I reverted those changes sorry for the overhead.

harshavardhana (approved):

vadmeste (approved):
LGTM

klauspost:
Test is failing:
```
2024-07-11T18:04:06.2797859Z ##[error]    utils_test.go:136: Test 18: Expected to fail with <ERROR> "Google Cloud Storage endpoint should be 'storage.googleapis.com'.", but passed instead
2024-07-11T18:04:06.2808886Z --- FAIL: TestGetEndpointURL (0.00s)
```

ammarhusain-sol:
@klauspost found the issue in test18 it has been written that the GCP endpoint with a port number should fail, since we made that change it is not failing.
Thanks for sharing

harshavardhana:
FWIW I fixed it locally, waiting for the build will be merged.

Issue #1552, opened 2021-09-18
tests: Add functional tests for bucket tagging


vadmeste:
Add functional tests for bucket tagging, PUT/GET/DELETE.

romanchechyotkin:
Hello, this task is 3 years old, but i can do it, wanna start my opensource adventure. 
Is this still valid issue?

klauspost:
@romanchechyotkin Good first topic indeed! I don't see it added to the repo.

It should be added to [functional_tests.go](https://github.com/minio/minio-go/blob/14b3aa6d56708fdf2071346aadcdb085069d8937/functional_tests.go#L13663). Add an entry. When I do it, I just add it at the top and add a temporary "return". To start out maybe copy an existing like `testMakeBucketError` and modify that.

I then run it against a local minio install with a simple script. You can adapt that to your needs:
```
SET ACCESS_KEY=minio
SET SECRET_KEY=minio123
SET SERVER_ENDPOINT=127.0.0.1:9001
SET MINT_DATA_DIR=
SET MINT_MODE=full
SET ENABLE_HTTPS=false
SET ENABLE_KMS=true

go build functional_tests.go && functional_tests
```

romanchechyotkin:
#1986 here is my PR

harshavardhana:
Task completed vi #1986

PR #1986, opened 2024-07-16
#1552 add functional tests for bucket tagging

romanchechyotkin:


harshavardhana (approved):

klauspost:
LGTM, but just merge them into a single test. It seems reasonable to test them together. 

You can explore the API limitations if you'd like and check if we are being strict with the limits:

> The tag key is a case-sensitive string that can contain 1 to 128 Unicode characters.

> The tag value. The tag value is a required string.  [...] The tag value is a case-sensitive string that can contain from 0 to 256 Unicode characters.

I am not too sure we actually check these - which should be done [here](https://github.com/minio/minio-go/blob/aadf94623806731617fda2c8a8624028f157a154/pkg/tags/tags.go#L256)

Maybe also check Get and Delete before any tagging has been set.

klauspost:
@harshavardhana Quickest draw of the West :D

harshavardhana:
> LGTM, but just merge them into a single test. It seems reasonable to test them together.
> 
> You can explore the API limitations if you'd like and check if we are being strict with the limits:
> 
> > The tag key is a case-sensitive string that can contain 1 to 128 Unicode characters.
> 
> > The tag value. The tag value is a required string.  [...] The tag value is a case-sensitive string that can contain from 0 to 256 Unicode characters.
> 
> I am not too sure we actually check these - which should be done [here](https://github.com/minio/minio-go/blob/aadf94623806731617fda2c8a8624028f157a154/pkg/tags/tags.go#L256)
> 

These API limits are server driven, basically AWS S3 magically changes these things whenever there is AWS Reinvent in Vegas. Adding API checks on client side can break things easily.

This happened in the past multiple times.

A test to return errors from server is a good thing to add. You can explore adding  these things 

https://github.com/minio/mint/blob/efe113e7aed81ed6ceb200e5e644198ea06e0e27/build/versioning/tagging.go#L34

https://github.com/minio/mint/blob/efe113e7aed81ed6ceb200e5e644198ea06e0e27/run/core/aws-sdk-go/main.go#L539

This was the main reason we didn't bother to finish these tests in this SDK as it's already kind of tested for the server.

klauspost:
@harshavardhana This code is actually used in MinIO server as well.

harshavardhana:
> @harshavardhana This code is actually used in MinIO server as well.

Oh then we can definitely add.

PR #1989, opened 2024-08-06
Adjust functional tests for bucket cors, to detect NotImplemented response via logError

marktheunissen:
This is needed to get https://github.com/minio/minio/pull/20220 working. 

Slight adjustment to the function call so that we can detect NotImplemented

### DIFF
--- a/functional_tests.go
+++ b/functional_tests.go
@@ -166,7 +166,7 @@ func logError(testName, function string, args map[string]interface{}, startTime 
 	}
 }
 
-// log failed test runs
+// Log failed test runs, do not call this directly, use logError instead, as that correctly stops the test run
 func logFailure(testName, function string, args map[string]interface{}, startTime time.Time, alert, message string, err error) {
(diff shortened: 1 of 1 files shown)


harshavardhana (approved):

marktheunissen:
Tests will fail until https://github.com/minio/minio/pull/20220 is merged and a release is made

marktheunissen:
I've pushed up a bigger change replacing all my calls to `logFailure` with `logError`. I mistakenly thought that calling `logFailure` would fail the CI job when running in CI, however it's not the case. The correct function to call is `logError` as that will panic and cause the CI job to fail.

The net result of this PR is that the CI jobs are currently failing for this PR, until https://github.com/minio/minio/pull/20220 is merged.

marktheunissen:
Tests are passing!

klauspost (approved):

harshavardhana (approved):

PR #1992, opened 2024-08-16
Fix tests

shtripat:
As per https://github.com/minio/minio/pull/20181 now AbortMultipartUpload call would remove uploadID entry from cache, so dont expect listing to return.


harshavardhana:
is this still needed? @shtripat

shtripat:
> is this still needed? @shtripat 

I don't think it's needed anymore. Closing.

PR #1996, opened 2024-09-02
Adjust testPresignedPostPolicy to ensure that GetObject returns the correct checksum. Add checksum headers to policy

marktheunissen:


functional_tests.go
```diff
@@ -5495,13 +5497,6 @@ func testPresignedPostPolicy() {
 		return
 	}
 
-	// Save the data
-	_, err = c.PutObject(context.Background(), bucketName, objectName, bytes.NewReader(buf), int64(len(buf)), minio.PutObjectOptions{ContentType: "binary/octet-stream"})
-	if err != nil {
-		logError(testName, function, args, startTime, "", "PutObject failed", err)
-		return
-	}
-
```
marktheunissen, line 5504: I think this is just a copy paste mistake, not sure why we were doing a PutObject here, as we are running the PostPolicy presigned request via the http client further down.

post-policy.go
```diff
@@ -302,11 +302,31 @@ func (p *PostPolicy) SetUserMetadata(key, value string) error {
...
 
 // SetChecksum sets the checksum of the request.
-func (p *PostPolicy) SetChecksum(c Checksum) {
+func (p *PostPolicy) SetChecksum(c Checksum) error {
 	if c.IsSet() {
 		p.formData[amzChecksumAlgo] = c.Type.String()
 		p.formData[c.Type.Key()] = c.Encoded()
+
+		// Needed for S3 compatibility. MinIO ignores the checksum keys in the policy.
+		// https://github.com/minio/minio/blob/RELEASE.2024-08-29T01-40-52Z/cmd/postpolicyform.go#L60-L65
```
marktheunissen, line 311: As the comment mentions, this is needed for S3 compat, but I really am not sure why we ignore these keys in the MinIO implementation.

klauspost: Probably just a missing implementation, or it may have added later. Doesn't really matter. Should be trivial to add.

klauspost: (also , I don't know if AWS added that, but adding trailing checksums in forms should also be doable)

klauspost, line 311: @jiuker You added this in https://github.com/minio/minio/pull/18074

I don't understand why these should be ignored, which seems like a bug.

marktheunissen: I spoke to Jiuker and he agrees it is a bug, so I will remove the ignores in a PR to minio/minio

marktheunissen:
FYI to myself: https://github.com/minio/minio/pull/18074#discussion_r1772687735

marktheunissen:
Converted to draft temporarily while I complete the MinIO side.

retry.go
```diff
@@ -45,7 +45,7 @@ var DefaultRetryCap = time.Second
 
 // newRetryTimer creates a timer with exponentially increasing
 // delays until the maximum retry attempts are reached.
-func (c *Client) newRetryTimer(ctx context.Context, maxRetry int, unit, cap time.Duration, jitter float64) <-chan int {
+func (c *Client) newRetryTimer(ctx context.Context, maxRetry int, unit, maxSleep time.Duration, jitter float64) <-chan int {
```
marktheunissen, line 48: linter: don't redefine `cap` function

post-policy.go
```diff
@@ -226,27 +226,44 @@ func (p *PostPolicy) SetContentDisposition(contentDisposition string) error {
...
+	if err := p.addNewPolicy(policyCond); err != nil {
+		return err
+	}
+	p.formData["Content-Encoding"] = contentEncoding
+	return nil
+}
+
 // SetContentLengthRange - Set new min and max content length
 // condition for all incoming uploads.
-func (p *PostPolicy) SetContentLengthRange(min, max int64) error {
```
marktheunissen, line 231: linter: don't redefine `max` and `min` functions

functional_tests.go
```diff
@@ -5689,50 +5688,22 @@ func testPresignedPostPolicy() {
...
 	policy := minio.NewPostPolicy()
-
-	if err := policy.SetBucket(""); err == nil {
```
marktheunissen, line 5701: I moved these negative tests into post-policy_test.go, and then added some more test cases in that file.

functional_tests.go
```diff
@@ -5840,9 +5811,206 @@ func testPresignedPostPolicy() {
...
-	want := checksum.Encoded()
-	if got := res.Header.Get("X-Amz-Checksum-Crc32c"); got != want {
-		logError(testName, function, args, startTime, "", fmt.Sprintf("Want checksum %q, got %q", want, got), nil)
+	wantChecksumCrc32c := checksum.Encoded()
+	if got := res.Header.Get("X-Amz-Checksum-Crc32c"); got != wantChecksumCrc32c {
+		logError(testName, function, args, startTime, "", fmt.Sprintf("Want checksum %q, got %q", wantChecksumCrc32c, got), nil)
+		return
+	}
+
+	// Ensure that when we subsequently GetObject, the checksum is returned
```
marktheunissen, line 5820: This was the original user request - to have the checksum value returned when doing a GetObject, after doing a Post Policy. That functionality was already merged in MinIO weeks ago so this just adds the test for it.

functional_tests.go
```diff
@@ -5641,9 +5643,6 @@ func testPresignedPostPolicy() {
 		"policy": "",
 	}
 
-	// Seed random based on current time.
-	rand.Seed(time.Now().Unix())
```
marktheunissen, line 5645: Deprecated, we can remove this call from every test, I might make a follow up PR to do the rest.

marktheunissen:
@klauspost ready for a re-review when you get a moment :)

harshavardhana (approved):

klauspost (approved):

PR #1999, opened 2024-09-12
Add header for lifecycle config expiry to ignore replication

poornas:


vadmeste (commented):
Isn't bucket replication configuration a better place where we put this config ?

poornas:
> Isn't bucket replication configuration a better place where we put this config ?

this setting is for ILM to expire unreplicated versions or not - it belongs here

constants.go
```diff
@@ -127,4 +127,7 @@ const (
 	minioTgtReplicationReady = "X-Minio-Replication-Ready"
 	// Header asks if delete marker replication request can be sent by source now.
 	isMinioTgtReplicationReady = "X-Minio-Check-Replication-Ready"
+
+	// Header indicating if ilm expiry ignores replication status
```
harshavardhana, line 131: what does it mean by ignoring replication status? can you explain

pkg/lifecycle/lifecycle.go
```diff
@@ -496,8 +496,9 @@ type Rule struct {
 
 // Configuration is a collection of Rule objects.
 type Configuration struct {
-	XMLName xml.Name `xml:"LifecycleConfiguration,omitempty" json:"-"`
-	Rules   []Rule   `xml:"Rule"`
+	XMLName                 xml.Name `xml:"LifecycleConfiguration,omitempty" json:"-"`
+	Rules                   []Rule   `xml:"Rule"`
+	ExpiryIgnoreReplication string   `xml:"-,omitempty" json:"-"`
```
harshavardhana, line 501: is this name appropriate - how does `mc` change look like it?

harshavardhana (changes requested):

pkg/lifecycle/lifecycle.go
klauspost, line 501: > `xml:"-,omitempty"`

Dash implies ignore.  Maybe you meant `xml:",omitempty"`?

klauspost (commented):
I don't really understand. Why are we sending a config value as a header?

pkg/lifecycle/lifecycle.go
poornas: should have used `-`, the new field is a custom setting for minio and will break `get` calls for the lifecycle config with other sdk clients that are aws compliant. 

Further, was thinking about  site replication and  `mc admin cluster bucket export/import that are on older release - sending this as a header allows backward compliance because the header can be ignored

poornas: the setting is `mc ilm default <alias>/<bucket> --expire-ignore-replication on|off`

constants.go
harshavardhana, line 131: Header indicates if ILM expiration must trigger after replication

harshavardhana (approved):

PR #2007, opened 2024-10-10
Add `credentials.WithPolicy` method to narrow down policy

ramondeklein:
The `NewSTSWebIdentity` allows fetching credentials that are narrowed down to the specified policy. This is useful when a client requires less rights then it actually has. The returned credentials will have the intersection of the requested policy and the assigned policies.

The `credentials.NewSTSWebIdentity` function is updated to use the options pattern (backward compatible). By adding the `credentials.WithPolicy(...)` method the credentials can be scoped.

To allow using the default Kubernetes service account that is assigned to the pod, the `credentials.NewKubernetesIdentity`  function is added.

### DIFF
--- a/pkg/credentials/sts_web_identity.go
+++ b/pkg/credentials/sts_web_identity.go
@@ -25,6 +25,7 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"os"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


harshavardhana (approved):

pkg/credentials/sts_web_identity.go
```diff
@@ -85,29 +85,44 @@ type STSWebIdentity struct {
...
+	Policy string
+
 	// roleSessionName is the identifier for the assumed role session.
 	roleSessionName string
 }
 
 // NewSTSWebIdentity returns a pointer to a new
 // Credentials object wrapping the STSWebIdentity.
-func NewSTSWebIdentity(stsEndpoint string, getWebIDTokenExpiry func() (*WebIdentityToken, error)) (*Credentials, error) {
+func NewSTSWebIdentity(stsEndpoint string, getWebIDTokenExpiry func() (*WebIdentityToken, error), opts ...func(*STSWebIdentity)) (*Credentials, error) {
```
harshavardhana, line 97:
```suggestion
func NewSTSWebIdentity(stsEndpoint string, getWebIDTokenExpiry func() (*WebIdentityToken, error), opts ...func(*STSWebIdentity)) (*Credentials, error) {
```

Can we not change the existing function signature? instead create a new one as 

```
NewSTSWebIdentityWithOpts
```

?

harshavardhana (changes requested):

pkg/credentials/sts_web_identity.go
ramondeklein: What's the point of doing that? I guess this won't break existing code (I did have a NewSTSWebIdentityWithPolicy` before, but I though this would be cleaner.

harshavardhana (approved):

PR #2013, opened 2024-10-23
retry: make max retries configurable

ashwanthgoli:
Adds an option to configure the max retries. This would come in handy for users that want to control retries at a higher level.

### DIFF
--- a/api.go
+++ b/api.go
@@ -283,7 +283,10 @@ func privateNew(endpoint string, opts *Options) (*Client, error) {
 	// healthcheck is not initialized
 	clnt.healthStatus = unknown
 
-	clnt.maxRetries = opts.MaxRetries
+	clnt.maxRetries = MaxRetry
+	if opts.MaxRetries > 0 {
+		clnt.maxRetries = opts.MaxRetries
+	}
 
 	// Return.
 	return clnt, nil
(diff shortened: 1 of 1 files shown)


klauspost (approved):

api.go
```diff
@@ -278,6 +283,8 @@ func privateNew(endpoint string, opts *Options) (*Client, error) {
 	// healthcheck is not initialized
 	clnt.healthStatus = unknown
 
+	clnt.maxRetries = opts.MaxRetries
```
harshavardhana, line 286: If MaxRetries is set to 0 set it to 3 to retain the current behavior, always default to MaxRetry global value

harshavardhana (changes requested):

api.go
ashwanthgoli: i was checking it [here](https://github.com/minio/minio-go/pull/2013/files#diff-f7bb71ca856c07b3329265f26b1e577e00b71e4792550cf14346bdbe0e7bd18dR604) before, but your suggestion is cleaner. updated in https://github.com/minio/minio-go/pull/2013/commits/fef61933ab1f75afb9b1b02a6d58ab1912fc4763

ashwanthgoli:
@klauspost @harshavardhana would really appreciate if you could take another look and merge it 🙇

harshavardhana (approved):

PR #2014, opened 2024-10-28
Add support for AllVersionsExpiration ilm rule

dhananjaykrutika:
This is a minio-only rule for removing all versions of an object upon expiration. The rule takes two parameters: Days and DeleteMarker.
"Days" indicates the number of days after which the object and all its versions must be expired.
The optional "DeleteMarker" flag indicates that the expiration be applied if the latest version of this object is a delete marker.

These specific changes in minio-go are for consumption by ec in a different PR.

### DIFF
--- a/pkg/lifecycle/lifecycle.go
+++ b/pkg/lifecycle/lifecycle.go
@@ -441,21 +441,18 @@ type AllVersionsExpiration struct {
 	DeleteMarker ExpireDeleteMarker `xml:"DeleteMarker,omitempty" json:"DeleteMarker,omitempty"`
 }
 
-// IsDaysNull returns true if days field is null
-func (e AllVersionsExpiration) IsDaysNull() bool {
+// IsNull returns true if days field is 0
+func (e AllVersionsExpiration) IsNull() bool {
 	return e.Days == 0
 }
 
-func (e AllVersionsExpiration) IsNull() bool {
-	return e.IsDaysNull()
-}
-
-func (ave AllVersionsExpiration) MarshalXML(enc *xml.Encoder, start xml.StartElement) error {
-	if ave.IsNull() {
+// MarshalXML satisfies xml.Marshaler to provide custom encoding
+func (e AllVersionsExpiration) MarshalXML(enc *xml.Encoder, start xml.StartElement) error {
+	if e.IsNull() {
 		return nil
 	}
 	type allVersionsExp AllVersionsExpiration
-	return enc.EncodeElement(allVersionsExp(ave), start)
+	return enc.EncodeElement(allVersionsExp(e), start)
 }
 
 // MarshalJSON customizes json encoding by omitting empty values
--- a/pkg/lifecycle/lifecycle.go
+++ b/pkg/lifecycle/lifecycle.go
@@ -434,12 +434,37 @@ func (de DelMarkerExpiration) MarshalXML(enc *xml.Encoder, start xml.StartElemen
 	return enc.EncodeElement(delMarkerExp(de), start)
 }
 
+// AllVersionsExpiration represents AllVersionsExpiration actions element in an ILM policy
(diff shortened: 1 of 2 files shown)


pkg/lifecycle/lifecycle.go
```diff
@@ -434,12 +434,37 @@ func (de DelMarkerExpiration) MarshalXML(enc *xml.Encoder, start xml.StartElemen
...
+	Days         int                `xml:"Days,omitempty" json:"Days,omitempty"`
+	DeleteMarker ExpireDeleteMarker `xml:"DeleteMarker,omitempty" json:"DeleteMarker,omitempty"`
+}
+
+// IsDaysNull returns true if days field is null
+func (e AllVersionsExpiration) IsDaysNull() bool {
+	return e.Days == 0
+}
+
+func (e AllVersionsExpiration) IsNull() bool {
```
klauspost, line 449: Do we need both?

pkg/lifecycle/lifecycle.go
```diff
@@ -434,12 +434,37 @@ func (de DelMarkerExpiration) MarshalXML(enc *xml.Encoder, start xml.StartElemen
...
 }
 
+// AllVersionsExpiration represents AllVersionsExpiration actions element in an ILM policy
+type AllVersionsExpiration struct {
+	XMLName      xml.Name           `xml:"AllVersionsExpiration" json:"-"`
+	Days         int                `xml:"Days,omitempty" json:"Days,omitempty"`
+	DeleteMarker ExpireDeleteMarker `xml:"DeleteMarker,omitempty" json:"DeleteMarker,omitempty"`
+}
+
+// IsDaysNull returns true if days field is null
```
klauspost, line 444:
```suggestion
// IsDaysNull returns true if days field is 0.
```

pkg/lifecycle/lifecycle.go
```diff
@@ -434,12 +434,37 @@ func (de DelMarkerExpiration) MarshalXML(enc *xml.Encoder, start xml.StartElemen
...
+func (e AllVersionsExpiration) IsDaysNull() bool {
+	return e.Days == 0
+}
+
+func (e AllVersionsExpiration) IsNull() bool {
+	return e.IsDaysNull()
+}
+
+func (ave AllVersionsExpiration) MarshalXML(enc *xml.Encoder, start xml.StartElement) error {
```
klauspost, line 453:
```suggestion
// MarshalXML satisfies xml.Marshaler to provide custom encoding.
func (e AllVersionsExpiration) MarshalXML(enc *xml.Encoder, start xml.StartElement) error {
```
ave -> e below.

pkg/lifecycle/lifecycle.go
klauspost, line 449:
```suggestion
// IsNull returns whether the expiration is enabled.
func (e AllVersionsExpiration) IsNull() bool {
```

pkg/lifecycle/lifecycle.go
```diff
@@ -434,12 +434,37 @@ func (de DelMarkerExpiration) MarshalXML(enc *xml.Encoder, start xml.StartElemen
...
+	if ave.IsNull() {
+		return nil
+	}
+	type allVersionsExp AllVersionsExpiration
+	return enc.EncodeElement(allVersionsExp(ave), start)
```
klauspost, line 458: Why is this cast needed?

klauspost (commented):
A few minor things

pkg/lifecycle/lifecycle.go
klauspost: Ah, to avoid the recursive call 👍🏼

klauspost (approved):
LGTM

harshavardhana (approved):

PR #2015, opened 2024-11-01
PromptObject API support

sidharthrajaram:
### promptObject API support

`PromptObject` is the function.

`PromptObjectOptions` can (optionally) contain the target `LambdaArn` and any `PromptArgs` for inference (similar to `kwargs` in the Python SDK for any arguments like `temperature`, `response_format`, `max_tokens`, etc.).`

### DIFF
--- a/api-prompt-object.go
+++ b/api-prompt-object.go
@@ -1,0 +1,78 @@
+/*
+ * MinIO Go Library for Amazon S3 Compatible Cloud Storage
+ * Copyright 2015-2024 MinIO, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package minio
+
+import (
+	"bytes"
+	"context"
+	"io"
+	"net/http"
+
+	"github.com/goccy/go-json"
+	"github.com/minio/minio-go/v7/pkg/s3utils"
+)
+
+// PromptObject performs language model inference with the prompt and referenced object as context.
+// Inference is performed using a Lambda handler that can process the prompt and object.
+// Currently, this functionality is limited to certain MinIO servers.
+func (c *Client) PromptObject(ctx context.Context, bucketName, objectName, prompt string, opts PromptObjectOptions) (io.ReadCloser, error) {
+	// Input validation.
+	if err := s3utils.CheckValidBucketName(bucketName); err != nil {
+		return nil, ErrorResponse{
+			StatusCode: http.StatusBadRequest,
+			Code:       "InvalidBucketName",
+			Message:    err.Error(),
+		}
+	}
+	if err := s3utils.CheckValidObjectName(objectName); err != nil {
+		return nil, ErrorResponse{
+			StatusCode: http.StatusBadRequest,
+			Code:       "XMinioInvalidObjectName",
+			Message:    err.Error(),
+		}
+	}
+
+	opts.AddLambdaArnToReqParams(opts.LambdaArn)
+	opts.SetHeader("Content-Type", "application/json")
+	opts.AddPromptArg("prompt", prompt)
+	promptReqBytes, err := json.Marshal(opts.PromptArgs)
+	if err != nil {
+		return nil, err
+	}
+
+	// Execute POST on bucket/object.
+	resp, err := c.executeMethod(ctx, http.MethodPost, requestMetadata{
+		bucketName:       bucketName,
+		objectName:       objectName,
+		queryValues:      opts.toQueryValues(),
+		customHeader:     opts.Header(),
+		contentSHA256Hex: sum256Hex(promptReqBytes),
+		contentBody:      bytes.NewReader(promptReqBytes),
+		contentLength:    int64(len(promptReqBytes)),
+	})
+	if err != nil {
+		return nil, err
+	}
+	if resp != nil {
+		if resp.StatusCode != http.StatusOK {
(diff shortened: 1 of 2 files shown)


api-prompt-object.go
```diff
@@ -0,0 +1,84 @@
...
+package minio
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"github.com/goccy/go-json"
+	"github.com/minio/minio-go/v7/pkg/s3utils"
+	"io"
+	"net/http"
```
harshavardhana, line 27:
```suggestion
	"bytes"
	"context"
	"fmt"
	"io"
	"net/http"
	
	"github.com/goccy/go-json"
	"github.com/minio/minio-go/v7/pkg/s3utils"
```

api-prompt-object.go
```diff
@@ -0,0 +1,84 @@
...
+
+	// Read the response body
+	bodyBytes, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return "", err
+	}
+
+	// Convert to string and return
+	responseString := string(bodyBytes)
+	return responseString, nil
```
harshavardhana, line 86: return the response body entirely, caller would interpret the results, instead of returning a string.

api-prompt-object.go
```diff
@@ -0,0 +1,84 @@
...
+	"context"
+	"fmt"
+	"github.com/goccy/go-json"
+	"github.com/minio/minio-go/v7/pkg/s3utils"
+	"io"
+	"net/http"
+)
+
+// PromptObject wrapper function that accepts a request context
+func (c *Client) PromptObject(ctx context.Context, bucketName, objectName string, prompt string, opts PromptObjectOptions) (string, error) {
```
harshavardhana, line 31:
```suggestion
func (c *Client) PromptObject(ctx context.Context, bucketName, objectName string, prompt string, opts PromptObjectOptions) (io.ReadCloser, error) {
```

api-prompt-object.go
```diff
@@ -0,0 +1,84 @@
...
+		queryValues:      opts.toQueryValues(),
+		customHeader:     opts.Header(),
+		contentSHA256Hex: sum256Hex(promptReqBytes),
+		contentBody:      bytes.NewReader(promptReqBytes),
+		contentLength:    int64(len(promptReqBytes)),
+	})
+	if err != nil {
+		return "", err
+	}
+	// Return string that is returned as HttpResponse
```
harshavardhana, line 69: Check for `resp.StatusCode != http.StatusOK` and return an error.

harshavardhana (changes requested):

api-prompt-object.go
sidharthrajaram: Got it. Responsibility for closing the body is passed on to caller.

api-prompt-object.go
```diff
@@ -0,0 +1,76 @@
...
+	"bytes"
+	"context"
+	"io"
+	"net/http"
+
+	"github.com/goccy/go-json"
+	"github.com/minio/minio-go/v7/pkg/s3utils"
+)
+
+// PromptObject wrapper function that accepts a request context
```
klauspost, line 30:
```suggestion
// PromptObject performs a lambda LLM prompt with the referenced object as context.
// Currently this functionality is limited to certain MinIO servers.
```

api-prompt-options.go
```diff
@@ -0,0 +1,110 @@
...
+		return
+	}
+	if o.reqParams == nil {
+		o.reqParams = make(url.Values)
+	}
+	o.reqParams.Set(key, value)
+}
+
+// AddReqParam - add request query string parameter
+// supported key: see supportedQueryValues and allowedCustomQueryPrefix.
```
klauspost, line 85: Not really nice if not exported....

api-prompt-options.go
```diff
@@ -0,0 +1,110 @@
...
+func (o PromptObjectOptions) Header() http.Header {
+	headers := make(http.Header, len(o.headers))
+	for k, v := range o.headers {
+		headers.Set(k, v)
+	}
+	return headers
+}
+
+// AddKwarg Add a key value pair to the kwargs where the key is a string but the value can be any type.
+func (o *PromptObjectOptions) AddKwarg(key string, value interface{}) {
```
klauspost, line 45:
```suggestion
func (o *PromptObjectOptions) AddKwarg(key string, value any) {
```

api-prompt-object.go
```diff
@@ -0,0 +1,76 @@
...
+		return nil, ErrorResponse{
+			StatusCode: http.StatusBadRequest,
+			Code:       "XMinioInvalidObjectName",
+			Message:    err.Error(),
+		}
+	}
+
+	opts.AddLambdaArnToReqParams(opts.LambdaArn)
+	opts.SetHeader("Content-Type", "application/json")
+	opts.AddKwarg("prompt", prompt)
```
klauspost, line 50: While it makes sense to add to "KwArgs", there is a good chance of this colliding with a custom value.

Either make `"prompt"`  less likely to occur or note on KwArgs that "prompt" is reserved.

klauspost (commented):
Mostly trying to get stuff ironed out before the API is fixed.

api-prompt-options.go
harshavardhana: I think he copy pasted, we might have to fix it else where as well and document it properly starting with here..

harshavardhana (changes requested):

api-prompt-object.go
```diff
@@ -0,0 +1,76 @@
...
+	"context"
+	"io"
+	"net/http"
+
+	"github.com/goccy/go-json"
+	"github.com/minio/minio-go/v7/pkg/s3utils"
+)
+
+// PromptObject wrapper function that accepts a request context
+func (c *Client) PromptObject(ctx context.Context, bucketName, objectName string, prompt string, opts PromptObjectOptions) (io.ReadCloser, error) {
```
fwessels, line 31: `bucketName, objectName string, prompt string,` --> `bucketName, objectName, prompt string`

api-prompt-options.go
sidharthrajaram: Yes, based on GetObjectOptions. 🙂

sidharthrajaram: Removing AddReqParam and SetReqParam to keep things minimal for now. We will address the unexported supportedQueryValues issue in a separate change.

harshavardhana (approved):

PR #2019, opened 2024-11-16
fix: get ETag from resp.Body rather than resp.Header

xiangcheng09:
When I use the CopyObject function with AWS S3, I find the ETag in the response is always an empty string. 
According to AWS's latest API Reference[^1], the variable ETag is in the CopyObjectResult part, not in the header part. 
AWS

Therefore, we should modify the source code to get the ETag from cpObjRes rather than resp.Header.Get("ETag").
I did so and got the correct ETag value, which is not an empty string.

I hope you can take this into consideration. If you have any questions, please feel free to ask.

**Reference:**
[^1]:https://docs.aws.amazon.com/AmazonS3/latest/API/API_CopyObject.html

### DIFF
--- a/api-copy-object.go
+++ b/api-copy-object.go
@@ -68,7 +68,7 @@ func (c *Client) CopyObject(ctx context.Context, dst CopyDestOptions, src CopySr
 		Bucket:           dst.Bucket,
 		Key:              dst.Object,
 		LastModified:     cpObjRes.LastModified,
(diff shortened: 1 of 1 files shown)


klauspost (approved):
LGTM. I will send a fix for the linter issues (unrelated to your PR)

vadmeste (approved):

xiangcheng09:
Thank you all. This is very exciting for a beginner. ♥

Issue #2016, opened 2024-11-08
Allow redirection with configurable parameter 


shivakumargowda12:
Its more of a suggestion than any issue/improvement.

Handling HTTP Redirect error is an important feature and widely used.
We clearly understand the intention to remove redirect handling support completely - https://github.com/minio/minio/issues/13918.

Although, it fixed the issue faced during AWS S3 resource access across regions.
Minio Client SDK is being widely used for other Object Store vendors as well, which is not completely dependent on Regions.

My suggestion here is, why cant we make **Redirection Handling** as an optional configurable feature in minio-go (**Setting Disabled by Default**), so that whoever wants to make use of redirection they can continue using that feature without any hassle.

So, the changes needed would be simple
 - Reverting back the changeset - https://github.com/minio/minio-go/pull/1601
 - Add a configuration setting in minio to enable/disable redirect handling. (Disabled by default)
 - Add a check at redirection flow and proceed based on the config parameter

harshavardhana:
Redirection support is not part of S3 API and is out of scope for this SDK. Let me know if this supported in AWS SDK if yes we can bring it back otherwise sorry.

PR #2021, opened 2024-11-19
Enable optional Redirect Handling Support

shivakumargowda12:
**Handle Issue** - https://github.com/minio/minio-go/issues/2016

**References**:
  - https://github.com/minio/minio-go/pull/1601
  - https://github.com/minio/minio/issues/13918

### DIFF
--- a/api.go
+++ b/api.go
@@ -203,6 +203,67 @@ func (r *lockedRandSource) Seed(seed int64) {
 	r.lk.Unlock()
 }
 
+// Redirect requests by re signing the request.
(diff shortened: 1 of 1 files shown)


klauspost (changes requested):
I don't want this controlled by an env var, since it triggers re-signing to a different host, it would make requests be signed to a remote controlled host. 

While the security aspect is very minimal, it should be determined by the caller application if that risk is acceptable and not be accepted based purely external factors.

harshavardhana:
Redirect is not supported in S3 API, it simply doesn't work correctly.

PR #2026, opened 2024-12-03
Add Full Object Checksum API

klauspost:
Add support for full object checksums as described here:

https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity.html

To enable, use `ChecksumCRC64NVME`, `ChecksumFullObjectCRC32` or `ChecksumFullObjectCRC32C` as checksum type when uploading.

Mint tests updated, but can be disabled with `MINT_NO_FULL_OBJECT=anything` env var.

PR will fail against community MinIO without above env var.

Validated against AWS S3.

### DIFF
--- a/functional_tests.go
+++ b/functional_tests.go
@@ -2006,9 +2006,13 @@ func testPutObjectWithChecksums() {
 		{cs: minio.ChecksumCRC32},
 		{cs: minio.ChecksumSHA1},
 		{cs: minio.ChecksumSHA256},
+		{cs: minio.ChecksumCRC64NVME},
 	}
 
 	for _, test := range tests {
+		if os.Getenv("MINT_NO_FULL_OBJECT") != "" && test.cs.FullObjectRequested() {
+			continue
+		}
 		bufSize := dataFileMap["datafile-10-kB"]
 
 		// Save the data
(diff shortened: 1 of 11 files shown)


utils.go
```diff
@@ -698,3 +699,146 @@ func (h *hashReaderWrapper) Read(p []byte) (n int, err error) {
 	}
 	return n, err
 }
+
+// Followis is ported from C to Go in 2016 by Justin Ruggles, with minimal alteration.
```
fwessels, line 703: Followis --> Following

functional_tests.go
```diff
@@ -2373,15 +2391,26 @@ func testPutMultipartObjectWithChecksums(trailing bool) {
...
 
 	for _, test := range tests {
-		bufSize := dataFileMap["datafile-129-MB"]
+		if os.Getenv("MINT_NO_FULL_OBJECT") != "" && test.cs.FullObjectRequested() {
+			continue
+		}
 
+		args["section"] = "prep"
+		bufSize := dataFileMap["datafile-129-MB"]
+		if false && test.cs.Is(minio.ChecksumFullObjectCRC32) {
```
fwessels, line 2410: `if false &&` -- left over ?

harshavardhana (approved):

fwessels (approved):
LGTM

PR #2041, opened 2024-12-26
Use proper HTTP client for fetching credentials

### LINKED ISSUE #2040: Custom HTTP transport is not used for STS
When `MinioClient` is using a custom HTTP client and STS, then the custom HTTP client isn't used for performing the STS request. It will always revert to the default HTTP transport when using `credentials.NewSTSAssumeRole` ([source](https://github.com/minio/minio-go/blob/master/pkg/credentials/assume_role.go#L119)).

Instead of using `credentials.NewSTSAssumeRole`, the provider can be created by passing the same HTTP client explicitly, like this:
```go
tr := customHTTPTransport()
stsCreds := credentials.New(&credentials.STSAssumeRole{
	Client:      &http.Client{Transport: tr},
	STSEndpoint: stsEndPoint,
	Options: credentials.STSAssumeRoleOptions{/*...*/},
})
client, err := minio.New(endPoint, &minio.Options{
	Creds:     stsCreds,
	Transport: tr,
})
```

Although a workaround is possible, this behavior is non-intuitive. Most people would expect the MinIO client to use the specified HTTP transport for all requests that are made on behalf of `MinioClient`.

ramondeklein:
This PR ensures that the HTTP client that is set on the `MinioClient` will also be used when the client needs to fetch credentials via an HTTP request.

- Instead of passing the HTTP client through all layers, I have choosen to use a `CredContext` structure that abstracts the HTTP client and may also be used to add more context (when needed).
- The `STSCertificateIdentity` method passes certificates via the `TLSClientConfig`, so in this case the HTTP transport is cloned and it sets the certificate explicitly. The original implementation did set additional `http.Transport` fields, but this has been removed.
- The `credentials.(*Credentials).Get()` function is deprecated and `Provider.GetWithCredContext()` should be used instead. It's still there to provide backward compatibility.
- When this PR is merged, then we should also update `madmin` to also pass the custom HTTP client.

Fixes #2040.

### DIFF
--- a/api.go
+++ b/api.go
@@ -600,9 +600,9 @@ func (c *Client) executeMethod(ctx context.Context, method string, metadata requ
 		return nil, errors.New(c.endpointURL.String() + " is offline.")
 	}
 
-	var retryable bool          // Indicates if request can be retried.
-	var bodySeeker io.Seeker    // Extracted seeker from io.Reader.
-	var reqRetry = c.maxRetries // Indicates how many times we can retry the request
+	var retryable bool       // Indicates if request can be retried.
+	var bodySeeker io.Seeker // Extracted seeker from io.Reader.
+	reqRetry := c.maxRetries // Indicates how many times we can retry the request
 
 	if metadata.contentBody != nil {
 		// Check if body is seekable then it is retryable.
@@ -808,7 +808,7 @@ func (c *Client) newRequest(ctx context.Context, method string, metadata request
 	}
 
 	// Get credentials from the configured credentials provider.
-	value, err := c.credsProvider.Get()
+	value, err := c.credsProvider.GetWithContext(c.CredContext())
 	if err != nil {
 		return nil, err
 	}
@@ -1017,4 +1017,15 @@ func (c *Client) isVirtualHostStyleRequest(url url.URL, bucketName string) bool 
 	// default to virtual only for Amazon/Google  storage. In all other cases use
 	// path style requests
 	return s3utils.IsVirtualHostSupported(url, bucketName)
+}
+
+// CredContext returns the context for fetching credentials
+func (c *Client) CredContext() *credentials.CredContext {
+	httpClient := c.httpClient
+	if httpClient == nil {
+		httpClient = http.DefaultClient
+	}
+	return &credentials.CredContext{
+		Client: httpClient,
+	}
 }
--- a/pkg/credentials/sts_tls_identity.go
+++ b/pkg/credentials/sts_tls_identity.go
@@ -20,8 +20,8 @@ import (
 	"crypto/tls"
 	"encoding/xml"
 	"errors"
+	"fmt"
 	"io"
-	"net"
 	"net/http"
 	"net/url"
 	"strconv"
@@ -32,12 +32,6 @@ import (
 // parameter - e.g. a custom HTTP transport configuration or S3 credental
 // livetime.
 type CertificateIdentityOption func(*STSCertificateIdentity)
-
-// CertificateIdentityWithTransport returns a CertificateIdentityOption that
-// customizes the STSCertificateIdentity with the given http.RoundTripper.
-func CertificateIdentityWithTransport(t http.RoundTripper) CertificateIdentityOption {
-	return CertificateIdentityOption(func(i *STSCertificateIdentity) { i.Client.Transport = t })
-}
 
 // CertificateIdentityWithExpiry returns a CertificateIdentityOption that
 // customizes the STSCertificateIdentity with the given livetime.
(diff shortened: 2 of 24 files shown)


api.go
```diff
@@ -600,9 +600,9 @@ func (c *Client) executeMethod(ctx context.Context, method string, metadata requ
 		return nil, errors.New(c.endpointURL.String() + " is offline.")
 	}
 
-	var retryable bool          // Indicates if request can be retried.
-	var bodySeeker io.Seeker    // Extracted seeker from io.Reader.
-	var reqRetry = c.maxRetries // Indicates how many times we can retry the request
+	var retryable bool       // Indicates if request can be retried.
+	var bodySeeker io.Seeker // Extracted seeker from io.Reader.
+	reqRetry := c.maxRetries // Indicates how many times we can retry the request
```
ramondeklein, line 605: This was changed by my IDE to follow coding guidelines.

pkg/credentials/sts_tls_identity.go
```diff
@@ -33,12 +33,6 @@ import (
 // livetime.
 type CertificateIdentityOption func(*STSCertificateIdentity)
 
-// CertificateIdentityWithTransport returns a CertificateIdentityOption that
-// customizes the STSCertificateIdentity with the given http.RoundTripper.
-func CertificateIdentityWithTransport(t http.RoundTripper) CertificateIdentityOption {
-	return CertificateIdentityOption(func(i *STSCertificateIdentity) { i.Client.Transport = t })
-}
-
```
ramondeklein, line 41: Specifying a custom transport doesn't make much sense. If we do want to be able to use custom transports for STS, then I need to make a generic fix for all STS options.

pkg/credentials/sts_tls_identity.go
```diff
@@ -145,7 +115,19 @@ func (i *STSCertificateIdentity) Retrieve() (Value, error) {
...
 	req.Form.Add("DurationSeconds", strconv.FormatUint(uint64(livetime.Seconds()), 10))
 
-	resp, err := i.Client.Do(req)
+	tr, ok := cc.Client.Transport.(*http.Transport)
+	if !ok {
+		return Value{}, fmt.Errorf("CredContext should contain an http.Transport value")
+	}
+
+	trCopy := tr.Clone()
+	trCopy.TLSClientConfig.Certificates = []tls.Certificate{i.Certificate}
```
ramondeklein, line 124: This was the best alternative that I could come up with. We don't want to change the `TLSClientConfig` of the original HTTP transport, because that may have security and parallelization issues.

pkg/credentials/sts_tls_identity.go
ramondeklein: I have brought back the `http.Client` fields in all providers. Instead of being required, they are now optional. When set, they override the `http.Client` in `CredsContext`.

pkg/credentials/sts_tls_identity.go
```diff
@@ -36,7 +36,12 @@ type CertificateIdentityOption func(*STSCertificateIdentity)
 // CertificateIdentityWithTransport returns a CertificateIdentityOption that
 // customizes the STSCertificateIdentity with the given http.RoundTripper.
 func CertificateIdentityWithTransport(t http.RoundTripper) CertificateIdentityOption {
-	return CertificateIdentityOption(func(i *STSCertificateIdentity) { i.Client.Transport = t })
+	return CertificateIdentityOption(func(i *STSCertificateIdentity) {
+		if i.Client == nil {
+			i.Client = &http.Client{}
+		}
```
ramondeklein, line 42: Previously, the `i.Client` was probably set when `STSCertificateIdentity` was created using `NewSTSCertificateIdentity`. So we need to check for `nil` here.

pkg/credentials/sts_tls_identity.go
```diff
@@ -145,7 +130,25 @@ func (i *STSCertificateIdentity) Retrieve() (Value, error) {
...
+		return Value{}, fmt.Errorf("CredContext should contain an http.Transport value")
+	}
+
+	// Clone the HTTP transport (patch the TLS client certificate)
+	trCopy := tr.Clone()
+	trCopy.TLSClientConfig.Certificates = []tls.Certificate{i.Certificate}
+
+	// Clone the HTTP client (patch the HTTP transport)
+	clientCopy := *client
+	clientCopy.Transport = trCopy
```
ramondeklein, line 149: Although not the most elegant solution, I guess this is the safest method to ensure we get a copy of the HTTP client.

ramondeklein:
@harshavardhana I think I addressed your comments. If you prefer, I could add some unit-tests to ensure that the custom transport is being invoked. Not sure if that's needed.

harshavardhana (approved):

klauspost:
@ramondeklein I though this code was on the "Unesco World Heritage List" - nobody has been touching it for years :)

LGTM :)

PR #2047, opened 2025-01-12
Allow status code 200 for removeBucketPolicy

tiran133:
I'm using Hetzner Object Storage, and they return a status code of 200 on successfully deleting bucket policy.

This pull request fixes the issue of getting an error even through the policy was successfully deleted.

The error handling was introduced here 
https://github.com/minio/minio-go/pull/1609

Kind regards

### DIFF
--- a/api-bucket-policy.go
+++ b/api-bucket-policy.go
@@ -88,7 +88,7 @@ func (c *Client) removeBucketPolicy(ctx context.Context, bucketName string) erro
 		return err
 	}
 
-	if resp.StatusCode != http.StatusNoContent {
+	if resp.StatusCode != http.StatusNoContent && resp.StatusCode != http.StatusOK {
 		return httpRespToErrorResponse(resp, bucketName, "")
 	}
 


klauspost:
Maybe Hetzner should be S3 compatible instead? 

https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteBucketPolicy.html#API_DeleteBucketPolicy_ResponseSyntax

tiran133:
I guess. Sorry, didn't read the specs. I opened a ticket with Hetzner now see what they say.

I know is the AWS specs but does it really make a difference?

klauspost:
We generally don't add workarounds for third party S3 compatibility. We would end up with an unmaintainable mess.

When the spec is unclear or AWS behaves differently than the spec (yes, that happens more often than you think), we can make exceptions. Keeping it open in case my colleagues don't agree on this.

tiran133:
Fair enough. I wait for Hetzner''s answer for now. Thank you!

tiran133:
I heard back from Hetzner, and they say they're trying to fix it in a timely manner. So I guess for now I live with it and wait for the fix on Hetzner's side.
Thank you!

PR #2053, opened 2025-01-21
Update s3-endpoints.go with ISO endpoints

### LINKED ISSUE #2049: Missing ISO endpoints
`s3-endpoints.go` misses 3 isolated endpoints & regions that are present accroding to aws s3 docs [1], namely:

-  US_ISO_EAST_1
- US_ISOB_EAST_1
- US_ISO_WEST_1 

[1]: https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/model/Region.html

vnzongzna:
Fixes #2049

### DIFF
--- a/s3-endpoints.go
+++ b/s3-endpoints.go
@@ -32,6 +32,18 @@ var awsS3EndpointMap = map[string]awsS3Endpoint{
 		"s3.us-east-2.amazonaws.com",
 		"s3.dualstack.us-east-2.amazonaws.com",
 	},
+	"us-iso-east-1": {
+		"s3.us-iso-east-1.c2s.ic.gov",
+		"s3.dualstack.us-iso-east-1.c2s.ic.gov",
+	},
+	"us-isob-east-1": {
+		"s3.us-isob-east-1.c2s.ic.gov",
+		"s3.dualstack.us-isob-east-1.c2s.ic.gov",
+	},
+	"us-iso-west-1": {
+		"s3.us-iso-west-1.c2s.ic.gov",
+		"s3.dualstack.us-iso-west-1.c2s.ic.gov",
+	},
 	"us-west-2": {
 		"s3.us-west-2.amazonaws.com",
 		"s3.dualstack.us-west-2.amazonaws.com",
--- a/s3-endpoints.go
+++ b/s3-endpoints.go
@@ -37,8 +37,8 @@ var awsS3EndpointMap = map[string]awsS3Endpoint{
 		"s3.dualstack.us-iso-east-1.c2s.ic.gov",
 	},
 	"us-isob-east-1": {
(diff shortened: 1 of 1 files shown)


s3-endpoints.go
```diff
@@ -32,6 +32,18 @@ var awsS3EndpointMap = map[string]awsS3Endpoint{
 		"s3.us-east-2.amazonaws.com",
 		"s3.dualstack.us-east-2.amazonaws.com",
 	},
+	"us-iso-east-1": {
+		"s3.us-iso-east-1.c2s.ic.gov",
+		"s3.dualstack.us-iso-east-1.c2s.ic.gov",
```
klauspost, line 37:
```
λ ping s3.dualstack.us-iso-east-1.c2s.ic.gov
Ping request could not find host s3.dualstack.us-iso-east-1.c2s.ic.gov. Please check the name and try again.
```

We can't really have non-existing domains in here. 

Leave empty and add an error if attempting to use this region with dual stack.

klauspost (changes requested):

s3-endpoints.go
vnzongzna: do you see similar result with `s3.us-iso-east-1.c2s.ic.gov` as well?

klauspost: Yes. Just run ping or nslookup yourself - it is a bit faster than waiting for me to do it ;)

vnzongzna: Thank you, I tested out the endpoints from the jumpbox (Amazon Workspace) for `us-iso-east-1` and `us-iso-west-1` & they were reachable. However, I didn't check them outside of the jumpbox (my mistake here).

vnzongzna: would you mind approving @klauspost since the suggestion to throw error/panic won't be needed in case the region selected is one of the isolated region.

klauspost: @vnzongzna Would you please explain. Are these not expected to be publicly reachable?

vnzongzna: Correct @klauspost 
isolated endpoints are only reachable from within the environment.
Got the reference from [here](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/model/Region.html) 
image

klauspost: okie

klauspost (approved):
I can't validate these, but I suppose adding them will not hurt.

harshavardhana (approved):

vnzongzna:
Apologies to ping on weekend, but can we merge this, I'll like to use this updated module in
* https://github.com/argoproj/argo-events
* https://github.com/argoproj/argo-workflows

PR #2057, opened 2025-01-27
allow users to extend DNS style buckets

harshavardhana:
this is to allow for custom extensions, since
we keep getting new requests to add new URLs
as default for bucket DNS style requests. This
is not scalable for this SDK, allowing a caller
specific function allows callers can write their
own function and provide us when they wish to use
path style or bucket DNS style.

### DIFF
--- a/api.go
+++ b/api.go
@@ -92,6 +92,9 @@ type Client struct {
 	// default to Auto.
 	lookup BucketLookupType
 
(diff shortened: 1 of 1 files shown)


harshavardhana:
PTAL @klauspost

klauspost (approved):

Issue #2056, opened 2025-01-27
Any reason minio.RemoveObjectError does not implement error interface?


wuhaochen:
I'm trying to send the errors from `RemoveObjects` to a general `chan error` for further processing and it seems `minio.RemoveObjectError` does not implement `error` interface. Is there any specific design consideration for being so? If not I can send a PR.

I'm not sure if this pattern applies to any other Error structs in this package either.

harshavardhana:
Because it is a composite struct with multiple fields, `error` is an inflexible interface with limited methods. 

S3 APIs return a lot more context and it is this context that you must consume not just "err"

wuhaochen:
@harshavardhana I'm not sure that's a good reason. A struct can implement the `error` interface and keep all the internal state. It then can be treated as an `error` in other context if the internal state doesn't matter.

For example [the official blog](https://go.dev/blog/error-handling-and-go) used [json](https://pkg.go.dev/encoding/json#SyntaxError) package as an example exactly similar to this struct.

I'm suggesting merely adding
```
func (e RemoveObjectError) Error() string{
    return e.Err.Error()
}
```
to satisfy the error interface since I do think this struct is intended as an error.

harshavardhana:
We are not changing it, sorry.

klauspost:
Honestly wrapping it is trivial on your side.

On the other hand you are welcome to send a PR that does. Not really seeing any problem with it, but also not really a great reason to do so.

PR #2058, opened 2025-01-28
Implement error interface for RemoveObjectError

wuhaochen:
See discussion in #2056

### DIFF
--- a/api-remove.go
+++ b/api-remove.go
@@ -214,6 +214,10 @@ type RemoveObjectError struct {
 }
 
 func (err *RemoveObjectError) Error() string {
+	// This should never happen as we will have a non-nil error with no underlying error.
+	if err.Err == nil {
+		return "inconsistant remove object error result"
+	}
 	return err.Err.Error()
 }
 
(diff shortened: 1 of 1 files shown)


api-remove.go
```diff
@@ -213,6 +213,10 @@ type RemoveObjectError struct {
 	Err        error
 }
 
+func (err *RemoveObjectError) Error() string {
```
harshavardhana, line 216:
```suggestion
func (err *RemoveObjectError) Error() string {
     if err.Err != nil {
         return err.Err.Error()
     }
     return "<nil>"
```

harshavardhana (changes requested):

api-remove.go
wuhaochen: `err.Err` currently would never be nil and should not be nil in the future if we implement it as an error, but agree with we should check about it.

api-remove.go
```diff
@@ -214,6 +214,10 @@ type RemoveObjectError struct {
 }
 
 func (err *RemoveObjectError) Error() string {
+	// This should never happen as we will have a non-nil error with no underlying error.
+	if err.Err == nil {
+		return "inconsistant remove object error result"
```
harshavardhana, line 219:
```suggestion
		return "unexpected remove object error result"
```

PR #2063, opened 2025-02-07
fix: api-list ListBuckers possible deref

wooffie:
I saw we checking resp for nil, but after we use resp.Body without check.

You can offer another nil handling here =)


harshavardhana:
resp is never nil that code is old, we never bothered to fix it - these PRs and others don't really add any value #2062

PR #2064, opened 2025-02-10
support customHeader for client

jiuker:
support customHeader for client
mc will use this to support `customheader`


harshavardhana (changes requested):
we shouldn't do this @jiuker we should just add it via `mc` custom transport RoundTrip implementation

PR #2069, opened 2025-02-20
fix: decode tags key without PathUnescape

jiuker:
non idempotent
```
func Test_Main(t *testing.T) {
	tagsOMap := map[string]string{"A": "B", "C": "D", "D:A": "A", "E": "PA"}
	// putObject - mock mc cp file remote/path --tags="A=B&C=D&D:W=A&E=PA"
	tagStr := TagEncode(tagsOMap)
	// getObject - mock client getObject with tags return
	tagsNMap := TagDecode(tagStr)
	// putObject - mock put file to target
	tagNStr := TagEncode(tagsNMap)
	_, err := tags.ParseObjectTags(tagNStr)
	if err != nil {
		t.Log("TestTagsTagDecode: we got error here:", err)
	}

	// we must call url.PathUnescape here <--------------------
	unescapeTags, err := url.PathUnescape(tagNStr)
	if err != nil {
		return
	}
	_, err = tags.ParseObjectTags(unescapeTags)
	if err != nil {
		t.Log("TestTagsTagDecode: we got error here:", err)
	}
}
```

### DIFF
--- a/pkg/s3utils/utils_test.go
+++ b/pkg/s3utils/utils_test.go
@@ -501,7 +501,17 @@ func TestTagsTagDecode(t *testing.T) {
 	tagNStr := TagEncode(tagsNMap)
 	_, err := tags.ParseObjectTags(tagNStr)
 	if err != nil {
-		t.Log("TestTagsTagDecode: we got error here", err)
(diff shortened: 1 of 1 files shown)


harshavardhana:
what is the change here?

PR #2072, opened 2025-02-27
feat: add reverse version for listObjectVersions

jiuker:
feat: add reverse version for listObjectVersions
We need a sorted a versions by time for some case.

### DIFF
--- a/api-list.go
+++ b/api-list.go
@@ -22,7 +22,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
-	"sort"
+	"slices"
 	"time"
 
 	"github.com/minio/minio-go/v7/pkg/s3utils"
@@ -429,18 +429,7 @@ func (c *Client) listObjectVersions(ctx context.Context, bucketName string, opts
 		)
 		send := func(vers []Version) {
 			if opts.ReverseVersions {
-				sort.Slice(vers, func(i, j int) bool {
-					if vers[i].IsLatest {
-						return false
-					}
-					if vers[j].IsLatest {
-						return true
-					}
-					if vers[i].LastModified.Equal(vers[j].LastModified) {
-						return true
-					}
-					return vers[i].LastModified.Before(vers[j].LastModified)
-				})
+				slices.Reverse(vers)
 				numVersions = len(vers)
 			}
 			for _, version := range vers {
--- a/api-list.go
+++ b/api-list.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"sort"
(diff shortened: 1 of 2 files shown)


api-list.go
```diff
@@ -421,20 +422,23 @@ func (c *Client) listObjectVersions(ctx context.Context, bucketName string, opts
...
-
-		for {
-			// Get list of objects a maximum of 1000 per request.
-			result, err := c.listObjectVersionsQuery(ctx, bucketName, opts, keyMarker, versionIDMarker, delimiter)
-			if err != nil {
-				sendObjectInfo(ObjectInfo{
-					Err: err,
+		send := func(vers []Version) {
+			if opts.ReverseVersions {
+				sort.Slice(vers, func(i, j int) bool {
```
klauspost, line 431: Instead of re-sorting, just reverse the order. There are internal tie-breakers for versions with the same timestamp, and we should respect that.  Just be sure you only reverse once :)

You can use [slices.Reverse](https://pkg.go.dev/slices#Reverse).

klauspost (commented):
Just a minor change.

klauspost (approved):
lgtm

api-list.go
```diff
@@ -421,20 +422,17 @@ func (c *Client) listObjectVersions(ctx context.Context, bucketName string, opts
...
-		for {
-			// Get list of objects a maximum of 1000 per request.
-			result, err := c.listObjectVersionsQuery(ctx, bucketName, opts, keyMarker, versionIDMarker, delimiter)
-			if err != nil {
-				sendObjectInfo(ObjectInfo{
-					Err: err,
-				})
-				return
+		send := func(vers []Version) {
+			if opts.ReverseVersions {
```
harshavardhana, line 431:
```suggestion
			if opts.WithVersion && opts.ReverseVersions {
```

api-list.go
```diff
@@ -457,6 +456,38 @@ func (c *Client) listObjectVersions(ctx context.Context, bucketName string, opts
...
+		for {
+			// Get list of objects a maximum of 1000 per request.
+			result, err := c.listObjectVersionsQuery(ctx, bucketName, opts, keyMarker, versionIDMarker, delimiter)
+			if err != nil {
+				sendObjectInfo(ObjectInfo{
+					Err: err,
+				})
+				return
+			}
+			if opts.ReverseVersions {
```
harshavardhana, line 469:
```suggestion
			if opts.WithVersions && opts.ReverseVersions {
```

harshavardhana (changes requested):

harshavardhana (approved):

PR #2075, opened 2025-03-08
Adds `TokenRevokeType` field to credential providers.

taran-p:
miniohq/eos#498

### DIFF
--- a/pkg/credentials/assume_role.go
+++ b/pkg/credentials/assume_role.go
@@ -105,7 +105,7 @@ type STSAssumeRoleOptions struct {
 	RoleSessionName string
 	ExternalID      string
 
-	TokenType string // Optional, used for token revokation
+	TokenRevokeType string // Optional, used for token revokation
 }
 
 // NewSTSAssumeRole returns a pointer to a new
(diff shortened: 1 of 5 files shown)


pkg/credentials/assume_role.go
```diff
@@ -104,6 +104,8 @@ type STSAssumeRoleOptions struct {
 	RoleARN         string
 	RoleSessionName string
 	ExternalID      string
+
+	TokenRevokeType string // Optional, used for token revokation
```
harshavardhana, line 108:
```suggestion
	TokenRevokeType string // Optional, used for token revokation (MinIO only extension)
```

harshavardhana (changes requested):

pkg/credentials/assume_role.go
taran-p: @harshavardhana Should this comment be on any of the other APIs as well?

harshavardhana: this should be fine for now.

harshavardhana (approved):

PR #2079, opened 2025-03-18
make downtime info as map to denote per target info

Praveenrajmani:


dhananjaykrutika (approved):

pkg/replication/replication.go
```diff
@@ -988,10 +988,10 @@ func (q ReplQueueStats) QStats() (r ReplQStats) {
...
 // MetricsV2 represents replication metrics for a bucket.
 type MetricsV2 struct {
-	Uptime       int64          `json:"uptime"`
-	CurrentStats Metrics        `json:"currStats"`
-	QueueStats   ReplQueueStats `json:"queueStats"`
-	DowntimeInfo DowntimeInfo   `json:"downtimeInfo"`
+	Uptime       int64                   `json:"uptime"`
+	CurrentStats Metrics                 `json:"currStats"`
+	QueueStats   ReplQueueStats          `json:"queueStats"`
+	DowntimeInfo map[string]DowntimeInfo `json:"downtimeInfo"`
```
harshavardhana, line 994: We cannot break APIs and Structs in this manner, add it as a new entry - you are even retaining the JSON key which will cause all sorts of problems.

harshavardhana (changes requested):

pkg/replication/replication.go
shtripat: +1. Rather have a new field say `DowntimeInfoV2`

Praveenrajmani: this field was recently added and new.  Not yet used by any APIs. This was added in https://github.com/minio/minio-go/pull/2071 

I had to change this to address per target downtime info in my WIP - https://github.com/miniohq/ec/pull/37

Praveenrajmani: consider this as a continuation of https://github.com/minio/minio-go/pull/2071

shtripat: Oh ok, then it should be fine

shtripat (approved):

pkg/replication/replication.go
harshavardhana: Unfortunately we have made a release already, please use a different filed https://github.com/minio/minio-go/releases/tag/v7.0.88 @Praveenrajmani

Praveenrajmani: the server changes which sets this field was never merged or released @harshavardhana do we still need a new field?

poornas: should be fine to leave this as is

poornas (approved):

pkg/replication/replication.go
harshavardhana: Okay

harshavardhana: Please make sure to avoid these in the future.

harshavardhana (approved):

PR #2082, opened 2025-03-25
Implement AppendObject() API

harshavardhana:


HJLebbink (approved):
LGTM, works as advertised

api-append-object.go
```diff
@@ -0,0 +1,202 @@
...
+	}, nil
+}
+
+// AppendObject - S3 Express Zone https://docs.aws.amazon.com/AmazonS3/latest/userguide/directory-buckets-objects-append.html
+func (c *Client) AppendObject(ctx context.Context, bucketName, objectName string, reader io.Reader, objectSize int64,
+	opts AppendObjectOptions,
+) (info UploadInfo, err error) {
+	if objectSize < 0 && opts.ChunkSize == 0 {
+		return UploadInfo{}, errors.New("object size must be provided when no chunk size is provided")
+	}
```
klauspost, line 158: Why? Can't we just use 5GB, or is that too much memory?

Maybe just document on the options.

api-append-object.go
```diff
@@ -0,0 +1,202 @@
...
+
+	if err = opts.validate(c); err != nil {
+		return UploadInfo{}, err
+	}
+
+	oinfo, err := c.StatObject(ctx, bucketName, objectName, StatObjectOptions{})
```
klauspost, line 164: Probably you want...

```suggestion
	oinfo, err := c.StatObject(ctx, bucketName, objectName, StatObjectOptions{Checksum: true})
```

api-append-object.go
```diff
@@ -0,0 +1,202 @@
...
+	if err != nil {
+		return UploadInfo{}, err
+	}
+	opts.setWriteOffset(oinfo.Size) // First append must set the current object size as the offset.
```
klauspost, line 168: Transfer checksum type.

api-append-object.go
```diff
@@ -0,0 +1,202 @@
...
+		ChecksumSHA1:      h.Get(ChecksumSHA1.Key()),
+		ChecksumSHA256:    h.Get(ChecksumSHA256.Key()),
+		ChecksumCRC64NVME: h.Get(ChecksumCRC64NVME.Key()),
+	}, nil
+}
+
+// AppendObject - S3 Express Zone https://docs.aws.amazon.com/AmazonS3/latest/userguide/directory-buckets-objects-append.html
+func (c *Client) AppendObject(ctx context.Context, bucketName, objectName string, reader io.Reader, objectSize int64,
+	opts AppendObjectOptions,
+) (info UploadInfo, err error) {
```
klauspost, line 155: I kinda expect we would need `c.trailingHeaderSupport` to be set for checksums to always work.

klauspost (commented):
Nothing major.

api-append-object.go
```diff
@@ -0,0 +1,202 @@
...
+	"io"
+	"net/http"
+	"strconv"
+
+	"github.com/minio/minio-go/v7/pkg/s3utils"
+)
+
+// AppendObjectOptions https://docs.aws.amazon.com/AmazonS3/latest/userguide/directory-buckets-objects-append.html
+type AppendObjectOptions struct {
+	Progress             io.Reader
```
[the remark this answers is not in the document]
harshavardhana: It needs to be wrapped around to provide progress, will hook it up looks like forgot.

api-append-object.go
harshavardhana: fixed

api-append-object.go
harshavardhana: done

harshavardhana:
This PR fixes the checksum behavior. @klauspost - I also added a few missing things in the checksum behavior as per AWS S3 behavior. 

Server side is also changing in the Append() PR itself.

klauspost (approved):
<3

PR #2086, opened 2025-04-03
API for checking object management permissions

shtripat:
Check if given user's IAM policies allow s3:PutObject/s3:DeleteObject on given bucket

### DIFF
--- a/api-bucket-policy.go
+++ b/api-bucket-policy.go
@@ -145,3 +145,32 @@ func (c *Client) getBucketPolicy(ctx context.Context, bucketName string) (string
 	policy := string(bucketPolicyBuf)
 	return policy, err
 }
+
+// CheckObjectManagePermissions verifies if user has object permissions for given bucket
+func (c *Client) CheckObjectManagePermissions(ctx context.Context, bucket string, user string) error {
+	urlValues := make(url.Values)
+	urlValues.Set("user", user)
+
+	// Execute HEAD on bucketName.
+	resp, err := c.executeMethod(ctx, http.MethodGet, requestMetadata{
+		bucketName:  bucket,
+		queryValues: urlValues,
+	})
+	defer closeResponse(resp)
+	if err != nil {
+		if ToErrorResponse(err).Code == "NoSuchBucket" {
+			return nil
+		}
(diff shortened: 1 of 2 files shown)


api.go
```diff
@@ -1071,3 +1071,15 @@ func (c *Client) CredContext() *credentials.CredContext {
 		Endpoint: c.endpointURL.String(),
 	}
 }
+
+// GetCreds returns the access creds for the client
+func (c *Client) GetCreds() (string, string, error) {
```
harshavardhana, line 1076: Credentials are always 3 entities, you should return the value()

But what is the point of this function?

api-bucket-policy.go
```diff
@@ -145,3 +145,32 @@ func (c *Client) getBucketPolicy(ctx context.Context, bucketName string) (string
 	policy := string(bucketPolicyBuf)
 	return policy, err
 }
+
+// CheckObjectManagePermissions verifies if user has object permissions for given bucket
+func (c *Client) CheckObjectManagePermissions(ctx context.Context, bucket string, user string) error {
```
harshavardhana, line 150: And also why is this an S3 SDK API? This is specific to MinIO and must be in madmin-go

harshavardhana (changes requested):

api.go
shtripat: This is not needed with new changes. Will remove.

api-bucket-policy.go
shtripat: This we are checking with replicated bucket target clients. We need to reach all targets of bucket replication and see target has got PutObject/DeleteObject etc. Is there a way to achieve it? I may be missing something here.

harshavardhana (changes requested):

api-bucket-policy.go
poornas: this should be in madmin-go

shtripat: ok, will better move to madmin-go

shtripat:
Closing as https://github.com/minio/minio-go/pull/2089 takes care creds and API is added as admin API.

PR #2087, opened 2025-04-03
Rety iterators

tlyons-cs:
With the transition to go 1.23 the internal retry goroutine+channel can be replaced with an iter.Seq[int]. This reduces the memory allocation of a new goroutine and the extra overhead for waiting on the receipt/done channels.

### DIFF
--- a/api-bucket-notification.go
+++ b/api-bucket-notification.go
@@ -157,13 +157,6 @@ func (c *Client) ListenBucketNotification(ctx context.Context, bucketName, prefi
 			return
 		}
 
-		// Continuously run and listen on bucket notification.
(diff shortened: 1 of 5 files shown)


tlyons-cs:
For some added context: when using this client in a high volume service I noticed the number of goroutines was roughly double what I expected. This was because for every routine executing an operation, another routine had spun up handling the retries. Adding this change would eliminate the need for a separate goroutine for the retry delays and improve performance

klauspost (approved):
Looks like a solid refactor!

harshavardhana:
@tlyons-cs can you fix the linter errors?

harshavardhana (approved):

PR #2088, opened 2025-04-04
feat: make JSON lib configurable

secDre4mer:
Currently, a mix of goccy/json and encoding/json is used. Both use the same interface; goccy/json is faster, however, it also does questionable things with go:linkname (problematic with e.g. https://github.com/traefik/yaegi) and has issues with marshaling some structs.
Create a common package that can switch (via build tags) from one implementation to another, and use that in all packages.

### DIFF
--- a/api-bucket-notification.go
+++ b/api-bucket-notification.go
@@ -26,7 +26,7 @@ import (
 	"net/url"
 	"time"
 
-	"github.com/goccy/go-json"
+	"github.com/minio/minio-go/v7/internal/json"
 	"github.com/minio/minio-go/v7/pkg/notification"
 	"github.com/minio/minio-go/v7/pkg/s3utils"
 )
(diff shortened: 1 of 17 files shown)


harshavardhana (changes requested):

pkg/json/json_goccy.go
klauspost, line 1: Move to `internal/json` - we shouldn't be exporting this.

secDre4mer: Done.

pkg/json/json_goccy.go
```diff
@@ -0,0 +1,17 @@
+//go:build !stdlibjson
```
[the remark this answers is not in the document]
secDre4mer: Done.

klauspost (approved):

secDre4mer:
@harshavardhana Are there any additional changes you'd like to see for this PR?

harshavardhana (approved):

PR #2089, opened 2025-04-05
Add function for getting creds of remote client

shtripat:
This is needed while checking PutObject/DeleteObject permissions on bucket replicated remote targets to create admin client and invoke API.

### DIFF
--- a/api.go
+++ b/api.go
@@ -1071,9 +1071,5 @@ func (c *Client) GetCreds() (credentials.Value, error) {
 	if c.credsProvider == nil {
 		return credentials.Value{}, errors.New("no credentials provider")
 	}
-	value, err := c.credsProvider.GetWithContext(c.CredContext())
-	if err != nil {
-		return credentials.Value{}, err
-	}
-	return value, nil
+	return c.credsProvider.GetWithContext(c.CredContext())
 }
--- a/api.go
+++ b/api.go
@@ -1066,7 +1066,6 @@ func (c *Client) CredContext() *credentials.CredContext {
 	}
 }
 
-
 // GetCreds returns the access creds for the client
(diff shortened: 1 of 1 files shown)


api.go
```diff
@@ -1065,3 +1065,15 @@ func (c *Client) CredContext() *credentials.CredContext {
...
+// GetCreds returns the access creds for the client
+func (c *Client) GetCreds() (*credentials.Value, error) {
+	if c.credsProvider == nil {
+		return nil, errors.New("no credentials provider")
+	}
+	value, err := c.credsProvider.GetWithContext(c.CredContext())
+	if err != nil {
+		return nil, err
+	}
+	return &value, nil
```
harshavardhana, line 1078:
```suggestion
	return &value, nil
```

why do you need a pointer return?

harshavardhana (changes requested):

api.go
shtripat: OOPs copy paste from previous function. correcting.

harshavardhana (approved):

api.go
```diff
@@ -1065,3 +1065,15 @@ func (c *Client) CredContext() *credentials.CredContext {
...
+// GetCreds returns the access creds for the client
+func (c *Client) GetCreds() (credentials.Value, error) {
+	if c.credsProvider == nil {
+		return credentials.Value{}, errors.New("no credentials provider")
+	}
+	value, err := c.credsProvider.GetWithContext(c.CredContext())
+	if err != nil {
+		return credentials.Value{}, err
+	}
+	return value, nil
```
harshavardhana, line 1078:
```suggestion
	return c.credsProvider.GetWithContext(c.CredContext())
```

harshavardhana (approved):

PR #2100, opened 2025-05-07
feat: add support for tencent cos bucket lookup auto

okhowang:


klauspost:
@okhowang Doesn't Tencent work with path-style URLs?

harshavardhana:
No need for automation, chose what you want per client.

okhowang:
> @okhowang Doesn't Tencent work with path-style URLs?
> 
> Unless you can provide a rally good reason we are not adding more, since this would be a never-ending task.

Yes, reference https://www.tencentcloud.com/document/product/436/57456

PR #2103, opened 2025-05-11
support aws s3 express zone behavior

harshavardhana:


klauspost (approved):

pkg/s3utils/utils.go
```diff
@@ -344,6 +345,56 @@ func CheckValidBucketName(bucketName string) (err error) {
...
+		return errors.New("Bucket name cannot be shorter than 3 characters for S3 Express")
+	}
+
+	if len(bucketName) > 63 {
+		return errors.New("Bucket name cannot be longer than 63 characters for S3 Express")
+	}
+
+	// Check if the bucket matches the regex
+	if !validBucketNameS3Express.MatchString(bucketName) {
+		return errors.New("Bucket name contains invalid characters")
```
vadmeste, line 369: what about this?

s/Bucket name has unexpected format or contains invalid characters/

api.go
```diff
@@ -818,14 +825,21 @@ func (c *Client) newRequest(ctx context.Context, method string, metadata request
...
-	// Initialize a new HTTP request for the method.
-	req, err = http.NewRequestWithContext(ctx, method, targetURL.String(), nil)
+	// make sure to de-dup calls to credential services, this reduces
+	// the overall load to the endpoint generating credential service.
+	value, err, _ := c.credsGroup.Do(metadata.bucketName, func() (credentials.Value, error) {
+		if s3utils.IsS3ExpressBucket(metadata.bucketName) {
+			return c.CreateSession(ctx, metadata.bucketName, SessionReadWrite)
+		}
+		// Get credentials from the configured credentials provider.
+		return c.credsProvider.GetWithContext(c.CredContext())
```
vadmeste, line 835: If I am not mistaken, this function will only be called once, right ? but I think this is unexpected because credentials sometimes need to be renewed automatically and it seems nothing will trigger the renewal after this change.

vadmeste (changes requested):

vadmeste (commented):
LGTM, one minor thing

vadmeste (approved):

PR #2107, opened 2025-05-14
Add msgpack serializers to stringset

klauspost:
Add binary (de)serializers using msgp to stringsets.

Should we move this to to `minio/pkg` instead?

### DIFF
--- a/pkg/set/msgp.go
+++ b/pkg/set/msgp.go
@@ -1,0 +1,149 @@
+/*
+ * MinIO Go Library for Amazon S3 Compatible Cloud Storage
+ * Copyright 2015-2025 MinIO, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package set
+
+import "github.com/tinylib/msgp/msgp"
+
+// EncodeMsg encodes the message to the writer.
+// Values are stored as a slice of strings or nil.
+func (s StringSet) EncodeMsg(writer *msgp.Writer) error {
+	if s == nil {
+		return writer.WriteNil()
+	}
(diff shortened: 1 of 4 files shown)


pkg/set/stringset.go
```diff
@@ -37,6 +37,30 @@ func (set StringSet) ToSlice() []string {
...
+
+	// Add keys to this...
+	keys := make([][]byte, 0, len(set))
+	for k := range set {
+		n := copy(dst, k)
+		keys = append(keys, dst[:n])
+		dst = dst[n:]
+	}
+	sort.Slice(keys, func(i, j int) bool {
+		return string(keys[i]) < string(keys[j])
```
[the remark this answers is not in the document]
klauspost: Bro. It is exactly the same.

harshavardhana (approved):

pkg/set/msgp.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+	if len(s) == 0 {
+		return msgp.AppendArrayHeader(bytes, 0), nil
+	}
+	bytes = msgp.AppendArrayHeader(bytes, uint32(len(s)))
+	sorted := s.ToByteSlices()
+	for _, k := range sorted {
+		bytes = msgp.AppendStringFromBytes(bytes, k)
+	}
+	return bytes, nil
```
anjalshireesh, line 56:
```suggestion
	w := b.NewBuffer(bytes)
	err := s.EncodeMsg(msgp.NewWriter(w))
	return w.Bytes(), err
```

Can `EncodeMsg` be re-used here? Will something like this work?

anjalshireesh (approved):
LGTM, one minor question

pkg/set/msgp.go
klauspost: Yes, though I am pretty sure you would need to flush the `msgp.Writer`, plus it does an alloc for its internal buffer. But let me compare it. After all, it is a simplification.

klauspost: Nothing crazy.

```
Benchmark100StringSetMarshalMsg/before-32         	  156386	      7539 ns/op	    4192 B/op	       5 allocs/op
Benchmark100StringSetMarshalMsg/after-32          	  151599	      8146 ns/op	    6344 B/op	       8 allocs/op
```

Using...

```
func (s StringSet) MarshalMsg2(b []byte) ([]byte, error) {
	w := bytes.NewBuffer(b)
	mw := msgp.NewWriter(w)
	err := s.EncodeMsg(mw)
	mw.Flush()
	return w.Bytes(), err
}
```

If we didn't already have it, I wouldn't mind it. But now we have a slightly better might as well keep it. Does that make sense?

anjalshireesh: sure @klauspost , thanks :+1:

PR #2115, opened 2025-05-27
Add support of error responses in 200 OK body

vadmeste:
An S3 server can respond with 200 OK with an S3 Error response, for long standing calls, 
such as S3 listing when a prefix contains a lot of entries.

This commit will support this use case. This can only be enabled with specific calls, 
that returns XML as a success response, so this will not be confused with GetObject
returning a valid document that contains an S3 error response.

Supported calls: CompleteMultipartUpload and DeleteMultiObjects.

### DIFF
--- a/200OKwithError_test.go
+++ b/200OKwithError_test.go
@@ -1,0 +1,209 @@
+/*
+ * MinIO Go Library for Amazon S3 Compatible Cloud Storage
+ * Copyright 2017-2020 MinIO, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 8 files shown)


vadmeste:
@harshavardhana I wrote those tests. 

I also added more fixes to the original code to make those tests passing. minio-go was not retrying SlowDownRead and SlowDownWrite, it is only recognizing SlowDown error code.

harshavardhana:
> I also added more fixes to the original code to make those tests passing. minio-go was not retrying SlowDownRead and SlowDownWrite, it is only recognizing SlowDown error code.

ah we forgot to change support for differentiation here, even AWS SDKs might not retry, perhaps we should change it back and add another field to the error type?

harshavardhana:
> > I also added more fixes to the original code to make those tests passing. minio-go was not retrying SlowDownRead and SlowDownWrite, it is only recognizing SlowDown error code.
> 
> ah we forgot to change support for differentiation here, even AWS SDKs might not retry, perhaps we should change it back and add another field to the error type?

Talking about server side.

harshavardhana:
can you fix the build @vadmeste ?

PR #2119, opened 2025-06-04
test: add test for PresignedPostPolicy with empty fileName

jiuker:
test: add test for PresignedPostPolicy with empty fileName
fix https://github.com/minio/minio/pull/21323#pullrequestreview-2869695342
```
We should add a test for this in mint.
```

### DIFF
--- a/functional_tests.go
+++ b/functional_tests.go
@@ -5586,6 +5586,161 @@ func testPresignedPostPolicyWrongFile() {
 	logSuccess(testName, function, args, startTime)
 }
 
+// testPresignedPostPolicyEmptyFileName tests that an empty file name in the presigned post policy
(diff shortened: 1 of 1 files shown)


klauspost:
@jiuker What you posted looks more like a bug than a test failure to me. Consider failing rather than crashing.

klauspost:
@jiuker How can a remote package cause a ` runtime error: invalid memory address or nil pointer dereference` in a local test?

klauspost:
@jiuker The test *in this package* should not panic, just because it doesn't get the response it expects. It should log the error and move on.

harshavardhana:
> Panic happend at minio code, not this package.

are you sending a fix?

harshavardhana (approved):

PR #2124, opened 2025-06-25
fix: prefix should use stringLike when new a bucketPolicy (2)

mannreis:
Addressing https://github.com/minio/minio/issues/21244 

Is there a reason why #2105  was closed? Does it not fix the issue or is there more to it?

This a new attempt to fix it!


mannreis:
After further checking I think we do need the wildcard appended to the prefix other wise it will behave as an exact match. There's an example in the docs: https://min.io/docs/minio/linux/administration/identity-access-management/policy-based-access-control.html#minio-policy-variables-internal

Other wise removing the last two commits would address you suggestion.Thanks @jiuker

harshavardhana (approved):

PR #2133, opened 2025-07-14
test: add unit test for OptimalPartInfo

micbar:
# Description

Addes some unit tests to clarify / confirm the multipart upload behavior.

## Details

It adds some table tests with different object sizes when the `configuredPartSize=0`.

### Object Size Table

Object Size | Total Size | Start Offset | Number of Parts | Part Size | Last Part Size
-- | -- | -- | -- | -- | --
16MB object size | 16 MB | 0 | 1 | 16 × 1024 × 1024 | 16 × 1024 × 1024
100MB object size | 100 MB | 0 | 7 | 16 × 1024 × 1024 | 4 × 1024 × 1024
1GB object size | 1 GB | 0 | 64 | 16 × 1024 × 1024 | 16 × 1024 × 1024
10GB object size | 10 GB | 0 | 640 | 16 × 1024 × 1024 | 16 × 1024 × 1024
100GB object size | 100 GB | 0 | 6,400 | 16 × 1024 × 1024 | 16 × 1024 × 1024
500GB object size | 500 GB | 0 | 8,000 | 64 × 1024 × 1024 | 64 × 1024 × 1024
1TB object size | 1 TB | 0 | 9,363 | 112 × 1024 × 1024 | 32 × 1024 × 1024
2TB object size | 2 TB | 0 | 9,363 | 224 × 1024 × 1024 | 64 × 1024 × 1024
3TB object size | 3 TB | 0 | 9,831 | 320 × 1024 × 1024 | 128 × 1024 × 1024
5TB object size (max) | 5 TB | 0 | 9,930 | 528 × 1024 × 1024 | 368 × 1024 × 1024

### DIFF
--- a/api_unit_test.go
+++ b/api_unit_test.go
@@ -170,6 +170,52 @@ func TestPartSize(t *testing.T) {
 	}
 }
 
+// TestOptimalPartInfoWithZeroConfiguredPartSize - test for OptimalPartInfo() with configuredPartSize = 0
+func TestOptimalPartInfoWithZeroConfiguredPartSize(t *testing.T) {
+	testCases := []struct {
+		name                 string
+		objectSize           int64
+		configuredPartSize   uint64
+		expectedTotalParts   int
+		expectedPartSize     int64
+		expectedLastPartSize int64
+	}{
+		{"16MB object size", 16 * 1024 * 1024, 0, 1, 16 * 1024 * 1024, 16 * 1024 * 1024},
+		{"100MB object size", 100 * 1024 * 1024, 0, 7, 16 * 1024 * 1024, 4 * 1024 * 1024},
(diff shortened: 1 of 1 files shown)


micbar:
> There are no guarantees we wont change this in the future. As such the test isn't really worth much. Just seems like added maintenance.

Interesting argument 😅

Isn’t that the point of unit testing?

Why should we plan ahead to do an obscure change in a vital part of this Projekt in the near future?

klauspost:
@micbar You shouldn't add tests that go beyond the guarantees of the API. Then you are just testing the implementation, which means it is worthless if you ever make changes. So it adds no value by itself. Having the chart you did is valuable in so far as it reflects the current choices made.

PR #2138, opened 2025-07-22
Small change to functional test, check FULL_OBJECT mode is returned properly

marktheunissen:


marktheunissen:
This can be merged since it still tests against open source edition, and once AIStor PR 1189 is in, the test will pass against AIStor too.

klauspost (approved):

PR #2144, opened 2025-08-08
fix: FGetObject fails on local file names longer than 180 chars

stuchl4n3k:
Hi there good folks. I'm adding a quick fix with some tests for `FGetObject` which fails on most systems that have 255 chars file name limit. The function appends 75 chars to the user-supplied `filePath` to create a temp file. If the user-supplied filename is longer than 180 chars, the call results in `ENAMETOOLONG` error.

Note: intentionally left 2 commits to PoC this. Should I squash?

### DIFF
--- a/api-get-object-file.go
+++ b/api-get-object-file.go
@@ -69,7 +69,7 @@ func (c *Client) FGetObject(ctx context.Context, bucketName, objectName, filePat
 	}
 
 	// Write to a temporary file "fileName.part.minio" before saving.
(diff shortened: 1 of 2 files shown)


klauspost:
`Error: ./api-get-object-file_test.go:80:46: undefined: strings (typecheck)` You need to add the import.

klauspost (approved):

harshavardhana (approved):

PR #2147, opened 2025-08-21
Add inventory APIs

donatello:


api-inventory-ext.go
```diff
@@ -0,0 +1,218 @@
...
+)
+
+// This file contains the inventory API extension for MinIO server. It is not
+// compatible with AWS S3.
+
+func makeInventoryReqMetadata(bucket string, urlParams ...string) requestMetadata {
+	urlValues := make(url.Values)
+	urlValues.Set("minio-inventory", "")
+	if len(urlParams)%2 != 0 {
+		panic("invalid urlParams passed - this should not happen")
```
harshavardhana, line 40: No panics in SDKs please.

api-inventory-ext.go
```diff
@@ -0,0 +1,218 @@
...
+
+// DeleteBucketInventoryConfiguration deletes the given inventory configuration.
+func (c *Client) DeleteBucketInventoryConfiguration(ctx context.Context, bucket, id string) error {
+	reqMeta := makeInventoryReqMetadata(bucket, "id", id)
+	resp, err := c.executeMethod(ctx, http.MethodDelete, reqMeta)
+	defer closeResponse(resp)
+	if err != nil {
+		return err
+	}
+	if resp.StatusCode != http.StatusOK {
```
harshavardhana, line 117: This API should be very similar to what DeleteBucketLifecycle has

api-inventory-ext.go
```diff
@@ -0,0 +1,218 @@
...
+		return "", httpRespToErrorResponse(resp, bucket, "")
+	}
+	buf := new(strings.Builder)
+	_, err = io.Copy(buf, resp.Body)
+	return buf.String(), err
+}
+
+// PutBucketInventoryConfiguration - calls the inventory configuration
+// endpoint to create or update an inventory configuration for a bucket.
+func (c *Client) PutBucketInventoryConfiguration(ctx context.Context, bucket string, id string, yamlDef []byte) error {
```
harshavardhana, line 71: if you are going to return `yamlDef string` then it is better to take yamlDef as string input here as well. 

Also add some options pattern so that we can allow extensions to this API when it becomes apparent in future

api-inventory-ext.go
```diff
@@ -0,0 +1,218 @@
...
+	reqMeta := makeInventoryReqMetadata(bucket, "id", id)
+	reqMeta.contentBody = bytes.NewReader(yamlDef)
```
harshavardhana, line 73:
```suggestion
	reqMeta.contentBody = strings.NewReader(yamlDef)
```

api-inventory-ext.go
donatello: I've done this, but I can't really think of future extensions here, everything would likely happen in the yaml definition itself.

allanrogerr (approved):
I performed some basic testing on https://github.com/miniohq/ec/pull/148 and https://github.com/miniohq/eos/pull/1357. Nothing further to add right now.

donatello:
@harshavardhana I think we can merge, but could you take a look and let me know if you have any questions.

PR #2158, opened 2025-09-17
add new AWS region - ap-southeast-6

benclapp:
Adding the new ap-southeast-6 region to awsS3EndpointMap

### DIFF
--- a/endpoints.go
+++ b/endpoints.go
@@ -244,6 +244,10 @@ var awsS3EndpointMap = map[string]awsS3Endpoint{
 		"s3.ap-east-2.amazonaws.com",
 		"s3.dualstack.ap-east-2.amazonaws.com",
 	},
+	"ap-southeast-6": {
+		"s3.ap-southeast-6.amazonaws.com",
+		"s3.dualstack.ap-southeast-6.amazonaws.com",
+	},
 }
 
 // getS3ExpressEndpoint get Amazon S3 Express endpoing based on the region


harshavardhana:
> This endpoint is active now. Unfortunately, some software using minio-go is unable to access AWS endpoints unless they are in here. There was a PR that included this change 5 months ago, but these lines were rejected because the domains didn't resolve yet.
> 
> Can someone with write access please approve this so it can get into release, and in turn get into the next release of applications.

@shaneday as an application you should do your own thing if you want more dynamism in your application. 

```
  func geographicRegionMapper(u url.URL) string {
      hostname := u.Hostname()

      // Map geographic identifiers to AWS-compatible regions
      regionMap := map[string]string{
          "us-storage.example.com":     "us-east-1",
          "eu-storage.example.com":     "eu-west-1",
          "asia-storage.example.com":   "ap-southeast-1",
          "local-storage.example.com":  "us-west-2",
      }

      if region, found := regionMap[hostname]; found {
          return region
      }

      return "us-east-1" // default
  }

  func main() {
      minioClient, err := minio.New("eu-storage.example.com", &minio.Options{
          Creds:              credentials.NewStaticV4("key", "secret", ""),
          CustomRegionViaURL: geographicRegionMapper,
      })
      // Client will use "eu-west-1" region
  }
```

You don't have to rely on us anymore to keep this updated.

benclapp:
Hey @harshavardhana , thanks for having a look! Apologies for the delayed response - life 😅 

For context I was adding this new region to eventually update the use of this client for the Thanos project. However to my understanding this issue might be the same with other consumers too.

I had a spin with your suggested alternative, using a custom `CustomRegionViaURL` function. However I couldn't get this to work. I think it is because `CustomRegionViaURL` is only used if the `Region` is not set in the minio options when creating the client. https://github.com/minio/minio-go/blob/master/api.go#L278-L285 

With Thanos, both the region and endpoint are provided and the client is successfully created. It's the first request with the new client that actually fails. Tracing through it's because `getS3Endpoint()` is eventually called as part of the `executeMethod()` for Amazon endpoints. 

Given this, would you be open to reconsidering this PR?

allanrogerr:
Reviewed this against current `master`. The change itself is correct, but master already carries it, so this PR can be closed with nothing lost.

**Why it is superseded:** #2234 (commit `291e83f`, merged 2026-06-09) added a character-identical `ap-southeast-6` entry to `awsS3EndpointMap`. A merge simulation of this branch into today's master (`git merge-tree --write-tree upstream/master HEAD`, run on this branch's checkout with `upstream` = minio/minio-go) produces a tree with the same OID as `upstream/master^{tree}` — merging this PR would not change a single byte.

**Checks run on this branch** (head `421b604`), all passing:

- `go vet ./...`, `go build ./...`, `gofumpt -d endpoints.go` — clean.
- `go test -race -count=1 -run 'TestValidBucketLocation|TestMakeTargetURL' .` — pass, 3 of 3 iterations.
- Live probe of `s3.ap-southeast-6.amazonaws.com` and `s3.dualstack.ap-southeast-6.amazonaws.com` — both resolve (A records on the plain endpoint, AAAA on dualstack) and answer HTTPS 307, the same response the existing `ap-southeast-5` endpoint gives; a fabricated `ap-southeast-99` control fails DNS as expected.
- Merge simulation into master — merged tree identical to master's, byte for byte.

The live probe also retires the 2025 objection that these domains did not resolve yet: they resolve now, and #2234 has since shipped the same entry.

@benclapp thanks for the contribution — the entry you wrote is exactly what shipped in #2234; only the PR itself is now redundant.

PR #2186, opened 2025-12-07
upgrade CI to go1.25.x

harshavardhana:


klauspost:
Do we need to be this upgrade-happy on SDKs? Following the Go supported versions seems the reasonable move.

.github/workflows/go-windows.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        go-version: [1.24.x, 1.25.x]
+        go-version: [1.25.x]
```
ferhatelmas: might want to stick to `stable` and `oldstable` aliases which would reduce churn https://github.com/actions/setup-go#version-aliases

harshavardhana:
Thought go1.26 is coming in Feb, why not move the default version? but can keep 1.24.x if needed.

PR #2216, opened 2026-03-26
Add parallel range multipart get support

joshfriend:
It is pretty common for s3 clients to use range multipart requests to s3 to [increase throughput](https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpuoverview.html). This PR adds that support.


api-get-object.go
```diff
@@ -704,3 +742,116 @@ func (c *Client) getObject(ctx context.Context, bucketName, objectName string, o
...
 }
+
+const (
+	// DefaultParallelGetChunkSize is the default size of each parallel
+	// range-GET chunk (32 MiB).
+	DefaultParallelGetChunkSize int64 = 32 << 20
+
+	// DefaultParallelGetWorkers is the default number of concurrent
+	// range-GET workers.
+	DefaultParallelGetWorkers = 8
```
joshfriend: For the use case I developed this for I was uploading large (1-3GB) objects and these settings were a good balance between memory consumption and worker count that could still saturate the network and disk IOPS I had available.

Smaller chunk sizes make more requests, contributing more to the prefix ratelimit.

Open to feedback on preferred defaults!

api-put-object.go
```diff
@@ -377,6 +378,14 @@ func (c *Client) PutObject(ctx context.Context, bucketName, objectName string, r
 		return c.putObject(ctx, bucketName, objectName, reader, size, opts)
 	}
 
+	// When NumThreads > 1 and the reader supports ReadAt, use the parallel
+	// multipart upload path for maximum throughput.
+	if opts.NumThreads > 1 {
+		if ra, ok := reader.(io.ReaderAt); ok {
+			return c.parallelPut(ctx, bucketName, objectName, ra, size, opts)
+		}
+	}
```
klauspost: `c.putObjectMultipartStream` already does parallel uploads. Seems you are changing behaviour and duplicating functionality.

api-get-options.go
```diff
@@ -48,6 +48,14 @@ type GetObjectOptions struct {
 	// https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity.html
 	Checksum bool
 
+	// ParallelChunkSize is the byte size of each range-GET request when using
+	// parallel download. Defaults to DefaultParallelGetChunkSize (32 MiB) if zero.
+	ParallelChunkSize int64
+
+	// ParallelWorkers is the number of concurrent range-GET workers when using
+	// parallel download. Defaults to DefaultParallelGetWorkers (8) if zero.
+	ParallelWorkers int
```
klauspost: Don't change defaults.

api-get-object.go
klauspost: I guess it mostly depends on the server. Usually we don't see any benefit from concurrent requests - and  it just slows down requests, but YMMV.

klauspost (changes requested):
Remove the upload, clearly your AI didn't pick up on the nuances and it breaks way too many things.  I am ok to keep the read as an opt-in.

api-get-options.go
```diff
@@ -48,6 +48,14 @@ type GetObjectOptions struct {
 	// https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity.html
 	Checksum bool
 
+	// ParallelChunkSize is the byte size of each range-GET request when using
+	// parallel download. Defaults to minPartSize if zero.
```
joshfriend, line 52: Found the size default from the PUT side and used that instead

api-get-object.go
joshfriend: I cannot get more than 85MB/s from a single-stream get, but multi-stream gets me into the ~3400MB/s range. You're right, milage will definitely vary!

joshfriend (commented):
Alright, no more PUT stuff here, and the CI lint issue is also fixed.

allanrogerr:
Resolve conflicts please @joshfriend

allanrogerr:
klauspost's third point stands as a caution rather than a change request: concurrent range GETs often give no benefit and can slow a read, depending on the server. I benchmarked against a local server and agree it is workload-dependent. That is an argument for keeping the feature opt-in, which it is.

Must fix
- Workers share one header map: `getOpts := opts` copies the struct, but `headers` is a map, so all workers share it and write to it concurrently. Give each worker its own copy of the map.

- A `Seek` during a download races the same map: The download outlives the request that started it, while the request loop keeps mutating `opts.headers` for later requests. Copy the map once, before the workers start.

- No backpressure: Each per-chunk channel is buffered and has one sender, so a send never blocks. Against a slow reader the workers run ahead and buffer the whole object instead of `numWorkers × chunkSize`. It needs a `select` on the errgroup context alongside it — once a send can block, one worker's error would otherwise leave `eg.Wait` hanging.

- Chunk length is never checked: Nothing verifies that a range GET returned the bytes that were asked for. 

- A caller-set `Range` is dropped: The parallel branch deletes it, so the same options return different bytes depending on `ParallelWorkers`: `bytes=100-199` gave me 4096 bytes on the parallel path and 100 on the single-stream path. Send ranged reads down the single-stream path instead.

- `PartNumber` plus `Range` is rejected by S3: The chunk GETs add a `Range` header on top of the `partNumber` query parameter. With `PartNumber: 1` and `ParallelWorkers: 4` I get `Cannot specify both Range header and partNumber query parameter` and zero bytes; the same read without `ParallelWorkers` returns the part. The gate needs to exclude `PartNumber` too.

Worth fixing

- `ParallelChunkSize` is not range-checked, and the per-chunk channels are allocated up front. An 8 MiB object with `ParallelChunkSize: 1` allocates millions of channels before any bytes move. Clamp to `minPartSize`, or reject more than `maxPartsCount` chunks.

- An offset change closes the pipe but does not cancel the download. Dispatched chunk requests keep running until the head-of-line chunk lands. Reading a header and closing can pull a lot of data you never wanted.

- Small objects pay a HEAD and then a full GET where the single-stream path needs one round trip.

- The concurrency test does not test concurrency: `maxConcurrent` is tracked and never asserted, so the test still passes with `numWorkers := 1`.

- Two inline test handlers drop the `fmt.Sscanf` error, while the shared `newParallelGetTestServer` checks it. An open-ended `bytes=N-`, which the fallback path does send, silently becomes one byte instead of a 4xx.

- `snowball` is not threaded into `parallelGet`. Both other `If-Match` sites skip the pin when it is set. This path pins every chunk, so a snowball endpoint would fail here but succeed on the single-stream path.

Naming and docs

- The PUT side calls the same knobs `NumThreads` and `ConcurrentStreamParts`. Worth aligning before these become public API.

- The field docs should say the fields apply to `GetObject` only, and only when the first call on the returned `Object` is a `Read` at offset 0. A leading `Stat` or `ReadAt`, or any `Seek`, falls back to a single stream. `FGetObject` and `Core.GetObject` ignore both fields, and because `StatObjectOptions` aliases this type they appear, inert, on `StatObject`. `minPartSize` is unexported, so name the actual size (16 MiB) in the comment.

- Peak memory is roughly `(ParallelWorkers + 1) × ParallelChunkSize`. Callers picking worker and chunk values are picking a memory footprint, so state it.

- The title and description say "multipart get", but the code issues HTTP byte-range GETs, not S3 `partNumber` reads. The linked page documents multipart *upload*. The relevant guidance is "Use byte-range fetches" in the S3 performance guidelines.

allanrogerr (changes requested):
PTAL @joshfriend

PR #2218, opened 2026-04-03
feat: add ObjectLocking field to BucketVersioningConfiguration

varun0630:
## Summary
- Add `ObjectLocking bool` field to `BucketVersioningConfiguration` struct
- Allows clients to enable object locking on an existing bucket via a single `PutBucketVersioning` call
- This is a MinIO extension field (`xml:",omitempty"`) so it does not affect existing S3-compatible behavior

## Test plan
- [ ] Verify existing versioning operations are unaffected
- [ ] Test `mc version enable --with-lock` enables object locking on an existing bucket

### DIFF
--- a/api-bucket-versioning.go
+++ b/api-bucket-versioning.go
@@ -91,6 +91,9 @@ type BucketVersioningConfiguration struct {
 	ExcludedPrefixes []ExcludedPrefix `xml:",omitempty"`
 	ExcludeFolders   bool             `xml:",omitempty"`
 	PurgeOnDelete    string           `xml:",omitempty"`
+	// MinIO extension - enables object locking on the bucket when set to true.
+	// Requires versioning to be enabled.
+	ObjectLocking bool `xml:",omitempty"`
 }
 
 // Various supported states
--- a/api-bucket-versioning.go
+++ b/api-bucket-versioning.go
@@ -92,13 +92,8 @@ type BucketVersioningConfiguration struct {
 	ExcludeFolders   bool             `xml:",omitempty"`
 	PurgeOnDelete    string           `xml:",omitempty"`
 	// MinIO extension - enables object locking on the bucket.
-	// Pointer so callers and the wire format can distinguish three
-	// states: nil (field absent / server does not support the
-	// extension — fall back to GetBucketObjectLockConfiguration),
-	// &false (server explicitly reports lock disabled), and &true
-	// (lock enabled). omitempty on a pointer omits only when nil, so
-	// explicit false is preserved on the wire.
-	// Requires versioning to be Enabled.
+	// Pointer so explicit false is preserved on the wire, distinct
+	// from unset. Requires versioning to be Enabled.
 	ObjectLocking *bool `xml:",omitempty"`
 }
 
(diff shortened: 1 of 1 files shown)


api-bucket-versioning.go
```diff
@@ -91,6 +91,9 @@ type BucketVersioningConfiguration struct {
 	ExcludedPrefixes []ExcludedPrefix `xml:",omitempty"`
 	ExcludeFolders   bool             `xml:",omitempty"`
 	PurgeOnDelete    string           `xml:",omitempty"`
+	// MinIO extension - enables object locking on the bucket when set to true.
+	// Requires versioning to be enabled.
+	ObjectLocking bool `xml:",omitempty"`
```
klauspost, line 96: There should be a way to determine if the field was sent, but false or if it was omitted - and we don't know if we need to make an additional call to determine Object Locking.

Maybe this is enough... You'd have to try it out - don't know enough about XML.

```suggestion
	ObjectLocking *bool `xml:",omitempty"`
```

klauspost (changes requested):

klauspost (approved):
Maybe see if you can cook down the comment. A bit excessive.

varun0630:
@kannappanr @allanrogerr would appreciate some reviews on this.

klauspost (approved):

allanrogerr:
Reviewed this as requested. My recommendation is to close this PR: the server-side change it depends on was abandoned, so the field has nothing to talk to.

**The server never reads this field.** The versioning config struct in both miniohq/aistor and minio/minio has no `ObjectLocking` field, and the `PutBucketVersioning` handler decodes the request body via `versioning.ParseConfig` into that struct, so elements it does not recognise are silently discarded. The server-side counterpart (miniohq/aistor#3788) and the mc flag (miniohq/ac#462) were both closed without merging (on 2026-05-17 and 2026-05-18 — after the approval here landed on 2026-05-14) — with the reasoning that current behavior already matches AWS.

**Verified against a live server on 2026-08-03** (`quay.io/minio/aistor/minio:latest`, image built 2026-07-24), running this branch via a `go.mod` replace:

| Probe | Result |
|---|---|
| `SetBucketVersioning{Status: Enabled, ObjectLocking: &true}` on an unlocked bucket | Accepted with no error — and object locking stays **off**: `GetObjectLockConfig` still returns "Object Lock configuration does not exist" |
| `GetBucketVersioning` after that call | `ObjectLocking` is always `nil` — the server never returns the element |
| `SetObjectLockConfig` on the same existing, versioned bucket | Succeeds — `GetObjectLockConfig` now returns `Enabled` |

The last row is the point: enabling object locking on an existing bucket already works through the standard `PutObjectLockConfiguration` API, which this SDK already exposes as `SetObjectLockConfig` — the same flow AWS uses. That is why the server PR was closed.

Merging as-is would add a permanent exported field to `v7` that no server reads, with a doc comment telling users it "enables object locking on the bucket". A caller who sets it gets a success response and no locking — a silent no-op on a data-protection control. The PR description is also out of date: it says `ObjectLocking bool`, and the code is now `*bool`.

The Go mechanics themselves are sound — I verified the pointer semantics empirically (a nil pointer is omitted from the XML; explicit `true` and `false` are both sent). If the versioning-config approach is ever revived server-side, this can reopen with a test pinning that behavior. Today there is nothing for it to talk to.

PR #2194, opened 2026-01-11
Update dependencies: replace go-ini/ini with gopkg.in/ini.v1 and adju…

Suhaibinator:
…st go.mod/go.sum

PR #2220, opened 2026-04-18
replace github.com/go-ini/ini with  gopkg.in/ini.v1

h8liu:
upgrade from 1.67.0 to 1.67.1 and run `go mod tidy`

otherwise, when this library is being used, `go mod tidy` will print warning of:

````
module declares its path as: gopkg.in/ini.v1
        but was required as: github.com/go-ini/ini
restoring github.com/go-ini/ini@v1.67.0
````


h8liu:
seems to be a duplicate of https://github.com/minio/minio-go/pull/2194/changes

PR #2221, opened 2026-04-24
Add new AWS checksums

klauspost:
"Reference": https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity-upload.html

Validated against AWS S3.


klauspost:
Obviously running mint fails since it tests against old minio version.

harshavardhana (approved):

harshavardhana:
@klauspost we can skip the error in the test if we get "invalid checksum" we can skip the test since this test is downloading OSS minio

klauspost:
Yeah. We have more tests on the server side, so it wouldn't be too much of a loss. I guess we can deal with scenarios like this later.

klauspost:
@harshavardhana Ignoring "InvalidArgument" specifically. Added a const we can flip later if we start testing on "AIStor Free".

harshavardhana (approved):

PR #2225, opened 2026-05-17
fix: accept endpoint URLs that already include a scheme

### LINKED ISSUE #2157: Error thrown when passing an s3-like endpoint URL with https:// for storage config
https://github.com/minio/minio-go/blob/9b927a4c27f9a3b3fc3cc98ccf39a51ebae7d261/utils.go#L163C2-L165C3

I am setting up a monitoring k3s cluster using grafana, mimir, loki, and alloy. For mimir, I am configuring it to use openstack ceph RGW as the backend object storage. My mimir pods were failing because I was passing my ceph RGW endpoint which was similar to` https://openstack.com:6781`, but the conditional check above was failing due to https:// passed in with the URL. This took a while to troubleshoot so I am suggesting a fix be applied here. Passing the endpoint with https:// to the loki service worked just fine.

wyf027:
## Summary
- When the endpoint string already starts with `http://` or `https://`, parse it directly instead of prefixing another scheme

Fixes #2157

## Test plan
- [x] `go test -run TestGetEndpointURL`


utils_test.go
```diff
@@ -126,6 +126,8 @@ func TestGetEndpointURL(t *testing.T) {
 		{"[::1]:443", true, "https://[::1]:443", nil, true},
 		{"[::1]:9000", false, "http://[::1]:9000", nil, true},
 		{"[::1]:9000", true, "https://[::1]:9000", nil, true},
+		{"https://ceph.example.com:6781", true, "https://ceph.example.com:6781", nil, true},
```
klauspost: No need for product names here.

wyf027:
Addressed the active review feedback in commit a4fa7ba:\n\n- removed the product-specific endpoint name from the test case\n- accept URL schemes case-insensitively and normalize the parsed scheme\n- reject endpoint URL schemes that disagree with the secure option, so the client cannot end up with request URL, signing, and default transport state out of sync\n- added mismatch coverage for https+secure=false and http+secure=true\n\nLocal formatting checks passed with gofmt and git diff --check. Go test commands are currently stalling locally without child process output after dependency download, so I did not claim a local test pass. The remaining Linux CI failure I see is the golangci-lint installer checksum mismatch, which appears unrelated to this PR code.

PR #2227, opened 2026-05-17
fix: treat HTTP 202 Accepted as success in executeMethod

### LINKED ISSUE #2223: S3 Restore Misreports Success as Failure
Hi,

I believe I've found an issue with minio's implementation of the S3 restore command. Have tested with both AWS S3 and a third party S3 implementation.

It's reproducible by executing the provided example code.

When configuring with keys / bucket and providing an s3 deep archive file to be restored upon execution it reports the following:

```
phil@Phils-MacBook-Pro examples % ./restoreobject 
2026/05/03 13:23:48 202 Accepted
```

Logging the fatal error rather than reporting "Restore request succeeded." https://github.com/minio/minio-go/blob/f29e5687f3234ca088afedc03f4096ea50d4e10b/examples/s3/restoreobject.go#L53-L58

`202 Accepted` is the expected response according to the AWS documentation: https://docs.aws.amazon.com/AmazonS3/latest/API/API_RestoreObject.html

Upon executing a second time a genuine error is returned, as the first request was accepted:

```
phil@Phils-MacBook-Pro examples % ./restoreobject
2026/05/03 13:30:48 Object restore is already in progress
```

It appears that https://github.com/minio/minio-go/pull/1621 attempted to address the 200 / 202 being acceptable status codes.

It seems however that the restore function isn't making it as far as that check, instead returning here: https://github.com/minio/minio-go/blob/f29e5687f3234ca088afedc03f4096ea50d4e10b/api-restore.go#L175-L177

Seemingly because the executeMethod function doesn't consider 202 to be one of its success codes
https://github.com/minio/minio-go/blob/f29e5687f3234ca088afedc03f4096ea50d4e10b/api.go#L729
and
https://github.com/minio/minio-go/blob/f29e5687f3234ca088afedc03f4096ea50d4e10b/api.go#L643-L647

Adding `http.StatusAccepted` to the `successStatus` list resolves the issue and generates the expected response on first / second calls.

```
phil@Phils-MacBook-Pro examples % ./restoreobject                                
Restore request succeeded.
phil@Phils-MacBook-Pro examples % ./restoreobject
2026/05/03 13:36:43 Object restore is already in progress
```

Happy to create a PR with that change, but wasn't sure if it could cause side effects with other requests.

Thanks,

Phil

wyf027:
## Summary

`RestoreObject` returns HTTP 202 Accepted on success, but `executeMethod` only treated 200/204/206 as success and surfaced `202 Accepted` as an error before `RestoreObject` could handle the response.

## Changes

- Add `http.StatusAccepted` to `successStatus`
- Add `TestSuccessStatusIncludesAccepted`

Fixes #2223


api_success_status_test.go
klauspost, line 1: Move into `api_unit_test.go`

klauspost (changes requested):

wyf027:
Addressed the review feedback in commit 2c1b367:\n\n- moved the Accepted success-status regression test into api_unit_test.go\n- made trace-errors-only and dumpHTTP use the shared successStatus set, so 202 is handled consistently with the executeMethod success path\n\nI attempted local validation with go test ./... and then a focused package test, but the local go command stalled without child process output after dependency downloads, so I stopped it rather than leave it hanging. The remaining Linux CI failure I see is the golangci-lint installer checksum mismatch, which appears unrelated to this PR code.

klauspost (approved):
lgtm

pgreenland:
@wyf027 Think you may have closed this before it got merged?

PR #2231, opened 2026-05-20
add support to http3

nathanlaceyraft:
HTTP/3 (quic) utilizes UDP instead of TCP, which makes web performance more resilient and consistent, especially for users on unreliable connections.

S3 doesn't natively support HTTP/3
To use HTTP/3 for S3 objects, you must place an [Amazon CloudFront](https://aws.amazon.com/blogs/aws/new-http-3-support-for-amazon-cloudfront/) distribution in front of your S3 bucket.

Alternatively, you can self host Minio with ISTIO to add HTTP/3 support.

minio-go currently only supports http/1.

This PR adds support for for HTTP/3 (QUIC)

It does so if the Env Variable ENABLE_HTTP3_CLIENT = true, and if you are connecting to https server.

pkg/protocolswitchingclient/client.go will async make a http/1 and http/3 request at the same time, if http/3 returns, it will default to alway's using http/3 going forward, unless there is a problem, like a firewall change, at which point it will fall back to http/1 mode.

if HTTP/3 isn't requested, protocolswitchingclient will escape and alway's just using the http/1 client


pkg/protocolswitchingclient/client_test.go
```diff
@@ -0,0 +1,246 @@
...
+		log.Fatal(err)
+	}
+
+	keyFile.Write(keyOut.Bytes())
+
+	// Don't forget to close it!
+	if err := keyFile.Close(); err != nil {
+		log.Fatal(err)
+	}
+	fmt.Println("cert", certFile.Name(), "key", keyFile.Name())
```
klauspost, line 74:
```suggestion
```

pkg/protocolswitchingclient/client.go
```diff
@@ -0,0 +1,240 @@
...
+	// StateUnknown when we don't currently know if server will support http or http3
+	StateUnknown = iota // 0
+	// StateHttp server is currently known to support http
+	StateHttp // 1
+	// StateHttp3 server is currently known to support http3
+	StateHttp3 // 2
+)
+
+// HttpState one of the 'enum' values above
+type HttpState int
```
klauspost, line 24: No global state.

api.go
```diff
@@ -46,9 +46,16 @@ import (
...
+	"github.com/minio/minio-go/v7/pkg/protocolswitchingclient"
 	"golang.org/x/net/publicsuffix"
 )
 
+var enableHttp3 bool
+
+func Init(){
+	enableHttp3 = strings.EqualFold(os.Getenv("ENABLE_HTTP3_CLIENT"),"true")
+}
+
```
klauspost, line 58: No global options. No env vars. Must be configured in `Options`.

klauspost (changes requested):
Seems to require formatting.

Since we allow you to provide a custom `http.Roundtripper` and http3 provides it in [http3.Transport](https://pkg.go.dev/github.com/quic-go/quic-go/http3#Transport) why can't you just do that?

klauspost:
Honestly, this is so low quality I think you should just start over.

Closing to not waste other peoples time.

PR #2233, opened 2026-05-27
Add RDMA / NVIDIA GPU Direct Storage support

harshavardhana:
## Summary

- Adds opt-in RDMA dispatch to `*Client.PutObject` and `*Client.GetObject` via [minio-cpp](https://github.com/minio/minio-cpp)'s stable C ABI (`libminiocpp.so`). No new public methods, no separate `Client` class. Strictly opt-in: existing callers see zero behavior change.
- Gated behind the `rdma` Go build tag — default builds carry no cgo dependency and no need for `libminiocpp.so`.
- New CI workflow (`go-rdma.yml`) installs `libminiocpp.so` from minio-cpp `main` on Linux amd64 + arm64, then compiles minio-go and `cmd/rdma-test` with `-tags=rdma`. Validates that the cgo bindings stay in lockstep with minio-cpp's C ABI.

To use:

```go
client, _ := minio.New(endpoint, &minio.Options{
    Creds:      ...,
    EnableRDMA: true,
})

// Buffer-mode PUT — selects RDMA, falls back to HTTP-from-buf on decline.
buf := make([]byte, size)
_, err := client.PutObject(ctx, bucket, object,
    bytes.NewReader(buf), int64(size), minio.PutObjectOptions{RDMABuffer: unsafe.Pointer(&buf[0])})

// Buffer-mode GET — writes directly into the caller's buffer.
n, err := client.GetObject(ctx, bucket, object,
    minio.GetObjectOptions{RDMABuffer: unsafe.Pointer(&buf[0]), RDMABufferLen: size})
```

`libminiocpp.so` must be loadable at runtime (`LD_LIBRARY_PATH` or `MINIOCPP_LIB`). Without it, calling the buffer-mode path returns `ErrRDMANotCompiled` / `ErrRDMANotConnected`. Default `go build` (no `-tags=rdma`) compiles `rdma_stub.go` and has no native dependency.

## Test plan

- [x] Default `go.yml` matrix (Go 1.24, 1.25 × Linux) still passes
- [x] New `go-rdma.yml` matrix (Linux amd64, Linux arm64) builds minio-go and `cmd/rdma-test` against the freshly built `libminiocpp.so`
- [x] `go build ./...` succeeds without `-tags=rdma`

### DIFF
--- a/.github/workflows/go-rdma.yml
+++ b/.github/workflows/go-rdma.yml
@@ -1,0 +1,73 @@
+name: Build (RDMA)
+
+on:
+  pull_request:
+    branches:
+    - master
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    name: Build with -tags=rdma against minio-cpp (${{ matrix.config.arch }})
+    runs-on: ${{ matrix.config.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        config:
+        - { os: ubuntu-latest,    arch: "amd64" }
+        - { os: ubuntu-24.04-arm, arch: "arm64" }
+
+    steps:
+    - name: Checkout minio-go
+      uses: actions/checkout@v4
+      with:
+        path: "minio-go"
+
+    - name: Checkout minio-cpp
+      uses: actions/checkout@v4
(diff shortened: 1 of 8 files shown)


klauspost (approved):
Nice and clean! 👍🏼

Issue #1761, opened 2023-01-20
credential_process works differently than AWS cli


yan12125:
## NOTE
If this case is urgent, please subscribe to [Subnet](https://min.io/pricing) so that our 24/7 support team may help you faster.

## Expected Behavior

minio-go reads `credential_process` from `~/.aws/config`, like what the official AWS cli does: https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-sourcing-external.html

## Current Behavior

minio-go reads `credential_process` from `~/.aws/credentials`, not `~/.aws/config`. See https://github.com/minio/minio-go/pull/1716.

## Possible Solution

Read `~/.aws/config` for `credential_process`

## Steps to Reproduce (for bugs)

1. Configure `credential_process` following AWS docs
2. Try to use restic (a backup tool using minio-go) to access a backup repository on S3

## Context

As mebtioned above, I'd like to use restic with `credential_process`

## Regression

Not a regression

## Your Environment

* Version used (`minio --version`): minio-go 7.0.46
https://github.com/restic/restic/blob/v0.15.0/go.mod#L17
* Server setup and configuration: AWS S3
* Operating System and version (`uname -a`): 
```
Linux archlinux 5.15.89-1-lts #1 SMP Wed, 18 Jan 2023 11:24:46 +0000 x86_64 GNU/Linux
```

harshavardhana:
Feel free to send a fix, we should support both files.

umka1332:
Also credential process can work without parameters, but implementation in https://github.com/minio/minio-go/blob/master/pkg/credentials/file_aws_credentials.go#L111 fails when there is no parameters.

umka1332:
Here is [official AWS documentation](https://docs.aws.amazon.com/sdkref/latest/guide/feature-process-credentials.html#:~:text=credential_process%20%3D%20/Users/username/process.sh) where you can see an example of credential process without parameters.

jiuker:
fixed by #2257

PR #2236, opened 2026-06-06
fix: changed to config and fixed arg len #1761

Amr-Shams:


klauspost:
Without any more context this just seems like a breaking change.

Issue #2246, opened 2026-07-03
GetObject / FGetObject don't verify downloaded data integrity


Amr-Shams:
So I was reading through the download path and noticed that neither `GetObject` nor `FGetObject` verify that what we actually received matches what the server said we'd get.

The server sends back `ETag` and sometimes `x-amz-checksum-*` headers- i know not all of them support headers but can be configured in a standard way , but we never check the downloaded bytes against them. If there's corruption mid-transfer ,flipped bits, truncated response, whatever. we just silently accept it and the caller has no idea.

The fix would be something like wrapping the reader with a hash-computing layer and comparing against the response header after the transfer completes. Shouldn't require touching the public API, could be opt-in via `GetObjectOptions` to avoid breaking anything.

klauspost:
Note that Etag can't be used reliably (and MD5 is slow anyway). Checksum sometimes, as many will be "compound" and therefore not checkable.

Amr-Shams:
fair point, etag on multipart is a hash of hashes so yeah not reliable, and compound checksums can't be verified client side, makes sense.

but the case i'm thinking about is simpler, single-part objects where a full-object `x-amz-checksum-sha256` or `x-amz-checksum-crc32` actually lands in the response headers. those are verifiable. the proposal would just skip silently when the header isn't there or when the etag looks like a multipart hash (`-N` suffix), and only verify when we have something we can actually check.

klauspost:
Yes. Not saying it isn't possible. Just that having checksums doesn't automatically mean it can be validated.

PR #2247, opened 2026-07-03
feat: add request/response middleware pipeline

Amr-Shams:
## Summary

Adds a middleware pipeline to the minio-go client. You hook into
four phases: Initialize, Serialize, Finalize, Deserialize.

Pulled from AWS SDK's middleware.Stack. Same shape.

## Architecture

The pipeline lives inside `executeMethod()` and `newRequest()`.
No public API changes. Each middleware implements `Middleware`
(`ID() string`) plus one or more phase interfaces:

InitializeMiddleware  → Initialize(ctx, execCtx)
SerializeMiddleware   → Serialize(ctx, execCtx, req)
FinalizeMiddleware    → Finalize(ctx, execCtx, req)
DeserializeMiddleware → Deserialize(ctx, execCtx, resp, err)

`ExecutionContext` carries `Operation`, `BucketName`,
`ObjectName`. Scope middleware logic to specific S3 operations
(verify integrity on GetObject, skip on ListBuckets).

### Execution order

Initialize  (once, before retry loop)
retry loop
  Serialize    once per attempt, before signing
  Finalize     once per attempt, before do()
  do()
  Deserialize  once per attempt, after response

### Error handling

- Deserialize middleware errors stack with the transport error
  via `errors.Join`. Any error aborts the request, even on 2xx.
  Use this for response validation.
- Initialize, Serialize, and Finalize errors short-circuit.

## Why this instead of the old way

Before this, if you wanted custom headers, request logging, or
response verification, you had three options:

1. Wrap the whole client. Hundreds of methods to duplicate.
2. Swap `http.RoundTripper`. You lose bucket name, operation
   type, anything S3-specific.
3. Fork the SDK.

## Related

Prerequisite for #2246 (GetObject/FGetObject integrity verification).
The verifier will be a DeserializeMiddleware that wraps resp.Body
with a hash reader and checks against ETag or x-amz-checksum-*
headers.

## Summary by CodeRabbit

* **New Features**
  * Added configurable client-side middleware for the S3/HTTP request lifecycle, with hooks for initialization, request serialization/signing, finalization, and response deserialization.
  * Middleware receives an execution context (HTTP method, bucket/object info, and query parameters); initialization can adjust the request context.
  * Middleware can be registered via client options or `AddMiddleware`.
* **Bug Fixes**
  * Deserialize middleware failures are now treated as request failures, even if the HTTP status indicates success.
* **Tests**
  * Added an integration test covering hook order, single invocation per call, execution context capture, and header injection.

### DIFF
--- a/api.go
+++ b/api.go
@@ -563,9 +563,6 @@ type requestMetadata struct {
 	trailer          http.Header // (http.Request).Trailer. Requires v4 signature.
 
 	expect200OKWithError bool
-
-	// s3 operation name
-	s3Operation string
 }
 
 // dumpHTTP - dump HTTP request and response.
@@ -721,16 +718,11 @@ func (c *Client) executeMethod(ctx context.Context, method string, metadata requ
 		metadata.trailer.Set(metadata.addCrc.Key(), base64.StdEncoding.EncodeToString(crc.Sum(nil)))
 	}
 
-	opt := metadata.s3Operation
-	if opt == "" {
-		opt = resolveS3Operation(method, metadata)
-	}
-	metadata.s3Operation = opt
-
 	execCtx := ExecutionContext{
-		Operation:  opt,
-		BucketName: metadata.bucketName,
-		ObjectName: metadata.objectName,
+		Method:      method,
+		BucketName:  metadata.bucketName,
+		ObjectName:  metadata.objectName,
+		QueryValues: metadata.queryValues,
 	}
 
 	for _, m := range c.middlewares {
@@ -782,18 +774,19 @@ func (c *Client) executeMethod(ctx context.Context, method string, metadata requ
 			}
 			return nil, err
 		}
+
 		var mwError error
 		for _, m := range c.middlewares {
 			if deserializeM, ok := m.(DeserializeMiddleware); ok {
-				if dErr := deserializeM.Deserialize(ctx, execCtx, res, err); dErr != nil {
+				if dErr := deserializeM.Deserialize(ctx, execCtx, res); dErr != nil {
 					mwError = errors.Join(mwError, dErr)
 				}
 			}
 		}
+
 		if mwError != nil {
 			err = errors.Join(err, mwError)
 		}
-
 		success := successStatus.Contains(res.StatusCode)
 		if success && !metadata.expect200OKWithError {
 			if mwError != nil {
@@ -946,12 +939,10 @@ func (c *Client) newRequest(ctx context.Context, method string, metadata request
 		return nil, err
 	}
 	execCtx := ExecutionContext{
-		Operation:  metadata.s3Operation,
-		BucketName: metadata.bucketName,
-		ObjectName: metadata.objectName,
-	}
-	if execCtx.Operation == "" {
-		execCtx.Operation = resolveS3Operation(method, metadata)
+		Method:      method,
+		BucketName:  metadata.bucketName,
+		ObjectName:  metadata.objectName,
+		QueryValues: metadata.queryValues,
 	}
 	for _, m := range c.middlewares {
 		if serializeM, ok := m.(SerializeMiddleware); ok {
@@ -1234,152 +1225,4 @@ func (c *Client) GetCreds() (credentials.Value, error) {
 		return credentials.Value{}, errors.New("no credentials provider")
 	}
 	return c.credsProvider.GetWithContext(c.CredContext())
-}
-
-func resolveS3Operation(method string, metadata requestMetadata) string {
-	if metadata.s3Operation != "" {
-		return metadata.s3Operation
-	}
-
-	hasQuery := func(key string) bool {
-		return metadata.queryValues.Get(key) != "" || metadata.queryValues.Has(key)
-	}
-
-	isObject := metadata.objectName != ""
-
-	switch method {
-	case http.MethodGet:
-		if isObject {
-			if hasQuery("tagging") {
-				return "GetObjectTagging"
-			}
-			if hasQuery("acl") {
-				return "GetObjectAcl"
-			}
-			if hasQuery("legal-hold") {
-				return "GetObjectLegalHold"
-			}
-			if hasQuery("retention") {
-				return "GetObjectRetention"
-			}
-			return "GetObject"
-		} else {
-			if hasQuery("lifecycle") {
-				return "GetBucketLifecycle"
-			}
-			if hasQuery("policy") {
-				return "GetBucketPolicy"
-			}
-			if hasQuery("notification") {
-				return "GetBucketNotification"
-			}
-			if hasQuery("cors") {
-				return "GetBucketCors"
-			}
-			if hasQuery("encryption") {
-				return "GetBucketEncryption"
-			}
-			if hasQuery("tagging") {
-				return "GetBucketTagging"
-			}
-			if hasQuery("versioning") {
-				return "GetBucketVersioning"
-			}
-			if metadata.bucketName != "" {
-				return "ListObjectsV2"
-			}
-			return "ListBuckets"
-		}
-
-	case http.MethodPut:
-		if isObject {
-			if hasQuery("tagging") {
-				return "PutObjectTagging"
-			}
-			if hasQuery("acl") {
-				return "PutObjectAcl"
-			}
-			if hasQuery("legal-hold") {
-				return "PutObjectLegalHold"
-			}
-			if hasQuery("retention") {
-				return "PutObjectRetention"
-			}
-			return "PutObject"
-		} else {
-			if hasQuery("lifecycle") {
-				return "PutBucketLifecycle"
-			}
-			if hasQuery("policy") {
-				return "PutBucketPolicy"
-			}
-			if hasQuery("notification") {
-				return "PutBucketNotification"
-			}
-			if hasQuery("cors") {
-				return "PutBucketCors"
-			}
-			if hasQuery("encryption") {
-				return "PutBucketEncryption"
-			}
-			if hasQuery("tagging") {
-				return "PutBucketTagging"
-			}
-			if hasQuery("versioning") {
-				return "PutBucketVersioning"
-			}
-			return "CreateBucket"
-		}
-
-	case http.MethodDelete:
-		if isObject {
-			if hasQuery("tagging") {
-				return "DeleteObjectTagging"
-			}
-			return "DeleteObject"
-		} else {
-			if hasQuery("lifecycle") {
-				return "DeleteBucketLifecycle"
-			}
-			if hasQuery("policy") {
-				return "DeleteBucketPolicy"
-			}
-			if hasQuery("cors") {
-				return "DeleteBucketCors"
-			}
-			if hasQuery("encryption") {
-				return "DeleteBucketEncryption"
-			}
-			if hasQuery("tagging") {
-				return "DeleteBucketTagging"
-			}
-			return "DeleteBucket"
-		}
-
-	case http.MethodHead:
-		if isObject {
-			return "HeadObject"
-		}
-		return "HeadBucket"
-
-	case http.MethodPost:
-		if isObject {
-			if hasQuery("uploads") {
-				return "CreateMultipartUpload"
-			}
-			if hasQuery("restore") {
-				return "RestoreObject"
-			}
-			if hasQuery("select") {
-				return "SelectObjectContent"
-			}
-			return "PostObject"
-		} else {
-			if hasQuery("delete") {
-				return "DeleteObjects"
-			}
-		}
-	}
-
-	return method
 }
--- a/api.go
+++ b/api.go
@@ -118,6 +118,9 @@ type Client struct {
 	rdmaOnce    sync.Once         //nolint:unused
 	rdmaHandle  *rdmaClientHandle //nolint:unused
 	rdmaInitErr error             //nolint:unused
+
+	// Middlewares
+	middlewares []Middleware
 }
 
 // Options for New method
@@ -170,6 +173,9 @@ type Options struct {
 	// when the caller supplies PutObjectOptions.RDMABuffer / GetObjectOptions.RDMABuffer.
 	// No-op unless built with -tags=rdma.
 	EnableRDMA bool
+
+	// Middlewares to run on s3 operations.
+	Middlewares []Middleware
 }
 
 // Global constants.
@@ -218,6 +224,11 @@ func New(endpoint string, opts *Options) (*Client, error) {
 	}
 
 	return clnt, nil
+}
+
+// AddMiddleware adds a middleware to the chain.
+func (c *Client) AddMiddleware(m Middleware) {
+	c.middlewares = append(c.middlewares, m)
 }
 
 // EndpointURL returns the URL of the S3-compatible endpoint that this client connects to.
@@ -340,6 +351,7 @@ func privateNew(endpoint string, opts *Options) (*Client, error) {
 		clnt.maxRetries = opts.MaxRetries
 	}
 
+	clnt.middlewares = opts.Middlewares
 	// Return.
 	return clnt, nil
 }
@@ -551,6 +563,9 @@ type requestMetadata struct {
 	trailer          http.Header // (http.Request).Trailer. Requires v4 signature.
 
 	expect200OKWithError bool
+
+	// s3 operation name
+	s3Operation string
 }
 
 // dumpHTTP - dump HTTP request and response.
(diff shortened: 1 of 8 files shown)


api.go
```diff
@@ -1164,3 +1235,151 @@ func (c *Client) GetCreds() (credentials.Value, error) {
 	}
 	return c.credsProvider.GetWithContext(c.CredContext())
 }
+
+func resolveS3Operation(method string, metadata requestMetadata) string {
```
Amr-Shams: so i don't need to break the apis

api.go
```diff
@@ -740,10 +782,24 @@ func (c *Client) executeMethod(ctx context.Context, method string, metadata requ
...
+			err = errors.Join(err, mwError)
+		}
 
 		success := successStatus.Contains(res.StatusCode)
 		if success && !metadata.expect200OKWithError {
-			// We do not expect 2xx to return an error return.
+			if mwError != nil {
+				closeResponse(res)
+				return nil, err
+			}
```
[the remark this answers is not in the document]
Amr-Shams: this is not a short circuit and the more lean error handling

klauspost:
Honestly, this seems overengineered for what it does. Usually I'd leave the overengineering to AWS.

Also the list-of-operations is a maintenance chore that I'd rather avoid.

Amr-Shams:
> Honestly, this seems overengineered for what it does. Usually I'd leave the overengineering to AWS.
> 
> Also the list-of-operations is a maintenance chore that I'd rather avoid.

yea i see what you're pointing at. the 4-phase interface stack + type assertions is a lot of ceremony for what it does. but tbh the concept is useful, RoundTripper hacks lose context (bucket name, operation type, etc.).
what if i cut it down to just two interfaces: PreRequest and PostResponse? still gets the job done but the boilerplate drops by ~70%.
smt like:
type PreRequest interface {
    PreRequest(ctx, ExecutionContext, *http.Request) error
}
type PostResponse interface {
    PostResponse(ctx, ExecutionContext, *http.Response, error) error
}
slice of each in Options, call them in order. think that'd fly?

middleware.go
```diff
@@ -0,0 +1,60 @@
+package minio
```
harshavardhana, line 18: fix the copyright headers

harshavardhana (changes requested):

harshavardhana:
@coderabbitai help

Amr-Shams:
@harshavardhana PTAL

api.go
```diff
@@ -340,6 +351,7 @@ func privateNew(endpoint string, opts *Options) (*Client, error) {
 		clnt.maxRetries = opts.MaxRetries
 	}
 
+	clnt.middlewares = opts.Middlewares
```
allanrogerr, line 354: The client keeps the caller's slice header rather than a copy, so if the caller reuses or edits its own `Middlewares` slice after `New` the edit lands in this client's active chain — and it lands while in-flight requests may already be ranging over it at 728, 760, 779 and 947.

```suggestion
	clnt.middlewares = append([]Middleware(nil), opts.Middlewares...)
```

api.go
```diff
@@ -220,6 +226,11 @@ func New(endpoint string, opts *Options) (*Client, error) {
 	return clnt, nil
 }
 
+// AddMiddleware adds a middleware to the chain.
+func (c *Client) AddMiddleware(m Middleware) {
+	c.middlewares = append(c.middlewares, m)
+}
```
allanrogerr, line 232: `AddMiddleware` appends to the same slice that `executeMethod` and `newRequest` range over (728, 760, 779, 947), with no synchronisation on either side, so calling it once the client is live is a data race. Documenting it as setup-only is the cheap option; a mutex would mean taking a lock on all four read paths. Consider which you'd rather have.

```suggestion
// AddMiddleware adds a middleware to the chain.
//
// It is not safe for concurrent use: call it during client setup, before the
// client issues any request.
func (c *Client) AddMiddleware(m Middleware) {
	c.middlewares = append(c.middlewares, m)
}
```

middleware.go
```diff
@@ -0,0 +1,77 @@
+/*
+ * MinIO Go Library for Amazon S3 Compatible Cloud Storage
+ * Copyright 2015-2017 MinIO, Inc.
```
allanrogerr, line 3: This file is new in this PR, so a 2015-2017 range doesn't describe it. Other files added to the repo recently use the current year.

```suggestion
 * Copyright 2026 MinIO, Inc.
```

middleware.go
```diff
@@ -0,0 +1,77 @@
...
+
+package minio
+
+import (
+	"context"
+	"net/http"
+	"net/url"
+)
+
+// ExecutionContext provides context metadata about the S3 operation currently running.
```
allanrogerr, line 26: The struct is copied per call, which makes it look inert, but `QueryValues` is a map — it points at the same values `makeTargetURL` reads, and the request metadata is reused across retries. A middleware calling `Set` or `Del` on it is editing shared request state, which is surprising enough to be worth a line here.

```suggestion
// ExecutionContext provides context metadata about the S3 operation currently running.
//
// It is passed by value, but QueryValues is the same url.Values the request
// URL is built from, not a copy: treat it as read-only.
```

middleware.go
```diff
@@ -0,0 +1,77 @@
...
+	QueryValues url.Values // e.g. ?policy, ?tagging, ?versioning
+}
+
+// Middleware is the base interface for all pipeline interceptors.
+type Middleware interface {
+	ID() string
+}
+
+// InitializeMiddleware runs BEFORE the HTTP request is built.
+// It can mutate the context or alter metadata.
```
allanrogerr, line 40: `execCtx` arrives by value, so assigning to `Method`, `BucketName` or `ObjectName` inside `Initialize` is discarded — only the returned context survives, which makes "alter metadata" misleading. Separately: a middleware that returns `(nil, nil)` here puts a nil context straight into `newRetryTimer`, which calls `Done()` on it and panics the request. Stating the non-nil requirement is cheaper than a runtime check.

```suggestion
// InitializeMiddleware runs BEFORE the HTTP request is built.
// It can derive a new context. execCtx is a copy, so assigning to its fields
// has no effect outside the middleware. Implementations must return a non-nil
// context; returning ctx unchanged is valid.
```

middleware.go
```diff
@@ -0,0 +1,77 @@
...
+// NOTE: This fires on every retry attempt. Only Initialize runs once.
+type SerializeMiddleware interface {
+	Middleware
+	Serialize(ctx context.Context, execCtx ExecutionContext, req *http.Request) error
+}
+
+// FinalizeMiddleware runs AFTER the request is signed and ready to go out.
+// Ideal for logging request sizes, outbound traffic, or tracing.
+//
+// NOTE: This fires on every retry attempt. Only Initialize runs once.
```
allanrogerr, line 61: Signing happens inside `newRequest` (around 956-1001) and `Finalize` runs after it returns, at 760. Since the phase hands out a mutable `*http.Request`, it is easy to set a header here and get a signature mismatch back from the server with nothing pointing at the cause.

```suggestion
// FinalizeMiddleware runs AFTER the request is signed and ready to go out.
// Ideal for logging request sizes, outbound traffic, or tracing.
//
// NOTE: This fires on every retry attempt. Only Initialize runs once.
// The request is already signed, so mutating headers here invalidates the
// signature; use Serialize for changes that need to be signed.
```

middleware.go
```diff
@@ -0,0 +1,77 @@
...
+	Finalize(ctx context.Context, execCtx ExecutionContext, req *http.Request) error
+}
+
+// DeserializeMiddleware runs AFTER the HTTP response is received from the server.
+// It can inspect status codes, headers, or modify/log errors.
+//
+// NOTE: This fires on every retry attempt. Only Initialize runs once.
+// Errors from all Deserialize middleware stack via errors.Join with the transport
+// error. If any middleware returns a non-nil error, the request aborts even on
+// 2xx responses.
```
allanrogerr, line 73: Two corrections and an addition. Transport failures are handled at api.go:770-776, before this loop, so `Deserialize` never sees one — a middleware registered against a refused dial records zero calls, and the transport error joined at 788 is always nil. The `resp.Body` note is new: the body is still unread here, so a middleware that reads it consumes it for the caller. I also dropped "the request aborts even on 2xx responses" rather than reword it, because which paths abort is exactly the open question in my other comment — worth restating once you've settled that.

```suggestion
// NOTE: This fires on every attempt that produced a response. Only Initialize
// runs once; a transport failure short-circuits before this phase, so
// Deserialize never observes one. resp.Body has not been read yet, so
// consuming it here consumes it for the caller as well. Errors from all
// Deserialize middleware are joined via errors.Join.
```

middleware_test.go
```diff
@@ -0,0 +1,133 @@
+/*
+ * MinIO Go Library for Amazon S3 Compatible Cloud Storage
+ * Copyright 2015-2017 MinIO, Inc.
```
allanrogerr, line 3: Same as `middleware.go`: the file is new in this PR, so the 2015-2017 range doesn't describe it.

```suggestion
 * Copyright 2026 MinIO, Inc.
```

middleware_test.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	if len(mw.finalized) != 1 {
+		t.Fatalf("Expected 1 finalized call, got %d", len(mw.finalized))
+	}
+
+	if len(mw.deserialized) != 1 {
+		t.Fatalf("Expected 1 deserialized call, got %d", len(mw.deserialized))
+	}
+}
```
allanrogerr, line 133: The error contract the description sets out — the first three phases stopping the request before it goes out, and a `Deserialize` error aborting an otherwise successful response — isn't asserted anywhere, so a regression in any of it would stay green. This table covers all four phases, passes against the branch as it stands, and needs no import changes. It fails if the `Finalize` hook is removed, so it pins the behaviour rather than just exercising it. It deliberately has no non-2xx row, because that case does not pass today (see my comment on `executeMethod`) — worth adding once that is settled. Consider and test this.

```suggestion
}

type phaseError struct{ phase string }

func (e phaseError) Error() string { return "middleware " + e.phase + " failed" }

type phaseErrMiddleware struct {
	mu     sync.Mutex
	failOn string
	phases []string
}

func (m *phaseErrMiddleware) ID() string { return "phase-err-middleware" }

func (m *phaseErrMiddleware) record(phase string) error {
	m.mu.Lock()
	defer m.mu.Unlock()
	m.phases = append(m.phases, phase)
	if m.failOn == phase {
		return phaseError{phase: phase}
	}
	return nil
}

func (m *phaseErrMiddleware) Initialize(ctx context.Context, _ ExecutionContext) (context.Context, error) {
	return ctx, m.record("initialize")
}

func (m *phaseErrMiddleware) Serialize(_ context.Context, _ ExecutionContext, _ *http.Request) error {
	return m.record("serialize")
}

func (m *phaseErrMiddleware) Finalize(_ context.Context, _ ExecutionContext, _ *http.Request) error {
	return m.record("finalize")
}

func (m *phaseErrMiddleware) Deserialize(_ context.Context, _ ExecutionContext, _ *http.Response) error {
	return m.record("deserialize")
}

func TestMiddlewareErrorPropagation(t *testing.T) {
	for _, tc := range []struct {
		failOn       string
		wantPhases   string
		wantRequests int
	}{
		{failOn: "initialize", wantPhases: "initialize", wantRequests: 0},
		{failOn: "serialize", wantPhases: "initialize,serialize", wantRequests: 0},
		{failOn: "finalize", wantPhases: "initialize,serialize,finalize", wantRequests: 0},
		{failOn: "deserialize", wantPhases: "initialize,serialize,finalize,deserialize", wantRequests: 1},
	} {
		t.Run(tc.failOn, func(t *testing.T) {
			var requestsMu sync.Mutex
			requests := 0
			server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
				requestsMu.Lock()
				requests++
				requestsMu.Unlock()
				w.WriteHeader(http.StatusOK)
			}))
			defer server.Close()

			mw := &phaseErrMiddleware{failOn: tc.failOn}
			client, err := New(strings.TrimPrefix(server.URL, "http://"), &Options{
				Creds:       credentials.NewStaticV4("mockkey", "mocksecret", ""),
				Secure:      false,
				Region:      "us-east-1",
				Middlewares: []Middleware{mw},
			})
			if err != nil {
				t.Fatalf("Failed to create minio client: %v", err)
			}

			_, err = client.GetBucketPolicy(context.Background(), "mybucket")
			if err == nil || !strings.Contains(err.Error(), phaseError{phase: tc.failOn}.Error()) {
				t.Fatalf("Expected the %s error to reach the caller, got %v", tc.failOn, err)
			}

			mw.mu.Lock()
			defer mw.mu.Unlock()
			if got := strings.Join(mw.phases, ","); got != tc.wantPhases {
				t.Errorf("Expected phases %q, got %q", tc.wantPhases, got)
			}
			requestsMu.Lock()
			defer requestsMu.Unlock()
			if requests != tc.wantRequests {
				t.Errorf("Expected %d server request(s), got %d", tc.wantRequests, requests)
			}
		})
	}
}
```

api.go
```diff
@@ -741,9 +775,24 @@ func (c *Client) executeMethod(ctx context.Context, method string, metadata requ
...
+			err = errors.Join(err, mwError)
+		}
 		success := successStatus.Contains(res.StatusCode)
 		if success && !metadata.expect200OKWithError {
-			// We do not expect 2xx to return an error return.
+			if mwError != nil {
+				closeResponse(res)
+				return nil, err
+			}
 			return res, nil
```
allanrogerr, line 796: A `Deserialize` error only reaches the caller on a plain 2xx. Everywhere else it is discarded: the value joined here is overwritten by `io.ReadAll` at 801 and then by `err = errResponse` at 824, and the `apiErr == nil` branch at 818 returns `res, nil`. Two measurements against this branch —

- 404 with a middleware returning a sentinel: `errors.Is(err, sentinel)` is false; the caller sees only `NoSuchBucket`.
- `RemoveObjects` (which sets `expect200OKWithError`, api-remove.go:638) returning 200: the hook runs once and `RemoveObjects` reports no error at all, so the failure is invisible on a 2xx too.

Worth knowing before picking a direction: the obvious repair — joining the middleware error into `errResponse` — has a trap. `ToErrorResponse` (api-error-response.go:79) is a bare type switch on `ErrorResponse` with a `default` returning an empty struct, so it does not unwrap, and `errors.Join` wraps even a single error. Any join on that path makes `ToErrorResponse(err).Code` come back `""` for every caller that inspects it. That is already observable on the 2xx path today: with a failing `Deserialize` middleware and a 200, the returned error is an `*errors.joinError` and the code reads empty. Note this also makes the join at 788 redundant in its own right — `err` is provably nil here, since `c.do`'s error paths at 770-776 either retry or return.

So the choice looks like: surface the middleware error only where there is no S3 error to report, or teach `ToErrorResponse` to try `errors.As` first and then join freely. Both seem like calls for you to make rather than something to patch inline.

middleware.go
allanrogerr: Correction to my own suggestion above — the "use Serialize for changes that need to be signed" half is incomplete, and I would rather flag that than let it bite you.

`Serialize` runs at `api.go:949`, but several `Header.Set` calls come after it in the same function and overwrite unconditionally: `x-amz-s3session-token` (963), `setUserAgent` (1005), the `metadata.customHeader` loop (1009), `Content-Md5` (1031), and `X-Amz-Content-Sha256` (1037 and 1077). A Serialize middleware that sets any of those header names is silently discarded before signing. Serialize also sees `req.Body == nil` and `ContentLength == 0`, since both are assigned at 1017-1023, after the hook runs.

So the accurate version is narrower than what I wrote: Serialize is the right phase for headers the SDK does not set itself, and neither phase is a place to touch the SDK-managed ones or to inspect the body. Worth folding into the doc if you take the suggestion.

api.go
allanrogerr: One case I left out of the list above, and it happens to be the one that most constrains the fix, so it is worth adding.

A retryable non-2xx discards the middleware error on every attempt, not just once. Measured on this branch with a 503/`SlowDown` handler and `MaxRetries: 3`: 3 server hits, 3 `Deserialize` calls, and `errors.Is(err, sentinel)` false — the caller sees only `slow down`.

That rules out the shape I would otherwise have suggested, hoisting the abort above the `success` check, because it would turn a retryable throttle into an immediate failure. Whatever direction you pick needs to keep the retry path intact and still not lose the error once retries are exhausted.

PR #2251, opened 2026-07-17
fix: enable TraceOn will cause OOM when putting a large object

### LINKED ISSUE #1771: Enable TraceOn will cause OOM
Re-produce:

```go
package main

import (
	"context"
	log "github.com/sirupsen/logrus"
	"net/http"
	_ "net/http/pprof"
	"os"
	"time"
)

func main() {
	var cliOpts = minio.Options{Creds: credentials.NewEnvMinio(), Secure: false}
	//cli, err := minio.New("oss-cn-shanghai.aliyuncs.com", &cliOpts)
	cli, err := minio.New("127.0.0.1:9000", &cliOpts)
	if err != nil {
		log.Errorf("connection error %v", err)
		return
	}
	cli.TraceOn(os.Stdout)
	opts := minio.PutObjectOptions{DisableMultipart: true}
	info, err := cli.FPutObject(context.Background(), "oom-poc", "random.bin", "/Users/mingyang/Downloads/random.bin", opts)
	if err != nil {
		log.Errorf("put error %v", err)
		return
	}
	log.Infof("put success %v", info)
	log.Println(http.ListenAndServe("localhost:6060", nil))
	time.Sleep(1 * time.Hour)
	return
}
```

If I turn on `TraceOn(...)`, it can consume a lot of memory, resulting in an OOM error, when I `PutObject` with a large file. I think this is because when we use the `DumpRequestOut` function, it loads the entire request body into memory, even though only the HTTP header is needed. Check out this line of code in the dump.go file on: 
https://github.com/golang/go/blob/master/src/net/http/httputil/dump.go#L157-L167

Do you think we could use a more stream-based approach instead of the heavy-duty DumpRequestOut to avoid the memory consumption peak?

jiuker:
fix: enable TraceOn will cause OOM when putting a large object
fix #1771
before pr:
image
we can see max memory is 10GB and CPU is 100%
with this pr:
image
we can see max memory is 26MB and CPU is 34%

## Summary

* **Bug Fixes**
  * Improved HTTP tracing to avoid handling/suspending large request bodies during trace output, ensuring trace logs focus on request headers without buffering upload-sized content.

* **Tests**
  * Added a regression test to confirm tracing with very large payloads keeps memory allocations under control and still includes the request line in the trace output.

### DIFF
--- a/api.go
+++ b/api.go
@@ -567,7 +567,12 @@ func (c *Client) dumpHTTP(req *http.Request, resp *http.Response) error {
 		req.Header.Set("Authorization", redactSignature(origAuth))
 	}
 
-	// Only display request header.
+	//// Only display request header.
+	if req.Body != nil {
+		// httputil.DumpRequestOut call io.Copy(io.Discard, req.Body), then that would call the singer.getSigningKey.
+		// This is useless action.
+		req.Body = http.NoBody
+	}
 	reqTrace, err := httputil.DumpRequestOut(req, false)
 	if err != nil {
 		return err
--- a/api.go
+++ b/api.go
@@ -567,10 +567,8 @@ func (c *Client) dumpHTTP(req *http.Request, resp *http.Response) error {
 		req.Header.Set("Authorization", redactSignature(origAuth))
 	}
 
-	//// Only display request header.
+	// Only display request header.
 	if req.Body != nil {
-		// httputil.DumpRequestOut call io.Copy(io.Discard, req.Body), then that would call the singer.getSigningKey.
-		// This is useless action.
 		req.Body = http.NoBody
 	}
 	reqTrace, err := httputil.DumpRequestOut(req, false)
--- a/api_test.go
+++ b/api_test.go
@@ -1,0 +1,52 @@
+/*
+ * MinIO Go Library for Amazon S3 Compatible Cloud Storage
+ * Copyright 2015-2026 MinIO, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package minio
+
+import (
+	"bytes"
+	"net/http"
+	"runtime"
+	"strings"
+	"testing"
+)
+
+// TestDumpHTTPLargeBodyDoesNotAllocate tests that dumpHTTP does not allocate
+func TestDumpHTTPLargeBodyDoesNotAllocate(t *testing.T) {
+	var trace bytes.Buffer
+	c := &Client{traceOutput: &trace, isTraceEnabled: true}
+	req, err := http.NewRequest(http.MethodPut, "http://example.com/bucket/object", strings.NewReader("x"))
+	if err != nil {
+		t.Fatal(err)
+	}
+	req.ContentLength = 1 << 30 // pretend a 1 GiB upload was sent
+	resp := &http.Response{StatusCode: http.StatusOK, Proto: "HTTP/1.1", ProtoMajor: 1, ProtoMinor: 1, Header: http.Header{}, Body: http.NoBody}
+
+	var m0, m1 runtime.MemStats
+	runtime.GC()
+	runtime.ReadMemStats(&m0)
+	if err := c.dumpHTTP(req, resp); err != nil {
+		t.Fatal(err)
+	}
+	runtime.ReadMemStats(&m1)
+	if got := m1.TotalAlloc - m0.TotalAlloc; got > 64<<20 {
+		t.Fatalf("dumpHTTP allocated %d bytes for a 1 GiB Content-Length; the dump is buffering a body-sized dummy", got)
+	}
+	if !bytes.Contains(trace.Bytes(), []byte("PUT /bucket/object")) {
+		t.Fatal("trace output missing the request line")
+	}
+}


api.go
```diff
@@ -567,7 +567,12 @@ func (c *Client) dumpHTTP(req *http.Request, resp *http.Response) error {
 		req.Header.Set("Authorization", redactSignature(origAuth))
 	}
 
-	// Only display request header.
+	//// Only display request header.
```
klauspost:
```suggestion
	// Only display request header.
```

api.go
```diff
@@ -567,7 +567,12 @@ func (c *Client) dumpHTTP(req *http.Request, resp *http.Response) error {
...
+	if req.Body != nil {
+		// httputil.DumpRequestOut call io.Copy(io.Discard, req.Body), then that would call the singer.getSigningKey.
+		// This is useless action.
```
klauspost: I don't understand these comments.

```suggestion
		// Remove Body
```

api.go
```diff
@@ -567,7 +567,12 @@ func (c *Client) dumpHTTP(req *http.Request, resp *http.Response) error {
...
+		req.Body = http.NoBody
```
allanrogerr, line 572: Nothing in the test suite pins this behavior down, so a future refactor of the tracing path could quietly reintroduce the buffering. A small in-package test locks it in without needing a server — this one fails on the pre-fix code (it allocates ~2.6 GB tracing a request with a 1 GiB Content-Length) and passes with this change in milliseconds:

```go
func TestDumpHTTPLargeBodyDoesNotAllocate(t *testing.T) {
	var trace bytes.Buffer
	c := &Client{traceOutput: &trace, isTraceEnabled: true}
	req, err := http.NewRequest(http.MethodPut, "http://example.com/bucket/object", strings.NewReader("x"))
	if err != nil {
		t.Fatal(err)
	}
	req.ContentLength = 1 << 30 // pretend a 1 GiB upload was sent
	resp := &http.Response{StatusCode: http.StatusOK, Proto: "HTTP/1.1", ProtoMajor: 1, ProtoMinor: 1, Header: http.Header{}, Body: http.NoBody}

	var m0, m1 runtime.MemStats
	runtime.GC()
	runtime.ReadMemStats(&m0)
	if err := c.dumpHTTP(req, resp); err != nil {
		t.Fatal(err)
	}
	runtime.ReadMemStats(&m1)
	if got := m1.TotalAlloc - m0.TotalAlloc; got > 64<<20 {
		t.Fatalf("dumpHTTP allocated %d bytes for a 1 GiB Content-Length; the dump is buffering a body-sized dummy", got)
	}
	if !bytes.Contains(trace.Bytes(), []byte("PUT /bucket/object")) {
		t.Fatal("trace output missing the request line")
	}
}
```

Consider adding it (name/placement to taste) — worth a local run to confirm it fits the suite.

api.go
allanrogerr: The comments are wrong about the mechanism, but there is a real reason worth keeping here: with `body=false`, `DumpRequestOut` never reads the original body at all (no signing code is involved) — it swaps in a `ContentLength`-sized dummy body of `x` bytes and buffers that whole dummy into the dump before slicing it off (`net/http/httputil/dump.go`, the `neverEnding('x')` reader; tracing a 100 MiB PUT allocates ~326 MiB on current master, ~6 MiB with this change). `http.NoBody` makes the dump skip the dummy entirely; the dumped `Content-Length` reads `0` as a side effect. Wording that keeps the why:

```suggestion
		// DumpRequestOut(req, false) never reads this body, but buffers a
		// ContentLength-sized dummy body into the dump before discarding it,
		// which can exhaust memory on large uploads. Nothing reads req.Body
		// after this point, so swapping in http.NoBody skips the dummy; the
		// dumped Content-Length reads 0 as a side effect.
```

allanrogerr:
The title reads fine as-is; two small wording fixes would make the description accurate:

```md
fix: enable TraceOn will cause OOM when putting a large object
fix #1771
before pr:
<img width="1355" height="384" alt="image" src="https://github.com/user-attachments/assets/8043328c-bbec-43a5-b2e3-e244ceaa0fcd" />
we can see max memory is 10GB and CPU is 100%
with this pr:
<img width="553" height="746" alt="image" src="https://github.com/user-attachments/assets/8e7b6cde-3dc9-47b5-8c92-f2e10411092a" />
we can see max memory is 26MB and CPU is 34%
```

- "memery" → "memory" in both measurement lines.
- The auto-generated release-note bullet says tracing "no longer displays request body contents" — request bodies were never shown in traces: `DumpRequestOut` is called with `body=false` and always trimmed the body from the returned dump. What this PR actually changes is that the dump no longer buffers a body-sized dummy in memory while producing the header-only trace. Something like "HTTP tracing no longer buffers a request-body-sized dummy in memory when tracing uploads" would describe it accurately for release notes.

allanrogerr (commented):
@jiuker Please add the in-package test

api.go
klauspost: Pointless and way too long, except maybe the content-length.

jiuker: done

api.go
jiuker: done

api.go
jiuker: added

allanrogerr:
Took another pass over this against `master` and it looks solid — nothing to change.

The mechanism holds up end to end. With `body=false`, `httputil.DumpRequestOut` asks `outgoingLength(req)` for the size, and when the body is non-nil with a set `Content-Length` it installs a `Content-Length`-sized `neverEnding('x')` dummy reader and buffers the whole thing into the dump before slicing it off (`net/http/httputil/dump.go`). Swapping in `http.NoBody` drives `outgoingLength` to 0, so no dummy body is generated and only the header is dumped — which is exactly the goal.

Two things I checked specifically so this can't bite later:

- **The mutation is harmless to the actual transfer.** `dumpHTTP` runs only after `httpClient.Do(req)` has returned, and `executeMethod` builds a fresh `*http.Request` via `newRequest()` on every retry attempt — so setting `req.Body = http.NoBody` can't truncate the upload that was sent, nor a retried one.
- **The response side is already safe.** `DumpResponse(resp, false)` uses `failureToReadBody{}`/`emptyBody` rather than a `Content-Length`-sized dummy, so large GETs don't have the symmetric problem. The request-only fix is complete.

And the regression test does what it claims — I ran it both ways:

| | Command | Result |
|---|---|---|
| pre-fix (guard removed) | `go test -run TestDumpHTTPLargeBodyDoesNotAllocate` | **FAIL** — dumpHTTP allocated 2,684,481,184 bytes, 6.47s |
| this PR | `go test -race -run TestDumpHTTPLargeBodyDoesNotAllocate` | **PASS** — 0.01s |

`go vet`, `gofumpt -d`, and a full-package `go test -c` are all clean. 👍

allanrogerr (approved):

api.go
allanrogerr: PTAL again @klauspost

PR #2253, opened 2026-07-17
fix: disable retries for non-seekable PutObject readers

### LINKED ISSUE #2078: PutObject with `io.Reader` which doesn't implement `Seek` always retries but fails
When you use `PutObject` with an `io.Reader` which doesn't implement `Seek` (i.e. not a `io.ReadSeeker`) like `bytes.Buffer`, if you receive any kind of retry-able error on uploading the object you would see errors like:

```
net/http: HTTP/1.x transport connection broken: http: ContentLength=XXX with Body length 0
```

As far as I can tell this is because the `Progress` option even when nil is used to create a `hookReader` wrapping the original `io.Reader` which does implement `io.ReadSeeker` and results in attempts to `Seek` using [the fall through logic](https://github.com/minio/minio-go/blob/master/hook-reader.go#L66) of the `hookReader` which if wrapping a source which doesn't implement `io.Seeker` cannot actually seek to the appropriate offset (usually the beginning) used in the retry logic.

I took a look and I believe the original intention was to just disallow retries for `io.Reader` input which doesn't implement `io.Seeker` but with the progress wrapper, effectively the code path that disallows retries for non-`io.Seeker` implementations does not ever get triggered.

I think there are 3 problems here:

1. It's not clear to a user (while admittedly being somewhat obvious after thinking about it for a bit) that using a non-`io.Seeker` as input **should** result in no retries.
2. Using an input that doesn't implement `io.Seeker` in fact does result in a retry which could never succeed and throws the very cryptic error message mentioned above.
3. Using the `SendContentMd5` option wraps any `io.Reader` which doesn't implement `io.Seeker` with a `io.ReadSeeker` implementation which effectively papers over the problem as a side-effect.

Reproduction example:
<details>

```golang
// This code has to be run against something that will serve up a retry-able error for the PutObject calls to demonstrate the problem.

// this will retry the upload operation
func UploadWithReadSeeker(client *minio.Client, bucketName string, rawData []byte) error {
	readSeeker, size := bytes.NewReader(rawData), int64(len(rawData))
	_, err := client.PutObject(context.Background(), bucketName, "with-read-seeker", readSeeker, size, minio.PutObjectOptions{})
	return errors.Wrap(err, "failed to upload object")
}

// this will attempt to retry the upload operation and immediately fail
func UploadWithReader(client *minio.Client, bucketName string, rawData []byte) error {
	reader, size := bytes.NewBuffer(rawData), int64(len(rawData))
	_, err := client.PutObject(context.Background(), bucketName, "with-reader", reader, size, minio.PutObjectOptions{})
	return errors.Wrap(err, "failed to upload object")
}

// this will retry the upload operation
func UploadWithReaderWMd5Hash(client *minio.Client, bucketName string, rawData []byte) error {
	reader, size := bytes.NewBuffer(rawData), int64(len(rawData))
	_, err := client.PutObject(context.Background(), bucketName, "with-reader", reader, size, minio.PutObjectOptions{
		SendContentMd5: true,
	})
	return errors.Wrap(err, "failed to upload object")
}
```

</details>

Investigation notes:

Perhaps introduced accidentally in https://github.com/minio/minio-go/pull/1673
Having and not having progress hook shouldn't really influence the retry behavior. Perhaps it's best to implement two `hookReader` types, one which is a pure one-shot `io.Reader` and another which is a `io.ReadSeeker` and match according to the input type?

allanrogerr:
## Description

Fixes #2078. `PutObject` with a non-seekable `io.Reader` (e.g. `*bytes.Buffer`) retried after a retryable server error with an empty body, exhausting all 10 attempts and returning `http: ContentLength=N with Body length 0` instead of the server's real error.

- Root cause: `newHook` wrapped every source in `hookReader`, whose `Seek` silently returned `(0, nil)` for non-seekable sources, so `executeMethod`'s non-seekable ⇒ single-attempt gate (`api.go`) never fired and each retry re-sent a drained reader.
- Fix: split the wrapper into `hookReader` (Read only) and `hookReadSeeker` (embeds `hookReader` + `Seek`); `newHook` returns the seeker variant only when the source implements `io.Seeker`, so the retry gate now sees the truth.
- Neither wrapper implements `io.Closer`, so caller-supplied readers are still shielded from the transport's body close (#2137 behavior preserved).
- Progress-hook accounting and all seekable-source behavior are unchanged; all 9 `newHook` call sites audited, none degrades.

## Motivation and Context

A non-seekable upload that hit a transient error (e.g. 503 SlowDown) burned ~10 doomed retries and surfaced a cryptic transport error that hid the actual server response, misleading users and hammering servers with empty-body requests (see grafana/mimir#10952 for a downstream workaround this makes unnecessary).

## How to test this PR?

Side-by-side repro: httptest server returns `503 SlowDown` on the first PUT per object key, then 200; harness records per-attempt `Content-Length` and received body bytes. Identical client code run against base `ccfaaed` (master) and this branch via a `replace` directive, payload 1198 bytes.

| Case | master (`ccfaaed`) — failing | this PR — fixed |
|---|---|---|
| `bytes.Buffer` (non-seekable) | 10 attempts; attempts 2–10 sent **0 body bytes** (server read `unexpected EOF`); returns `http: ContentLength=1198 with Body length 0`; the 503 SlowDown is never surfaced | **1 attempt, full body**; returns the server's `SlowDown` ErrorResponse |
| `bytes.NewReader` (seekable) | 2 attempts, both full body; succeeds | unchanged: 2 attempts, both full body; succeeds |
| `bytes.Buffer` + `SendContentMd5: true` | 2 attempts; succeeds (buffering masks the bug) | unchanged: 2 attempts; succeeds |

Tests and gates run on this branch:

- New regression tests: `go test -race -count=1 -v -run 'TestNewHookSeekability|TestHookReadSeekerSeek|TestPutObjectRetryNonSeekable' .` — PASS ×3 iterations (16 PASS lines each; the `-run` pattern also matches the two newer test functions). Covers: `newHook` seekability matrix (incl. `io.NopCloser` source and a guard that the wrapper never implements `io.Closer`), Seek + progress-hook parity (including a non-seekable progress hook, and all three Seek error paths), and the retry behavior table above (non-seekable ⇒ exactly 1 attempt + `SlowDown` code asserted; seekable ⇒ 2 attempts each carrying the full identical body — asserted as a lower bound plus cross-attempt equality, since the wire body includes aws-chunked signature framing; non-seekable with `Progress` set ⇒ 1 attempt).
- `go test -short -race -count=1 ./...` — all packages ok (root package 57s).
- `make lint` — 0 issues; `gofumpt -l` clean; `typos` clean.

### Live-server validation (AIStor and MinIO community)

Beyond the httptest mock, the fix was validated end-to-end against real servers — both licensed AIStor (ENTERPRISE-PLUS license) and MinIO community — with the SDK built from this branch driven through a fault-injecting reverse proxy in front of each server (first PUT per object key answered `503 SlowDown` without forwarding; every other request forwarded verbatim with the Host header preserved so the V4 signature stays valid), recording per-attempt `Content-Length` and body bytes on the wire.

| Target | Result |
|---|---|
| AIStor `DEVELOPMENT.2026-07-08T16-17-04Z` (licensed, Edition: AIStor) | 15/15 checks pass |
| AIStor built from master `85ddefac` on 2026-07-17 (licensed, Edition: AIStor) | 15/15 checks pass |
| MinIO community built from master `7aac2a2c` | 15/15 checks pass |

Checks per target:

- Fault-injection through the proxy: seekable reader retries with a byte-identical full body and succeeds; non-seekable reader makes **exactly 1 attempt** and surfaces the server's real `SlowDown` `ErrorResponse` (never the `ContentLength=N with Body length 0` transport error); non-seekable + `SendContentMd5` retries successfully (SDK buffers to hash, so the request body is seekable) with a full body.
- Direct happy path (no proxy): non-seekable uploads at 1 KiB (single PUT), 20 MiB (multipart), and unknown size (streaming) all succeed, with downloaded content verified by SHA-256 against the original payload in every case, including the objects stored via post-fault retries.

## Types of changes

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved upload retry behavior using request body seekability: non-seekable bodies are sent once and fail fast; seekable bodies are retried with an identical wire body.
  * Updated seeking so the reader only exposes `Seek` when the underlying source is seekable, and keeps hook/source offsets synchronized when the hook is also seekable.
  * Enhanced error reporting for seek failures and offset mismatches.
* **Tests**
  * Added coverage for seekability detection, hook/source synchronization behavior, seek error propagation, and `PutObject` retry semantics (including progress hooks).

### DIFF
--- a/hook-reader.go
+++ b/hook-reader.go
@@ -27,7 +27,7 @@ import (
 // notified about the exact number of bytes read from the primary
 // source on each Read operation. It deliberately implements neither
 // io.Seeker nor io.Closer: retry logic treats a seekable body as
-// rewindable, and the transport layer closes bodies that implement
+// rewindable, and executeMethod closes request bodies that implement
 // io.Closer — a caller-supplied reader must be shielded from both.
 type hookReader struct {
 	source io.Reader
@@ -41,18 +41,13 @@ type hookReader struct {
 // retrying over a drained reader.
 type hookReadSeeker struct {
 	hookReader
+	seeker io.Seeker
 }
 
 // Seek implements io.Seeker. Seeks source first, and if necessary
 // seeks hook if Seek method is appropriately found.
 func (hr *hookReadSeeker) Seek(offset int64, whence int) (n int64, err error) {
-	sourceSeeker, ok := hr.source.(io.Seeker)
-	if !ok {
-		// Unreachable by construction: newHook only builds a
-		// hookReadSeeker around a seekable source.
-		return 0, fmt.Errorf("source reader %T is not seekable", hr.source)
-	}
-	n, err = sourceSeeker.Seek(offset, whence)
+	n, err = hr.seeker.Seek(offset, whence)
 	if err != nil {
 		return 0, err
 	}
@@ -99,8 +94,8 @@ func (hr *hookReader) Read(b []byte) (n int, err error) {
 // when the source does.
 func newHook(source, hook io.Reader) io.Reader {
 	hr := hookReader{source: source, hook: hook}
-	if _, ok := source.(io.Seeker); ok {
-		return &hookReadSeeker{hookReader: hr}
+	if seeker, ok := source.(io.Seeker); ok {
+		return &hookReadSeeker{hookReader: hr, seeker: seeker}
 	}
 	return &hr
 }
--- a/hook-reader.go
+++ b/hook-reader.go
@@ -25,22 +25,36 @@ import (
 // hookReader hooks additional reader in the source stream. It is
 // useful for making progress bars. Second reader is appropriately
 // notified about the exact number of bytes read from the primary
-// source on each Read operation.
+// source on each Read operation. It deliberately implements neither
+// io.Seeker nor io.Closer: retry logic treats a seekable body as
+// rewindable, and the transport layer closes bodies that implement
+// io.Closer — a caller-supplied reader must be shielded from both.
 type hookReader struct {
 	source io.Reader
 	hook   io.Reader
 }
 
+// hookReadSeeker extends hookReader with seeking support. It is
+// constructed only when the source implements io.Seeker, so a wrapped
+// reader exposes Seek if and only if it can actually rewind. This lets
+// retry logic disable retries for non-seekable bodies instead of
+// retrying over a drained reader.
+type hookReadSeeker struct {
+	hookReader
+}
+
 // Seek implements io.Seeker. Seeks source first, and if necessary
 // seeks hook if Seek method is appropriately found.
-func (hr *hookReader) Seek(offset int64, whence int) (n int64, err error) {
-	// Verify for source has embedded Seeker, use it.
+func (hr *hookReadSeeker) Seek(offset int64, whence int) (n int64, err error) {
 	sourceSeeker, ok := hr.source.(io.Seeker)
-	if ok {
-		n, err = sourceSeeker.Seek(offset, whence)
-		if err != nil {
-			return 0, err
-		}
+	if !ok {
+		// Unreachable by construction: newHook only builds a
+		// hookReadSeeker around a seekable source.
+		return 0, fmt.Errorf("source reader %T is not seekable", hr.source)
+	}
+	n, err = sourceSeeker.Seek(offset, whence)
+	if err != nil {
+		return 0, err
 	}
 
 	if hr.hook != nil {
@@ -53,7 +67,7 @@ func (hr *hookReader) Seek(offset int64, whence int) (n int64, err error) {
 				return 0, err
 			}
 			if n != m {
-				return 0, fmt.Errorf("hook seeker seeked %d bytes, expected source %d bytes", m, n)
+				return 0, fmt.Errorf("hook seeker sought %d bytes, expected source %d bytes", m, n)
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


hook-reader.go
```diff
@@ -25,22 +25,36 @@ import (
...
 	hook   io.Reader
 }
 
+// hookReadSeeker extends hookReader with seeking support. It is
+// constructed only when the source implements io.Seeker, so a wrapped
+// reader exposes Seek if and only if it can actually rewind. This lets
+// retry logic disable retries for non-seekable bodies instead of
+// retrying over a drained reader.
+type hookReadSeeker struct {
+	hookReader
```
klauspost, line 43: Instead of asserting on seeks (and having the impossible check), just store the interface:
```suggestion
	hookReader
	seeker io.Seeker
```

allanrogerr: Done — `hookReadSeeker` now stores the `io.Seeker` at construction and `Seek` uses it directly; the unreachable branch and its force-constructed subtest are removed, and the two remaining error-path subtests construct via `newHook`. Fixed in 45d43ab; `go test -race -run 'TestNewHookSeekability|TestHookReadSeekerSeek|TestPutObjectRetryNonSeekable' .` passes 3/3 iterations.

klauspost (approved):

hook-reader.go
```diff
@@ -80,14 +89,13 @@ func (hr *hookReader) Read(b []byte) (n int, err error) {
...
-		return &hookReader{source: source}
-	}
-	return &hookReader{
-		source: source,
-		hook:   hook,
+	hr := hookReader{source: source, hook: hook}
+	if seeker, ok := source.(io.Seeker); ok {
+		return &hookReadSeeker{hookReader: hr, seeker: seeker}
 	}
+	return &hr
```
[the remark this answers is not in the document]
allanrogerr: Not a behavior change in this PR — the std-stream guard has never fired through `newHook`, before or after. At the merge base (ccfaaed), `newHook` already wrapped the source even with a nil hook (`if hook == nil { return &hookReader{source: source} }`), and the old `hookReader` also had a `Seek` method, so `executeMethod`'s `bodySeeker, retryable = metadata.contentBody.(io.Seeker)` (api.go:678) always received the wrapper — `switch bodySeeker { case os.Stdin, os.Stdout, os.Stderr }` compares interface identity and never matched a wrapped stream in either version. For `os.Stdin` (an `*os.File`, hence an `io.Seeker`) this PR builds a `hookReadSeeker` that delegates to the file's real `Seek` — exactly what the old wrapper did when the source was seekable — so retryability and error propagation on unseekable pipes are byte-identical pre/post PR. The guard fires only when the body IS the raw std stream, a path this PR does not touch.

PR #2256, opened 2026-07-17
fix: add UserMetadataStripped to list results

### LINKED ISSUE #2054: ListObjects doesn't decode UserMetadata correctly
When ListObjects is used with the `WithMetadata: true` option the returned UserMetadata is not decoded to what was set via PutObject and what GetObject and StatObject would return.

```go
package main

import (
	"bytes"
	"context"
	"fmt"
)

func main() {
	client, err := minio.New("127.0.0.1:9000", &minio.Options{
		Creds:  credentials.NewStaticV4("minio-user", "minio-password", ""),
		Secure: false,
	})
	if err != nil {
		panic(err)
	}

	bucket := "usermeta-test"

	client.MakeBucket(context.Background(), bucket, minio.MakeBucketOptions{})
	defer func() {
		client.RemoveBucket(context.Background(), bucket)
	}()

	content := []byte("hello world")

	_, err = client.PutObject(context.Background(), bucket, "hello.txt", bytes.NewReader(content), int64(len(content)), minio.PutObjectOptions{
		UserMetadata: map[string]string{
			"Hello": "World",
		},
	})
	if err != nil {
		panic(err)
	}

	info, err := client.StatObject(context.Background(), bucket, "hello.txt", minio.StatObjectOptions{})
	if err != nil {
		panic(err)
	}
	fmt.Println("StatObject >", info.Key, info.UserMetadata)

	objs := client.ListObjects(context.Background(), bucket, minio.ListObjectsOptions{
		WithMetadata: true,
	})
	for obj := range objs {
		fmt.Println("ListObjects >", obj.Key, obj.UserMetadata)
	}
}
```

Output:
```
StatObject > hello.txt map[Hello:World]
ListObjects > hello.txt map[X-Amz-Meta-Hello:World content-type:application/octet-stream expires:Mon, 01 Jan 0001 00:00:00 GMT]
```

For StatObject UserMetadata contains exactly the same entries that were saved via PutObject. The UserMetadata returned by ListObjects still has a `X-Amz-Meta-`prefix for the keys and an additional `content-type` and `expires` entries.

allanrogerr:
## Description

Adds `UserMetadataStripped StringMap` to `ObjectInfo` and `Version`, populated when listing with `ListObjectsOptions{WithMetadata: true}` (V2 and versioned listings). It carries the keyed form `StatObject`/`GetObject` return in `UserMetadata`: `X-Amz-Meta-*` keys with the prefix stripped, system entries (`content-type`, `expires`, …) dropped, values passed through verbatim — list responses carry the stored values, so the SDK applies no decoding. The existing `UserMetadata` in list results is unchanged — it keeps the exact values stored on the object, per maintainer direction in #2054 ("a separate field would be needed").

- New helper `stripUserMetadata` (utils.go); keys are canonicalized before the prefix match so server casing does not matter.
- Populated in `listObjectsV2Query` and `listObjectVersionsQuery`; copied into the `ObjectInfo` emitted by versioned listings.
- Stale doc comments on both `UserMetadata` fields corrected (they claimed keys were stripped; list results carry raw keys).
- Regression test `TestListObjectsUserMetadataStripped` asserts, for both list paths, that `UserMetadata` stays byte-identical raw AND `UserMetadataStripped` equals the stripped map — including that an RFC 2047-looking value is passed through verbatim, NOT MIME-decoded.

Review updates (2026-07-22), per @harshavardhana's review:

- `X-Amz-Meta-` lifted to `const amzMetaPrefix` (fc9556e).
- MIME (RFC 2047) decoding of values removed — list XML carries the stored values verbatim; decoding is the application's responsibility since the SDK never encodes these values (4895775).
- Field renamed `UserMetadataDecoded` → `UserMetadataStripped` to match the decode removal (0fcdf85). The field is new in this PR (zero occurrences in upstream master), so the rename is not a breaking change.
- Branch merged with upstream master to resolve an `ObjectInfo` conflict with the new `Headers` field (89bf1d2).

Fixes #2054

## Motivation and Context

`ListObjects(WithMetadata: true)` and `StatObject` return the same user metadata in incompatible shapes, forcing callers to probe both key forms (see #2054). Changing `UserMetadata` in place would break existing consumers, so the stripped form is exposed as a separate additive field.

## How to test this PR?

**Server test matrix (summary):** the same live reproduction of #2054 was run against **both** server lines, each built from source at master on 2026-07-17. Output was identical on both servers, for both the failing (master) client and the fixed (PR) client. (This A/B predates the 2026-07-22 rename — the field was then `UserMetadataDecoded` and included a decode step; the observed values are unchanged by its removal because list XML already carries the stored values verbatim, as the tables show.)

| Server | Build | License | Result |
|---|---|---|---|
| MinIO AIStor master | `miniohq/aistor` @ `85ddefa`, built from source | valid AIStor license (startup log: `License: MinIO Enterprise License`) | ✅ identical to community — raw `UserMetadata` unchanged, stripped field matches `StatObject` |
| MinIO community master | `minio/minio` @ `7aac2a2`, built from source | AGPLv3 | ✅ same |

Client-side A/B against each of those servers, running the issue's reproduction program (object stored with `Hello: World` and a Unicode value that forces RFC 2047 on the PUT wire) — the only variable is the client library:

| | client @ `master` (c30a92d) — failing | client @ this PR — fixed |
|---|---|---|
| `StatObject` `UserMetadata` | `map[Hello:World Unicode:smörgåsbord]` | `map[Hello:World Unicode:smörgåsbord]` |
| `ListObjects` V2 `UserMetadata` | `map[X-Amz-Meta-Hello:World X-Amz-Meta-Unicode:smörgåsbord content-type:application/octet-stream]` | same (unchanged) |
| `ListObjects` V2 stripped field | (field does not exist) | `map[Hello:World Unicode:smörgåsbord]` |
| `ListObjects` versions `UserMetadata` | `map[X-Amz-Meta-Hello:World X-Amz-Meta-Unicode:smörgåsbord content-type:application/octet-stream]` | same (unchanged) |
| `ListObjects` versions stripped field | (field does not exist) | `map[Hello:World Unicode:smörgåsbord]` |

Testing performed, in full:

- Live-server A/B above on **AIStor master (licensed)** and **community master**: raw `UserMetadata` byte-identical before/after on both; stripped field correct on both list paths on both servers.
- `go test -run TestListObjectsUserMetadataStripped -count=1 .` — PASS at head 0b12391 (canned-XML httptest covering V2 + versions paths, raw-unchanged + stripped-map + verbatim-RFC-2047 assertions).
- `go test -short -count=1 .` — package suite PASS at merge commit 89bf1d2 (post-merge with upstream master).
- `go test -race -count=1 -short ./...` — full module suite PASS (12 packages ok, pre-review head 72ba76d).
- Scale benchmark (1000-object list page, 2 meta keys + content-type each, pre-review head with the since-removed decode step): XML unmarshal alone 72.84ms/76,119 allocs vs unmarshal+strip 73.76ms/78,118 allocs (≈1.3% of local parse cost, unobservable against the network round trip); removing the decode only lowers this.
- `make lint` (golangci-lint) 0 issues, `go vet` clean, `gofumpt -d` clean (pre-review head; `go vet` + `gofumpt` re-run clean on every review-round commit).

## Types of changes

## Checklist:

- [x] Internal documentation updated (godoc on the affected fields)

## Summary by CodeRabbit

* **New Features**
  * Added a new `userMetadataStripped` field to object and version listing results, returning prefix-stripped user metadata (values preserved) in JSON and omitting it from XML.
* **Bug Fixes**
  * Improved consistency of user-metadata handling across standard listings and version listings, including correct normalization of metadata header key casing.
* **Tests**
  * Added regression coverage for both raw and prefix-stripped metadata in listing and version-listing scenarios.
* **Documentation**
  * Clarified how user metadata is represented in the public metadata fields.


klauspost (approved):
Nice!

utils.go
```diff
@@ -249,6 +250,32 @@ func extractObjMetadata(header http.Header) http.Header {
...
+// decodeUserMetadata converts raw object metadata, as returned inside the
+// <UserMetadata> element of MinIO list responses, into the decoded form that
+// StatObject and GetObject return in ObjectInfo.UserMetadata: only
+// "X-Amz-Meta-*" entries are kept, with the prefix stripped and RFC 2047
+// encoded values decoded. Returns nil if raw contains no user metadata.
+func decodeUserMetadata(raw StringMap) StringMap {
+	var decoded StringMap
+	for k, v := range raw {
+		k = textproto.CanonicalMIMEHeaderKey(k)
+		if !strings.HasPrefix(k, "X-Amz-Meta-") {
```
harshavardhana: Make this string a constant `const amzMetaPrefix = "X-Amz-Meta-"`

utils.go
```diff
@@ -249,6 +250,32 @@ func extractObjMetadata(header http.Header) http.Header {
...
+			continue
+		}
+		if strings.HasPrefix(v, "=?") {
+			decoder := mime.WordDecoder{}
+			if d, err := decoder.DecodeHeader(v); err == nil {
+				v = d
+			}
+		}
```
harshavardhana: the value is never encoded in x-amz-meta-encoded: - doing decoding is application responsibility not this SDKs since we don't encode it ourselves.

harshavardhana (changes requested):

utils.go
allanrogerr: Done in fc9556e — added `const amzMetaPrefix = "X-Amz-Meta-"` and replaced all uses of the literal in utils.go with it.

utils.go
allanrogerr: Agreed — removed the MIME decoding in 4895775. Values from the list response are now passed through verbatim; the helper only strips the X-Amz-Meta- prefix from keys so the map matches the shape StatObject returns. The test now pins that an RFC 2047-looking value is NOT decoded. Happy to rename the field from UserMetadataDecoded (e.g. to UserMetadataStripped) if you'd prefer, since it no longer decodes values.

allanrogerr: Follow-up: renamed the field UserMetadataDecoded → UserMetadataStripped (and the helper to stripUserMetadata) in 0fcdf85, since it no longer decodes anything. The field is new in this PR — zero occurrences in upstream master — so the rename is not a breaking change. Branch is also merged with upstream master (89bf1d2) to resolve the ObjectInfo conflict with the new Headers field.

allanrogerr:
PTAL @rraulinio

jiuker (approved):

rraulinio (approved):

allanrogerr:
PR looks good to merge @harshavardhana - let me know

PR #2257, opened 2026-07-17
feat: add AWS SSO credential provider to FileAWSCredentials

### LINKED ISSUE #1930: Add ability to use AWS SSO profile as a credential provider
I have a feature request, I want the ability to use AWS SSO profile stored in `.aws/sso/cache` to retrieve temporary access keys and session tokens to my S3 bucket. 

Here is the corresponding AWS SDK SSO credential provider in golang for reference. 
https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/credentials/ssocreds

allanrogerr:
## Description

Adds AWS IAM Identity Center (SSO) support to `FileAWSCredentials`, dependency-free (stdlib + the existing `gopkg.in/ini.v1` only — no aws-sdk-go):

- When a profile defines `sso_role_name`, the provider reads the token cached by `aws sso login` under `~/.aws/sso/cache/<sha1>.json` and exchanges it at the region's SSO portal (`https://portal.sso.<region>.<partition-suffix>/federation/credentials`, partition-aware: `amazonaws.com.cn` for `cn-*`, the iso-partition suffixes for `us-iso*`/`eu-isoe-*`, `amazonaws.eu` for `eusc-*`, `amazonaws.com` otherwise including GovCloud) for temporary S3 credentials (access key, secret, session token, expiration).
- Both config layouts supported: modern `sso_session` + `[sso-session <name>]` section (cache key = SHA1 of the session name) and legacy inline `sso_start_url` / `sso_region` (cache key = SHA1 of the start URL).
- With no explicit `Filename`, the provider now resolves profiles the way the AWS SDK/CLI does: it merges the AWS config file (`AWS_CONFIG_FILE` or `~/.aws/config`) with the shared credentials file (`AWS_SHARED_CREDENTIALS_FILE` or `~/.aws/credentials`), credentials-file values winning per profile and the static key pair moving atomically, so profiles written by `aws configure sso` are discovered by default (`FileAWSCredentials{}` + `AWS_PROFILE` just works). An explicit `Filename` still reads exactly that one file, where `loadProfile` also resolves AWS config-style `[profile <name>]` sections.
- Expired cached token returns an actionable error telling the user to run `aws sso login`. Token refresh (SSO-OIDC) is deliberately out of scope.
- The portal call goes through the `CredContext` HTTP client with a bounded (1 minute) request context; non-200 responses surface the portal's error body; a 200 with empty `roleCredentials` is rejected instead of silently yielding empty credentials; the configured `sso_region` takes precedence over the cached token's region (aws-sdk-go-v2 parity), with the token region as fallback; profile `sso_region`/`sso_start_url` values that contradict the `[sso-session]` section are rejected, matching the SDK's validation.

Revives the approach of #1911 (credit @jchorl), rebased onto the current `RetrieveWithCredContext` provider contract. Intentional behavior notes: the `[profile <name>]` fallback only changes lookups that previously errored; if a single profile defines both a complete static key pair and `sso_*`, the static keys win, matching aws-sdk-go-v2's static-first profile resolution (changed from #1911's SSO-first ordering per review); a profile carrying any SSO marker (`sso_session`, `sso_start_url`, `sso_account_id`, or `sso_region`) without `sso_role_name` now errors instead of silently yielding empty anonymous or half-populated credentials — unless a complete static key pair is present, which then wins (a lone access key or lone secret is instead rejected as partial static credentials before any resolution — including when the profile also carries complete SSO configuration — on both the default-discovery and explicit-Filename paths, matching the SDK's partial-credentials validation; the one deliberate divergence here is the default-path merge, which lets a complete config-file pair survive a partial credentials-file override of the same profile instead of failing the profile outright the way aws-sdk-go-v2's mergeSections does); within a profile, static keys and SSO both take precedence over `credential_process`, matching aws-sdk-go-v2's resolution order; and a default-path AWS file that exists but fails to parse aborts resolution (only a missing file is tolerated), matching the SDK's loader. The `credential_process` single-word-command fix is intentionally not bundled here — #2250 already carries it; the same deferral covers quoted-argument command parsing and the caching of process output that omits `Expiration` (currently re-executed per retrieval, consistent with this provider's existing zero-expiry semantics).

## Motivation and Context

Fixes #1930. Users authenticating via `aws sso login` (e.g. restic users, see restic/restic#4688) cannot use minio-go without exporting credentials through side channels. The maintainer position on record (#1911) is that an aws-sdk-go dependency is unacceptable, so this is a from-scratch implementation.

## How to test this PR?

```
go test -race -run 'TestFileAWSSSO|TestFileAWSDefaultDiscovery|TestSSOPortalBaseURL' ./pkg/credentials/
```

`TestFileAWSSSO`: 18 subtests covering the SSO happy paths and error paths (expired/missing/malformed cache token, non-200 portal response, empty roleCredentials, missing/incomplete SSO config including partial static keys, missing account id, region resolution), portal mocked via httptest, no external calls. `TestFileAWSDefaultDiscovery`: 24 subtests covering the default two-file merge (config-file-only SSO discovery, `AWS_PROFILE`/`AWS_CONFIG_FILE` env handling, credentials-file precedence, atomic static-pair merge, invalid section-name filtering, `[profile default]` wholesale replacement, missing-files error shape, static/SSO precedence over `credential_process`, malformed-file abort in both directions, incomplete-SSO marker guard, profile-vs-session `sso_region`/`sso_start_url` mismatch rejection, partial-static-pair rejection). `TestSSOPortalBaseURL`: all eight AWS partitions.

Side-by-side against the pre-fix SDK — same probe program (`NewFileAWSCredentials(config, profile).Get()`), same `aws configure sso`-style config file, hermetic `HOME`:

| Scenario | master `c30a92df` (pre-fix) | this PR (`287c851`, an earlier head — later commits only tighten precedence/validation and do not change these scenarios) |
|---|---|---|
| SSO profile, no cached token | `err=section "sso-user" does not exist` — config-file layout unreadable, SSO never engages | `err=reading cached SSO token ".../b28b7af...json" (try aws sso login): no such file or directory` |
| SSO profile, valid-format cached token with an invalid bearer, against the live AWS portal | same `section does not exist` failure | `err=fetching SSO role credentials: 401 Unauthorized: {"message":"Session token not found or invalid","__type":"com.amazonaws.switchboard.portal#UnauthorizedException"}` — the real `portal.sso.us-east-1.amazonaws.com` endpoint accepted the request shape and read the bearer header |
| Static-keys profile in a `[profile static-user]` section | `err=section "static-user" does not exist` | `accessKeyID="AKIASTATICEXAMPLE" err=<nil>` |

Full validation on this branch: `go test -race -count=1 ./pkg/credentials/` 25/25 test functions PASS; the SSO/discovery/portal-URL tests (18 + 24 subtests plus 8 portal partition cases) PASS across 3 consecutive `-race` runs; `golangci-lint run` (repo config) 0 issues; `gofumpt -d` and `go vet` clean.

Live-AWS status: the full path — including the `200 → roleCredentials` **success** response — is now live-verified against a real AWS IAM Identity Center account (see "Live end-to-end SSO validation" below). The earlier error-path checks (invalid bearer → `401`) remain covered too, and the success-response shape also stays pinned by the httptest suite (matches aws-sdk-go-v2 `ssocreds`).

## Live end-to-end SSO validation (real AWS IAM Identity Center)

Exercised the complete chain against a real, freshly-provisioned AWS IAM Identity Center account — no mocks anywhere in the path:

- **IdC instance** `d-9a675a5d1f` in `us-east-2`; account `776956157281`; permission set `AdministratorAccess`.
- Config written by-hand in `aws configure sso` style (`[sso-session]` + `[profile]`); token minted by a real `aws sso login` (Authorization Code + PKCE) landing under `~/.aws/sso/cache/`.
- A small probe calls `NewFileAWSCredentials(config, "ssotest").GetWithContext(cc)`, then builds a `minio.Client` from the resolved creds and does a real S3 round-trip against `s3.amazonaws.com`.

```
config=/home/.../.aws/config profile=ssotest
RESULT=OK resolved in 633ms
  SignerType      = S3v4
  AccessKeyID     = ASIA…              # STS temp creds (ASIA prefix), from the live SSO portal
  SecretAccessKey = ****
  SessionToken    = IQoJ… (len=1080)

-- real AWS S3 round-trip: bucket=allanrogerr region=us-east-2 --
  PutObject OK -> minio-go-sso-probe/<ts>.txt
  GetObject OK -> "minio-go PR#2257 live SSO probe" (31 bytes)   # byte-exact read-back
RESULT=OK S3 round-trip complete                                 # object removed after
```

What this proves that the httptest suite could not: the provider resolves the cached `aws sso login` token, exchanges it at the **live** `portal.sso.us-east-2.amazonaws.com/federation/credentials` endpoint, receives and decodes a genuine `roleCredentials` payload, and the resulting SigV4-signed `minio.Client` successfully Puts/Gets/Removes on real AWS S3.

## Dual-target server validation (AIStor + MinIO community master)

The full SDK test suite and the live-server functional suite were run on this branch (at the earlier head `8fb1bd6`; the branch-health conclusion is unchanged by the later credential-resolution commits) against **both** server flavors, using the same client, same env (`ENABLE_HTTPS=0 MINT_MODE=full`), single-node local deployment:

- **MinIO AIStor** — commit `7e1c0794` (Enterprise License, licensed via `MINIO_LICENSE`)
- **MinIO community master** — [minio/minio](https://github.com/minio/minio) `7aac2a2c` (GNU AGPLv3), current `master` HEAD

| Target | `go test -race -count=1 ./...` | `functional_tests` (`MINT_MODE=full`) |
|---|---|---|
| AIStor `7e1c0794` (Enterprise) | ✅ all 12 packages PASS, 0 fail, exit 0 | ✅ 159 PASS / 0 FAIL / 12 N/A, exit 0 |
| Community master `7aac2a2c` (AGPLv3) | ✅ all 12 packages PASS, 0 fail, exit 0 | ✅ 109 PASS / 0 FAIL / 30 N/A, exit 0 |

Zero failures and clean exit on both. The only difference is the N/A (`NotImplemented`) count — community master leaves more features unimplemented/disabled in a single-node deployment (SSE, trailing checksums, object annotations, bucket CORS), which is orthogonal to this change.

Scope note: this provider is **client-side and server-agnostic**. Its logic — read `~/.aws/config` + `~/.aws/sso/cache/*.json`, exchange the bearer token at AWS's `portal.sso.<region>.amazonaws.com`, emit temp credentials — never contacts the S3 server, and the resulting SigV4 requests are identical regardless of server flavor. The `TestFileAWSSSO` suite that exercises the SSO path itself is httptest-mocked (it passes in the `pkg/credentials` package against both targets above, but by construction touches neither server). The dual-target run above is therefore a **branch-health / no-regression gate** — proving the SSO branch builds clean and the whole SDK + functional suite pass end-to-end against both AIStor and community master — not an SSO-specific server test. The SSO path itself is instead proven live against real AWS (see "Live end-to-end SSO validation" above).

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## Checklist:

- [x] Internal documentation updated (godoc on `FileAWSCredentials` documents the default config/credentials-file merge and the explicit-Filename single-file contract)

## Summary by CodeRabbit

* **New Features**
  * Added AWS SSO role credential retrieval using cached `aws sso login` tokens, with both `sso_session` and legacy SSO support.
  * Added an INI-style AWS credentials sample showcasing correctly configured and intentionally broken SSO profiles.
* **Bug Fixes**
  * Enforced precedence: complete static access keys override SSO resolution.
  * Added clearer failures for incomplete SSO configuration (including missing required SSO details) and region/cache issues.
  * Improved default AWS config and credentials discovery/merging semantics.
* **Tests**
  * Added end-to-end SSO portal and cached-token tests, plus broader default discovery/precedence coverage.

### DIFF
--- a/pkg/credentials/file_aws_credentials.go
+++ b/pkg/credentials/file_aws_credentials.go
@@ -181,9 +181,13 @@ func (p *FileAWSCredentials) retrieve(cc *CredContext) (Value, error) {
 		}, nil
 	}
 
+	// A complete static key pair takes precedence over SSO configuration in
+	// the same profile, matching aws-sdk-go-v2's resolution order.
+	hasStaticKeys := id.String() != "" && secret.String() != ""
+
 	// If the profile is configured for AWS SSO, obtain credentials from the
 	// token cached by `aws sso login`.
-	if iniProfile.Key("sso_role_name").String() != "" {
+	if !hasStaticKeys && iniProfile.Key("sso_role_name").String() != "" {
 		ssoCreds, err := p.getSSOCredentials(cc, iniConfig, iniProfile)
 		if err != nil {
 			return Value{}, err
--- a/pkg/credentials/file_aws_credentials.go
+++ b/pkg/credentials/file_aws_credentials.go
@@ -18,8 +18,14 @@
 package credentials
 
 import (
+	"context"
+	"crypto/sha1"
+	"encoding/hex"
 	"encoding/json"
 	"errors"
+	"fmt"
+	"io"
+	"net/http"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -28,6 +34,9 @@ import (
 
 	"gopkg.in/ini.v1"
 )
+
+// ssoPortalRequestTimeout bounds the AWS SSO portal role-credentials request.
+const ssoPortalRequestTimeout = time.Minute
 
 // A externalProcessCredentials stores the output of a credential_process
 type externalProcessCredentials struct {
@@ -38,10 +47,42 @@ type externalProcessCredentials struct {
 	Expiration      time.Time
 }
 
+// A ssoCredentials stores the response of fetching role credentials for
+// an AWS SSO role from the SSO portal.
+type ssoCredentials struct {
+	RoleCredentials ssoRoleCredentials `json:"roleCredentials"`
+}
+
+// A ssoRoleCredentials stores the role-specific credentials portion of
+// an SSO role credentials response. Expiration is milliseconds since
+// the Unix epoch.
+type ssoRoleCredentials struct {
+	AccessKeyID     string `json:"accessKeyId"`
+	Expiration      int64  `json:"expiration"`
+	SecretAccessKey string `json:"secretAccessKey"`
+	SessionToken    string `json:"sessionToken"`
+}
+
+func (s ssoRoleCredentials) expirationTime() time.Time {
+	return time.Unix(0, s.Expiration*int64(time.Millisecond))
+}
+
+// ssoCachedToken is the subset of the cached token file written by
+// `aws sso login` under ~/.aws/sso/cache/ that we need.
+type ssoCachedToken struct {
+	AccessToken string    `json:"accessToken"`
+	ExpiresAt   time.Time `json:"expiresAt"`
+	Region      string    `json:"region"`
+}
+
 // A FileAWSCredentials retrieves credentials from the current user's home
 // directory, and keeps track if those credentials are expired.
 //
 // Profile ini file example: $HOME/.aws/credentials
+//
+// Profiles configured for AWS SSO (sso_session / sso_* keys) live in the AWS
+// config file ($HOME/.aws/config), whose non-default sections are named
+// "profile <name>"; point Filename at that file to use them.
 type FileAWSCredentials struct {
 	Expiry
 
@@ -60,6 +101,14 @@ type FileAWSCredentials struct {
 
 	// retrieved states if the credentials have been successfully retrieved.
 	retrieved bool
+
+	// overrideSSOCacheDir overrides, for tests, the directory holding the
+	// cached SSO tokens (defaults to ~/.aws/sso/cache).
+	overrideSSOCacheDir string
+
+	// overrideSSOPortalURL overrides, for tests, the AWS SSO portal URL
+	// serving role credentials.
+	overrideSSOPortalURL string
 }
 
 // NewFileAWSCredentials returns a pointer to a new Credentials object
@@ -71,7 +120,7 @@ func NewFileAWSCredentials(filename, profile string) *Credentials {
 	})
 }
 
-func (p *FileAWSCredentials) retrieve() (Value, error) {
+func (p *FileAWSCredentials) retrieve(cc *CredContext) (Value, error) {
 	if p.Filename == "" {
 		p.Filename = os.Getenv("AWS_SHARED_CREDENTIALS_FILE")
 		if p.Filename == "" {
@@ -91,7 +140,7 @@ func (p *FileAWSCredentials) retrieve() (Value, error) {
 
 	p.retrieved = false
 
-	iniProfile, err := loadProfile(p.Filename, p.Profile)
+	iniConfig, iniProfile, err := loadProfile(p.Filename, p.Profile)
 	if err != nil {
 		return Value{}, err
 	}
@@ -131,6 +180,26 @@ func (p *FileAWSCredentials) retrieve() (Value, error) {
 			SignerType:      SignatureV4,
 		}, nil
 	}
+
+	// If the profile is configured for AWS SSO, obtain credentials from the
+	// token cached by `aws sso login`.
+	if iniProfile.Key("sso_role_name").String() != "" {
+		ssoCreds, err := p.getSSOCredentials(cc, iniConfig, iniProfile)
+		if err != nil {
+			return Value{}, err
+		}
+		expiration := ssoCreds.RoleCredentials.expirationTime()
+		p.retrieved = true
+		p.SetExpiration(expiration, DefaultExpiryWindow)
+		return Value{
+			AccessKeyID:     ssoCreds.RoleCredentials.AccessKeyID,
+			SecretAccessKey: ssoCreds.RoleCredentials.SecretAccessKey,
+			SessionToken:    ssoCreds.RoleCredentials.SessionToken,
+			Expiration:      expiration,
+			SignerType:      SignatureV4,
+		}, nil
+	}
+
 	p.retrieved = true
 	return Value{
 		AccessKeyID:     id.String(),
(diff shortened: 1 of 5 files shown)


pkg/credentials/file_aws_credentials.go
```diff
@@ -131,6 +180,34 @@ func (p *FileAWSCredentials) retrieve() (Value, error) {
...
+	}
+
+	// A profile carrying SSO configuration without sso_role_name cannot
+	// engage the SSO flow above; if it also has no static keys, returning
+	// the empty values below would silently yield anonymous credentials.
+	if id.String() == "" && secret.String() == "" &&
+		(iniProfile.Key("sso_session").String() != "" || iniProfile.Key("sso_start_url").String() != "") {
+		return Value{}, errors.New("profile has SSO configuration but no sso_role_name, and no static credentials")
+	}
+
```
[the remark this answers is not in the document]
allanrogerr: Fixed in 20c561a: a complete static key pair now takes precedence over SSO configuration in the same profile, matching aws-sdk-go-v2's resolution order (`resolveCredsFromProfile` checks `Credentials.HasKeys()` before `hasSSOConfiguration()`). Added profile `p9-mixed-complete` and the `mixed-complete-static-precedence` test to pin the behavior — the test's empty SSO cache dir would fail the SSO flow, so returning the static keys proves SSO was never engaged.

pkg/credentials/file_aws_credentials.go
```diff
@@ -140,28 +217,147 @@ func (p *FileAWSCredentials) retrieve() (Value, error) {
...
+	if portalURL == "" {
+		portalURL = fmt.Sprintf("https://portal.sso.%s.amazonaws.com", ssoRegion)
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), ssoPortalRequestTimeout)
+	defer cancel()
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, portalURL+"/federation/credentials", nil)
+	if err != nil {
+		return ssoCredentials{}, fmt.Errorf("creating SSO role credentials request: %w", err)
+	}
+	req.Header.Set("x-amz-sso_bearer_token", cachedToken.AccessToken)
```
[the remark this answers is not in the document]
allanrogerr: Agreed this is a real limitation, but threading a caller context here requires adding one to the `Provider`/`CredContext` API in credentials.go — a public cross-file API change out of this PR's scope, as the finding itself notes. The bounded `context.Background()` + `ssoPortalRequestTimeout` pattern matches the package's existing precedent (iam_aws.go:346 does the same for the IMDS token request), so the SSO provider stays consistent with the rest of the chain; a follow-up that adds context to `CredContext` could then pick it up here mechanically.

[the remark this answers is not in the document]
allanrogerr: Yes

klauspost (approved):
lgtm

pkg/credentials/file_aws_credentials.go
```diff
@@ -38,10 +47,42 @@ type externalProcessCredentials struct {
...
+}
+
 // A FileAWSCredentials retrieves credentials from the current user's home
 // directory, and keeps track if those credentials are expired.
 //
 // Profile ini file example: $HOME/.aws/credentials
+//
+// Profiles configured for AWS SSO (sso_session / sso_* keys) live in the AWS
+// config file ($HOME/.aws/config), whose non-default sections are named
+// "profile <name>"; point Filename at that file to use them.
```
rraulinio: **Discover SSO profiles from the standard AWS config path**

When `FileAWSCredentials` has no explicit filename, it still reads only `AWS_SHARED_CREDENTIALS_FILE` or `~/.aws/credentials`. However, `aws configure sso` writes profiles to `AWS_CONFIG_FILE` or `~/.aws/config`, so setting only `AWS_PROFILE` never reaches the new SSO logic.

This also means the motivating restic workflow remains broken by default: restic constructs `FileAWSCredentials{}` without a filename. Users would have to repurpose `AWS_SHARED_CREDENTIALS_FILE` to point at `~/.aws/config`, which is not standard AWS behavior.

Should load/merge the standard config file when resolving profiles, preserving the normal credentials-file precedence, or introduce a separate config-file provider. A regression test should use an empty provider with `HOME` containing only `.aws/config`.

pkg/credentials/file_aws_credentials.go
```diff
@@ -140,28 +221,147 @@ func (p *FileAWSCredentials) retrieve() (Value, error) {
...
+	if ssoRegion == "" {
+		ssoRegion = cachedToken.Region
+	}
+	if ssoRegion == "" {
+		return ssoCredentials{}, errors.New("unable to determine AWS SSO region from profile, sso-session or cached token")
+	}
+
+	portalURL := p.overrideSSOPortalURL
+	if portalURL == "" {
+		portalURL = fmt.Sprintf("https://portal.sso.%s.amazonaws.com", ssoRegion)
```
rraulinio: **Resolve the SSO portal endpoint using the AWS partition**

`https://portal.sso.<region>.amazonaws.com` is not valid for every AWS partition. For example, the official endpoint for `cn-north-1` is `portal.sso.cn-north-1.amazonaws.com.cn`, and the isolated partitions use other DNS suffixes.

As written, a valid SSO profile in AWS China will call the wrong endpoint and cannot retrieve credentials. Should make endpoint construction partition-aware and add at least a custom specific test case (example `aws-cn`).

pkg/credentials/file_aws_credentials.go
```diff
@@ -131,6 +180,38 @@ func (p *FileAWSCredentials) retrieve() (Value, error) {
...
+			SessionToken:    ssoCreds.RoleCredentials.SessionToken,
+			Expiration:      expiration,
+			SignerType:      SignatureV4,
+		}, nil
+	}
+
+	// A profile carrying SSO configuration without sso_role_name cannot
+	// engage the SSO flow above; if it also has no static keys, returning
+	// the empty values below would silently yield anonymous credentials.
+	if id.String() == "" && secret.String() == "" &&
```
rraulinio: **Reject partial static credentials in incomplete SSO profiles**

`hasStaticKeys` correctly requires both the access key and secret, but this validation only rejects the profile when both fields are empty. If an incomplete SSO profile has just one static field, it bypasses this error and returns a credential value containing only half of the key pair.

This is especially problematic in `Chain`, which accepts the provider whenever either field is non-empty, so the malformed value stops credential discovery and later fails during signing. Should use `!hasStaticKeys` here and cover profiles containing only an access key or only a secret.

rraulinio (changes requested):

pkg/credentials/file_aws_credentials.go
allanrogerr: Fixed in `286f374`. With no explicit Filename, the provider now loads the AWS config file (`AWS_CONFIG_FILE` or `~/.aws/config`) first and overlays the shared credentials file (`AWS_SHARED_CREDENTIALS_FILE` or `~/.aws/credentials`), matching aws-sdk-go-v2's LoadSharedConfigProfile: config-file section-name rules, credentials-file precedence, atomic static key pair. A default-constructed `FileAWSCredentials{}` with only `AWS_PROFILE` set now reaches SSO profiles; an explicit Filename still reads exactly that one file. Covered by `TestFileAWSDefaultDiscovery` (11 subtests), including the config-file-only `HOME` shape you suggested.

pkg/credentials/file_aws_credentials.go
allanrogerr: Fixed in `286f374`. `ssoPortalBaseURL` now selects the portal DNS suffix by partition from the region prefix, per aws-sdk-go-v2's endpoint metadata: `cn-*` → `amazonaws.com.cn`, the four iso partitions → their suffixes, `eusc-*` → `amazonaws.eu`. `us-gov-*` stays on `amazonaws.com`, which the SDK metadata lists as GovCloud's portal suffix. `TestSSOPortalBaseURL` pins all eight partitions, including your cn-north-1 example.

pkg/credentials/file_aws_credentials.go
allanrogerr: Fixed in `286f374`. The guard now keys on `!hasStaticKeys`, so an SSO-marked profile with only one static field errors ("no complete static credentials") instead of returning half a key pair; sample profiles p10/p11 plus two subtests cover only-access-key and only-secret. The default-path merge also moves the pair atomically — a credentials-file section carrying half a pair cannot clobber a complete config-file pair — matching the SDK's mergeSections.

allanrogerr:
@rraulinio PTAL

jiuker (approved):

pkg/credentials/file_aws_credentials.go
```diff
@@ -131,6 +182,38 @@ func (p *FileAWSCredentials) retrieve() (Value, error) {
 			SignerType:      SignatureV4,
 		}, nil
 	}
+
+	// A complete static key pair takes precedence over SSO configuration in
+	// the same profile, matching aws-sdk-go-v2's resolution order.
+	hasStaticKeys := id.String() != "" && secret.String() != ""
```
rraulinio, line 161: **Resolve static/SSO credentials before running `credential_process`**

`credential_process` is executed before the new static-key and SSO precedence checks. With the new two-file merge, a valid setup can have `credential_process` in `~/.aws/config` and a complete static key pair in `~/.aws/credentials`. Before this PR, `FileAWSCredentials{}` read only the credentials file and returned that static pair; now it runs the process instead, potentially failing credential lookup or returning credentials from the wrong source.

This also differs from `aws-sdk-go-v2`, whose profile resolution order is static credentials, SSO, then `credential_process`. I reproduced this with a failing process in the config file and valid static credentials in the credentials file: retrieval returned the process error. Should apply the precedence checks before executing `credential_process` and add coverage for this two-file case.

pkg/credentials/file_aws_credentials.go
```diff
@@ -140,28 +223,295 @@ func (p *FileAWSCredentials) retrieve() (Value, error) {
...
+	}
+	if credsFilename != "" {
+		if config, err := ini.Load(credsFilename); err != nil {
+			loadErrs = append(loadErrs, err)
+		} else {
+			loaded = true
+			mergeAWSCredentialsSections(merged, config)
+		}
+	}
+	if !loaded {
```
rraulinio, line 471: **Do not silently ignore a malformed AWS file when the other file loads**

Errors from both `ini.Load` calls are collected, but they are returned only when neither file loads. Consequently, if `~/.aws/config` is valid while `~/.aws/credentials` has an INI parse error, the malformed credentials file is silently ignored and credentials from the config file are returned instead. That can select stale or unintended credentials even though the higher-precedence file is invalid.

I reproduced this with a valid config profile and a malformed credentials file; retrieval succeeded using the config credentials. `aws-sdk-go-v2` aborts on parser errors, even though it tolerates files that cannot be opened. Should surface parse errors and add a one-valid/one-malformed regression test.

rraulinio (changes requested):

pkg/credentials/file_aws_credentials.go
allanrogerr: Fixed in f95114d. `credential_process` now runs only after the static-pair and SSO checks, matching aws-sdk-go-v2's `resolveCredsFromProfile` order (static → SSO → process). Your two-file repro is pinned by `static-over-credential-process-across-files`; `sso-over-credential-process` and `incomplete-sso-not-rescued-by-credential-process` pin SSO-before-process, and `credential-process-without-static-or-sso` keeps the process path working via default discovery.

pkg/credentials/file_aws_credentials.go
allanrogerr: Fixed in f95114d. `loadDefaultProfiles` now aborts on any `ini.Load` failure other than `fs.ErrNotExist`, so only a missing file is tolerated while a file that exists but fails to parse surfaces its error — matching aws-sdk-go-v2's `loadIniFiles` (skip unreadable, abort on parse). One-valid/one-malformed regression tests cover both directions (`malformed-credentials-file-aborts`, `malformed-config-file-aborts`).

allanrogerr:
PTAL @rraulinio

harshavardhana (approved):

PR #2255, opened 2026-07-17
fix: return io.EOF when reading at or beyond object size

allanrogerr:
## Description

Reading a `minio.Object` at or beyond its known size surfaced the server's 416 `InvalidRange` error instead of `io.EOF`, and the saved error permanently poisoned the object for every subsequent `Seek`/`Read`.

- `Object.Read`: return `io.EOF` locally — no request issued — when the current offset is already known to be at or past `objectInfo.Size`
- `Object.Read` / `Object.ReadAt`: translate a server 416 `InvalidRange` into `io.EOF`; `Read` stores `prevErr = io.EOF` (recoverable via `Seek`), `ReadAt` stores nothing, mirroring the pre-existing at-EOF guard. On `Read` the translation applies only to ranges the SDK generated from a non-zero read offset — a caller-supplied `GetObjectOptions` range on the first read keeps surfacing `InvalidRange` untranslated
- The request goroutine survives an `InvalidRange` fetch error instead of ending the stream, so a translated `io.EOF` is genuinely recoverable: `Seek` back and read, or issue another in-range `ReadAt`, on the same object (previously any fetch error ended the stream and later calls returned `context canceled`)
- `Seek` boundary semantics deliberately untouched (that is #2155 scope), so the `functional_tests.go` seek tables are unaffected

## Motivation and Context

Fixes #2166. `Seek(size, io.SeekStart)` — or `Seek(0, io.SeekEnd)` — followed by `Read` sent `Range: bytes=<size>-`, which the server correctly rejects per RFC 9110; `os.File` returns `(0, io.EOF)` for the same sequence. Covers the read-side portion of the closed #2226/#2228.

## How to test this PR?

Regression tests added in this PR: `go test -race -run 'TestObjectSeekAtObjectSizeAllowsSubsequentReadEOF|TestGetObjectReadAtEOFReturnsEOF|TestGetObjectReadUserRangeInvalidRangeSurfaces|TestGetObjectReadStaleSizeReturnsEOF' .` — the second replays the issue's wire capture (HEAD 200, then ranged GET answered 416) against an httptest server, asserting zero GETs are issued for the at-EOF read and the object remains usable afterward — including after a genuine round-trip 416 (`ReadAt` past the end, or a stale-high cached size); the third pins that a caller-supplied unsatisfiable range still surfaces `InvalidRange`; the fourth drives the server-416 translation via a stale-high `HEAD` size and asserts the object stays usable. Full suite `go test -short -race ./...` and `make lint` both pass.

[...]

PR #2228, opened 2026-05-17
fix: allow SeekCurrent with negative offset

wyf027:
## Summary

`Object.Seek` rejected any negative offset unless `whence` was `SeekEnd`, which breaks `io.Seeker` semantics for `SeekCurrent` (e.g. stepping back from EOF).

## Changes

- Only reject negative offsets up front for `SeekStart`
- After applying whence, reject if the computed offset is before the object start
- Add `TestObjectSeekCurrentNegativeOffset`

Fixes #2155

## Release Notes

* **Bug Fixes**
  * Improved end-of-file behavior for reads when the current position is at/after the known object size.
  * Refined seek validation and EOF handling across seek modes, including consistent errors for invalid negative absolute offsets and corrected behavior for seeking relative to the end.

* **Tests**
  * Added unit tests for edge seek cases at the object size and negative relative seeks.
  * Updated functional seek tests to expect successful `SEEK_END` seeks with positive offsets, including `io.EOF` on subsequent reads past the end.

PR #2258, opened 2026-07-17
fix: honor io.Seeker contract in Object.Seek

### LINKED ISSUE #2155: `minio.Object` violates `io.Seeker` contract
## Violation of `io.Seeker` contract
According to Go's `io.Seeker` documentation:
```
Seeking to an offset before the start of the file is an error. Seeking to any positive offset may be allowed, but if the new offset exceeds the size of the underlying object, the behavior of subsequent I/O operations is implementation-dependent.
```
Currently, minio.Object does not respect this contract (its not possible to seek to any positive offset), making it incompatible with standard Go I/O expectations.

## How to reproduce
Use tests from standard library `testing/iotest`. `minio.Object` returned by `GetObject` will fail on those tests.

Currently, it fails with:
```
Seek(-1, 1) from EOF = 0, Negative position not allowed for 1, want 39, nil
```

The issue likely originates here:

https://github.com/minio/minio-go/blob/60b85ef162eb9faa881c431229d3c9fc40919c7f/api-get-object.go#L539

Since `testing/iotest` is part of the standard Go library, this behavior prevents `minio.Object` from being safely used in other libraries that rely on standard I/O behavior.

## Proposed solution:
```golang
	// Negative offset not valid for whence of '0'.
	if offset < 0 && whence == 0 {
		return 0, errInvalidArgument("Negative position not allowed for 0")
	}
```

### LINKED ISSUE #2166: Seeking to offset=length results in The requested range 'bytes=N--1' is not satisfiable error
If you seek to an offset exactly equal to length, subsequent read will return rather unexpected "Requested Range Not Satisfiable" error instead of either `io.EOF` or empty data.

I think these strict inequalities should be replaced with greater-or-equal to fix this bug:
https://github.com/minio/minio-go/blob/f4c4350a79636314092251ec5cdd22fff5a0934d/api-get-object.go#L567-L576

Somewhat related to #2155: changing the library to allow past-the-end seeks will fix this as well.

```
HEAD /XXX/logs/1 HTTP/1.1
Host: 127.0.0.1:9000
User-Agent: MinIO (linux; amd64) minio-go/v7.0.74
Authorization: AWS4-HMAC-SHA256 Credential=XXX/20251009/us-east-1/s3/aws4_request, SignedHeaders=host;x-amz-content-sha256;x-amz-date, Signature=ac2c05657ab592d24c7d43d3de32fd7cfd716746bf2dcce364aa53cd4ac11a3f
X-Amz-Content-Sha256: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
X-Amz-Date: 20251009T231603Z


HTTP/1.1 200 OK
Accept-Ranges: bytes
Content-Length: 6306
Content-Type: text/plain
ETag: "2d9df643507c5e8a783a4a57818e305f"
Last-Modified: Thu, 09 Oct 2025 17:39:48 GMT
Server: MinIO
Strict-Transport-Security: max-age=31536000; includeSubDomains
Vary: Origin
Vary: Accept-Encoding
X-Amz-Id-2: dd9025bab4ad464b049177c95eb6ebf374d3b3fd1af9251148b658df7ac2e3e8
X-Amz-Request-Id: 186CF5C11188BB95
X-Content-Type-Options: nosniff
X-Ratelimit-Limit: 5592
X-Ratelimit-Remaining: 5592
X-Xss-Protection: 1; mode=block
Date: Thu, 09 Oct 2025 23:16:03 GMT


GET /XXX/logs/1 HTTP/1.1
Host: 127.0.0.1:9000
User-Agent: MinIO (linux; amd64) minio-go/v7.0.74
Authorization: AWS4-HMAC-SHA256 Credential=XXX/20251009/us-east-1/s3/aws4_request, SignedHeaders=host;if-match;range;x-amz-content-sha256;x-amz-date, Signature=ae7d08cfd2339e23fed2d5e8b25dd34bc38c02b776d70e6add61462c3332e02a
If-Match: "2d9df643507c5e8a783a4a57818e305f"
Range: bytes=6306-
X-Amz-Content-Sha256: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
X-Amz-Date: 20251009T231603Z


HTTP/1.1 416 Requested Range Not Satisfiable
Accept-Ranges: bytes
Content-Length: 448
Content-Type: application/xml
Server: MinIO
Strict-Transport-Security: max-age=31536000; includeSubDomains
Vary: Origin
Vary: Accept-Encoding
X-Amz-Id-2: dd9025bab4ad464b049177c95eb6ebf374d3b3fd1af9251148b658df7ac2e3e8
X-Amz-Request-Id: 186CF5C1119FAF1C
X-Content-Type-Options: nosniff
X-Ratelimit-Limit: 5592
X-Ratelimit-Remaining: 5592
X-Xss-Protection: 1; mode=block
Date: Thu, 09 Oct 2025 23:16:03 GMT

<Error><Code>InvalidRange</Code><Message>The requested range &#39;bytes=6306--1&#39; is not satisfiable</Message><Key>logs/1</Key><BucketName>XXX</BucketName><Resource>/XXX/logs/1</Resource><RequestId>186CF5C1119FAF1C</RequestId><HostId>dd9025bab4ad464b049177c95eb6ebf374d3b3fd1af9251148b658df7ac2e3e8</HostId><ActualObjectSize>6306</ActualObjectSize><RangeRequested>6306--1</RangeRequested></Error>
```

allanrogerr:
## Description

`Object` did not behave like a standard Go reader/seeker at the end-of-object boundary (#2155, #2166):

- Negative offsets were rejected for `io.SeekCurrent` even when the resulting absolute position was still within the object.
- Seeking to exactly `offset == size` succeeded, but the subsequent `Read` issued an unsatisfiable range request (`bytes=N--1`) that the server rejects with 416 `InvalidRange`, instead of returning a clean `io.EOF` — and the saved error permanently poisoned the object for every subsequent `Seek`/`Read`.

This PR (now also carrying the read-side fixes from #2255, which it supersedes):

**Seek semantics**

- Allows negative offsets for `SeekCurrent`/`SeekEnd` as long as the computed absolute position is `>= 0`; positions before the start of the object are rejected for any whence, as are negative absolute offsets for `SeekStart`.
- Allows seeking to exactly the end of the object; the subsequent `Read` reports `io.EOF` (the `Read` path short-circuits at a known EOF instead of issuing an unsatisfiable range request).
- Keeps the long-standing rejection of seeks **past** the end: `Seek` returns `(0, io.EOF)` and leaves the offset unchanged. Per review feedback, the `io.Seeker` contract's "may be allowed" wording permits rejecting past-end seeks, and the existing behavior is kept for compatibility.
- Rewrites the `Seek` doc comment to document the actual behavior instead of attributing it to the `io.Seeker` contract.

**Read/ReadAt EOF robustness** (from #2255)

- `Object.Read`: returns `io.EOF` locally — no request issued — when the current offset is already known to be at or past `objectInfo.Size`.
- `Object.Read` / `Object.ReadAt`: translate a server 416 `InvalidRange` into `io.EOF`. `Read` stores `prevErr = io.EOF` (recoverable via `Seek`); `ReadAt` stores nothing. On `Read` the translation applies only to ranges the SDK generated from a non-zero read offset — a caller-supplied `GetObjectOptions` range on the first read keeps surfacing `InvalidRange` untranslated.
- The request goroutine survives an `InvalidRange` fetch error instead of ending the stream, so a translated `io.EOF` is genuinely recoverable: `Seek` back and read, or issue another in-range `ReadAt`, on the same object (previously any fetch error ended the stream and later calls returned `context canceled`).

**Tests**

- Adds `TestObjectSeekWithSetRange` pinning the current `SetRange`+`Seek` interplay: a `Seek` issued before any `Read` stats the object without the range header, so seek offsets and the end-of-object boundary refer to the full object, and a `Read` after a `Seek` requests a fresh range from the seek offset, discarding the caller's range (cross-ref #2267, which proposes changing the related `Stat` behavior).
- The three functional seek tables and the V2 flow assert that a rejected seek leaves the cursor unchanged (probed via `Seek(0, io.SeekCurrent)`) and that a `Read` after seeking to exactly the end returns zero bytes and `io.EOF`.

## Motivation and Context

Fixes #2155
Fixes #2166
Supersedes #2255 and the approved-but-abandoned #2228.

## How to test this PR?

Unit tests in `api-get-object_test.go`:

```
go test -race -run 'TestObjectSeek|TestGetObjectRead' -v .
```

- Seek: `TestObjectSeekAtObjectSizeAllowsSubsequentReadEOF`, `TestObjectSeekPastObjectSizeReturnsEOF`, `TestObjectSeekCurrentNegativeOffset`, `TestObjectSeekEndUnknownSize`, `TestObjectSeekInvalidWhence`, `TestObjectSeekWithSetRange`
- Read/ReadAt: `TestGetObjectReadAtEOFReturnsEOF`, `TestGetObjectReadUserRangeInvalidRangeSurfaces`, `TestGetObjectReadStaleSizeReturnsEOF`

Plus the extended `functional_tests.go` seek coverage (three seek tables + V2 flow).

### Proof of validity — pre-fix vs fix (unit tests run against both trees)

| Test | Pre-fix (`ccfaaed`) | Fix (this PR) |
| --- | --- | --- |
| `TestObjectSeekAtObjectSizeAllowsSubsequentReadEOF` | ❌ FAIL — `Read` after seeking to `size` issues an unsatisfiable range request (surfaces as a panic in the unit harness) | ✅ PASS |
| `TestObjectSeekPastObjectSizeReturnsEOF` | ❌ FAIL — past-end rejection itself is unchanged, but the final recovery step (`Seek(size)` then `Read`) hits the same at-EOF `Read` defect | ✅ PASS |
| `TestObjectSeekCurrentNegativeOffset` | ❌ FAIL — `expected SeekCurrent with negative offset to succeed, got Negative position not allowed for 1` | ✅ PASS |
| `TestObjectSeekEndUnknownSize`, `TestObjectSeekInvalidWhence`, `TestObjectSeekWithSetRange` | ✅ PASS (pin behavior that is intentionally unchanged) | ✅ PASS |

### Validation (2026-07-21, this exact tree)

| Check | Result |
| --- | --- |
| `go vet ./...`, `gofumpt -d`, `make lint` (golangci-lint) | ✅ clean |
| Unit: `go test -race -run 'TestObjectSeek\|TestGetObject' .` | ✅ 13/13 PASS |
| Functional against a live server: `testGetObjectReadSeekFunctional`, `testSSECEncryptedGetObjectReadSeekFunctional`, `testSSES3EncryptedGetObjectReadSeekFunctional`, `testGetObjectReadSeekFunctionalV2` | ✅ 4/4 PASS × 3 iterations |

Functional runs used the same server setup as this repo's CI (`registry.k5.min.dev/aistor/minio:edge` with TLS certs and `MINIO_KMS_SECRET_KEY`), so the SSE-C and SSE-S3 seek variants executed rather than being skipped.

## Types of changes

## Checklist:

- [ ] Create a documentation update request [here](https://github.com/miniohq/aistor-object-store-docs/issues/new)
- [x] No internode changes / version interoperability introduced

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved object reading when requests reach or pass the end of an object.
  * Reads and `ReadAt` now consistently return `io.EOF` for end-of-object conditions.
  * Prevented unnecessary network requests after seeking to the object’s end.
  * Improved recovery after invalid range responses so subsequent reads can continue.
  * Updated seeking behavior for positions before, at, and beyond the object boundaries.
  * Expanded coverage for seeking, range handling, stale size information, and EOF behavior.

### DIFF
--- a/api-get-object.go
+++ b/api-get-object.go
@@ -548,9 +548,9 @@ func (o *Object) ReadAt(b []byte, offset int64) (n int, err error) {
 // and 2 means relative to the end.
 // Seek returns the new offset and an error, if any.
 //
-// Seeking to a negative offset is an error. Seeking to any positive
-// offset is legal, subsequent io operations succeed until the
-// underlying object is not closed.
+// Seeking to a position before the start of the object is an error.
+// Seeking to or past the end of the object is legal per the io.Seeker
+// contract; a subsequent Read reports io.EOF.
 func (o *Object) Seek(offset int64, whence int) (n int64, err error) {
 	if o == nil {
 		return 0, errInvalidArgument("Object is nil")
--- a/api-get-object.go
+++ b/api-get-object.go
@@ -115,6 +115,12 @@ func (c *Client) GetObject(ctx context.Context, bucketName, objectName string, o
 					httpReader, objectInfo, _, err = c.getObject(gctx, bucketName, objectName, opts)
 					if err != nil {
 						resCh <- getResponse{Error: err}
+						// An unsatisfiable range is recoverable: the caller
+						// may Seek or ReadAt within bounds next, so keep
+						// serving requests instead of ending the stream.
+						if ToErrorResponse(err).Code == InvalidRange {
+							continue
+						}
 						return
 					}
 					etag = objectInfo.ETag
@@ -196,7 +202,10 @@ func (c *Client) GetObject(ctx context.Context, bucketName, objectName string, o
 				// if the object has been read or not to only initialize
 				// new ones when they haven't been already.
 				// All readAt requests are new requests.
-				if req.DidOffsetChange || !req.beenRead {
+				// A nil httpReader means the previous fetch failed and
+				// left no stream, so a new one must be established
+				// regardless of the offset bookkeeping.
+				if req.DidOffsetChange || !req.beenRead || httpReader == nil {
 					// Check whether this is snowball
 					// if yes do not use If-Match feature
 					// it doesn't work.
@@ -221,6 +230,12 @@ func (c *Client) GetObject(ctx context.Context, bucketName, objectName string, o
 					if err != nil {
 						resCh <- getResponse{
 							Error: err,
+						}
+						// An unsatisfiable range is recoverable: the caller
+						// may Seek or ReadAt within bounds next, so keep
+						// serving requests instead of ending the stream.
+						if ToErrorResponse(err).Code == InvalidRange {
+							continue
 						}
 						return
 					}
@@ -414,6 +429,14 @@ func (o *Object) Read(b []byte) (n int, err error) {
 	// Send and receive from the first request.
 	response, err := o.doGetRequest(readReq)
 	if err != nil && err != io.EOF {
+		// An InvalidRange response to a range generated from a non-zero
+		// read offset means the position is at or past EOF for the object
+		// as it currently exists. Offset zero sends only a caller-supplied
+		// range, whose InvalidRange must surface untranslated.
+		if o.currOffset > 0 && ToErrorResponse(err).Code == InvalidRange {
+			o.prevErr = io.EOF
+			return 0, io.EOF
+		}
 		// Save the error for future calls.
 		o.prevErr = err
 		return response.Size, err
@@ -518,6 +541,12 @@ func (o *Object) ReadAt(b []byte, offset int64) (n int, err error) {
 	// Send and receive from the first request.
 	response, err := o.doGetRequest(readAtReq)
 	if err != nil && err != io.EOF {
+		// Reading at an offset at or beyond the object size yields
+		// InvalidRange from the server: report io.EOF, matching the
+		// io.ReaderAt contract for reads past the end.
+		if ToErrorResponse(err).Code == InvalidRange {
+			return 0, io.EOF
+		}
 		// Save the error.
 		o.prevErr = err
 		return response.Size, err
@@ -549,8 +578,9 @@ func (o *Object) ReadAt(b []byte, offset int64) (n int, err error) {
 // Seek returns the new offset and an error, if any.
 //
 // Seeking to a position before the start of the object is an error.
-// Seeking to or past the end of the object is legal per the io.Seeker
-// contract; a subsequent Read reports io.EOF.
+// Seeking to the end of the object is legal; the subsequent Read reports
+// io.EOF. When the object size is known, seeking past the end returns
+// io.EOF from Seek itself and leaves the offset unchanged.
 func (o *Object) Seek(offset int64, whence int) (n int64, err error) {
 	if o == nil {
 		return 0, errInvalidArgument("Object is nil")
@@ -608,10 +638,15 @@ func (o *Object) Seek(offset int64, whence int) (n int64, err error) {
 		newOffset = o.objectInfo.Size + offset
 	}
 	// Seeking to a position before the start of the object is not allowed for
-	// any whence. Seeking to or past the end is allowed per the io.Seeker
-	// contract; the subsequent Read reports io.EOF.
+	// any whence.
 	if newOffset < 0 {
 		return 0, errInvalidArgument(fmt.Sprintf("Seeking at negative offset not allowed for %d", whence))
+	}
+	// Seeking past the end of the object is rejected with io.EOF, preserving
+	// long-standing behavior. Seeking to exactly the end is legal and the
+	// subsequent Read reports io.EOF.
+	if o.objectInfo.Size > -1 && newOffset > o.objectInfo.Size {
+		return 0, io.EOF
 	}
 	// Reset the saved error since we successfully seeked, let the Read
 	// and ReadAt decide.
--- a/api-get-object.go
+++ b/api-get-object.go
@@ -387,6 +387,13 @@ func (o *Object) Read(b []byte) (n int, err error) {
 	if o.prevErr != nil || o.isClosed {
 		return 0, o.prevErr
 	}
+	// If the current offset is at or beyond the known object size, there is
+	// nothing more to read. Return io.EOF directly instead of issuing a range
+	// request from EOF (which the server rejects as an unsatisfiable range).
+	if o.objectInfoSet && o.objectInfo.Size > -1 && o.currOffset >= o.objectInfo.Size {
+		o.prevErr = io.EOF
+		return 0, io.EOF
+	}
 
 	// Create a new request.
 	readReq := getRequest{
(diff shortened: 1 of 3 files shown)


api-get-object.go
```diff
@@ -541,9 +548,9 @@ func (o *Object) ReadAt(b []byte, offset int64) (n int, err error) {
 // and 2 means relative to the end.
 // Seek returns the new offset and an error, if any.
 //
-// Seeking to a negative offset is an error. Seeking to any positive
-// offset is legal, subsequent io operations succeed until the
-// underlying object is not closed.
+// Seeking to a position before the start of the object is an error.
+// Seeking to or past the end of the object is legal per the io.Seeker
+// contract; a subsequent Read reports io.EOF.
```
klauspost: That is strictly wrong: [ref](https://pkg.go.dev/io#Seeker)

> Seeking to any positive offset may be allowed, but if the new offset exceeds the size of the underlying object the behavior of subsequent I/O operations is implementation-dependent.

Instead of trying to "blame" `io.Seeker` it should document the actual behavior.

klauspost (commented):
Issue #2155 is wrong - and ignores "may" in the contract, which allows for the current behavior.

* Seeking to EOF should be ok and should return `0, io.EOF` on next read. Yes, that is a bug. 
* Seeking *past* EOF may as well be rejected in the Seek call. It is pointless to issue a call for this. I would prefer it to return `io.ErrUnexpectedEOF`, but I'd rather keep the current behavior for compatibility.

Add tests with SetRange. I am not really sure how it behaves. Crossreference with #2267

api-get-object.go
allanrogerr: Appreciate the review. Checking

allanrogerr: Implemented in 30eaee1: seeks past a known object size are rejected again with io.EOF from Seek (offset unchanged); seeking to exactly the end succeeds and the next Read returns io.EOF; and the doc comment now describes this behavior directly instead of citing io.Seeker.

Added `TestObjectSeekWithSetRange` pinning the SetRange interplay: a Seek issued before any Read stats the object without the range header, so seek offsets and the end boundary resolve against the full object, and a Read after a Seek requests a fresh range from the seek offset, discarding the caller's range. Cross-referenced with #2267, which changes the related Stat behavior but not these paths.

This PR now also carries the read-side 416-to-io.EOF fixes from #2255 and supersedes it.

klauspost (approved):

harshavardhana (approved):

PR #1379, opened 2020-08-31
Add opts for replication deletes

poornas:
Enhancing StatObjectOptions, RemoveObjectOptions
for propagating replication deletes.

This PR also enhances replication pkg to
toggle DeleteMarker replication flag, intended
for `mc replicate set/add` commands

PR #2273, opened 2026-08-01
fix: return delete-marker ObjectInfo from StatObject alongside the error

### LINKED ISSUE #2260: StatObject no longer returns delete-marker ObjectInfo (VersionID/IsDeleteMarker) alongside the error — regression since v7.0.93 (#2115)
## Summary

`Client.StatObject`'s delete-marker/error handling block (`api-stat.go:98-121`) is unreachable since commit 7a9dff0 (PR #2115, first released in **v7.0.93**). On a HEAD that returns a non-2xx status, `StatObject` now always returns a **zero `ObjectInfo`**, so the `VersionID`, `IsDeleteMarker`, and `ReplicationReady` fields that the block was written to populate alongside the error (added in #1379/#1397 and 8257e70) are silently lost.

## Mechanism

Before #2115, `executeMethod` returned `(res, nil)` for a non-retryable non-2xx response — callers like `StatObject` were expected to check `resp.StatusCode` themselves, which is exactly what the block at `api-stat.go:98-121` does.

PR #2115 added `err = errResponse` in `executeMethod` (`api.go:778` on current master `c30a92d`), so `executeMethod` now returns a non-nil error for **every** non-2xx. `StatObject` therefore early-returns at `api-stat.go:94-96`:

```go
if err != nil {
    return ObjectInfo{}, err
}
```

…before the delete-marker block can run. The only status that can still enter the block is HTTP 204 (a member of `successStatus`, `api.go:661-665`, that is neither 200 nor 206), which no S3 implementation returns for HEAD Object — so the block is dead code for its intended 405/404 paths.

## User-visible regressions (verified against master `c30a92d`)

1. Stat of a delete marker (`StatObject` with `VersionID` set, server responds 405 + `x-amz-delete-marker: true`): callers previously received `ObjectInfo{VersionID: ..., IsDeleteMarker: true}` alongside the error; they now receive a zero `ObjectInfo` (`IsDeleteMarker == false`, empty `VersionID`).
2. The error code for that case changed: the block returned `ErrorResponse{Code: "MethodNotAllowed"}`; callers now get `executeMethod`'s parsed error, which for an empty-body 405 falls into `httpRespToErrorResponse`'s default case — `Code: "405 Method Not Allowed"` (the raw status line). Code-based checks for `MethodNotAllowed` no longer match.

## Reproduction

```go
func TestStatObjectDeleteMarkerInfo(t *testing.T) {
	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
		w.Header().Set("x-amz-delete-marker", "true")
		w.Header().Set("x-amz-version-id", "test-version-id")
		w.WriteHeader(http.StatusMethodNotAllowed)
	}))
	defer srv.Close()

	clnt, err := New(srv.Listener.Addr().String(), &Options{Region: "us-east-1"})
	if err != nil {
		t.Fatal(err)
	}

	objInfo, err := clnt.StatObject(context.Background(), "bucketName", "objectName",
		StatObjectOptions{VersionID: "test-version-id"})
	if err == nil {
		t.Fatal("expected error for delete marker, got nil")
	}
	if !objInfo.IsDeleteMarker {
		t.Error("expected IsDeleteMarker to be true") // FAILS on master
	}
	if objInfo.VersionID != "test-version-id" {
		t.Errorf("VersionID = %q, want %q", objInfo.VersionID, "test-version-id") // FAILS on master
	}
}
```

Both assertions fail on master; on v7.0.92 (pre-#2115) they pass.

## Possible fixes

- Populate the delete-marker `ObjectInfo` in the `err != nil` branch when `resp != nil` (moving the intent of the now-dead block there), restoring both the populated value and the `MethodNotAllowed` code; or
- have `executeMethod` keep the old contract for callers that do their own status handling.

Either way the dead block should not stay as-is — it reads as if the behavior still works.

Related: #1993 (raw response headers) touches the same function but is independent; happy to send a PR for this one if the preferred direction is confirmed.

allanrogerr:
## Description

`StatObject` stopped returning delete-marker information. This PR restores it.

**What broke.** PR #2115, first released in v7.0.93, changed `executeMethod` to return a non-nil error for every response outside its success set. That set was 200/204/206; #2252 later added 202. `StatObject` returns early with a zero `ObjectInfo` whenever the error is non-nil, so its delete-marker block never runs. That block is what handled the 405 and 404 responses that carry delete-marker headers.

**What this PR does.** It moves that logic into the `err != nil` branch and removes the old block. `StatObject` now behaves as it did before v7.0.93:

- Stat of a delete marker — a HEAD with a `VersionID`, answered `405` with `x-amz-delete-marker: true` — returns `ObjectInfo{VersionID, IsDeleteMarker}` together with a `MethodNotAllowed` `ErrorResponse`. The SDK synthesizes that error exactly as before v7.0.93: it does not merge in the fields parsed from the server, and it does not merge in `ReplicationReady`. Tests pin both omissions.
- Every other error response returns the `ObjectInfo` built from the headers (`VersionID`, `IsDeleteMarker`, `ReplicationReady`) together with the error that `executeMethod` already parsed.
- When there is no response at all (`resp == nil`), `StatObject` returns a zero `ObjectInfo` with the error, as before.

**One behavior change beyond the restore.** `executeMethod` treats 202 and 204 as success, so a HEAD that returns either is now parsed like a 200. The old block turned them into errors. No known S3 server returns 202 or 204 for a HEAD.

The PR changes two files: `api-stat.go` and `api-stat_test.go`.

It also carried two `go-rdma.yml` CI fixes for a while — a vcpkg checkout pin and `persist-credentials: false` — because a broken arm64 job was blocking CI here. #2276 and #2279 have since landed the same fixes upstream, so this branch now takes the upstream version of that file and no longer changes it.

Fixes #2260

## Is this a breaking change?

No. This is a regression fix.

- **The API surface does not change.** `StatObject`'s signature, `ObjectInfo`'s fields, and every exported identifier are untouched. Existing callers compile unchanged.
- **The error contract holds.** The error path still returns a non-nil error of the same type, built by `httpRespToErrorResponse` inside `executeMethod`, so `minio.ToErrorResponse(err)` keeps working.
- **The behavior change restores the documented contract.** Returning `VersionID`, `IsDeleteMarker`, and `ReplicationReady` alongside the error is what `StatObject` did before v7.0.93, and losing them is what #2260 reports. #2115 made the change unintentionally; this PR undoes it. The only caller this affects is one that relied on `ObjectInfo` being zero whenever the error is non-nil. The docs never promised that, and the new doc comment now states the contract.
- **The 202/204 change is real but inert in practice.** Those statuses now parse like a 200 instead of becoming errors. No known S3 server returns them for a HEAD, and treating a success status as success is the more correct behavior.
- **Nothing else ships with it.** The only other change is the new unit tests in `api-stat_test.go`.

## Motivation and Context

Callers detect delete markers from `StatObject` errors. That behavior arrived in #1379 and #1397. Since v7.0.93 those callers have silently lost `VersionID`, `IsDeleteMarker`, and `ReplicationReady` — and, on 405 responses, the `MethodNotAllowed` error code as well.

## How to test this PR?

Run `go test -race -run TestStatObject .`. `api-stat_test.go` adds five regression tests. The first four fail on master and pass on this branch:

- `TestStatObjectDeleteMarker` — 405 with delete-marker headers returns a populated `ObjectInfo`, the `MethodNotAllowed` code, and status 405.
- `TestStatObjectMethodNotAllowedGeneric` — a 405 that is missing either half of the delete-marker shape (no marker header, or a stat that does not target a version) takes the generic path and reports the raw status code.
- `TestStatObjectNoContentSuccess` — 202 and 204 parse like a 200: nil error, fields taken from the headers.
- `TestStatObjectErrorHeaders` — a 404 carrying delete-marker and replication headers keeps all three fields alongside `NoSuchKey`.
- `TestStatObjectNoResponse` — an unreachable endpoint returns a zero `ObjectInfo` with the error.

`go test -short -race ./...` passes on this branch: 482 tests across 16 packages, run at merge commit `34e8078`, which merges the current `master`.

The same fix logic was validated live on 2026-07-17 against MinIO community master and AIStor master. The steps: create a versioned bucket, delete an object, then stat both the marker version and the latest version. On both servers the unfixed client loses the fields, and the fixed client returns `MethodNotAllowed` or `NoSuchKey` with the information populated. The repro test also passes on v7.0.92, which brackets the regression to #2115.

One detail worth knowing: `StatObject` sends a HEAD request, and Go's HTTP client never exposes a body for a HEAD response. `httpRespToErrorResponse` therefore always falls through to its `default:` arm and reports `Code` as the raw status line. The lost error code affects every server, not only servers that return an empty 405 body, and the synthesized `ErrorResponse` is the only source of `MethodNotAllowed` on this path. The unit tests pin this.

## Types of changes

## Checklist:

- [x] Fixes a regression (introduced by #2115, commit 7a9dff0)
- [ ] Create a documentation update request
- [x] No internode changes / version interoperability introduced

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved object metadata reporting when status requests encounter errors.
  - Preserved version and delete-marker details for relevant responses.
  - Added safer handling for missing responses and common HTTP error cases.
- **Tests**
  - Added regression coverage for delete markers, successful responses, failures, and error metadata.

### DIFF
--- a/.github/workflows/go-rdma.yml
+++ b/.github/workflows/go-rdma.yml
@@ -24,22 +24,27 @@ jobs:
     - name: Checkout minio-go
       uses: actions/checkout@v4
       with:
+        persist-credentials: false
         path: "minio-go"
 
     - name: Checkout minio-cpp
       uses: actions/checkout@v4
       with:
         repository: minio/minio-cpp
+        persist-credentials: false
         path: "minio-cpp"
 
     - name: Checkout vcpkg
       uses: actions/checkout@v4
       with:
         repository: microsoft/vcpkg
-        # Pinned: vcpkg master requires a newer CMake than the arm64
-        # runner image ships (scripts use string(JSON ... STRING_ENCODE)
-        # since microsoft/vcpkg@5397c5c9f), which fails every arm64 run.
+        # Pinned: vcpkg master needs CMake >= 4.3 (scripts use
+        # string(JSON ... STRING_ENCODE) since microsoft/vcpkg@5397c5c9f).
+        # vcpkg bootstraps its own CMake 4.4 on linux-amd64 but not on
+        # linux-arm64, which falls back to apt's CMake 3.28 and fails
+        # every arm64 run. Unpin once vcpkg supplies an arm64 CMake.
         ref: cd61e1e26a038e82d6550a3ebbe0fbbfe7da78e3 # tag 2026.06.24
+        persist-credentials: false
         path: "vcpkg"
 
     - name: Install system dependencies
--- a/.github/workflows/go-rdma.yml
+++ b/.github/workflows/go-rdma.yml
@@ -36,6 +36,10 @@ jobs:
       uses: actions/checkout@v4
       with:
         repository: microsoft/vcpkg
+        # Pinned: vcpkg master requires a newer CMake than the arm64
+        # runner image ships (scripts use string(JSON ... STRING_ENCODE)
+        # since microsoft/vcpkg@5397c5c9f), which fails every arm64 run.
+        ref: cd61e1e26a038e82d6550a3ebbe0fbbfe7da78e3 # tag 2026.06.24
         path: "vcpkg"
 
     - name: Install system dependencies
--- a/api-stat.go
+++ b/api-stat.go
@@ -57,7 +57,11 @@ func (c *Client) BucketExists(ctx context.Context, bucketName string) (bool, err
 }
 
 // StatObject verifies if object exists, you have permission to access it
-// and returns information about the object.
+// and returns information about the object. When the returned error is
+// non-nil but a response was received, the ObjectInfo still carries the
+// VersionID and IsDeleteMarker values parsed from the response headers,
+// plus ReplicationReady on every error path except the versioned
+// delete-marker 405.
 func (c *Client) StatObject(ctx context.Context, bucketName, objectName string, opts StatObjectOptions) (ObjectInfo, error) {
 	// Input validation.
 	if err := s3utils.CheckValidBucketName(bucketName); err != nil {
@@ -92,8 +96,9 @@ func (c *Client) StatObject(ctx context.Context, bucketName, objectName string, 
 	})
 	defer closeResponse(resp)
 	if err != nil {
-		// Surface the version, delete-marker, and replication-ready
-		// fields carried in the response headers alongside the error.
+		// executeMethod returns a non-nil error for every non-success
+		// status. When a response exists, its headers still carry the
+		// version and delete-marker fields — surface them with the error.
 		if resp == nil {
 			return ObjectInfo{}, err
 		}
@@ -115,7 +120,7 @@ func (c *Client) StatObject(ctx context.Context, bucketName, objectName string, 
 		return ObjectInfo{
 			VersionID:        resp.Header.Get(amzVersionID),
 			IsDeleteMarker:   deleteMarker,
-			ReplicationReady: replicationReady,
+			ReplicationReady: replicationReady, // whether delete marker can be replicated
 		}, err
 	}
 
--- a/api-stat.go
+++ b/api-stat.go
@@ -61,7 +61,8 @@ func (c *Client) BucketExists(ctx context.Context, bucketName string) (bool, err
 // non-nil but a response was received, the ObjectInfo still carries the
 // VersionID and IsDeleteMarker values parsed from the response headers,
 // plus ReplicationReady on every error path except the versioned
-// delete-marker 405.
+// delete-marker 405 (an asymmetry preserved from the pre-v7.0.93
+// behavior, which populated ReplicationReady on the generic branch only).
 func (c *Client) StatObject(ctx context.Context, bucketName, objectName string, opts StatObjectOptions) (ObjectInfo, error) {
 	// Input validation.
 	if err := s3utils.CheckValidBucketName(bucketName); err != nil {
--- a/api-stat.go
+++ b/api-stat.go
@@ -92,32 +92,31 @@ func (c *Client) StatObject(ctx context.Context, bucketName, objectName string, 
 	})
 	defer closeResponse(resp)
 	if err != nil {
-		return ObjectInfo{}, err
-	}
-
-	if resp != nil {
+		// Surface the version, delete-marker, and replication-ready
+		// fields carried in the response headers alongside the error.
+		if resp == nil {
+			return ObjectInfo{}, err
+		}
 		deleteMarker := resp.Header.Get(amzDeleteMarker) == "true"
 		replicationReady := resp.Header.Get(minioTgtReplicationReady) == "true"
-		if resp.StatusCode != http.StatusOK && resp.StatusCode != http.StatusPartialContent {
-			if resp.StatusCode == http.StatusMethodNotAllowed && opts.VersionID != "" && deleteMarker {
-				errResp := ErrorResponse{
-					StatusCode: resp.StatusCode,
-					Code:       MethodNotAllowed,
-					Message:    s3ErrorResponseMap[MethodNotAllowed],
-					BucketName: bucketName,
-					Key:        objectName,
-				}
-				return ObjectInfo{
-					VersionID:      resp.Header.Get(amzVersionID),
-					IsDeleteMarker: deleteMarker,
-				}, errResp
+		if resp.StatusCode == http.StatusMethodNotAllowed && opts.VersionID != "" && deleteMarker {
+			errResp := ErrorResponse{
+				StatusCode: resp.StatusCode,
+				Code:       MethodNotAllowed,
+				Message:    s3ErrorResponseMap[MethodNotAllowed],
+				BucketName: bucketName,
+				Key:        objectName,
 			}
 			return ObjectInfo{
-				VersionID:        resp.Header.Get(amzVersionID),
-				IsDeleteMarker:   deleteMarker,
-				ReplicationReady: replicationReady, // whether delete marker can be replicated
-			}, httpRespToErrorResponse(resp, bucketName, objectName)
+				VersionID:      resp.Header.Get(amzVersionID),
+				IsDeleteMarker: deleteMarker,
+			}, errResp
 		}
+		return ObjectInfo{
+			VersionID:        resp.Header.Get(amzVersionID),
+			IsDeleteMarker:   deleteMarker,
+			ReplicationReady: replicationReady,
+		}, err
 	}
 
 	return ToObjectInfo(bucketName, objectName, resp.Header)
--- a/api-stat_test.go
+++ b/api-stat_test.go
@@ -205,3 +205,20 @@ func TestStatObjectErrorHeaders(t *testing.T) {
 		t.Error("expected ReplicationReady to be true")
 	}
 }
+
+// Tests that the delete-marker branch is gated on the 405 status: a
+// non-405 error carrying the same delete-marker and version shape must
+// still take the generic path.
+func TestStatObjectDeleteMarkerNon405(t *testing.T) {
+	clnt := newTestStatClient(t, func(w http.ResponseWriter, _ *http.Request) {
+		w.Header().Set(amzDeleteMarker, "true")
+		w.Header().Set(amzVersionID, "test-version-id")
+		w.WriteHeader(http.StatusNotFound)
+	})
+
+	_, err := clnt.StatObject(context.Background(), "bucket-name", "object-name",
+		StatObjectOptions{VersionID: "test-version-id"})
+	if errResp := ToErrorResponse(err); errResp.Code != NoSuchKey {
+		t.Errorf("error code = %q, want %q", errResp.Code, NoSuchKey)
+	}
+}
--- a/api-stat_test.go
+++ b/api-stat_test.go
@@ -21,35 +21,40 @@ import (
 	"context"
 	"net/http"
 	"net/http/httptest"
-	"net/url"
+	"reflect"
 	"testing"
 
 	"github.com/minio/minio-go/v7/pkg/credentials"
 )
 
-// Tests that StatObject returns the delete-marker ObjectInfo fields
-// (VersionID, IsDeleteMarker) and the MethodNotAllowed error code when a
-// versioned HEAD hits a delete marker (HTTP 405).
-func TestStatObjectDeleteMarker(t *testing.T) {
-	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
-		w.Header().Set(amzDeleteMarker, "true")
-		w.Header().Set(amzVersionID, "test-version-id")
-		w.WriteHeader(http.StatusMethodNotAllowed)
-	}))
-	defer srv.Close()
+// newTestStatClient returns a Client pointed at an httptest server that
+// serves handler; the server is closed via t.Cleanup.
+func newTestStatClient(t *testing.T, handler http.HandlerFunc) *Client {
+	t.Helper()
+	srv := httptest.NewServer(handler)
+	t.Cleanup(srv.Close)
 
-	u, err := url.Parse(srv.URL)
-	if err != nil {
-		t.Fatal(err)
-	}
-
-	clnt, err := New(u.Host, &Options{
+	clnt, err := New(srv.Listener.Addr().String(), &Options{
 		Creds:  credentials.NewStaticV4("foo", "foo12345", ""),
 		Region: "us-east-1",
 	})
 	if err != nil {
 		t.Fatal(err)
 	}
+	return clnt
+}
+
+// Tests that StatObject returns the delete-marker ObjectInfo fields
+// (VersionID and IsDeleteMarker — ReplicationReady is deliberately not
+// merged into this return) and the MethodNotAllowed error code when a
+// versioned HEAD hits a delete marker (HTTP 405).
+func TestStatObjectDeleteMarker(t *testing.T) {
+	clnt := newTestStatClient(t, func(w http.ResponseWriter, _ *http.Request) {
+		w.Header().Set(amzDeleteMarker, "true")
+		w.Header().Set(amzVersionID, "test-version-id")
+		w.Header().Set(minioTgtReplicationReady, "true")
+		w.WriteHeader(http.StatusMethodNotAllowed)
+	})
 
 	objInfo, err := clnt.StatObject(context.Background(), "bucket-name", "object-name",
 		StatObjectOptions{VersionID: "test-version-id"})
@@ -63,11 +68,18 @@ func TestStatObjectDeleteMarker(t *testing.T) {
 	if errResp.StatusCode != http.StatusMethodNotAllowed {
 		t.Errorf("error status = %d, want %d", errResp.StatusCode, http.StatusMethodNotAllowed)
 	}
+	if errResp.BucketName != "bucket-name" || errResp.Key != "object-name" {
+		t.Errorf("error bucket/key = %q/%q, want %q/%q",
+			errResp.BucketName, errResp.Key, "bucket-name", "object-name")
+	}
 	if !objInfo.IsDeleteMarker {
 		t.Error("expected IsDeleteMarker to be true")
 	}
 	if objInfo.VersionID != "test-version-id" {
 		t.Errorf("VersionID = %q, want %q", objInfo.VersionID, "test-version-id")
+	}
+	if objInfo.ReplicationReady {
+		t.Error("expected ReplicationReady to stay false on the delete-marker return")
 	}
 }
 
@@ -75,6 +87,7 @@ func TestStatObjectDeleteMarker(t *testing.T) {
 // shape (the x-amz-delete-marker header, or a version-targeted stat)
 // falls through to the generic error path with the raw status code.
 func TestStatObjectMethodNotAllowedGeneric(t *testing.T) {
+	const wantCode = "405 Method Not Allowed"
 	tests := []struct {
 		name         string
 		deleteMarker bool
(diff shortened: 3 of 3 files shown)


api-stat.go
```diff
@@ -92,32 +92,31 @@ func (c *Client) StatObject(ctx context.Context, bucketName, objectName string,
...
+				VersionID:      resp.Header.Get(amzVersionID),
+				IsDeleteMarker: deleteMarker,
+			}, errResp
 		}
+		return ObjectInfo{
+			VersionID:        resp.Header.Get(amzVersionID),
+			IsDeleteMarker:   deleteMarker,
+			ReplicationReady: resp.Header.Get(minioTgtReplicationReady) == "true", // whether delete marker can be replicated
+		}, err
 	}
```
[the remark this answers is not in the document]
allanrogerr: Implemented in 2bd0f11: TestStatObjectNoContentSuccess covers 202 and 204 (both fail on the pre-fix base), and TestStatObjectMethodNotAllowedGeneric adds the remaining 405 fall-through cases.

api-stat.go
```diff
@@ -92,32 +92,31 @@ func (c *Client) StatObject(ctx context.Context, bucketName, objectName string,
...
-			}, httpRespToErrorResponse(resp, bucketName, objectName)
+				VersionID:      resp.Header.Get(amzVersionID),
+				IsDeleteMarker: deleteMarker,
+			}, errResp
 		}
+		return ObjectInfo{
+			VersionID:        resp.Header.Get(amzVersionID),
+			IsDeleteMarker:   deleteMarker,
+			ReplicationReady: resp.Header.Get(minioTgtReplicationReady) == "true", // whether delete marker can be replicated
+		}, err
```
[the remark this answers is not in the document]
allanrogerr: Keeping this inline. The branch is three flat cases with a single caller, and every path is pinned by tests that exercise the real request path — dropping either conjunct of the 405 condition now fails TestStatObjectMethodNotAllowedGeneric. A one-caller helper adds indirection without new coverage, and the inline shape matches the pre-v7.0.93 code this PR restores.

api-stat_test.go
```diff
@@ -0,0 +1,236 @@
...
+	if !objInfo.IsDeleteMarker {
+		t.Error("expected IsDeleteMarker to be true")
+	}
+	if objInfo.VersionID != "test-version-id" {
+		t.Errorf("VersionID = %q, want %q", objInfo.VersionID, "test-version-id")
+	}
+	if !objInfo.ReplicationReady {
+		t.Error("expected ReplicationReady to be true")
+	}
+}
```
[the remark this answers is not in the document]
allanrogerr: Extracted `newTestStatClient(t, handler)` in d3d5611, covering the four identical setups; it constructs the client from `srv.Listener.Addr().String()` directly, dropping the `url.Parse` step, and closes the server via `t.Cleanup`. `TestStatObjectNoResponse` keeps its own construction (closed listener + `MaxRetries`). Unlike the single-caller extraction declined earlier on this PR, this one removes four duplicates.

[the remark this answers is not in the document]
allanrogerr: d3d5611 is now the PR head (the push landed just after the reply above); `newTestStatClient` is at api-stat_test.go:32 with the four call sites.

.github/workflows/go-rdma.yml
```diff
@@ -24,18 +24,27 @@ jobs:
     - name: Checkout minio-go
       uses: actions/checkout@v4
       with:
+        persist-credentials: false
         path: "minio-go"
 
     - name: Checkout minio-cpp
       uses: actions/checkout@v4
```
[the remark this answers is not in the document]
allanrogerr: These steps are not added in this PR and will be addressed separately in https://github.com/minio/minio-go/pull/2276. THis is beyond the current PR scope

jiuker (approved):

shtripat (commented):
Went through this against master and ran the suite both ways. The fix holds up: I reproduced the regression on `802bd60` (four of the five new tests fail there), and the root cause is confirmed — #2115 added `err = errResponse` at `api.go:790`, and before that the named `err` was never assigned inside the retry loop, so `executeMethod` returned `(res, nil)` for non-success statuses and the old block was reachable. `go test -short -race ./...` is green (481 tests, 16 packages), `go tool golangci-lint run ./...` reports 0 issues, and `StatObject` coverage goes 59.1% -> 81.0% with no package dropping.

Two things worth fixing before merge, both one-liners:

- The `resp.StatusCode == http.StatusMethodNotAllowed` conjunct at `api-stat.go:107` isn't pinned by any test — I deleted it locally and all 9 tests still passed. Details and a test inline on `api-stat_test.go`.
- The XML-error-body note in the description doesn't hold for a HEAD request. The correction and a full replacement description are in a separate comment on this PR.

The rest is minor.

api-stat_test.go
```diff
@@ -0,0 +1,207 @@
...
```
shtripat, line 233: The `resp.StatusCode == http.StatusMethodNotAllowed` half of the guard at `api-stat.go:107` isn't pinned by anything — I deleted it locally and all 9 tests still pass. The two rows at `api-stat_test.go:96-97` cover the `deleteMarker` and `opts.VersionID` conjuncts, and `TestStatObjectErrorHeaders` at `api-stat_test.go:183` sends a 404 with the delete-marker header but leaves `opts.VersionID` empty, so nothing reaches the branch on a non-405 status. Drop that conjunct and a 404 for a missing version that still echoes `x-amz-delete-marker` gets a synthesized `MethodNotAllowed` instead of `NoSuchKey`, silently. This closes it — I checked it passes as-is and fails with the conjunct removed:

```suggestion
}

// Tests that the delete-marker branch is gated on the 405 status: a
// non-405 error carrying the same delete-marker and version shape must
// still take the generic path.
func TestStatObjectDeleteMarkerNon405(t *testing.T) {
	clnt := newTestStatClient(t, func(w http.ResponseWriter, _ *http.Request) {
		w.Header().Set(amzDeleteMarker, "true")
		w.Header().Set(amzVersionID, "test-version-id")
		w.WriteHeader(http.StatusNotFound)
	})

	_, err := clnt.StatObject(context.Background(), "bucket-name", "object-name",
		StatObjectOptions{VersionID: "test-version-id"})
	if errResp := ToErrorResponse(err); errResp.Code != NoSuchKey {
		t.Errorf("error code = %q, want %q", errResp.Code, NoSuchKey)
	}
}
```

api-stat.go
```diff
@@ -57,7 +57,11 @@ func (c *Client) BucketExists(ctx context.Context, bucketName string) (bool, err
 }
```
shtripat, line 57: `BucketExists` just above has the same shape you're fixing here: its `if resp != nil` block at `api-stat.go:47-55` is only reachable when `err == nil`, which after #2115 means the status is already in `successStatus` (`api.go:672-677`) — so the `resp.StatusCode != http.StatusOK` check at `api-stat.go:52` now converts successes into errors. Against an httptest server it returns `false, "202 Accepted"`, `false, "204 No Content"` and `false, "206 Partial Content"`, which is exactly the 202/204 behaviour your description calls out for `StatObject`. Fine to leave for a follow-up, but worth making it a deliberate call rather than an oversight.

.github/workflows/go-rdma.yml
```diff
@@ -24,18 +24,27 @@ jobs:
     - name: Checkout minio-go
       uses: actions/checkout@v4
       with:
+        persist-credentials: false
```
shtripat, line 27: `persist-credentials: false` on these three steps isn't in the PR description — the ride-along bullet covers only the vcpkg pin, and the breaking-change bullet says "`go-rdma.yml` checkout pin, new unit tests". It's a good change (this job runs `bootstrap-vcpkg.sh`, `vcpkg install` and `cmake` inside the workspace, so keeping the token out of `.git/config` matters), which is why it deserves a line in the description. Worth noting it's only applied here: `.github/workflows/go.yml:30`, `.github/workflows/vulncheck.yml:20` and `.github/workflows/go-windows.yml:30` still persist credentials — good companion for #2276.

api-stat.go
```diff
@@ -57,7 +57,11 @@ func (c *Client) BucketExists(ctx context.Context, bucketName string) (bool, err
...
 
 // StatObject verifies if object exists, you have permission to access it
-// and returns information about the object.
+// and returns information about the object. When the returned error is
+// non-nil but a response was received, the ObjectInfo still carries the
+// VersionID and IsDeleteMarker values parsed from the response headers,
+// plus ReplicationReady on every error path except the versioned
```
shtripat, line 63: The "except the versioned delete-marker 405" clause reads as arbitrary without the reason. I traced it: the asymmetry comes from #1707, which added `ReplicationReady` to the generic branch only, and the one real consumer — AIStor `cmd/bucket-replication.go:948-984` — handles the 405 in `case isErrMethodNotAllowed(serr):` while the `!toi.ReplicationReady` check sits in `default:`, so nothing is affected by the omission. Adding "(preserved from the pre-v7.0.93 behaviour)" saves the next reader that trace.

api-stat_test.go
```diff
@@ -0,0 +1,207 @@
...
+
+// Tests that StatObject surfaces the delete-marker and replication-ready
+// headers on a generic error response, e.g. HEAD on an object whose
+// latest version is a delete marker (HTTP 404).
+func TestStatObjectErrorHeaders(t *testing.T) {
+	clnt := newTestStatClient(t, func(w http.ResponseWriter, _ *http.Request) {
+		w.Header().Set(amzDeleteMarker, "true")
+		w.Header().Set(amzVersionID, "test-version-id")
+		w.Header().Set(minioTgtReplicationReady, "true")
```
shtripat, line 191: This sets the response header directly, so the request half of the round trip is never exercised: nothing in the suite sets `opts.Internal.IsReplicationReadyForDeleteMarker`, which makes `api-stat.go:85-87` the only part of `StatObject` the new tests leave at 0% in the coverage profile. Setting that option here and asserting the handler saw `X-Minio-Check-Replication-Ready: true` would pin the whole path for one extra line.

shtripat:
Three corrections to the description — two factual, one omission. Full replacement below, everything else is your wording verbatim.

`````md
## Description

Since #2115 (first released in v7.0.93), `executeMethod` returns a non-nil error for every response outside its success set — 200/204/206 then, 202 added later by #2252 — so `StatObject` early-returns a zero `ObjectInfo` before its delete-marker handling block can run — leaving the block unreachable for its intended 405/404 delete-marker paths. This PR moves that logic into the `err != nil` branch, restoring the pre-v7.0.93 behavior:

- Stat of a delete marker (HEAD with `VersionID`, answered 405 + `x-amz-delete-marker: true`) again returns `ObjectInfo{VersionID, IsDeleteMarker}` alongside a `MethodNotAllowed` `ErrorResponse` (synthesized exactly as before v7.0.93; the parsed server fields are not merged in, and `ReplicationReady` is likewise not merged into this return — pinned by the tests).
- Any other error-bearing response returns the header-populated `ObjectInfo` (`VersionID`, `IsDeleteMarker`, `ReplicationReady`) alongside the error `executeMethod` already parsed.
- When no response exists (`resp == nil`), `StatObject` still returns a zero `ObjectInfo` with the error. The old block is removed.
- One narrow behavior change beyond the restore: 202/204 HEAD responses, which the old block turned into errors, are now parsed like a 200 (`executeMethod` treats them as success). No known S3 server returns these for HEAD.
- Rides along, both in `.github/workflows/go-rdma.yml`: (a) pins the vcpkg checkout to an immutable commit (`cd61e1e2`, tag `2026.06.24`); (b) adds `persist-credentials: false` to all three `actions/checkout` steps, so the job token is not written into `.git/config` before the vcpkg bootstrap and CMake build run inside the workspace. On the pin: vcpkg master now requires CMake >= 4.3, which vcpkg's tool bootstrap downloads on linux-amd64 but not on linux-arm64 (falling back to apt's CMake 3.28), failing every arm64 run since 2026-07-30 — unrelated to this fix, but it blocks CI here.

Fixes #2260

## Is this a breaking change?

No — this is a regression-fix branch, not a breaking-change branch:

- **No API-surface change.** `StatObject`'s signature, `ObjectInfo`'s fields, and all exported identifiers are untouched; every existing caller compiles unchanged.
- **Error contract preserved.** The error path still returns a non-nil error of the same type — `executeMethod` builds it via `httpRespToErrorResponse`, so `minio.ToErrorResponse(err)` keeps working exactly as before.
- **The behavior change restores the documented contract.** Populating `VersionID`/`IsDeleteMarker`/`ReplicationReady` alongside the error is what `StatObject` did before v7.0.93 and what #2260 reports as broken; v7.0.93 (#2115) was the unintentional behavior change, and this PR undoes it. A caller is only affected if it relied on `ObjectInfo` being zero-valued whenever `err != nil` — an anti-pattern the docs never promised (the new doc comment now spells the contract out).
- **The 202/204 change is real but practically inert.** Those statuses now parse like a 200 instead of being converted to errors, but no known S3 server returns them for HEAD, and treating a success status as success is the more correct behavior.
- **The ride-along commits are CI hygiene and tests** (`go-rdma.yml` checkout pin, `persist-credentials: false`, new unit tests) — no public-API impact.

## Motivation and Context

Callers that detect delete markers from `StatObject` errors (behavior added in #1379/#1397) silently lost `VersionID`, `IsDeleteMarker`, and `ReplicationReady`, and — for 405 responses — the `MethodNotAllowed` error code.

## How to test this PR?

- `go test -race -run TestStatObject .` — five new regression tests in `api-stat_test.go`:
  - `TestStatObjectDeleteMarker`: 405 + delete-marker headers → populated `ObjectInfo`, `MethodNotAllowed` code and 405 status. Fails on master, passes here.
  - `TestStatObjectMethodNotAllowedGeneric`: 405 missing either half of the delete-marker shape (no marker header, or no version-targeted stat) → the generic path with the raw status code. Fails on master, passes here.
  - `TestStatObjectNoContentSuccess`: 202/204 → parsed like a 200 (nil error, header-derived fields). Fails on master, passes here.
  - `TestStatObjectErrorHeaders`: 404 with delete-marker/replication headers → all three fields survive alongside `NoSuchKey`. Fails on master, passes here.
  - `TestStatObjectNoResponse`: unreachable endpoint → zero `ObjectInfo` with the error.
- Full `go test -short -race ./...` passes (16 packages) on this branch (base `802bd60`).
- Live-validated (2026-07-17, against the same fix logic) on MinIO community master and AIStor master (versioned bucket → delete object → stat the marker version and the latest version): the unfixed client loses the fields on both servers; the fixed client returns `MethodNotAllowed`/`NoSuchKey` with populated info. The repro test also passes on v7.0.92, bracketing the regression to #2115. Note: `StatObject` issues a HEAD, and Go's HTTP client never exposes a body for a HEAD response, so `httpRespToErrorResponse` always falls through to its `default:` arm and reports `Code` as the raw status line. The error-code half of the regression therefore hits every server, not only empty-body 405s, and the synthesized `ErrorResponse` is the only source of `MethodNotAllowed` on this path — pinned by the unit tests.

## Types of changes

## Checklist:

- [x] Fixes a regression (introduced by #2115, commit 7a9dff0)
- [ ] Create a documentation update request
- [x] No internode changes / version interoperability introduced
`````

**What changed and why**

**"servers that send an XML error body (such as MinIO) supply the `MethodNotAllowed` code from the body, so the error-code half of the regression manifests only with empty-body 405s"** — this can't happen. `StatObject` sends `http.MethodHead` (`api-stat.go:90`), and Go's HTTP client never exposes a body for a HEAD response, so `xmlDecodeAndBody` at `api-error-response.go:134` always fails with `io.EOF` and `httpRespToErrorResponse` always takes the `default:` arm at `api-error-response.go:185-198`, which sets `Code = resp.Status`. I confirmed it with a raw listener replying `405` + a correct `Content-Length` + a full S3 XML error: the generic path returned `Code="405 Method Not Allowed"` and the body's `<Message>` never appeared. MinIO's own HEAD handler uses `writeErrorResponseHeadersOnly` and sends no body here at all, and there is no `x-minio-error-code` constant in AIStor's `internal/http/headers.go`, so the header override at `api-error-response.go:202-205` doesn't apply either. Worth getting right because as written it makes the regression sound narrower than it was, which is what someone sizing a backport will read — and it undersells the fix: the synthesized `ErrorResponse` at `api-stat.go:108-114` is the only source of `MethodNotAllowed` on this path, on every server.

**"and — for empty-body 405 responses — the `MethodNotAllowed` error code"** (Motivation and Context) — same cause; narrowed to "for 405 responses".

**`persist-credentials: false` is undocumented** — it's added to all three `actions/checkout` steps at `.github/workflows/go-rdma.yml:27,34,47`, but the ride-along bullet mentions only the vcpkg pin and the breaking-change bullet says only "checkout pin, new unit tests". Folded into both bullets. `d3d5611`'s subject does say "checkout hygiene", so `git log` records it — only the description was silent.

api-stat_test.go
allanrogerr: Fixed in f347e34 — added the suggested `TestStatObjectDeleteMarkerNon405` verbatim. Re-verified the mutation locally: with the `resp.StatusCode == http.StatusMethodNotAllowed` conjunct removed, the new test fails (`error code = "MethodNotAllowed", want "NoSuchKey"`); with the conjunct intact, all 10 StatObject tests pass under `-race`.

api-stat.go
allanrogerr: Confirmed, and deferring is the deliberate call. `err == nil` now implies the status is in `successStatus` (200/202/204/206 — `api.go:672-677`), so the `resp.StatusCode != http.StatusOK` check in `BucketExists` converts 202/204/206 into errors, exactly the shape this PR fixes for `StatObject`. Fixing it needs its own behavior call-out and tests, so it stays out of this regression fix and goes to a follow-up PR.

.github/workflows/go-rdma.yml
allanrogerr: Done — the PR description now documents the `persist-credentials: false` change: a new ride-along bullet explains why the token must not stay in `.git/config` while the build scripts run, and the breaking-change bullet names it. Verified the edit persisted by re-fetching the body. The other three workflows stay untouched here, as you suggest — that belongs with #2276.

api-stat.go
allanrogerr: Fixed in 2800b67 — the doc comment now reads "(an asymmetry preserved from the pre-v7.0.93 behavior, which populated ReplicationReady on the generic branch only)". Verified against the actual v7.0.92 blob (`git show v7.0.92:api-stat.go`): the old 405 delete-marker return built `ObjectInfo{VersionID, IsDeleteMarker}` without `ReplicationReady`, while the generic branch set it — same shape this PR restores.

api-stat_test.go
allanrogerr: Fixed in bbcf609 — `TestStatObjectErrorHeaders` now sets `Internal: AdvancedGetOptions{IsReplicationReadyForDeleteMarker: true}` and asserts the handler saw `X-Minio-Check-Replication-Ready: true`, so the request-side mapping at `api-stat.go:85-87` is exercised. All 10 StatObject tests pass under `-race`.

shtripat (commented):
Delta looks good — pulled `bbcf6094` and re-ran everything.

`f347e34` closes the conjunct gap: I re-ran the mutation and dropping `resp.StatusCode == http.StatusMethodNotAllowed` from `api-stat.go:108` now fails the suite, where it passed before. `bbcf609` does the same for the request side — deleting the `headers.Set(isMinioTgtReplicationReady, "true")` block at `api-stat.go:86-88` is now caught, and `StatObject` coverage goes 81.0% -> 85.7% with the package total flat. I checked `2800b67`'s justification against `git show v7.0.92:api-stat.go` and it's right: the old 405 return really did omit `ReplicationReady` while the generic branch set it. `go test -short -race ./...` is 482/482 across 16 packages, `-race -count=40` on the two new tests is clean, and `go tool golangci-lint run ./...` reports 0 issues. Agreed on deferring `api-stat.go:47-55` to a follow-up.

One thing from my description comment is still outstanding — the "servers that send an XML error body (such as MinIO) supply the `MethodNotAllowed` code from the body" sentence under "How to test this PR?", and the matching "for empty-body 405 responses" in "Motivation and Context". `StatObject` sends a HEAD (`api-stat.go:91`) and Go's client never exposes a body for a HEAD response, so `httpRespToErrorResponse` always takes the `default:` arm at `api-error-response.go:185-198` and reports the raw status line — the error-code half of the regression hit every server, not just empty-body ones. The replacement text is in https://github.com/minio/minio-go/pull/2273#issuecomment-5162432095; the `persist-credentials` bullets from that same comment already landed.

allanrogerr:
Done — both outstanding sentences from https://github.com/minio/minio-go/pull/2273#issuecomment-5162432095 are now in the description, verbatim: the "How to test this PR?" note now explains that `StatObject` issues a HEAD, Go's client never exposes a HEAD response body, and `httpRespToErrorResponse` therefore always reports the raw status line — so the error-code half of the regression hits every server; "Motivation and Context" now says "for 405 responses". Re-verified the mechanism at head `bbcf609` before editing: `api-stat.go` sends `http.MethodHead` and the `default:` arm at `api-error-response.go` sets `Code = resp.Status`. Confirmed the edit persisted by re-fetching the body.

shtripat (approved):

harshavardhana:
Please fix conflicts

allanrogerr:
@harshavardhana Conflicts fixed

PR #2274, opened 2026-08-01
fix: trim whitespace from bucket location

### LINKED ISSUE #2271: GetBucketLocation() should trim whitespace from LocationConstraint before caching bucket location
# Summary

When communicating with an S3-compatible provider that returns a whitespace-only `LocationConstraint`, `minio-go` caches the whitespace verbatim. That cached value is later used as the region component of the SigV4 credential scope, producing an invalid `Authorization` header containing embedded newlines.

Trimming the `LocationConstraint` before caching resolves the issue.

---

# Environment

- minio-go: **v7.2.1**
- Go: **1.26**
- Tested against a third-party S3-compatible provider.

---

# Minimal reproducer

```
client, err := minio.New(endpoint, &minio.Options{
    Creds:  credentials.NewStaticV4(accessKey, secretKey, ""),
    Secure: true,
})

client.TraceOn(os.Stdout)

_, err = client.FPutObject(
    context.Background(),
    bucket,
    "test.txt",
    "test.txt",
    minio.PutObjectOptions{},
)

```

(Full reproducer attached.)

---

# Investigation

I instrumented:

- `bucket-cache.go`
- `request-signature-v4.go`

The only functional change was:

```
location := strings.TrimSpace(locationConstraint)
```

Everything else was debug logging.

---

# Observed behavior (before patch)

The bucket location response is decoded as:

```
LocationConstraint: "\n  \n  \n  \n  \n  \n  \n    \n\n"
```

That value is cached and later becomes the region portion of the SigV4 credential scope.

The generated credential becomes conceptually:

```
ACCESSKEY/DATE/<whitespace>/s3/aws4_request
```

resulting in an `Authorization` header containing embedded newlines.

Go then rejects the request before it is sent with:

```
net/http: invalid header field value for "Authorization"
```

---

# Expected behavior

Whitespace surrounding `LocationConstraint` should not become part of the cached bucket location.

A whitespace-only value should behave the same as an empty value and ultimately resolve to the default region (`us-east-1`).

---

# Proposed fix

```
- location := locationConstraint
+ location := strings.TrimSpace(locationConstraint)

if location == "" {
    location = "us-east-1"
}
```

---

# Result after patch

With only the above change:

- identical credentials
- identical endpoint
- identical bucket
- identical application

the upload succeeds successfully.

---

# Notes

I believe the S3 provider is also behaving poorly by returning a whitespace-only `LocationConstraint` rather than an empty element or a region value.

However, trimming the decoded value makes `minio-go` more robust and prevents malformed `Authorization` headers from being generated.

[debug.output.txt](https://github.com/user-attachments/files/30442693/debug.output.txt)
[diff.txt](https://github.com/user-attachments/files/30442691/diff.txt)
[main.go.txt](https://github.com/user-attachments/files/30442692/main.go.txt)
[patched.output.txt](https://github.com/user-attachments/files/30442694/patched.output.txt)

vigneshakaviki:
## Description

Trim whitespace from decoded `LocationConstraint` values before applying existing empty-region and EU normalization. This prevents whitespace-only responses from being cached and inserted into SigV4 `Authorization` headers.

## Testing

- `go test -run '^TestProcessBucketLocationResponse$' -count=1 .`
- `go test -short -race ./...`
- `go test -race -v ./...` against TLS-enabled AIStor edge
- Full functional suite against TLS-enabled AIStor edge
- `make lint`
- `make examples`
- `go build ./...`
- `govulncheck ./...` with Go 1.26.5

Fixes #2271

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved bucket location handling by removing surrounding whitespace from responses.
  * Ensured blank or whitespace-only locations correctly use the default region.

* **Tests**
  * Added coverage for blank locations and regional values containing extra whitespace.

### DIFF
--- a/bucket-cache.go
+++ b/bucket-cache.go
@@ -23,6 +23,7 @@ import (
 	"net/http"
 	"net/url"
 	"path"
+	"strings"
 
 	"github.com/minio/minio-go/v7/pkg/credentials"
 	"github.com/minio/minio-go/v7/pkg/s3utils"
@@ -114,7 +115,7 @@ func processBucketLocationResponse(resp *http.Response, bucketName string) (buck
 		return "", err
 	}
 
-	location := locationConstraint
+	location := strings.TrimSpace(locationConstraint)
 	// Location is empty will be 'us-east-1'.
 	if location == "" {
 		location = "us-east-1"
--- a/bucket-cache_test.go
+++ b/bucket-cache_test.go
@@ -309,6 +309,8 @@ func TestProcessBucketLocationResponse(t *testing.T) {
 	}{
 		{"my-bucket", "", true, APIErrors[0], "us-east-1", nil, true},
 		{"my-bucket", "", false, APIError{}, "us-east-1", nil, true},
+		{"my-bucket", "\n  \n", false, APIError{}, "us-east-1", nil, true},
+		{"my-bucket", " \neu-central-1\t", false, APIError{}, "eu-central-1", nil, true},
 		{"my-bucket", "EU", false, APIError{}, "eu-west-1", nil, true},
 		{"my-bucket", "eu-central-1", false, APIError{}, "eu-central-1", nil, true},
 		{"my-bucket", "us-east-1", false, APIError{}, "us-east-1", nil, true},


bucket-cache_test.go
```diff
@@ -309,6 +309,8 @@ func TestProcessBucketLocationResponse(t *testing.T) {
 	}{
 		{"my-bucket", "", true, APIErrors[0], "us-east-1", nil, true},
 		{"my-bucket", "", false, APIError{}, "us-east-1", nil, true},
+		{"my-bucket", "\n  \n", false, APIError{}, "us-east-1", nil, true},
+		{"my-bucket", " \neu-central-1\t", false, APIError{}, "eu-central-1", nil, true},
```
allanrogerr, line 313: The description says the trim happens before the `EU` normalization, but neither new row exercises that ordering. The padded `eu-central-1` row trims to a value that never reaches the `EU` branch, and the plain `"EU"` row passes with or without trimming, so nothing here would notice if the two steps were ever reordered.

This row covers the interaction directly: it resolves to `eu-west-1` with the trim in place, and comes back as `" EU "` unchanged without it.

```suggestion
		{"my-bucket", " \neu-central-1\t", false, APIError{}, "eu-central-1", nil, true},
		{"my-bucket", " EU ", false, APIError{}, "eu-west-1", nil, true},
```

bucket-cache.go
```diff
@@ -114,7 +115,7 @@ func processBucketLocationResponse(resp *http.Response, bucketName string) (buck
 		return "", err
 	}
 
-	location := locationConstraint
+	location := strings.TrimSpace(locationConstraint)
```
allanrogerr, line 118: One branch up, the error-response path returns `errResp.Region` without the same treatment, and the `errResp.Region == ""` check guarding it has the identical gap this line just closed below: a whitespace-only region is not `""`, so it is returned as-is, cached by `getBucketLocation`, and ends up as the SigV4 region — the same newline-in-the-`Authorization`-header failure described in #2271.

The same value also reaches that cache and the signer from `executeMethod`'s retry path (`api.go:805`, `814`, `822`) without passing through this function at all, so trimming only here would not cover it. Trimming once where `errResp.Region` is populated covers all three callers, and has the nicer property that a whitespace-only region normalizes to `""` and falls through to the `x-amz-bucket-region` header instead of defaulting to `us-east-1`. `strings` is already imported in that file.

In `api-error-response.go`, immediately before the existing header fallback:

```go
	errResp.Region = strings.TrimSpace(errResp.Region)
	if errResp.Region == "" {
		errResp.Region = resp.Header.Get("x-amz-bucket-region")
	}
```

I have not seen a server actually emit a padded `<Region>` in an error body, so treat this as hardening rather than something you are hitting today — likely its own PR, and entirely your call.

allanrogerr (changes requested):
PTAL

PR #2278, opened 2026-08-04
rdma: reject buffers over the 4 GiB cuObject registration limit

harshavardhana:
## What

A single cuObject registration (`cuMemObjGetDescriptor`, inside libminiocpp) can pin at most **4 GiB**. `putObjectRDMA`/`getObjectRDMA` passed `RDMABufferSize` straight into the cgo call, so an oversized buffer failed **opaquely** in the C layer (`RDMA put/get: <last error>`).

This adds a fail-fast guard: reject `RDMABufferSize > 4 GiB` before the cgo call with a clear, actionable error telling the caller to split the transfer into parts ≤ 4 GiB (multipart upload / ranged read).

- Add `maxRDMABufferSize` (4 GiB) and guard both `putObjectRDMA` and `getObjectRDMA`.

The substantive handling lives upstream in libminiocpp (already merged); minio-go is a thin cgo wrapper, so its correct role is a clear guard rather than a workaround. Mirrors the equivalent guards in minio-cpp and minio-rs.

## Testing

`gofmt` clean. The `//go:build rdma` cgo build (which links libminiocpp) is exercised by the `go-rdma` CI workflow; the added code is plain Go (a const + two `fmt.Errorf` guards).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation to prevent RDMA transfers exceeding 4 GiB.
  * Oversized put and get operations now return clear guidance to split data into smaller transfers.

### DIFF
--- a/rdma.go
+++ b/rdma.go
@@ -26,6 +26,13 @@ import (
 
 var ErrRDMANotConnected = errors.New("RDMA infrastructure not connected")
 
+// maxRDMABufferSize is the largest buffer a single cuObject registration
+// (cuMemObjGetDescriptor, inside libminiocpp) can pin — 4 GiB. A larger buffer
+// cannot be RDMA-registered, so reject it here with a clear error instead of
+// letting the cgo call fail opaquely; the caller must split the transfer into
+// parts no larger than this (multipart upload / ranged read).
+const maxRDMABufferSize int64 = 4 << 30 // 4 GiB
+
 type rdmaClientHandle struct {
 	cptr *C.miniocpp_client
 }
@@ -73,6 +80,12 @@ func (c *Client) putObjectRDMA(_ context.Context, bucketName, objectName string,
 	objectC := C.CString(objectName)
 	defer C.free(unsafe.Pointer(objectC))
 
+	if int64(opts.RDMABufferSize) > maxRDMABufferSize {
+		return UploadInfo{}, fmt.Errorf(
+			"RDMA put: buffer size %d exceeds the cuObject registration limit of %d bytes (4 GiB); split into parts <= 4 GiB",
+			opts.RDMABufferSize, maxRDMABufferSize)
+	}
+
 	var etagBuf, checksumBuf [64]C.char
 	n := C.miniocpp_put_object(h.cptr, bucketC, objectC,
 		opts.RDMABuffer, C.size_t(opts.RDMABufferSize),
@@ -102,6 +115,12 @@ func (c *Client) getObjectRDMA(_ context.Context, bucketName, objectName string,
 	defer C.free(unsafe.Pointer(bucketC))
 	objectC := C.CString(objectName)
 	defer C.free(unsafe.Pointer(objectC))
+
+	if int64(opts.RDMABufferSize) > maxRDMABufferSize {
+		return 0, fmt.Errorf(
+			"RDMA get: buffer size %d exceeds the cuObject registration limit of %d bytes (4 GiB); split into parts <= 4 GiB",
+			opts.RDMABufferSize, maxRDMABufferSize)
+	}
 
 	n := C.miniocpp_get_object(h.cptr, bucketC, objectC,
 		opts.RDMABuffer, C.size_t(opts.RDMABufferSize), nil, nil)
--- a/rdma.go
+++ b/rdma.go
@@ -27,10 +27,12 @@ import (
 var ErrRDMANotConnected = errors.New("RDMA infrastructure not connected")
 
 // maxRDMABufferSize is the largest buffer a single cuObject registration
-// (cuMemObjGetDescriptor, inside libminiocpp) can pin — 4 GiB. A larger buffer
-// cannot be RDMA-registered, so reject it here with a clear error instead of
-// letting the cgo call fail opaquely; the caller must split the transfer into
-// parts no larger than this (multipart upload / ranged read).
+// (cuMemObjGetDescriptor, inside libminiocpp) can pin — 4 GiB. libminiocpp
+// itself does not fail on a larger buffer: it skips RDMA and transfers it over
+// one ordinary HTTP request. The guards below reject it instead, so an oversize
+// buffer does not silently become a much slower HTTP transfer on a path the
+// caller explicitly opted into for RDMA. The lower bound also stops a negative
+// size from wrapping to a huge C.size_t.
 const maxRDMABufferSize int64 = 4 << 30 // 4 GiB
 
 type rdmaClientHandle struct {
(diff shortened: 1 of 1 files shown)


allanrogerr:
RDMA failure fixed in https://github.com/minio/minio-go/pull/2276

rdma.go
```diff
@@ -26,6 +26,13 @@ import (
 
 var ErrRDMANotConnected = errors.New("RDMA infrastructure not connected")
 
+// maxRDMABufferSize is the largest buffer a single cuObject registration
+// (cuMemObjGetDescriptor, inside libminiocpp) can pin — 4 GiB. A larger buffer
+// cannot be RDMA-registered, so reject it here with a clear error instead of
+// letting the cgo call fail opaquely; the caller must split the transfer into
+// parts no larger than this (multipart upload / ranged read).
+const maxRDMABufferSize int64 = 4 << 30 // 4 GiB
```
allanrogerr, line 36: libminiocpp doesn't fail on a buffer over 4 GiB — it skips RDMA and transfers it over one ordinary HTTP request (`use_rdma = size <= kCuObjMaxMemoryRegSize && cuMemObjGetDescriptor(...) == 0` in `client.cc`), so the "fail opaquely" premise here isn't accurate, and the "split into parts" advice runs against the upstream header's own note that the SDK does not chunk registrations. Rejecting rather than silently downgrading may well be what you want — someone who bothered to allocate an aligned buffer may prefer a clear error to a quietly much slower path — so this only makes the comment describe what actually happens. Whether the guard itself should stay is your call.

```suggestion
// maxRDMABufferSize is the largest buffer a single cuObject registration
// (cuMemObjGetDescriptor, inside libminiocpp) can pin — 4 GiB. libminiocpp
// itself does not fail on a larger buffer: it skips RDMA and transfers it over
// one ordinary HTTP request. The guards below reject it instead, so an oversize
// buffer does not silently become an HTTP transfer.
const maxRDMABufferSize int64 = 4 << 30 // 4 GiB
```

rdma.go
```diff
@@ -73,6 +80,12 @@ func (c *Client) putObjectRDMA(_ context.Context, bucketName, objectName string,
 	objectC := C.CString(objectName)
 	defer C.free(unsafe.Pointer(objectC))
 
+	if int64(opts.RDMABufferSize) > maxRDMABufferSize {
+		return UploadInfo{}, fmt.Errorf(
+			"RDMA put: buffer size %d exceeds the cuObject registration limit of %d bytes (4 GiB); split into parts <= 4 GiB",
+			opts.RDMABufferSize, maxRDMABufferSize)
+	}
```
allanrogerr: `RDMABufferSize` is a signed `int`, so a negative value slips straight past this check and `C.size_t(opts.RDMABufferSize)` on the next line turns it into 18446744073709551615 — the same opaque C-layer failure the guard is here to prevent. Folding the lower bound in also lets the message state the limit once instead of three times, and drops the "split into parts" advice that the single-shot C API can't actually deliver.

```suggestion
	if opts.RDMABufferSize < 0 || int64(opts.RDMABufferSize) > maxRDMABufferSize {
		return UploadInfo{}, fmt.Errorf(
			"RDMA put: buffer size %d must be between 0 and the cuObject registration limit of %d bytes",
			opts.RDMABufferSize, maxRDMABufferSize)
	}
```

rdma.go
```diff
@@ -103,6 +116,12 @@ func (c *Client) getObjectRDMA(_ context.Context, bucketName, objectName string,
 	objectC := C.CString(objectName)
 	defer C.free(unsafe.Pointer(objectC))
 
+	if int64(opts.RDMABufferSize) > maxRDMABufferSize {
+		return 0, fmt.Errorf(
+			"RDMA get: buffer size %d exceeds the cuObject registration limit of %d bytes (4 GiB); split into parts <= 4 GiB",
+			opts.RDMABufferSize, maxRDMABufferSize)
+	}
```
allanrogerr: Same signed-`int` gap as the put path, but this one is worse: `GetObjectArgs::Validate()` upstream imposes no size bound, so a negative size wraps to 18446744073709551615 and reaches `pubsetbuf(sink, size)` in `client.cc`'s HTTP fallback, which then streams the response body past the end of the caller's real allocation. `miniocpp_get_object` returns `(ssize_t)size` = -1 afterwards, so the `n < 0` check below reports a stale `lastRDMAError()` rather than the real problem.

```suggestion
	if opts.RDMABufferSize < 0 || int64(opts.RDMABufferSize) > maxRDMABufferSize {
		return 0, fmt.Errorf(
			"RDMA get: buffer size %d must be between 0 and the cuObject registration limit of %d bytes",
			opts.RDMABufferSize, maxRDMABufferSize)
	}
```

harshavardhana:
## Review feedback — Applied (5fb0569)

Addressed all four threads:

- **@allanrogerr (put/get, signed-int gap):** `RDMABufferSize` is a signed `int`; a negative value slipped past the upper bound and `C.size_t(negative)` wrapped to a huge length — for GET reaching `pubsetbuf(sink, size)` in libminiocpp's HTTP fallback and streaming past the caller's buffer. Both guards now require `0 <= size <= 4 GiB`.
- **@allanrogerr (comment):** corrected — libminiocpp does not fail on >4 GiB; it downgrades to a single HTTP request. The guard rejects that so an RDMA-opt-in path doesn't silently become a slow HTTP transfer. Dropped the inapplicable "split into parts" advice (the single-shot C API can't chunk). Kept the reject (your call — clear error over silent downgrade).
- **@coderabbitai (fail-fast):** moved the guard to the top of both functions, before `c.rdma()` and the `CString` allocations.

gofmt clean; the cgo build is exercised by the `go-rdma` CI.

allanrogerr (approved):

rdma.go
allanrogerr: Implemented in `5fb0569` (`rdma.go:29-36`). The "fail opaquely" premise and the "split into parts (multipart upload / ranged read)" remedy are both gone — the comment now says libminiocpp "does not fail on a larger buffer: it skips RDMA and transfers it over one ordinary HTTP request."

Your two additions both hold up: the path does require an explicit opt-in (`EnableRDMA` on the client *and* a non-nil `RDMABuffer`), and the closing note about the lower bound stopping a negative size from wrapping to a huge `C.size_t` is accurate. Keeping the rejection is your call, and it now reads as the deliberate choice it is rather than as a workaround for a failure that doesn't happen.

rdma.go
allanrogerr: Implemented in `5fb0569` (`rdma.go:75-79`), and hoisted to the top of `putObjectRDMA` so it now runs before `c.rdma()` at `:81`:

```go
	if opts.RDMABufferSize < 0 || int64(opts.RDMABufferSize) > maxRDMABufferSize {
```

A negative size can no longer reach `C.size_t(opts.RDMABufferSize)`. Verified against the `5fb0569` blob; `gofmt`/`gofumpt` clean and all 8 checks green on this head.

rdma.go
allanrogerr: Implemented in `5fb0569` (`rdma.go:111-115`), also hoisted above `c.rdma()` at `:117`. This is the one that mattered most: a negative size can no longer wrap to 18446744073709551615 and reach `pubsetbuf(sink, size)` in libminiocpp's HTTP fallback, which had no upstream size bound to stop it streaming a response body past the end of the caller's allocation. Verified against the `5fb0569` blob.

PR #2281, opened 2026-08-06
Add AIStor Lifecycle Compression

klauspost:
Allow setting compression as a lifecycle for current and non-current versions. All lifecycle filters apply.

Will link AIStor/mc

## Summary by CodeRabbit

- **New Features**
  - Added lifecycle rule support for compressing current and noncurrent object versions.
  - Compression settings can specify immediate or delayed actions based on configured days.
  - Compression options are supported consistently in JSON and XML configurations.
  - Optional compression settings distinguish between an omitted action and an action configured with zero days.
  - Compression settings are preserved during configuration conversion and round trips.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -47,7 +47,7 @@ jobs:
           sudo apt install devscripts -y
           mkdir -p /tmp/certs-dir
           cp testcerts/* /tmp/certs-dir
-          # Test against the AIStor :edge image so the latest AIStor APIs
+          # Test against the AIStor :edge-daily image so the latest AIStor APIs
           # (object annotations) and checksum algorithms (xxhash3/xxhash128,
           # sha512, md5) are available to the functional tests.
           docker pull quay.io/minio/aistor/minio:edge-daily
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -50,7 +50,7 @@ jobs:
           # Test against the AIStor :edge image so the latest AIStor APIs
           # (object annotations) and checksum algorithms (xxhash3/xxhash128,
           # sha512, md5) are available to the functional tests.
-          docker pull registry.k5.min.dev/aistor/minio:edge
+          docker pull quay.io/minio/aistor/minio:edge-daily
           # Free-tier AIStor license so the server allows writes (without a
           # license the server boots read-only). This is a public FREE-plan
           # license, not a secret.
@@ -64,7 +64,7 @@ jobs:
             -e MINIO_CI_CD=true \
             -e MINIO_KMS_SECRET_KEY="${MINIO_KMS_SECRET_KEY}" \
             -e MINIO_LICENSE="${MINIO_LICENSE}" \
-            registry.k5.min.dev/aistor/minio:edge \
+            quay.io/minio/aistor/minio:edge-daily \
             server --quiet --certs-dir /certs /data/{1...4}
           for _ in $(seq 1 60); do
             if curl -sf --cacert /tmp/certs-dir/public.crt https://localhost:9000/minio/health/ready >/dev/null; then
--- a/pkg/lifecycle/lifecycle_test.go
+++ b/pkg/lifecycle/lifecycle_test.go
@@ -297,6 +297,24 @@ func TestLifecycleJSONRoundtrip(t *testing.T) {
 				ID:     "rule-10",
 				Status: "Enabled",
 			},
+			{
+				Compression: &Compression{
+					Days:           ExpirationDays(30),
+					SkipCompressed: true,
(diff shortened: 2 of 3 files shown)


.github/workflows/go.yml
```diff
@@ -50,7 +50,7 @@ jobs:
           # Test against the AIStor :edge image so the latest AIStor APIs
           # (object annotations) and checksum algorithms (xxhash3/xxhash128,
           # sha512, md5) are available to the functional tests.
-          docker pull registry.k5.min.dev/aistor/minio:edge
+          docker pull quay.io/minio/aistor/minio:edge-daily
```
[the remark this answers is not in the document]
klauspost: I think this is what we want.

.github/workflows/go.yml
```diff
@@ -50,7 +50,7 @@ jobs:
           # Test against the AIStor :edge image so the latest AIStor APIs
```
allanrogerr: The pull and run commands just below both use `:edge-daily`, so this line names an image the workflow no longer runs. Anyone reading it to find out which build CI tests against gets the wrong answer.

```suggestion
          # Test against the AIStor :edge-daily image so the latest AIStor APIs
```

pkg/lifecycle/lifecycle_test.go
```diff
@@ -331,6 +349,41 @@ func TestLifecycleJSONRoundtrip(t *testing.T) {
...
+	}
+
+	got, err := json.Marshal(lc)
+	if err != nil {
+		t.Fatalf("Failed to marshal due to %v", err)
+	}
+	if !bytes.Equal(expected, got) {
+		t.Fatalf("Expected %s but got %s", expected, got)
 	}
 }
```
allanrogerr, line 388: `MarshalJSON` copies `Rule` into its mirror struct one field at a time, so a field added to both structs but never assigned just vanishes from the JSON under `omitempty`. I tried it: adding a field to `Rule` and to the mirror without the assignment drops it from the output and the whole package still passes — `TestRuleWrapperCopyCompleteness` only covers the XML path, which is safe because `ruleWrapper` embeds the alias. This fills every `Rule` field and checks each one reaches the JSON. It passes on the branch as-is and fails on that mutation. Consider and test this.

```suggestion
	}
}

// TestRuleMarshalJSONCopyCompleteness fills every Rule field and requires each
// one to reach the JSON output. MarshalJSON copies Rule into a mirror struct
// field by field, so a field added to Rule and to the mirror but never assigned
// is dropped by omitempty with every other test still passing. RuleFilter is
// skipped for the same reason as in TestRuleWrapperCopyCompleteness.
func TestRuleMarshalJSONCopyCompleteness(t *testing.T) {
	skip := func(name string) bool { return name == "XMLName" || name == "RuleFilter" }

	var r Rule
	rv := reflect.ValueOf(&r).Elem()
	for i := range rv.NumField() {
		if skip(rv.Type().Field(i).Name) {
			continue
		}
		fillNonZero(t, rv.Field(i))
	}

	got, err := json.Marshal(r)
	if err != nil {
		t.Fatalf("could not marshal a filled Rule: %v", err)
	}
	for i := range rv.NumField() {
		f := rv.Type().Field(i)
		if skip(f.Name) {
			continue
		}
		name, _, _ := strings.Cut(f.Tag.Get("json"), ",")
		if !bytes.Contains(got, []byte(`"`+name+`":`)) {
			t.Fatalf("Rule.%s (JSON %q) is missing from MarshalJSON output: %s", f.Name, name, got)
		}
	}
}
```

allanrogerr:
Tested this branch against a real server: `quay.io/minio/aistor/minio:edge-daily`, booted exactly as this PR's `go.yml` does. The compression elements aren't implemented server-side yet.

- Any rule whose only action is `Compression` or `NoncurrentVersionCompression` → `400 InvalidArgument`, the same error a rule with no action at all returns.
- `Compression` alongside an `Expiration` → accepted, and the element is absent from the read-back.
- `Expiration` on its own (control) → accepted, round-trips.

That's just the unknown-element path: aistor `edge` (`8b438b16`, the exact commit this image reports) parses lifecycle with its own `internal/bucket/lifecycle`, whose `Rule` carries Expiration, Transition, DelMarkerExpiration, AllVersionsExpiration and the two noncurrent variants — no compression fields. Rebuilding aistor against this minio-go wouldn't change that; the SDK types aren't on the server's parse path.

Nothing to fix here, and I assume this is just ordering given "Will link AIStor/mc". Worth pinning the element names and shape against the server PR before merge, since exported types are a public wire contract. Happy to re-run once the server side lands.

rraulinio (approved):

poornas (approved):

PR #2284, opened 2026-08-15
feat(rdma): stream a PUT over RDMA instead of pinning the whole object

harshavardhana:
`PutObject` took the RDMA path only when the caller had already staged the **entire object** into pinned memory, and ignored any reader. Two consequences:

- an RDMA upload was capped at the **4 GiB** an RDMA descriptor can address
- the caller had to pin the object's size — 16 GiB of registered memory for a 16 GiB object

A reader alongside `RDMABuffer` now streams instead.

## How

libminiocpp already had all of this and nobody called it: one part buffer registered once for the whole upload, then read-part → RDMA `UploadPart` → repeat, with the per-part CRC64NVME the server requires on that path. Pinned memory becomes **one part rather than one object**, and object size stops being bounded by the descriptor.

Passing `buf` to the C API would ask for a single-shot transfer, so the streaming call passes `NULL` and lets libminiocpp own the part buffer. The caller's `RDMABuffer` is what *selects* the path, not the staging area.

Two details worth review:

- **The read callback loops** rather than trusting one `Read`. An `io.Reader` may return short without being at EOF, and libminiocpp reads a short return as end-of-part — so a partial fill mid-stream would upload a short part and truncate the object with no error anywhere.
- **The handle reaches C through `malloc`'d memory**, not as a Go pointer. A `cgo.Handle` is a `uintptr`, and converting one straight to `unsafe.Pointer` is the misuse `go vet` rejects.

Read errors are kept on the Go side and reported from there, because the C callback can only say "failed".

## Verified

Two-rail mlx5 host against a live 4-node cluster, driven through warp:

- **16 GiB objects upload** — past the descriptor limit that previously blocked them
- a server-side `mc admin trace` shows **320 of 320 `UploadPart` requests carrying an `x-amz-rdma-token`**, so nothing silently falls back to HTTP
- builds and vets clean both with `-tags=rdma` and without (hence the `rdma_stub.go` addition)

## Known limitation

Per-caller throughput is **below the HTTP streaming path** — 319 vs 1113 MiB/s on the same generator — though it scales linearly with concurrency (5341 MiB/s at 32 callers). Part of that is minio-cpp uploading stream parts one at a time; [minio-cpp#253](https://github.com/minio/minio-cpp/pull/253) fixes that half and is what the 319 figure already includes. The remainder is not yet explained, so this is worth taking for the capability — objects past 4 GiB, without pinning them — rather than as a throughput win.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added support for streaming RDMA uploads from readers while preserving existing buffer-based uploads.
  * Uploads now report metadata including ETag and CRC64NVME checksum.
  * Added cancellation handling, read-error reporting, and protection against stalled data sources.

* **Bug Fixes**
  * Improved handling of short reads and incomplete RDMA upload data.

* **Compatibility**
  * Environments without RDMA support continue to receive the appropriate unsupported-feature response.


harshavardhana:
## Fixes Applied Successfully

Fixed 5 file(s) based on 3 CodeRabbit feedback item(s) — all three applied.

**Files modified:**
- `rdma_callback.go`
- `rdma_stream.go`
- `api-put-object.go`
- `rdma_options.go` (new)
- `rdma_options_test.go` (new)

**Commit:** `2fe74f2`

### Do not convert a zero-byte read into end of stream

The sharpest of the three, and self-inflicted: the fill loop exists precisely so a short `Read` does not become a short part — then returned one anyway the moment a reader handed back `(0, nil)`, which `io.Reader` permits. libminiocpp reads a short return as end-of-object, so the upload **completed successfully with the rest of the data missing**. Consecutive empty reads are now tolerated up to `bufio`'s limit of 100 and then fail with `io.ErrNoProgress` — never truncating, never spinning.

### Propagate cancellation

The context was discarded outright. libminiocpp's put is synchronous with no cancellation hook, but the read callback runs once per part, so checking `ctx.Err()` there stops the upload at the next part boundary — bounded by one part rather than the whole object.

Verified live: a 6 GiB upload aborted mid-flight leaves **no partial object** (18 completed objects, all exactly 6.00 GiB).

### Refuse options the RDMA path cannot carry

The gate returns before `opts.validate`, and the C entry point takes a bucket, an object, and a buffer or reader — nothing else. Metadata, tags, encryption, retention, legal hold and the rest were dropped on the floor, storing an object quietly different from the one requested. They are now rejected up front, naming every offender so a caller fixes them in one pass.

Two deliberate exclusions:

- **Transfer shaping stays permitted** — part size and concurrency change how bytes move, not what is stored, and libminiocpp picks its own.
- **`ContentType` stays permitted and is still not applied.** That is a real remaining gap, but rejecting it would refuse the common case of a caller that sets it for the HTTP path and falls through to RDMA — warp does exactly this on every PUT.

Worth noting the same gap exists on the pre-existing single-shot path; the new check sits at the shared gate, so both are covered.

### Verified

`go vet` clean, `golangci-lint` 0 issues, `go test -short -race` passes, and 11 new subtests cover the rejection logic including that empty-but-allocated maps are not treated as set. Built and vetted with `-tags=rdma` on a two-rail mlx5 host, then exercised through warp against a live 4-node cluster: 4 MiB single-shot at 6892 MiB/s, 6 GiB streaming at 636 MiB/s.

The latest autofix changes are on the `feat/rdma-stream` branch.

PR #2286, opened 2026-08-17
feat: add ListUnsorted() for the ?unsorted=true listing extension

harshavardhana:
AIStor serves `ListObjectsV2` with `?unsorted=true` by concatenating the per-erasure-set streams instead of merging them into lexical order. Clients that build their own set client-side (a `readdir()` that de-duplicates into a directory entry set, hydration passes, existence probes) pay for ordering they never use — skipping the merge is a measurable latency win on multi-pool/multi-set deployments.

This is the Go counterpart of what minio-rs already sends (`.unsorted(true)`), against the server extension documented in `docs/extensions/unsorted-listing/README.md`.

## API

- `ListObjectsOptions.Unsorted` sets `?unsorted=true`, composing with `metadata=true`.
- `ListUnsorted(ctx, bucket, opts) iter.Seq[ObjectInfo]` is the entrypoint.

```go
for object := range c.ListUnsorted(ctx, "mybucket", minio.ListObjectsOptions{Prefix: "a/", Recursive: true}) {
    if object.Err != nil {
        // handle the error
    }
    fmt.Println(object.Key)
}
```

## Semantics

Pagination stays positional — the continuation token encodes a position rather than a key — so the existing `NextContinuationToken` loop is unchanged. `StartAfter` has no key ordering to resolve against and the server answers `501 NotSupported`, so it is rejected client-side, along with the V1 and versioned listing paths that do not implement the extension.

Callers must de-duplicate: common prefixes can repeat when a delimiter is used, and the same key can be returned twice while a pool is being decommissioned or rebalanced. Both are documented on `ListUnsorted`.

Servers without the extension ignore the parameter and keep returning a sorted listing.

## Testing

`functional_tests.go` gains `testListUnsorted()` covering plain, paginated (`MaxKeys` below the object count, to exercise the positional token), and `WithMetadata` listings, plus the three rejected option sets.

## Summary by CodeRabbit

* **New Features**
  * Added support for unsorted object listings through a new listing option and convenience method.
  * Unsorted listings support prefixes, pagination, and object metadata.
* **Bug Fixes**
  * Added validation to reject unsupported combinations, including V1, versioned, or `StartAfter` listings.
* **Tests**
  * Added coverage for unsorted listing behavior, pagination, metadata, and validation errors.

### DIFF
--- a/api-list.go
+++ b/api-list.go
@@ -761,6 +761,23 @@ type ListObjectsOptions struct {
 	headers http.Header
 }
 
+// checkUnsorted returns an error if the unsorted listing extension cannot be
+// honored for these options against a bucket in the given location. It is only
+// implemented by ListObjects V2, and its continuation token encodes a position
+// rather than a key, so there is nothing for 'StartAfter' to resolve against.
+func (o ListObjectsOptions) checkUnsorted(location string) error {
+	if !o.Unsorted {
+		return nil
+	}
+	if o.UseV1 || o.WithVersions || location == "snowball" {
+		return errInvalidArgument("unsorted listing is only supported by ListObjects V2")
+	}
+	if o.StartAfter != "" {
+		return errInvalidArgument("unsorted listing does not support StartAfter")
+	}
+	return nil
+}
+
 // Set adds a key value pair to the options. The
 // key-value pair will be part of the HTTP GET request
 // headers.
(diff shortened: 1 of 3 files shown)


harshavardhana:
## Fixes Applied Successfully

Fixed 2 file(s) based on 2 CodeRabbit feedback item(s).

**Files modified:**
- `api-list.go`
- `functional_tests.go`

**Commit:** `e2432e4`

The latest autofix changes are on the `feat/unsorted-listing` branch.

PR #2299, opened 2026-09-02
Allow setting custom upload limits

klauspost:
Adds to client configuration:

```go
// UploadLimits overrides the upload limits the client enforces before sending
// a request. The defaults are the limits Amazon S3 imposes; only raise them
// when the remote endpoint is known to accept the larger values.
//
// A zero field means "use the default", so the zero UploadLimits behaves
// exactly like Amazon S3 — except for MaxSinglePutObjectSize, whose default is
// deliberately not enforced by PutObject. See that field.
//
// New rejects limits it cannot derive a part layout from. No field may be
// negative; the remaining bounds are noted on each field.
type UploadLimits struct {
	// MinPartSize is the smallest size allowed for a part that is not the
	// last part of a multipart upload. Defaults to 5 MiB. May not exceed
	// MaxPartSize.
	MinPartSize int64

	// MaxPartSize is the largest size allowed for a single part.
	// Defaults to 5 GiB. MaxPartSize * MaxPartsCount must fit in an int64.
	MaxPartSize int64

	// MaxPartsCount is the maximum number of parts in a single multipart
	// upload. Together with MaxPartSize this caps the object size the client
	// is willing to upload. Defaults to 10000, and may not exceed 2^53
	// because the part layout is computed in float64.
	MaxPartsCount int64

	// MaxSinglePutObjectSize is the largest object the remote accepts in a
	// single PUT. Defaults to 5 GiB.
	//
	// Unlike the other fields, the 5 GiB default is not enforced by PutObject:
	// MinIO and AIStor accept single PUTs far above Amazon's limit, so gating on
	// the default would refuse uploads that work today. Setting it explicitly
	// does enforce it — PutObject then rejects a larger object outright when
	// PutObjectOptions.DisableMultipart is set, and otherwise sends it as a
	// multipart upload.
	//
	// The resolved value, default included, always bounds the single PUT that
	// PutObject falls back to when a multipart upload fails with AccessDenied.
	// Core.PutObject never checks it.
	MaxSinglePutObjectSize int64
}
```

## Summary by CodeRabbit

* **New Features**
  * Added configurable upload limits for part sizes, part counts, single-request uploads, and maximum object size.
  * Custom limits can be set when creating a client, with S3-compatible defaults applied automatically.
  * Multipart uploads and object composition now honor configured limits.

* **Bug Fixes**
  * Improved upload-size validation and part-size calculations.
  * Prevented incomplete unknown-length uploads from being stored as truncated objects.
  * Oversized uploads now use multipart uploads when supported.

* **Tests**
  * Added coverage for customized, invalid, and boundary upload-limit configurations.

### DIFF
--- a/api-append-object.go
+++ b/api-append-object.go
@@ -88,7 +88,7 @@ func (opts *AppendObjectOptions) setChecksumParams(info ObjectInfo) {
 }
 
 func (opts AppendObjectOptions) validate(c *Client) (err error) {
-	if int64(opts.ChunkSize) > c.limits.maxPartSize() {
+	if opts.ChunkSize > uint64(c.limits.maxPartSize()) {
 		return errInvalidArgument("Append chunkSize cannot be larger than max part size allowed")
 	}
 	switch {
--- a/api-append-object.go
+++ b/api-append-object.go
@@ -88,7 +88,7 @@ func (opts *AppendObjectOptions) setChecksumParams(info ObjectInfo) {
 }
 
 func (opts AppendObjectOptions) validate(c *Client) (err error) {
-	if opts.ChunkSize > maxPartSize {
+	if int64(opts.ChunkSize) > c.limits.maxPartSize() {
 		return errInvalidArgument("Append chunkSize cannot be larger than max part size allowed")
 	}
 	switch {
@@ -212,7 +212,7 @@ func (c *Client) AppendObject(ctx context.Context, bucketName, objectName string
 		if objectSize > 0 {
 			finalObjSize = info.Size + objectSize
 		}
-		totalPartsCount, partSize, lastPartSize, err := OptimalPartInfo(finalObjSize, opts.ChunkSize)
+		totalPartsCount, partSize, lastPartSize, err := c.optimalPartInfo(finalObjSize, opts.ChunkSize)
 		if err != nil {
 			return UploadInfo{}, err
 		}
--- a/api-put-object-common.go
+++ b/api-put-object-common.go
@@ -180,6 +180,13 @@ func (l UploadLimits) optimalPartInfo(objectSize int64, configuredPartSize uint6
 // errIfMoreData reports errUploadTooLarge when reader still holds data after
 // the last part allowed by the upload limits was consumed. Unknown length
 // uploads would otherwise silently complete a truncated object.
+//
+// This deliberately fails an upload whose bytes were all transferred: the probe
+// runs after the final part, so a reader that reports anything other than a
+// clean io.EOF — a closed file, a reset connection, a wrapper with its own
+// sentinel — aborts the multipart upload instead of completing it. Silently
+// storing a possibly truncated object is the worse outcome, but callers should
+// expect this error to arrive late and to look like a transport fault.
 func errIfMoreData(reader io.Reader, uploadedSize, totalPartsCount int64, bucketName, objectName string) error {
 	var b [1]byte
 	n, err := readFull(reader, b[:])
--- a/api-put-object-common.go
+++ b/api-put-object-common.go
@@ -105,6 +105,13 @@ func (l UploadLimits) optimalPartInfo(objectSize int64, configuredPartSize uint6
 
 	var partSizeFlt float64
 	if configuredPartSize > 0 {
+		// Compared unsigned and up front, so the int64 conversions below cannot
+		// wrap a caller-supplied part size into a negative that slips past them.
+		if configuredPartSize > uint64(l.maxPartSize()) {
+			err = errInvalidArgument(fmt.Sprintf("Input part size is bigger than allowed maximum of %s.", humanize.IBytes(uint64(l.maxPartSize()))))
+			return totalPartsCount, partSize, lastPartSize, err
+		}
+
 		if int64(configuredPartSize) > objectSize {
 			err = errEntityTooLarge(int64(configuredPartSize), objectSize, "", "")
 			return totalPartsCount, partSize, lastPartSize, err
@@ -119,11 +126,6 @@ func (l UploadLimits) optimalPartInfo(objectSize int64, configuredPartSize uint6
 
 		if int64(configuredPartSize) < l.minPartSize() {
 			err = errInvalidArgument(fmt.Sprintf("Input part size is smaller than allowed minimum of %s.", humanize.IBytes(uint64(l.minPartSize()))))
-			return totalPartsCount, partSize, lastPartSize, err
-		}
-
-		if int64(configuredPartSize) > l.maxPartSize() {
-			err = errInvalidArgument(fmt.Sprintf("Input part size is bigger than allowed maximum of %s.", humanize.IBytes(uint64(l.maxPartSize()))))
 			return totalPartsCount, partSize, lastPartSize, err
 		}
 
@@ -142,6 +144,11 @@ func (l UploadLimits) optimalPartInfo(objectSize int64, configuredPartSize uint6
 		// overflows during float64 to int64 conversions.
 		partSizeFlt = float64(objectSize / maxPartsCount)
 		partSizeFlt = math.Ceil(partSizeFlt/float64(configuredPartSize)) * float64(configuredPartSize)
+		// An object smaller than maxPartsCount rounds down to a zero part size,
+		// and a non-final part must never fall below MinPartSize either.
+		if minPS := float64(l.minPartSize()); partSizeFlt < minPS {
+			partSizeFlt = minPS
+		}
 		// Rounding up to a minPartSize multiple can overshoot a MaxPartSize
 		// that was lowered below, or is not a multiple of, minPartSize.
 		if maxPS := float64(l.maxPartSize()); partSizeFlt > maxPS {
@@ -163,8 +170,14 @@ func (l UploadLimits) optimalPartInfo(objectSize int64, configuredPartSize uint6
 // uploads would otherwise silently complete a truncated object.
 func errIfMoreData(reader io.Reader, uploadedSize int64, bucketName, objectName string) error {
 	var b [1]byte
-	if n, _ := readFull(reader, b[:]); n > 0 {
+	n, err := readFull(reader, b[:])
+	if n > 0 {
 		return errEntityTooLarge(uploadedSize+int64(n), uploadedSize, bucketName, objectName)
+	}
+	// Only a clean EOF proves the reader was drained; anything else has to
+	// surface rather than complete a possibly truncated object.
+	if err != nil && err != io.EOF {
+		return err
 	}
 	return nil
 }
--- a/api-put-object-common.go
+++ b/api-put-object-common.go
@@ -103,6 +103,12 @@ func (l UploadLimits) optimalPartInfo(objectSize int64, configuredPartSize uint6
 		return totalPartsCount, partSize, lastPartSize, err
 	}
 
+	// An empty object has no parts; the minimum part size below would otherwise
+	// report a part size and a last part size for a layout with no parts in it.
+	if objectSize == 0 && configuredPartSize == 0 {
+		return 0, 0, 0, nil
+	}
+
 	var partSizeFlt float64
 	if configuredPartSize > 0 {
 		// Compared unsigned and up front, so the int64 conversions below cannot
--- a/api-put-object-common.go
+++ b/api-put-object-common.go
@@ -87,14 +87,14 @@ func (l UploadLimits) optimalPartInfo(objectSize int64, configuredPartSize uint6
 	maxPartsCount := l.maxPartsCount()
 	maxObjectSize := l.maxObjectSize()
 
-	// When object size is unknown (-1), default to 5TiB to limit memory usage.
-	// This results in ~537MiB part sizes. For larger objects (up to the
-	// maximum object size), callers should set configuredPartSize explicitly
-	// to control memory usage.
+	// When object size is unknown (-1), default to 5TiB (or the maximum object
+	// size, when lower) to limit memory usage. This results in ~537MiB part
+	// sizes. For larger objects (up to the maximum object size), callers should
+	// set configuredPartSize explicitly to control memory usage.
 	var unknownSize bool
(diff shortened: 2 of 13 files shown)


api-append-object.go
```diff
@@ -212,7 +212,7 @@ func (c *Client) AppendObject(ctx context.Context, bucketName, objectName string
 		if objectSize > 0 {
 			finalObjSize = info.Size + objectSize
 		}
-		totalPartsCount, partSize, lastPartSize, err := OptimalPartInfo(finalObjSize, opts.ChunkSize)
+		totalPartsCount, partSize, lastPartSize, err := c.optimalPartInfo(finalObjSize, opts.ChunkSize)
```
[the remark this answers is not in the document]
klauspost: the suggested change to oinfo.Size + objectSize would break AppendObject. info is the named
  return and is zero there, so finalObjSize == objectSize, which is exactly what the loop below needs: it reads
  totalPartsCount * partSize bytes from the reader. Including the existing object's size would make it demand bytes the
  reader doesn't have and fail with io.ErrUnexpectedEOF. The info.Size + term is misleading dead weight, but it is not
  currently a bug — I left it rather than change append semantics under a review comment.

api-put-object.go
```diff
@@ -374,6 +377,17 @@ func (c *Client) PutObject(ctx context.Context, bucketName, objectName string, r
 		partSize = minPartSize
 	}
 
+	// A single PUT can only carry up to the endpoint's single PUT limit, so
+	// anything above it has to go out as multipart.
+	if maxSinglePut := c.limits.maxSinglePutObjectSize(); size > maxSinglePut {
```
poornas: This makes `PutObject` refuse uploads that work today against MinIO/AIStor, on default settings. Before this PR there was no single-PUT ceiling on this path — `DisableMultipart` fell straight through to `c.putObject`, and so did any `size <= opts.PartSize`. AIStor accepts a single PUT up to `globalMaxObjectSize = 5 * humanize.TiByte` (`cmd/utils.go:319`, whose comment calls the divergence from S3 deliberate). A 6GiB `PutObject` with `DisableMultipart: true` succeeds on master and returns `EntityTooLarge` here with zero requests sent; the same size with `PartSize: 6GiB` previously went out as one PUT and is now clamped into multipart at 386, where `optimalPartInfo` rejects it as `InvalidArgument: Input part size is bigger than allowed maximum of 5.0 GiB`. Deleting just this block restores both.

The part that makes it a blocker rather than a knob is `PutObjectsSnowball`: it sets `opts.Opts.DisableMultipart = true` itself at `api-putobject-snowball.go:237` and calls `PutObject` at 244, and its doc comment at 107 says "Total size should be < 5TB." A snowball tar over 5GiB now fails, and no caller of that method can opt out — the opt-out lives on the client, not the call.

```go
maxSinglePut := c.limits.maxSinglePutObjectSize()
if c.limits.MaxSinglePutObjectSize == 0 && !s3utils.IsAmazonEndpoint(*c.endpointURL) {
    maxSinglePut = maxMultipartPutObjectSize
}
```

This file already switches on endpoint identity for GCS at 371.

api-put-object.go
```diff
@@ -374,6 +377,17 @@ func (c *Client) PutObject(ctx context.Context, bucketName, objectName string, r
...
+		if opts.DisableMultipart {
+			return UploadInfo{}, errEntityTooLarge(size, maxSinglePut, bucketName, objectName)
+		}
+		if int64(partSize) > maxSinglePut {
```
poornas, line 388: Under default limits this clamp can only ever convert one error into a different error. It fires when `int64(partSize) > maxSinglePut`, and `partSize` is either `opts.PartSize` or 16MiB — so it requires `opts.PartSize > 5GiB`, which is exactly what `optimalPartInfo` rejects at `api-put-object-common.go:110`, since `defaultMaxPartSize` is 5GiB too. Both other exits behave the same: the V2 branch reaches `optimalPartInfo(-1, opts.PartSize)` and fails identically, and GCS returns at 371 before getting here. So the "select multipart upload when multipart is enabled" half of the suggestion this block answers isn't actually delivered; the clamp only becomes live under an asymmetric config such as `UploadLimits{MaxPartSize: 8GiB}` with the single-PUT limit left at 5GiB. Clamping what reaches `optimalPartInfo` rather than only the local routing variable would make it do the intended thing.

api-put-object-common.go
```diff
@@ -131,6 +171,23 @@ func OptimalPartInfo(objectSize int64, configuredPartSize uint64) (totalPartsCou
...
+// uploads would otherwise silently complete a truncated object.
+func errIfMoreData(reader io.Reader, uploadedSize, totalPartsCount int64, bucketName, objectName string) error {
+	var b [1]byte
+	n, err := readFull(reader, b[:])
+	if n > 0 {
+		return errUploadTooLarge(uploadedSize, totalPartsCount, bucketName, objectName)
+	}
+	// Only a clean EOF proves the reader was drained; anything else has to
+	// surface rather than complete a possibly truncated object.
+	if err != nil && err != io.EOF {
```
poornas, line 198: This turns a class of previously-successful uploads into failures. Before the PR the reader was never touched after the part loop; now, whenever the last allowed part exactly filled its buffer, one more read is issued and any error that isn't `io.EOF` fails the whole upload — a drained `*os.File` that a deferred `Close` already got to, a connection reset after the final byte, or a wrapper returning its own sentinel instead of `io.EOF`. Refusing to complete when EOF can't be proven is the right call, but it's worth saying so in the doc for the three call sites (`api-put-object.go:525`, `api-put-object-multipart.go:183`, `api-put-object-streaming.go:606`), because the failure arrives after every byte has already been uploaded and looks like a transport fault rather than a deliberate policy.

api-put-object.go
```diff
@@ -322,6 +322,9 @@ func (a completedParts) Less(i, j int) bool { return a[i].PartNumber < a[j].Part
 //     For larger objects (up to ~48.83TiB), set PutObjectOptions.PartSize
 //     to control memory usage and enable uploads beyond 5TiB.
 //
+//     The ~48.83TiB ceiling is the product of the client's max part size and
```
poornas, line 325: While this block is being updated: the list above it still says single PUT below 16MiB / multipart above, which no longer describes the routing after 382-389. The choice now also depends on `MaxSinglePutObjectSize`, and `DisableMultipart` above that limit is a hard error rather than a large single PUT.

api-put-object-common.go
```diff
@@ -131,6 +171,23 @@ func OptimalPartInfo(objectSize int64, configuredPartSize uint64) (totalPartsCou
 	return totalPartsCount, partSize, lastPartSize, nil
```
poornas, line 177: For `objectSize == 0` the automatic branch returns `totalPartsCount == 0` with `lastPartSize == l.minPartSize()` (5MiB), because line 170 computes `0 - int64(0-1)*partSize`. No in-repo caller reaches it — every internal call passes either `-1` or a nonzero `configuredPartSize`, and `putObjectMultipartStream` is gated by `size > partSize`. Since `OptimalPartInfo` is exported, returning `0` for an empty object would be less surprising.

api-put-object.go
```diff
@@ -415,7 +418,7 @@ func (c *Client) putObjectMultipartStreamNoLength(ctx context.Context, bucketNam
 	var complMultipartUpload completeMultipartUpload
 
 	// Calculate the optimal parts info for a given size.
-	totalPartsCount, partSize, _, err := OptimalPartInfo(-1, opts.PartSize)
+	totalPartsCount, partSize, _, err := c.optimalPartInfo(-1, opts.PartSize)
```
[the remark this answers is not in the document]
klauspost: What was done

  api-put-object-common.go:147:

  configuredPartSize = uint64(max(minPartSize, l.minPartSize()))

  The finding said use l.minPartSize() instead of the package-level minPartSize. I used max(...) rather than replacing
  outright.

  Why

  They are two different constants with two different jobs:

  - minPartSize = 16 MiB — an internal rounding/threshold heuristic, not an S3 limit.
  - defaultMinPartSize = 5 MiB — the actual S3 minimum, what l.minPartSize() resolves to when unset.

  So a literal replacement would drop the default rounding unit from 16 MiB to 5 MiB and change the part layout for
  every existing user on default limits — a 1 GiB object would go from 64 parts of 16 MiB to 205 parts of 5 MiB. Part
  boundaries determine multipart ETags, so that's a silent, observable change for everyone, in order to fix a bug that
  only bites clients who raise MinPartSize above 16 MiB. max() fixes the bug and leaves defaults untouched.

  Just re-verified:

  defaults:            1GiB → 16777216   (16 MiB, unchanged)
  MinPartSize=64MiB:   1GiB → 67108864   (64 MiB, honored)
  MinPartSize=1MiB:    1GiB → 16777216   (16 MiB, unchanged)

  The finding's stated goal — "ensure all existing callers honor UploadLimits.MinPartSize" — is met; only its prescribed
  mechanism differs.

  Since then

  Round 3 added a second mechanism in the same branch: an explicit floor at l.minPartSize() after rounding. The max()
  handles the rounding unit; the floor catches the case where objectSize < maxPartsCount makes the quotient round to
  zero. Both are needed.

  Coverage is TestOptimalPartInfoRaisedMinPartSize (64 MiB minimum at 1 GiB — the exact example from the round-3
  feedback) — it fails if either mechanism is reverted.

  One caller that deliberately still uses the 16 MiB constant

  api-put-object.go:377 — partSize = minPartSize when opts.PartSize == 0. That's the single-PUT-vs-multipart routing
  threshold, not a part size sent to the server, so it's correctly the internal heuristic and not a limit. Every path
  that produces real part sizes goes through c.optimalPartInfo → c.limits, as the caller list confirms. The one exported
  exception is OptimalPartInfo, which uses UploadLimits{} by design and documents that a client with custom limits may
  differ.

poornas (changes requested):

api-put-object.go
klauspost: ack

api-put-object.go
klauspost: ack

api-put-object-common.go
klauspost: Documented on `errIfMoreData` itself rather than repeated at the three call sites — they already carry a one-line "why we probe" comment that leads there, and I would rather not have the policy drift between three copies. Everything else as described: the late arrival and the "looks like a transport fault" framing are both called out.

api-put-object.go
klauspost: Resolved by dropping the 382-389 block rather than by rewording: with that gone the routing is master's again, so "single PUT below 16MiB / multipart above" is accurate and `MaxSinglePutObjectSize` no longer enters the decision. Left the list as it was.

api-put-object-common.go
klauspost: Already fixed — `optimalPartInfo` returns early for `objectSize == 0` with no configured part size, so `OptimalPartInfo(0, 0)` is `(0, 0, 0, nil)`. Scoped to the automatic branch so an explicit part size above a zero object still returns the existing error. Covered by `TestOptimalPartInfoEmptyObject`.

api-put-object.go
klauspost: Correction to the ack above — the final shape gates on the field being *explicitly set* rather than removing the check entirely:

```go
if maxSinglePut := c.limits.MaxSinglePutObjectSize; maxSinglePut > 0 && size > maxSinglePut {
```

Note the raw field, not the accessor: an unset `MaxSinglePutObjectSize` means no gate, so the snowball path and every other default-limits `DisableMultipart` upload behave exactly as on master. Callers who configure the limit opt in to enforcement. Did not take the `IsAmazonEndpoint` route — it makes the default endpoint-dependent, and AIStor sets the field explicitly anyway, so the gate is live for it either way.

One wrinkle worth knowing: `UploadLimits{}` and `UploadLimits{MaxSinglePutObjectSize: 5 << 30}` now differ, even though 5GiB is the documented default. Called out on the field and on the struct doc.

upload-limits.go
```diff
@@ -0,0 +1,145 @@
...
+	if l.minPartSize() > l.maxPartSize() {
+		return errInvalidArgument("UploadLimits.MinPartSize cannot be larger than UploadLimits.MaxPartSize")
+	}
+	if l.maxPartSize() > math.MaxInt64/l.maxPartsCount() {
+		return errInvalidArgument("UploadLimits.MaxPartSize multiplied by UploadLimits.MaxPartsCount overflows int64")
+	}
+	// The part layout is computed in float64. A MaxPartSize that rounds to or
+	// above 2^63 does not convert back into int64, and callers allocate buffers
+	// of the part size the layout reports.
+	if float64(l.maxPartSize()) >= math.MaxInt64 {
```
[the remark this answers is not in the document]
klauspost: I think we are getting so nitpicky it is pointless. I don't see how your proposal has any practical impact. Feel free to correct me.

klauspost:
@coderabbitai "Client.UploadLimits [...] does not accept context.Context". Adding context to a call that always finishes instantly is just plain bad practice, adding complication to the caller with no benefits. Context is for (potentially) long-running tasks, not for helper functions. Please record this.

klauspost:
@coderabbitai You are still showing as "Requesting Changes". Which ones?

klauspost:
@coderabbitai Then you should either clear your status or approve changes. You are blocking a potential merge.

klauspost:
@coderabbitai approve

api.go
```diff
@@ -110,6 +110,10 @@ type Client struct {
 	trailingHeaderSupport bool
 	maxRetries            int
 
+	// Upload limits enforced before sending a request. Always read through
+	// its accessors, which resolve zero fields to the S3 defaults.
```
poornas, line 114:
```suggestion
	// Upload limits enforced before sending a request. Read through its
       // accessors, which resolve zero fields to the S3 defaults — except
      // MaxSinglePutObjectSize on the PutObject path, where an unset field
     // deliberately means "no limit" (see api-put-object.go).
```

poornas (approved):
LGTM, minor comment nit

PR #2301, opened 2026-09-04
signer: fix streaming signer to include Content-Type in SignedHeaders (fixes #2300)

### LINKED ISSUE #2300: Streaming/chunked signer excludes Content-Type from SignedHeaders, causing rejections on hardened S3-compatible servers (Ceph RGW)
## Summary

The chunked/streaming SigV4 signer (`pkg/signer/request-signature-streaming.go`) unconditionally excludes `Content-Type` from `SignedHeaders` via `ignoredStreamingHeaders`, even when `Content-Type` is present as an actual header on the request. This differs from the non-streaming signer (`pkg/signer/request-signature-v4.go`), which correctly includes `Content-Type` in `SignedHeaders` whenever it's present.

Recently-hardened S3-compatible servers (e.g. Ceph RGW, post CVE-2026-54330) now reject any request where a present header is absent from `SignedHeaders`, causing streaming/chunked uploads from minio-go to fail with errors like:

```
Signature rejected: 'content-type' supplied but not in CanonicalHeaders.
```

## Environment

- minio-go version: v7.1.0 (confirmed same behavior on `master` as of 2026-09)
- Consumed via: `github.com/thanos-io/objstore` → used by Grafana Mimir (ingester block/meta.json uploads)
- S3 backend: Ceph RGW (Tentacle, post `rgw_sigv4_insecure` hardening for CVE-2026-54330)

## Root cause

Two separate header-exclusion maps exist for two separate signing paths:

**`pkg/signer/request-signature-v4.go`** (non-streaming `SignV4`):
```go
var v4IgnoredHeaders = map[string]bool{
	"Accept-Encoding": true,
	"Authorization":   true,
	"User-Agent":      true,
}
```
Content-Type is *not* excluded here — signed correctly.

**`pkg/signer/request-signature-streaming.go`** (`StreamingSignV4`):
```go
var ignoredStreamingHeaders = map[string]bool{
	"Authorization": true,
	"User-Agent":    true,
	"Content-Type":  true,
}
```
Content-Type *is* hardcoded as ignored here, regardless of whether it's present on the request.

Per the AWS SigV4 spec, if `Content-Type` is present on the request, it must be included in `SignedHeaders`:
> "If the Content-Type header is present in the request, you must add it to the CanonicalHeaders list."
(https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-create-signed-request.html)

The streaming signer violates this whenever Content-Type is set on a chunked upload.

## Reproduction

Captured via minio-go's own `trace.enable: true` against a real Ceph RGW endpoint, on a PUT of a small metadata object (triggers the streaming signer due to chunked transfer):

```
PUT /<bucket>/<path>/meta.json HTTP/1.1
Host: <redacted>
User-Agent: MinIO (linux; amd64) minio-go/v7.0.98 thanos-ingester/ (go1.26.4)
Content-Length: 827
Authorization: AWS4-HMAC-SHA256 Credential=<redacted>/20260903/us/s3/aws4_request,SignedHeaders=host;x-amz-content-sha256;x-amz-date;x-amz-decoded-content-length,Signature=<redacted>
X-Amz-Content-Sha256: STREAMING-AWS4-HMAC-SHA256-PAYLOAD
X-Amz-Date: 20260903T154549Z
X-Amz-Decoded-Content-Length: 653
Accept-Encoding: gzip
```

Note `Content-Type: application/octet-stream` is present as a header, but absent from `SignedHeaders=host;x-amz-content-sha256;x-amz-date;x-amz-decoded-content-length`.

## Expected vs actual

- **Expected:** `Content-Type` is either omitted from the request entirely, or included in `SignedHeaders` to match the non-streaming signer's behavior — consistent with the AWS spec.
- **Actual:** `Content-Type` is present on the wire but categorically excluded from `SignedHeaders` in the streaming signer, regardless of spec compliance.

## Suggested fix

Remove `Content-Type` from `ignoredStreamingHeaders` in `request-signature-streaming.go`, aligning its behavior with `v4IgnoredHeaders` in `request-signature-v4.go` (i.e. sign Content-Type when present, don't force-exclude it).

## Related context

- This surfaced after Ceph shipped stricter SigV4 verification in RGW (CVE-2026-54330, Tentacle v20.2.4 / Squid v19.2.6, Aug 19 2026), which now rejects any present-but-unsigned header rather than tolerating it.
- Ceph has since relaxed the *server-side* enforcement to only require Content-Type in SignedHeaders if the client's own `SignedHeaders` implies it should be there: https://github.com/ceph/ceph/pull/71364 — which will restore compatibility once released, but the underlying minio-go signer behavior still looks like a spec deviation worth fixing independently, since other S3-compatible backends may enforce this more strictly.
- A similar report against a different consumer of minio-go/objstore: https://github.com/restic/restic/issues/22033

jiuker:
## Summary
Fixes #2300

The streaming SigV4 signer (`StreamingSignV4`) was unconditionally excluding `Content-Type` from `SignedHeaders` via `ignoredStreamingHeaders`, even when `Content-Type` was present on the request. This violates the AWS SigV4 specification.

## Root Cause
```go
// Before (incorrect)
var ignoredStreamingHeaders = map[string]bool{
    "Authorization": true,
    "User-Agent":    true,
    "Content-Type":  true,  // ❌ Wrong: hardcoded exclusion
}
```

The non-streaming signer (`v4IgnoredHeaders`) correctly does **not** ignore `Content-Type`.

## Evidence
| Source | Finding | Link |
|--------|---------|------|
| **AWS SigV4 Spec** | Spec explicitly requires signing Content-Type when present in request | [docs.aws.amazon.com](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-create-signed-request.html) |
| **Non-streaming signer** | `v4IgnoredHeaders` omits Content-Type → it gets signed when present | [request-signature-v4.go](https://github.com/minio/minio-go/blob/master/pkg/signer/request-signature-v4.go#L61-L65) |
| **AWS SDK for .NET** | Fixed identical bug in streaming signer | [aws/aws-sdk-net#678](https://github.com/aws/aws-sdk-net/issues/678) |
| **Ceph RGW CVE-2026-54330** | Patched versions (20.2.4+, 19.2.6+) now reject requests with present-but-unsigned headers | [docs.ceph.com](https://docs.ceph.com/en/latest/security/CVE-2026-54330/) |
| **AWS CLI behavior** | Real streaming requests include `content-type` in `SignedHeaders` | [seaweedfs/seaweedfs#6713](https://github.com/seaweedfs/seaweedfs/issues/6713) |
| **StackOverflow** | Community confirms spec requires Content-Type in signature when present | [stackoverflow.com](https://stackoverflow.com/questions/79754009) |

## Impact
- Fixes streaming/chunked uploads to Ceph RGW (Tentacle 20.2.4+, Squid 19.2.6+)
- Fixes compatibility with any S3-compatible backend enforcing SigV4 strictly
- Aligns streaming signer with non-streaming signer behavior

## Testing
- All existing signer tests pass
- Verified `Content-Type` now appears in `SignedHeaders` when present on request

### DIFF
--- a/pkg/signer/request-signature-streaming.go
+++ b/pkg/signer/request-signature-streaming.go
@@ -48,10 +48,14 @@ const (
 
 // Request headers to be ignored while calculating seed signature for
 // a request.
+//
+// Per AWS SigV4 specification:
+//   "If the Content-Type header is present in the request, you must add it
+//   to the CanonicalHeaders list."
+//   https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-create-signed-request.html
 var ignoredStreamingHeaders = map[string]bool{
 	"Authorization": true,
 	"User-Agent":    true,
-	"Content-Type":  true,
 }
 
 // getSignedChunkLength - calculates the length of chunk metadata
--- a/pkg/signer/request-signature-streaming.go
+++ b/pkg/signer/request-signature-streaming.go
@@ -50,9 +50,10 @@ const (
 // a request.
 //
 // Per AWS SigV4 specification:
-//   "If the Content-Type header is present in the request, you must add it
-//   to the CanonicalHeaders list."
-//   https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-create-signed-request.html
+//
+//	"If the Content-Type header is present in the request, you must add it
+//	to the CanonicalHeaders list."
+//	https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-create-signed-request.html
 var ignoredStreamingHeaders = map[string]bool{
 	"Authorization": true,
 	"User-Agent":    true,


klauspost (approved):
LGTM. I don't see any reason *not* to sign it.

Only minor worry would be reverse proxies fiddling with it. Maybe @harshavardhana remembers why it was excluded - or maybe AWS just did at some point.

harshavardhana:
It was something quite old since then that list has been updated and content-type has been added as expected signedheaders

PR #2302, opened 2026-09-09
fix: honor the caller's TLS trust on the RDMA path

harshavardhana:
## What

RDMA transfers are driven by libminiocpp, which issues its own S3 requests rather than going through this package's `http.Client`. A
`TLSClientConfig` set on that client therefore governed the HTTP path alone: the C++ side verified against the OpenSSL trust store and
could not see it.

So an RDMA `PutObject`/`GetObject` failed against exactly the self-signed and private-CA endpoints where the same client's plain S3 calls
worked, and `--insecure` could not rescue it. There is no fallback to the HTTP data path once RDMA is selected
(`api-put-object.go:332`, `api-get-object.go:49` return the error), so the operation failed outright rather than degrading.

## How

The trust decision is carried across to libminiocpp via `miniocpp_client_new_tls`, released in
[minio-cpp v1.0.0](https://github.com/minio/minio-cpp/releases/tag/v1.0.0).

It is **derived from the transport** rather than exposed as a new option, so a caller's existing `--insecure` keeps working with no
call-site change. Only `*http.Transport` can be inspected; a caller that wraps its transport in another `RoundTripper` keeps verification
rather than having this guess its way into a weaker setting.

A private CA is still named by `SSL_CERT_FILE`, which libminiocpp reads per request — a `*x509.CertPool` cannot be turned back into the
file path that argument wants, so `nil` is passed for it.

## The other two commits

**`-lminio`.** minio-cpp 1.0.0 builds the library as `libminio` (with a stable `libminio.so.1` soname), so the link flag follows it. The
RDMA lane pins `v1.0.0`, which is also the first release carrying `miniocpp_client_new_tls` — `v0.7.0` has neither.

**apt hardening.** The lane's `apt-get update` fails the job when the runner image's third-party lists serve a bad index; this cost three
green lanes in minio-cpp within one minute on unrelated code. The lists nothing here installs from are dropped, the update retries once,
and `apt-get install` stays the gate.

## Test

`TestRDMASkipCertVerify` covers the derivation: insecure transport over https, a verifying transport, insecure without https (nothing to
skip), no `TLSClientConfig`, the default transport, a nil client, and a wrapped transport. Forcing the helper to return `false` fails the
insecure-over-https case, so the test guards the behaviour rather than restating it.

The helper is in an **untagged** file on purpose. `rdma_test.go` is `//go:build rdma` and only builds in the RDMA lane, which is why this
gap went unnoticed — the derivation is pure Go and now runs in the ordinary test lane. `go vet -tags rdma` also passes against the
minio-cpp 1.0.0 header, so the cgo call to the new symbol type-checks.

## Summary by CodeRabbit

* **Bug Fixes**
  * RDMA transfers now honor the client’s TLS certificate-verification settings.
  * Secure connections verify certificates by default, while explicitly configured insecure connections can skip verification as intended.
  * This aligns TLS behavior between regular and RDMA-based transfers.

* **Tests**
  * Added coverage for secure and insecure TLS configurations, missing or wrapped transports, and nil or default clients.

### DIFF
--- a/.github/workflows/go-rdma.yml
+++ b/.github/workflows/go-rdma.yml
@@ -31,11 +31,13 @@ jobs:
       uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       with:
         repository: minio/minio-cpp
-        # Pinned: this job copies vendor/s3rdma out of the checkout, so a
-        # floating main makes the build race whatever landed there. It already
-        # did -- a run started before minio-cpp vendored libs3rdma and failed
-        # on the missing directory. Bump this when picking up a minio-cpp fix.
-        ref: v1.0.0 # first release carrying miniocpp_client_new_tls, as libminio
+        # Pinned to a commit, like vcpkg and the actions above: this job copies
+        # vendor/s3rdma out of the checkout and installs what it builds, so a
+        # floating ref makes the build race whatever landed there. It already
+        # did -- a run started before minio-cpp vendored libs3rdma and failed on
+        # the missing directory. A tag would still leave that open, since a tag
+        # can be moved after review. Bump this when picking up a minio-cpp fix.
+        ref: 78e1db407d97bf5fc643858d9f4f482bff069696 # tag v1.0.0, first release carrying miniocpp_client_new_tls as libminio
         persist-credentials: false
         path: "minio-cpp"
 
--- a/.github/workflows/go-rdma.yml
+++ b/.github/workflows/go-rdma.yml
@@ -35,7 +35,7 @@ jobs:
         # floating main makes the build race whatever landed there. It already
         # did -- a run started before minio-cpp vendored libs3rdma and failed
         # on the missing directory. Bump this when picking up a minio-cpp fix.
-        ref: v0.7.0 # first release carrying the libs3rdma transport
+        ref: v1.0.0 # first release carrying miniocpp_client_new_tls, as libminio
         persist-credentials: false
         path: "minio-cpp"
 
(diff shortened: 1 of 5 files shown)


klauspost:
```
Error: ./rdma.go:85:10: could not determine what C.miniocpp_client_new_tls refers to
Error: Process completed with exit code 1.
```

harshavardhana:
@klauspost that build error came from the earlier push, when the RDMA lane still pinned minio-cpp `v0.7.0` — which predates
`miniocpp_client_new_tls`. minio-cpp [v1.0.0](https://github.com/minio/minio-cpp/releases/tag/v1.0.0) is now released with that entry
point, and with the library renamed to `libminio` (and given a stable `libminio.so.1` soname, which it never had — it used to put the full
version in the soname, so every release orphaned every already-linked binary). The lane pins that release and links `-lminio`, and both
RDMA lanes are green.

Also in the latest push, addressing review feedback:

- The minio-cpp checkout is pinned to the commit `v1.0.0` points at rather than the tag, matching how vcpkg and every action in this file
  are already pinned. This job copies `vendor/s3rdma` out of that checkout and installs what it builds, so the ref decides what lands on
  the runner.
- `Analysis (1.25.x)` was red for an unrelated reason: `setup-go` pins `GOTOOLCHAIN=local`, and `x/vuln` v1.8.0 declares `go 1.26.0`, so
  `go install govulncheck@latest` could not build the tool on that leg at all. It now builds the tool with `GOTOOLCHAIN=auto` while the
  scan still runs on the matrix leg. That one hits every PR in this repo, not just this branch — happy to split it into its own PR if
  you'd rather it landed separately.
- The RDMA lane's `apt-get update` no longer fails the job when the runner image's third-party lists serve a bad index; the lists nothing
  here installs from are dropped and the update retries once, with `apt-get install` still the gate.

Not addressed: zizmor flags this workflow for having no `permissions:` block. It's a fair point but predates this change, so I left it out
rather than widen the diff.

PR #2303, opened 2026-09-10
Report listing checksums consistently, add WithRestoreStatus

klauspost:
S3 returns `<ChecksumAlgorithm>` and `<ChecksumType>` on `ListObjects`, `ListObjectsV2` and `ListObjectVersions` for every object that has a checksum. It is ungated — no header or query parameter — and the checksum value itself is never returned. AiStor additionally returns the values when listing with `WithMetadata`.

## Changes

- **Fix: xxhash checksums were silently dropped from non-versioned listings.** `ObjectInfo` was missing the all-caps `ChecksumXXHASH64/3/128` xml tags that `Version`, `ObjectPart` and `CompletePart` already carry. `encoding/xml` matches element names case-sensitively, so `<ChecksumXXHASH64>` never matched. The fields were dead; no server has ever emitted the mixed-case spelling.
- Fill in `ObjectInfo.ChecksumAlgorithm` from the checksum value when the server does not name the algorithm. AWS never sends `x-amz-checksum-algorithm` on HEAD/GET, so `StatObject` left it empty while listings set it. A server-supplied value is never overwritten.
- Add `ObjectInfo.Checksum(ChecksumType) string`, matching the existing `ObjectPart.Checksum`.
- Add `ListObjectsOptions.WithRestoreStatus`, which sends `x-amz-optional-object-attributes: RestoreStatus` on all three list APIs and reports the reply in `ObjectInfo.Restore` — the field `StatObject` already fills from `x-amz-restore`. Caller headers set via `ListObjectsOptions.Set` are cloned, not mutated.

No API breakage: `ObjectInfo` is decode-only (`xml.Marshal` on it already fails on `StringMap`), JSON keys are unchanged, and the new fields and method are additive.

## Verified

Against AWS S3 (us-east-2) and AiStor.

- `x-amz-optional-object-attributes` accepts only `RestoreStatus`; `Checksum` is rejected with `InvalidArgument`. Checksum algorithm and type come back on plain listings with no header.
- Objects stored under CRC32/CRC32C/SHA1/SHA256/CRC64NVME/MD5/SHA512/XXHASH64/XXHASH3/XXHASH128 report the right algorithm and mode on all three list APIs; a multipart upload reports `COMPOSITE`. AiStor with `WithMetadata` returns matching values, xxhash included.
- `WithRestoreStatus` against a GLACIER object mid-restore returns `OngoingRestore: true`, and after an expedited restore completed, `OngoingRestore: false` with `ExpiryTime` set. Absent when not requested. Same on V1, V2 and versions. AiStor ignores the header and returns a normal listing.

New unit tests: `TestListObjectsChecksums`, `TestListObjectsWithRestoreStatus`, `TestToObjectInfoChecksumAlgorithm`. New functional test: `testListObjectsChecksums`, plus a `WithRestoreStatus` variant in `testListObjects`.

## Summary by CodeRabbit

* **New Features**
  * Added support for requesting restore status when listing objects and object versions.
  * Added restore status and expiry details to object metadata.
  * Added access to XXHash checksum values and automatic checksum algorithm detection.
  * Checksum information is now available across standard, versioned, multipart, and metadata-enabled listings.

* **Bug Fixes**
  * Improved parsing of restore and checksum fields from service responses.

* **Tests**
  * Added coverage for restore status requests and checksum handling across listing APIs.


klauspost:
@coderabbitai review

harshavardhana (approved):

PR #2306, opened 2026-09-14
fix CopyObject swallowing errors embedded in a 200 OK response

### LINKED ISSUE #2305: CopyObject returns nil error when S3 sends a 200 OK with an embedded error body
CopyObject can fail midway and S3 still answers 200 OK, with the actual error embedded as an `<Error>` document in the body: https://docs.aws.amazon.com/AmazonS3/latest/API/API_CopyObject.html

CopyObject doesn't set expect200OKWithError, so executeMethod returns early on the 200 and the body is decoded into copyObjectResult, which has no Error field. The caller gets back a nil error and an UploadInfo with an empty ETag for an object that was never copied.

The flag is already there for exactly this case — CompleteMultipartUpload and RemoveObjects both set it. CopyObject looks like the one 2xx-with-error path that was missed. aws-sdk-go-v2 handles this for CopyObject too (it rewrites the status to 500 when the root element is <Error>).

Happy to send a PR that sets the flag.

huanghaoyuanhhy:
Fixes #2305

Sets expect200OKWithError on CopyObject so executeMethod parses the body for an embedded <Error> instead of trusting the 200 status, matching what CompleteMultipartUpload and RemoveObjects already do.

Added Test200CopyObjectWithError, modeled on Test200MultipartUploadWithError: it fails before the change (CopyObject returns nil) and passes after, including the retry. Ran go test -short -race ./... and go build, both clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Copy operations now correctly detect and report errors embedded in otherwise successful responses.
  - Failed copy requests are retried according to the configured retry policy.

- **Tests**
  - Added coverage verifying error handling and retry behavior for copy operations.


klauspost (approved):

klauspost:
Windows CI unrelated, fixed in #2307

api-copy-object.go
```diff
@@ -41,6 +41,10 @@ func (c *Client) CopyObject(ctx context.Context, dst CopyDestOptions, src CopySr
 		bucketName:   dst.Bucket,
 		objectName:   dst.Object,
 		customHeader: header,
+		// S3 can answer a failed CopyObject with 200 OK and an embedded <Error>
+		// body, so this request must parse the body for an error instead of
+		// trusting the status code.
```
jiuker: Please add this document URL https://repost.aws/knowledge-center/s3-resolve-200-internalerror in the comments, and there's no need for such a lengthy explanation.

jiuker (changes requested):
Code is fine. Just for the comments.

jiuker (approved):
LGTM

